Issue #2, opened 2026-08-06
Day-1 gate: contracts/ and docs/decisions/ ADRs 0001-0004 are missing


tanifgit:
`CONTRIBUTING.md` §6 makes Day 1 a gate: nobody starts implementation until `contracts/proxy-api.md`, `contracts/healthscan-api.md`, and `contracts/samples/` are merged to `main`. It also asks for ADRs `0001`–`0004` in `docs/decisions/` recording the four open decisions from §7.4 of the MVP doc.

**Neither exists.** `contracts/` is absent and `origin/main` is still at the initial commit. Raising it rather than letting it stay implicit.

### Current state

Dev C has started the dashboard against a transcription of `apps/dashboard/CLAUDE.md` §2.3 (which matches §5 of the MVP docx) instead of a published contract — a deliberate call to avoid idling, but it means the gate is open.

Exposure is bounded. Of ~1,750 lines in `apps/dashboard/src`, ~300 are contract-coupled (types, guards, finding metadata) plus ~580 lines of fixtures. The remaining ~1,450 — shell, grid, list, tiles, formatting, severity, icons — are contract-independent and survive any renaming.

### The four ADRs (§7.4 of the MVP doc)

- `0001` — **Where the detection engine lives.** Recommendation: in the proxy, keeping MVP 1 free of custom IRIS code.
- `0002` — **Baseline strategy.** Recommendation: in-memory rolling window; persistence later.
- `0003` — **Thresholds.** Recommendation: configurable via a small config file with conservative defaults.
- `0004` — **Mock-first contracts.** All three confirm they're building against mocked contracts from Day 1.

**0004 is the one currently load-bearing without being recorded** — it's the agreement Dev C is relying on right now.

### Suggested close-out

1. Dev B answers the nine schema questions (#1) and lands `contracts/healthscan-*` + samples.
2. Dev A lands `contracts/proxy-*` + `samples/metrics-dump.txt`, `alerts.json`, `proxy-response.json`.
3. Someone writes the four ADRs — `docs/decisions/` is shared/PR-gated, so not one developer's call unilaterally.
4. Dev C reconciles the 13 `CONTRACT-Q` sites and re-points fixtures at the shared samples.

### Also missing from the §8 checklist

- `.github/CODEOWNERS`
- PR template
- `ci.yml` with the four per-path jobs

[...]

Issue #1, opened 2026-08-06
Contract: 9 questions on the findings/hosts schema before contracts/ is finalized


tanifgit:
The Health Scan dashboard's first pass is built (host grid, findings list, severity summary) against the schema in §5 of the MVP docx, mirrored in `apps/dashboard/CLAUDE.md` §2.3. It renders all four LABDEMO hosts and all eight finding types from fixtures.

`contracts/` isn't published yet, so I transcribed that spec section rather than wait — **my types are a spec-derived guess, not the contract.** Below is exactly what I assumed. Most are probably right; I need to know which aren't before the files are locked.

**Replying "all correct except N, M" is enough.**

| # | Question | What I coded | Cost if different |
|---|---|---|---|
| **4** | Is `id` stable across polls for the same ongoing condition? Do findings vanish when resolved, or gain `resolvedAt`/`status`? | ids stable while condition persists; findings disappear when cleared | **High** — new-finding highlight + drawer-survives-poll both break |
| **3** | How is baseline warm-up represented? Top-level flag, per-host field, or `baselineValue: null`? | `baselineValue: number \| null` | Medium — guards + detail view |
| **1** | Exact `status` enum for hosts | `OK \| Warning \| Error \| Inactive` | Low — unknown renders grey |
| **2** | Exact finding `type` strings | the eight snake_case names in §1.3 | Low — unknown gets neutral icon + humanized label |
| **6** | Are `avgProcessingTime` / `avgQueueingTime` in **seconds**? | seconds (`0.08` → displayed `80 ms`) | Medium — every duration off by 1000× |
| **7** | Zero findings → `[]` or `404`? What during engine startup? | *(no assumption — can't guess this one)* | Medium — blocks error/empty handling |
| **8** | Is `finding.host` always exactly a `host.host` value? | exact string match | Low — unmatched findings just don't tint a card |
| **5** | Is `findings` sorted `detectedAt` desc server-side? | UI sorts (newest first, severity tiebreak) | None — I sort regardless |
| **9** | Will the API send `Access-Control-Allow-Origin`? | using the Vite dev proxy, so CORS is moot | None either way |

**Q4 matters most.** If ids churn per poll, every finding looks new on every poll — the highlight animation has to go, and the detail drawer can't stay open across a refresh. It changes behavior, not just types.

Two asks beyond the answers:

[...]

PR #3, opened 2026-08-06
contracts: publish Health Scan findings/hosts contract (Day-1 gate)

Ari-Glikman:
Closes the Day-1 gate for the **Dev B → Dev C** contract. Addresses #1 (the nine schema questions) and the `contracts/` half of #2.

Publishes `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts`, and the two shared samples. Follows §5 of the MVP doc **exactly** — no fields added or removed.

## For Dev C: one answer contradicts your assumptions

**`status` has no `Warning`.** I read the real enum from IRIS source rather than guessing:

```ts
export type HostStatus =
  | 'OK' | 'Error' | 'Inactive' | 'Retry' | 'Stopped' | 'Unconfigured' | 'Disabled';
```

`Warning` doesn't exist, and there are three statuses you didn't have. That's a one-line edit to your `HostStatus` union — and because unknown statuses already render neutral per your `CLAUDE.md` §2.4, your current code degrades correctly even before you make it.

**The other eight you got right**, including Q4, the one you flagged as highest-cost:

| # | Answer |
|---|---|
| **Q4** | ids **stable** while the condition persists; findings **disappear** when cleared. Keep the highlight animation and the poll-surviving drawer. |
| **Q3** | `baselineValue: number \| null`, as you assumed. |
| **Q6** | **Seconds** — confirmed empirically, not assumed. Cloud API configured at 0.05s latency reports `avgProcessingTime: 0.05`; Lab Router reports `0.08`. Your `0.08 → "80 ms"` is correct. |
| **Q7** | `200` + `[]`, never `404`. Full table in §3. |
| Q2, Q5, Q8, Q9 | Confirmed as assumed. Sorting is server-side too, so your client sort is harmless — keep it. |

## Three questions you couldn't have known to ask

Surfaced from live IRIS metrics while standing up LABDEMO:

- **Q10 — `type` is normalized.** IRIS reports business processes as `actor`, not `process`. We map `actor → process` so the contract keeps the MVP doc's vocabulary; the IRIS word never reaches you.
- **Q11 — `lastActivity` is derived.** IRIS gives *elapsed seconds since last activity*, not a timestamp. We compute `now − elapsed`, so treat it as ±10s — correct for "2 minutes ago", not for sub-second ordering.
- **Q12 — `avgProcessingTime` is an aggregate.** IRIS emits these per `(host, messagetype)`, so one host has *several* series. We collapse them to one number, weighted by `iris_interop_sample_count`.

## Samples are real, not invented

Captured from a live LABDEMO production, including a genuinely induced degraded state — I disabled Cloud API and watched the queue reach 48 — rather than plausible-looking numbers. `findings-response.json` covers all eight finding types, all three severities, and two `baselineValue: null` cases so the warm-up path gets exercised.

## Verification

- Both samples validate against the schema (`ajv` + `ajv-formats`)
- Four **negative** cases correctly rejected: the retired `Warning` status, a string `baselineValue`, an unknown finding type, a non-`Z` timestamp. Without these, "valid" would mean nothing
- `healthscan.d.ts` clean under `tsc --noEmit --strict`

## Known gap — for Dev A, not a contract change

`iris_interop_queued` carries **no `host` label**; it emits once per production. But per-host depth *is* available from `Ens.Util.Statistics:EnumerateHostStatus` (verified: `Cloud API = 48` while disabled), so `Host.queued` stays a required number and no extra IRIS config is needed.

What this does mean: **the proxy must read host status, not only parse `/api/monitor/metrics`.** No contract impact if that holds.

I also have a real 1006-line `/api/monitor/metrics` capture from LABDEMO that would save time on `samples/metrics-dump.txt` — yours to land, happy to hand it over.

## Review

`contracts/` requires all three developers per `CONTRIBUTING.md` §4. `.github/CODEOWNERS` doesn't exist yet, so this won't auto-request — flagging @tanifgit and Dev A directly.

Still open from #2 and **not** in this PR: the four ADRs in `docs/decisions/` (shared, explicitly not one developer's unilateral call), `contracts/proxy-*`, and the CI/CODEOWNERS setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+node_modules/
+dist/
+build/
+.env
+.env.local
+.claude/settings.local.json
+*.mp4
+*.mov
+*.log
+.DS_Store
+Thumbs.db
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,41 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-08-09 — Schema fixes from Dev C's review (Dev B)
+
+Two **schema** defects found by Dev C reviewing PR #3, both with reproductions, both confirmed
+here before fixing. The prose contract was right in both cases; the schema disagreed with it. No
+change to any documented field, so **nothing to reconcile on the consumer side.**
+
+**1. The schema rejected `[]`.** The root `oneOf` failed on an empty array because it satisfied
+*both* `HostsResponse` and `FindingsResponse` vacuously, and `oneOf` requires exactly one match.
+§3 mandates `200` + `[]` for no findings, no hosts, *and* engine startup — so the CI job in
+`README.md` would have failed on the single most common healthy response.
+
+Fixed by removing the root `oneOf` entirely and validating against named definitions. That is
+strictly better than switching to `anyOf`: it also closes a hole where a **hosts array served in
+the findings position validated fine**, because the root schema accepted either and could not
+tell them apart. A check that silently was not happening.
+
+**2. The timestamp `pattern` rejected `toISOString()`.** It forbade fractional seconds, but JS
+always emits milliseconds (`.000Z`) and Python's `isoformat()` gives microseconds. Dev C's eight
+fixtures all failed on this for real. Fixed by allowing optional sub-second digits:
+`(\.[0-9]{1,6})?`. Second precision stays valid, so nothing that passed before now fails.
+
+Chose relaxing over demanding second precision because `format: date-time` already accepted what
+`pattern` rejected — only the stricter one bit, which made it an accident rather than a decision.
+Requiring emitters to slice digits off a native formatter is a tax with no benefit, given
+`lastActivity` is ±10s anyway (Q11).
+
+**Also added `validate.mjs` + `package.json`,** replacing the `ajv-cli` one-liners in `README.md`.
+Dev C found those degrade silently: `-c ajv-formats` resolves from the invocation directory and
+ignores `format` when run elsewhere, and the `#/definitions/...` argument is rewritten into a
+filesystem path by Git Bash on Windows. The script also asserts **five must-reject and four
+must-accept cases**, because both defects here were *structural* — the class that positive-only
+and value-level testing cannot reach.
+
+Verified: `npm run validate` → 11/11 checks pass; both committed samples still valid.
+
 ## 2026-08-06 — Health Scan contract published (Dev B)
 
 Initial publication of `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts`,
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -1,0 +1,41 @@
+# Contract changelog
+
+Every contract change, dated, with the reason. Newest first.
+
+---
+
+## 2026-08-06 — Health Scan contract published (Dev B)
+
+Initial publication of `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts`,
+`samples/hosts-response.json`, `samples/findings-response.json`.
+
+Closes the Day-1 gate for the Dev B → Dev C contract (issues #1, #2). Answers all nine of Dev C's
+schema questions inline in §4 of `healthscan-api.md`.
+
+Follows §5 of the MVP doc exactly — no fields added or removed. Three deviations from what the
+MVP doc *implies*, all forced by what live IRIS actually emits:
+
+- **`status` has no `Warning`.** The real enum, read from IRIS source, is `OK`, `Error`,
+  `Inactive`, `Retry`, `Stopped`, `Unconfigured`, plus `Disabled` from `EnumerateHostStatus`.
+  This is the one answer that contradicts a Dev C assumption — the `HostStatus` union needs
+  `Warning` removed. Existing code still degrades correctly because unknown statuses already
+  render neutral.
+- **`type` is normalized.** IRIS reports business processes as `actor`. We map
+  `actor → process` so the contract keeps the MVP doc's vocabulary.
+- **`avgProcessingTime` / `avgQueueingTime` are aggregates.** IRIS emits these per
+  `(host, messagetype)`, so one host has several series. We collapse them to one number,
+  weighted by `iris_interop_sample_count`.
+
+Units confirmed empirically rather than assumed: Cloud API configured at 0.05s latency reports
+`avgProcessingTime: 0.05`. Seconds, as Dev C assumed.
+
+Samples carry real measured values from the LABDEMO production, including a genuinely induced
+degraded state (Cloud API disabled → queue depth 48), not invented numbers.
+
+### Known gap, not a contract change
+
+`iris_interop_queued` carries no `host` label — it emits once per production. Per-host queue depth
+is available from `Ens.Util.Statistics:EnumerateHostStatus` (verified: `Cloud API = 48` while
+disabled), so `Host.queued` stays a required number. **This needs Dev A's proxy to read host
+status, not only the Prometheus metrics text.** Raised with Dev A separately; no contract impact
+if that holds.
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -37,6 +37,23 @@ The `contracts` CI job validates every file in `samples/` against the schemas. I
 the shared fixtures disagree, CI says so — rather than the Day-5 rehearsal.
 
 ```bash
-npx ajv-cli validate -s healthscan.schema.json -d samples/hosts-response.json
-npx ajv-cli validate -s healthscan.schema.json -d samples/findings-response.json
+cd contracts && npm install && npm run validate
 ```
+
+It checks three things, and the last two are what make the first mean anything:
+
+- each sample against **one named definition** (`HostsResponse` / `FindingsResponse`)
+- structural cases that must **pass** — `[]`, and sub-second timestamps from any language
+- cases that must **fail** — a retired `Warning` status, an unknown finding type, a hosts array
+  served in the findings position
+
+**Do not replace this with an `ajv-cli` one-liner.** Three reasons, all learned the hard way on
+PR #3:
+
+1. Validating against the root schema instead of a named definition cannot tell a hosts array
+   from a findings array. That check silently does not happen.
+2. `ajv validate -r '#/definitions/HostsResponse'` is not portable — Git Bash on Windows rewrites
+   the `#/...` argument into a filesystem path and the command fails.
+3. `-c ajv-formats` resolves relative to the invocation directory, so from the wrong place it
+   silently ignores `format` rather than failing. A validator that quietly gets weaker is worse
+   than one that breaks.
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -1,0 +1,42 @@
+# `contracts/` — the coordination point
+
+Two API contracts, their schemas, and the shared sample payloads. This directory is the only
+path all three developers read, which is why it is PR-gated.
+
+| File | Owner | Consumer |
+|---|---|---|
+| `proxy-api.md`, `proxy.schema.json` | Dev A | Dev B |
+| `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts` | Dev B | Dev C |
+| `samples/metrics-dump.txt`, `samples/alerts.json`, `samples/proxy-response.json` | Dev A | Dev B |
+| `samples/hosts-response.json`, `samples/findings-response.json` | Dev B | Dev C |
+
+## Why the samples matter
+
+`samples/` is the handoff currency. Dev A's `proxy-response.json` is *literally* Dev B's mock
+input; Dev B's `findings-response.json` is *literally* Dev C's mock input. Same bytes, which is
+what makes "works against the mock" predict "works against the real thing."
+
+Samples are captured from live IRIS wherever possible, not invented. The Health Scan samples in
+this directory use real measured values from the LABDEMO production — `avgProcessingTime: 0.08`
+for Lab Router is what IRIS actually reported, not a plausible-looking number.
+
+## Never edit a contract in place
+
+- **Never** edit a file here as part of an implementation task.
+- **Never** add a field to a local type to make code compile. That is a contract change request
+  to the owning developer.
+- A contract change is its own PR: the schema edit, a `CHANGELOG.md` entry with the reason, and
+  a heads-up to the consumer. CODEOWNERS requires all three reviewers.
+
+The reason for the heavy process on a 5-day project: a silent contract change is the failure mode
+that breaks the Day-5 integration, and the review takes 30 seconds.
+
+## Validating
+
+The `contracts` CI job validates every file in `samples/` against the schemas. If a contract and
+the shared fixtures disagree, CI says so — rather than the Day-5 rehearsal.
+
+```bash
+npx ajv-cli validate -s healthscan.schema.json -d samples/hosts-response.json
+npx ajv-cli validate -s healthscan.schema.json -d samples/findings-response.json
+```
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -31,6 +31,29 @@ for Lab Router is what IRIS actually reported, not a plausible-looking number.
 The reason for the heavy process on a 5-day project: a silent contract change is the failure mode
 that breaks the Day-5 integration, and the review takes 30 seconds.
 
+### Estimating what a change costs
+
+When proposing a change, do not estimate the cost from the size of the edit. **The question is not
+how many lines reference a value, it's whether anything *means* something in terms of it.** Cost
+scales with attached meaning, not with characters.
+
+The worked example from this sprint: removing `Warning` from `HostStatus` was one line in a union,
+and it was published as "roughly one line" of consumer impact. The actual reconciliation was **83
+insertions across 10 files** — because seven demo fixtures used `status: "Warning"` to *mean* "this
+host is degraded", so narrowing the enum made a concept the dashboard had built on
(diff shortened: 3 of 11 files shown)


tanifgit:
Reviewed as the consumer. **The contract is good and I've reconciled against it** — `devC/dashboard-scaffold` @ ce8773f. Q1 was the only genuine correction, exactly as you predicted, and reading the enum from IRIS source rather than guessing is the reason my transcription was wrong in a *recoverable* way.

Confirming your empirical answers landed: `0.08 → "80 ms"` unchanged (Q6), highlight animation and poll-surviving drawer kept (Q4), client sort kept as cheap insurance (Q5).

**Two schema defects though — both would fire on your own engine, not on mine.** Requesting changes on the schema only; the prose contract is right in both cases and the schema disagrees with it.

## 1. The schema rejects `[]` — the response §3 mandates

`oneOf` is the problem. An empty array satisfies *both* `HostsResponse` and `FindingsResponse` (vacuously — no items to fail), so `oneOf` fails for matching two branches:

```
$ ajv validate -c ajv-formats -s healthscan.schema.json -d empty.json
empty.json invalid
[ { instancePath: '', schemaPath: '#/oneOf', keyword: 'oneOf',
    params: { passingSchemas: [ 0, 1 ] },
    message: 'must match exactly one schema in oneOf' } ]
```

§3 says `200` + `[]` is the correct response for no findings, no hosts yet, *and* engine startup — so the CI job you'd wire up from `README.md` would fail on the single most common healthy response. `scenario-healthy.json` is exactly this case, and it's the one I care most about not breaking on stage.

Worth noting `oneOf` is also why a *hosts* array validates in the *findings* position — the root schema accepts either, so `-s healthscan.schema.json` can't tell them apart. That's the check silently not happening.

**Suggested fix** — drop the root `oneOf` and have CI target the definitions directly, which is stricter *and* fixes the empty case:

```bash
ajv validate -c ajv-formats -s healthscan.schema.json \
  -r '#/definitions/HostsResponse' -d samples/hosts-response.json
```

Or if you'd rather keep a root entry point, `anyOf` instead of `oneOf` fixes `[]` — but it keeps the hosts-vs-findings ambiguity, so I'd prefer per-definition validation. Your call, you own it.

## 2. The timestamp `pattern` rejects `toISOString()`

`^…T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$` forbids fractional seconds. But `Date.prototype.toISOString()` **always** emits milliseconds:

```
new Date(...).toISOString()  →  '2026-08-06T15:47:52.000Z'   ← pattern REJECTS
                                 '2026-08-06T15:47:52Z'      ← pattern accepts
```

Your samples are hand-written at second precision so they pass. Anything *generated* by a JS/TS engine won't, unless it explicitly slices the milliseconds off — there's no built-in second-precision ISO formatter.

This one caught me for real: **all 8 of my fixtures failed** validation on it, because `mockClient` resolves relative ages via `toISOString()`. I've fixed my side (`toContractIso`, seconds only) since second precision is valid under both readings. But if the engine is Node, you'll hit it too — and if it's Python, `datetime.isoformat()` gives microseconds, which also fails.

**Either is fine by me:** relax the pattern to allow optional `(\.[0-9]{1,3})?`, or keep it strict and treat second-precision as a contract requirement on emitters. Strict is defensible — `lastActivity` is ±10s anyway per Q11, so sub-second digits are noise. Just needs to be a decision rather than an accident, since right now `format: date-time` accepts what `pattern` rejects and only the stricter one bites.

## On your negative-case testing

The four negative cases are the right instinct and I got value from them — but note all four are *value*-level (bad enum, bad type). Both defects above are *structural*, which is the class positive-plus-value testing can't reach. `[]` and a `.000Z` timestamp are the two cheapest structural cases to add, and both are real traffic rather than hypotheticals.

Also: `-c ajv-formats` needs `ajv-formats` resolvable from the *invocation* directory — `npx ajv-cli` alone fails with `Cannot find module 'ajv-formats'` even when it's installed globally. Worth pinning both as devDependencies in whatever workspace the CI job runs from, or the job silently degrades to the weaker check you flagged.

## Answering your two open items

**`samples/metrics-dump.txt`** — yes please hand the 1006-line capture to Dev A. Not mine to land, but a real capture beats a synthesized one for exactly the reason ADR 0004 gives.

**The `:3002` stub** — understood, keeping my throwaway `dev-stub/` until you comment here. It now emits second-precision timestamps and CORS `*` to match your contract, so when I delete it the swap should be invisible. Verified its output validates against your schema.

## One heads-up on my fixtures, not a contract issue

Your real LABDEMO captures run ~0.2–1.2 msg/sec; my fixtures use ~18 msg/sec throughout, invented before any capture existed. Not a contract violation — `messagesPerSec` is just a number — but demo and live won't look continuous side by side, which §5 of my area rules explicitly wants. I'd rather rescale mine to your measured values than have the stage demo contradict the live one. Tracking on my side; flagging so you know the mismatch is known rather than missed.

---

Approving the contract itself once the `[]` case is fixed — that's the only one I'd call blocking, since it breaks the CI job that ADR 0004 calls the main defence. The timestamp one you could reasonably resolve either way, but silently is not one of the options.

tanifgit (changes requested):
Registering my earlier review formally — that was a comment, not a review state, so it wasn't counting toward the three-developer gate `CONTRIBUTING.md` §4 requires on `contracts/`.

**Requesting changes on one item only: the schema rejects `[]`.**

To be clear about what I'm *not* objecting to — the contract itself is good, the twelve answers are right, and Q10–Q12 were worth more than several of my original nine. I've already reconciled the dashboard against it (ce8773f, now in #5). This is a schema/prose disagreement, not a design objection.

**Blocking:** the root `oneOf` fails on `[]` because an empty array satisfies both `HostsResponse` and `FindingsResponse` vacuously. §3 of your own contract mandates `200` + `[]` for no findings, no hosts yet, *and* engine startup — so the schema rejects the response the prose calls normal, and the CI job from `README.md` fails on the most common healthy payload. `scenario-healthy.json` is exactly that case.

Blocking rather than nit because `contracts/` is the directory we've all agreed is expensive to change: landing it means the fix costs a second contract-change PR with a `CHANGELOG.md` entry and three more reviews. Cheaper now.

**Not blocking:** the timestamp `pattern` rejecting `toISOString()` output. Either resolution is fine by me — relax to allow optional milliseconds, or keep it strict and treat second precision as a requirement on emitters. Strict is defensible given `lastActivity` is ±10s anyway per Q11. It just needs to be a decision, since right now `format: date-time` accepts what `pattern` rejects and only the stricter one bites. I've already made my side emit second precision, which is valid either way.

Reproductions for both are in my earlier comment. Happy to re-review immediately.

Ari-Glikman:
Both defects confirmed and **fixed in 2e13515**. You were right on both counts, and both would have fired on our engine rather than yours — thanks for reproducing them instead of just reporting them.

I reproduced each before changing anything:

```
=== DEFECT 1: does the schema reject [] ? ===
empty.json invalid
  schemaPath: '#/oneOf', keyword: 'oneOf', params: { passingSchemas: [Array] }

=== DEFECT 2: does it reject toISOString() fractional seconds ? ===
ts-frac.json invalid
  schemaPath: '#/items/properties/detectedAt/pattern'
```

## 1. Root `oneOf` removed — took your preferred fix

Validating against named definitions, not `anyOf`. Your reasoning decided it: `anyOf` fixes `[]` but keeps the hosts-vs-findings ambiguity, and **that ambiguity is the more dangerous of the two**. A check that silently isn't happening is worse than one that fails loudly on `[]`, because the second gets noticed on day one.

That case is now an explicit must-reject:

```
ok    rejects: hosts array served in the findings position
```

## 2. Timestamp pattern relaxed to `(\.[0-9]{1,6})?`

Chose relaxing over demanding second precision, for the reason you identified — `format: date-time` already accepted what `pattern` rejected, so only the stricter one bit. That made it an accident, not a decision. Taxing every emitter to slice digits off its native formatter buys nothing when `lastActivity` is ±10s anyway (Q11).

Microseconds are allowed too, so a Python proxy works without special-casing. Second precision stays valid, so **your fix needs no revisiting** — and our engine emits second precision independently (`2026-08-09T11:11:37Z`).

## Your three process points, all acted on

**"All four negative cases are value-level."** That was the sharpest thing in your review. Both your defects were structural, which is precisely the class value-level testing can't reach — and I'd have kept missing them. `[]` and `.000Z` are now must-accept cases, plus three more must-rejects.

**`-c ajv-formats` degrades silently.** Confirmed, and worse than you found: `ajv validate -r '#/definitions/HostsResponse'` also fails on Git Bash, because Windows rewrites the `#/...` argument into a filesystem path — so the fix you suggested doesn't run portably as a CLI one-liner.

So the `ajv-cli` invocations are gone, replaced by **`contracts/validate.mjs`** with `ajv` and `ajv-formats` pinned as devDependencies:

```
$ cd contracts && npm run validate
ok    accepts: empty hosts array
ok    accepts: empty findings array
ok    accepts: JS toISOString() milliseconds
ok    accepts: Python isoformat() microseconds
ok    rejects: retired "Warning" host status
ok    rejects: hosts array served in the findings position
ok    rejects: timestamp with no timezone
ok    rejects: unknown finding type
ok    rejects: baselineValue as a string

```

Drop-in for `ci.yml` — exits non-zero on first failure. `README.md` now says *why* not to go back to a one-liner, so the reasoning survives us.

**No documented field changed**, so there's nothing to reconcile on your side. `CHANGELOG.md` has the full entry.

## Your `dead_host` rendering choice — agreed, and better than my table implies

`Error` critical while `Stopped`/`Disabled`/`Unconfigured` read neutral is the right call, and it's a distinction my rule table flattens. **A deliberately stopped host isn't a failing one**, and the finding carries the alarm either way. That's display semantics, squarely yours — I'd rather you make it than inherit mine.

Worth recording that our side reinforces it: `Retry` is deliberately *excluded* from the statuses that trip `dead_host`, because a retrying host is degraded but alive. So your "degraded" rendering for `Retry` matches the engine's own reasoning without either of us coordinating it.

## On the `Warning` fixture cost

> a wrong enum *value* is cheap while a wrong enum value you've built semantics on is not

Recording that, because "one-line edit" was my phrasing and it undersold the work — 62 insertions across 9 files, not one. I was measuring the type change and not the concept it supported. That distinction is worth carrying into any future contract change: the question isn't how many lines reference a value, it's whether anything *means* something in terms of it.

## The stub can go — endpoints are live

Opened as #8. `npm start` in `services/detection-engine/` serves both endpoints on `:3002` **with nothing on `:3001`**, replaying captured LABDEMO fixtures through healthy → degrading → dead → recovery so findings appear *and clear*.

Verified the thing you'd care about most: I captured live responses from the running engine and validated them against this schema. Not "same author wrote both" — actual output through the actual schema.

**Metrics dump** — will hand the 1006-line capture to Dev A as you suggested.

---

Ari-Glikman:
@tanifgit re-review please — this is still blocked on `CHANGES_REQUESTED` and only you can clear it.

Both defects are fixed in 2e13515 (pushed after your Aug 7 review, so GitHub still shows the block). Nothing else has changed since; no documented field moved, so there's nothing new for you to reconcile.

Everything downstream is waiting on this one merging:

| | |
|---|---|
| #3 (this) | blocked on your re-review |
| #8 (engine) | should rebase after this merges, so `src/types/healthscan.ts` and the real contract live in one tree |
| #1, #2 | you said you'd close from your side once this is on `main` |

`cd contracts && npm run validate` if you want to confirm before approving — 11/11, and the two structural cases you found are now explicit assertions in it.

tanifgit (approved):
Both defects fixed, verified against the actual schema rather than the diff. **Approving.**

I re-ran your validator and then re-ran my own two reproductions independently, because a validator written alongside a fix can encode the same blind spot as the fix:

```
$ cd contracts && npm run validate

$ node tools/validate-fixtures.mjs        # my 8 fixtures, against YOUR corrected schema
ok    scenario-baseline-warming.json   ok    scenario-queue-buildup.json
ok    scenario-dead-host.json          ok    scenario-slow-processing.json
ok    scenario-error-storm.json        ok    scenario-system-alert.json
ok    scenario-healthy.json            ok    scenario-throughput-drop.json
```

That second run is the one I cared about. All 8 previously failed on the fractional-seconds pattern; they now pass, and they pass without the `toContractIso` workaround being load-bearing — I confirmed `.000Z` is accepted directly.

## Removing the root `oneOf` was the better fix and I want to be precise about why

I'd have accepted `anyOf`. You were right that it was the weaker option: it fixes `[]` while leaving hosts-in-the-findings-position accepted, and that hole is worse than the bug that exposed it, because `[]` fails loudly on day one whereas a silently-absent check never surfaces at all. Verified the hole is actually closed in both directions:

```
ok   findings endpoint served a hosts array: rejected
ok   hosts endpoint served a findings array: rejected
```

## 22 adversarial probes on the corrected schema — no further holes

Since the timestamp pattern was hand-edited, I probed around the edit rather than just re-testing the reported case:

```
rejected: +00:00 offset instead of Z      rejected: 7 fractional digits
rejected: lowercase 'z'                   rejected: 2026-13-45 (impossible date)
rejected: 25:99:99 (impossible time)      rejected: timestamp with no timezone
accepted: .000Z (JS)                      accepted: .123456Z (Python)
```

Plus: `additionalProperties` rejects extra fields on both objects, float `queued` rejected, unknown `severity` rejected, missing `baselineValue` key rejected, empty `message` rejected, unknown host `type` string accepted (correct — open string per §2), `null` baseline accepted, non-array roots rejected.

Two accepted-by-design cases worth recording as deliberate, since neither is a defect but both would surprise someone reading the schema later:

- **Negative `currentValue` is accepted.** Correct as-is — `Finding.currentValue` has no `minimum` and shouldn't, since a delta-shaped metric could legitimately go negative. Noting it so nobody adds `minimum: 0` reflexively.
- **Duplicate `id`s within one response are accepted.** JSON Schema can't express set-uniqueness on a key, so this is a consumer concern, not a schema gap. I key React lists on `id` (Q4), so duplicates would collide in my render — my problem to guard, not yours to encode. Your registry is keyed by `(host, type)` so it can't produce them anyway; I verified no duplicates in 46 real findings from the running engine.

## `contracts/validate.mjs` over the CLI one-liner — agreed, and the reasoning being in the README is the valuable part

Your Git Bash finding is the sharper half of that. I'd only found the `-c ajv-formats` resolution problem; `#/definitions/...` being rewritten into a filesystem path by the shell is worse, because it fails in a way that looks like a schema error rather than a shell error. Having *why not to go back* written down beats the fix itself — the next person to see a 12-line script where a one-liner would do will otherwise "simplify" it.

One operational note now that CI depends on it: `contracts/` is its own npm workspace with its own lockfile, so the CI job needs `npm ci` in `contracts/` specifically. If it runs `npm ci` at the repo root only, `validate.mjs` fails on `Cannot find package 'ajv'` — I hit exactly that when I first ran your script from a directory without the deps installed. Not blocking, and possibly already handled in whatever `ci.yml` ends up being, but it's the failure mode your README is otherwise designed to prevent.

## On the enum-semantics point

> the question isn't how many lines reference a value, it's whether anything *means* something in terms of it

That's a better formulation than mine and I'd like it to outlive this PR — it belongs in `contracts/README.md` next to the change-process description, not just in a PR thread. Your call whether it's worth an edit; I'm not asking for one before merge.

Nothing further from me. `contracts/` on `main` unblocks closing #1, and my Q1–Q12 reconciliation (ce8773f) is already done against exactly these bytes.

Ari-Glikman:
Thanks for re-running your own reproductions independently rather than trusting my validator — *"a validator written alongside a fix can encode the same blind spot as the fix"* is right, and the 22 adversarial probes around the hand-edited pattern found things I wasn't looking for.

Two follow-ups, both done.

## The enum-semantics formulation is now in `contracts/README.md` (7ce8609)

You said it belonged next to the change process rather than in a PR thread, and you were right. It's under a new **"Estimating what a change costs"** heading, using the `Warning` removal as the worked example — including that "roughly one line" was my figure and was wrong by ~80×.

It also credits the two practices separately, per your correction on #7: markers made reconciliation *findable*, defensive rendering made it *bounded*, and only the second limits blast radius. Doc-only, no schema or sample change — `npm run validate` still 11/11.

## Your CI concern was already handled

> if it runs `npm ci` at the repo root only, `validate.mjs` fails on `Cannot find package 'ajv'`

Good catch, and it's the failure mode `validate.mjs` exists to prevent, so worth being specific: `ci.yml` in #9 sets `working-directory: contracts` on both the install and the validate step, so `npm ci` runs in the workspace with the lockfile. Verified from a clean worktree checkout:

```
$ cd contracts && (npm ci || npm install) && npm run validate
CONTRACTS_JOB_EXIT=0
```

#9 also adds a `detection-engine` job that asserts ADR 0004 directly — the engine must serve four hosts with nothing on `:3001`. Worth a look if you have a moment; it's the other shared file nobody had claimed.

## Your two accepted-by-design cases — agreed, both recorded

**Negative `currentValue`** — correct to leave unconstrained. A delta-shaped metric can legitimately go negative, and `minimum: 0` would be the reflexive wrong fix. Noted.

**Duplicate `id`s** — agreed it's a consumer concern JSON Schema can't express. Confirming from our side that the registry keys conditions by `(host, type)`, so it structurally cannot emit two findings with the same id in one response. Your 46-finding check matches. Not something you need to guard, though I understand why you'd rather not rely on my say-so.

## One thing from #8 that changes your Info tile

Chasing your `info`-unreachability note, I found `system_alert` was **structurally unable to fire at all** — it suppressed itself on the second poll, so it never met the registry's two-consecutive-verdicts bar. Fixed in 412f9e3 (#8), and the live engine now emits `system_alert|info`.

So your Info tile has an honest source, and the demo loop reaches 8/8 types rather than the 3 it actually managed. Your 46-findings observation is what surfaced it — thank you.

Nothing further from me either. Ready to merge whenever.

PR #4, opened 2026-08-06
docs(decisions): draft ADRs 0001-0004 (MVP §7.4)

Ari-Glikman:
Addresses the ADR half of #2, which flagged **0004 as load-bearing without being recorded**.

| ADR | Decision |
|---|---|
| [0001](docs/decisions/0001-detection-engine-location.md) | Detection engine runs **outside IRIS**, in `services/detection-engine/` |
| [0002](docs/decisions/0002-baseline-strategy.md) | Rolling **30-minute in-memory** window; nothing persisted |
| [0003](docs/decisions/0003-threshold-configuration.md) | **JSON config file**, hot-reloaded, conservative defaults |
| [0004](docs/decisions/0004-mock-first-contracts.md) | **Mock-first contracts** — needs all three to confirm |

All four follow the MVP doc's recommendations. Each carries a **Revisit when** section naming the trigger that would reopen it — these are the decisions §7.4 expects to be re-litigated on Day 4, so the point is to make that a lookup rather than an argument.

## 0004 is deliberately unfinished

I drafted it but left Dev A's and Dev C's confirmation boxes **unchecked**. §7.4 asks all three of us to *confirm* mock-first, and it's an agreement between people rather than a technical choice one developer can decide. Please each check your own box rather than inherit my summary.

It does record **evidence rather than assertion**: of Dev C's nine assumptions, eight were correct, so building against a transcription instead of idling cost about one line against ~1,750 written. The bet paid off close to exactly as §4.1 intended.

It also recommends promoting Dev C's `// CONTRACT-Q<n>` convention from accident to practice — greppable assumption markers are what made reconciliation a `grep` instead of an audit.

## For Dev A — one constraint in 0001

0001 records something found while standing up LABDEMO: **`iris_interop_queued` carries no `host` label**, so per-host queue depth isn't in the Prometheus text. It *is* available from `Ens.Util.Statistics:EnumerateHostStatus` (verified: `Cloud API = 48` while disabled), which means **the proxy needs to expose host status, not only parse `/metrics`.**

No extra IRIS configuration required, and no contract change — `Host.queued` stays a required number. But it's a real requirement on your component, and 0001 is where the reasoning is written down.

## Two places I chose the stricter option

**0002 — warm-up.** The MVP doc suggests seeding the baseline from the first N samples. I propose the stricter thing: report `baselineValue: null` and let comparative rules stay silent until 12 samples (~2 min) exist. A seeded baseline manufactures false findings in the first seconds — and §6 names false positives as a risk. A demo showing three spurious criticals on startup is worse than one saying "warming up".

**0003 — two conditions per rule.** `queue_buildup` needs both a baseline multiplier *and* an absolute floor. `5× baseline` alone fires on 1 → 5, which is noise. Both numbers configurable together is the single biggest false-positive lever.

Both are open to challenge — that's what review is for.

## Review

`docs/decisions/` is shared and PR-gated per `CONTRIBUTING.md` §2, so this needs review rather than a direct push. Related: #3 (the Health Scan contract), #2 (Day-1 gate).

### DIFF
--- a/docs/decisions/0001-detection-engine-location.md
+++ b/docs/decisions/0001-detection-engine-location.md
@@ -1,0 +1,82 @@
+# ADR 0001 — Where the detection engine lives
+
+- **Status:** proposed
+- **Date:** 2026-08-06
+- **Deciders:** Dev A, Dev B, Dev C
+- **Drafted by:** Dev B (implementer)
+- **Source:** §7.4 decision 1 of `production-guardian-healthscan-mvp1.docx`
+
+## Context
+
+Health Scan polls IRIS metrics, compares them to a rolling baseline, and emits findings. That
+detection work has to run somewhere. Two candidates:
+
+1. In the Node/Python service alongside the metrics proxy (Layer 3 of the §2.1 data flow)
+2. Inside IRIS as ObjectScript, reading the metrics natively
+
+The MVP doc recommends the proxy, "to keep MVP 1 free of custom IRIS code and reusable across
+environments."
+
+## Decision
+
+**The detection engine runs outside IRIS, in the `services/detection-engine/` service on port
+3002.** It consumes Dev A's proxy JSON over HTTP and never touches IRIS directly.
+
+MVP 1 ships **no custom ObjectScript for detection**.
+
+## Rationale
+
+**It uses only the built-in `/api/monitor/` API.** This is §1.2's stated reason Health Scan is
+first — lowest-risk module, no custom IRIS code required. Putting detection in ObjectScript would
+forfeit that property on day one.
+
+**Deployability is the real argument.** A customer evaluating Production Guardian can point it at
+an existing production without installing classes into their namespace. Detection in ObjectScript
+means a code deployment into a live interoperability namespace — a much harder sell, and a much
+harder rollback.
+
+**It matches the ownership split.** `iris/**` is Dev A's, `services/detection-engine/**` is Dev
+B's. An in-IRIS engine would put the engine inside Dev A's directory and make the two developers
+share a component, which is precisely what the three-way split exists to avoid.
+
+**Language freedom for the consumer.** Dev C binds to nothing language-specific — they read a base
+URL from an env var (`apps/dashboard/CLAUDE.md` §4.3). That stays true only while the engine is an
+HTTP service.
+
+## Consequences
+
+**Accepted costs:**
+
+- Every metric crosses a process boundary, so the engine sees data at proxy-poll granularity
+  (10s) rather than live. Acceptable: the bar is "updates within 10s of a change," and the
+  dashboard polls at 5s, so the proxy is the pacing item either way.
+- Baseline state lives in a process that can restart, losing warm-up. See ADR 0002.
+- Some IRIS state is not in the Prometheus text at all — `iris_interop_queued` has no `host`
+  label. Per-host queue depth comes from `Ens.Util.Statistics:EnumerateHostStatus` instead, which
+  means **the proxy must expose host status, not only parse `/metrics`.** This is a real
+  constraint on Dev A's component discovered while building LABDEMO, and the one place this
+  decision costs something concrete.
+
+**Gained:**
+
+- MVP 1 installs nothing into IRIS beyond enabling built-in metrics
+- The engine is testable against fixture JSON with no IRIS at all — which is what makes
+  mock-first (ADR 0004) work
+
+## Alternatives considered
+
+**Detection in ObjectScript inside IRIS.** Rejected for MVP 1. It has genuine advantages —
+sub-second access to host state, no serialization, direct `Ens.*` API access — and is worth
+revisiting for **Smart Resolve**, which must act on the production and will need in-IRIS presence
+anyway. But adopting it now trades away Health Scan's lowest-risk property for capability MVP 1
+does not use.
+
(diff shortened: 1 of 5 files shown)


tanifgit:
Reviewed all four. **0001–0003 look right to me** and I'm not going to second-guess decisions in your own component beyond the one place each touches mine. **0004 confirmed — Dev C box is mine to check, so:**

```
- [x] **Dev C** — dashboard
```

Confirming for real, not as a formality: the dashboard was built against a transcription of §5 from Day 1, never waited on `:3002`, and now runs against a throwaway local stub until your endpoints are up. That's mock-first as described.

You were right to leave the box unchecked rather than fill it in for me. Worth doing the same for Dev A.

## One correction to 0004's evidence — in the honest direction

0004 says the reconciliation cost "roughly one line against ~1,750." **The enum correction was one line; the fixture consequence was 62 insertions across 9 files.** Actual reconciliation commit: ce8773f.

The reason is worth recording, because it's the interesting part. Eight of my fixtures used `status: "Warning"` to mean "this host is degraded." Removing `Warning` didn't just narrow a union — it made a *concept* I'd relied on unrepresentable. Every affected fixture had to be rewritten to express degradation through findings instead of status.

So the bet still paid off, but the accurate lesson is sharper than "one line": **a wrong enum value is cheap; a wrong enum value you built semantics on top of is not.** What kept it cheap was that the semantics were only in fixtures and one tone map, not spread through components — and that's because §2.4's defensive rendering forced unknown statuses down a neutral path from the start. The mitigation that saved this wasn't the `CONTRACT-Q` markers, it was the defensive-rendering rule. The markers told me *where* to look; defensive rendering is why looking was all it took.

Suggest 0004's evidence section say that instead of the one-line figure. It's a better argument for mock-first because it's a real cost honestly accounted for, and it names the practice that actually contained it.

## On promoting `CONTRACT-Q<n>` to practice

Agreed, with one refinement from having just consumed my own markers: **a marker is only as good as its assumption being written down at the site.** The ones that reconciled fastest said what was assumed (`// CONTRACT-Q1: assumed OK | Warning | Error | Inactive`); the ones that just said `CONTRACT-Q5` sent me back to the question list. Worth putting in the convention: marker plus the assumption inline, not marker alone.

## 0002 — the one line that touches my component

> An engine restart loses all baselines, and the dashboard shows warming for ~2 minutes.

Confirming this is handled on my side and not a hidden dependency: `baselineValue: null` renders as an em dash, `scenario-baseline-warming.json` exercises it, and the warm-up path is one of the states §7.3 requires me to have designed. So a restart mid-demo degrades to "—" rather than to `0`, `NaN`, or a blank.

Two notes:

**Your stricter choice is the right one, and it's right for a display reason too.** A seeded baseline doesn't just manufacture findings — it manufactures a *ratio*, and my detail drawer's whole purpose is showing current vs. baseline side by side. `48 vs 0 → Infinity×` on a projector reads as a bug in the product. `48 vs —` reads as a system that knows what it doesn't know. `formatComparison` already refuses to print a ratio against a zero or null baseline for exactly that reason.

**The cue-sheet item is mine — taking it.** "Start the engine before rehearsal, not during it" goes in `docs/demo/cue-sheet.md` (my task 5). Good catch; that's precisely the kind of thing that gets discovered live otherwise.

## 0003 — no ask, one observation

Severity-as-threshold means the same condition can arrive as `warning` on one poll and `critical` on the next while `id` stays stable (0004/Q4). My list handles that — severity is read per-render and the highlight keys on id appearing, not on severity changing — so a mid-condition escalation re-colors in place without re-pulsing. Flagging only so it's recorded as intended behavior rather than something we discover looks odd on stage.

Also: `"Cloud API"` as a `hostOverrides` key means host *names* are load-bearing config. That's consistent with Q8 (`finding.host` is exactly `host.host`), so no issue — just noting a LABDEMO rename would now break config as well as the join.

## 0001 — the Dev A constraint

No opinion to offer, it's outside my area, but recording that I read it: per-host `queued` coming from `EnumerateHostStatus` rather than the metrics text is invisible to me as long as `Host.queued` stays a required number, which the contract keeps. My host cards render Queued for all four LABDEMO hosts unconditionally, so if that field ever went optional it *would* reach me. Worth Dev A knowing that's the one part of the constraint with a downstream consumer.

## Process

Approving. 0004 needs Dev A's box before it leaves `proposed` — I'd rather it merge as-is with two of three checked and Dev A tick theirs in a follow-up than sit open waiting, since the ADR being *recorded* is what issue #2 was about.

Not blocking on the evidence correction above — happy for you to amend it here, or I'll open it as a small follow-up PR if you'd rather keep this one moving.

tanifgit (approved):
Formal approval to register the review I left in detail above (that comment was a comment, not an approval — recording it properly here).

**0001–0003:** approved. They're Dev B's component and the reasoning is sound; my comments were confined to where each touches the dashboard.

**0004:** confirming the Dev C box as my own action, not an inherited summary — the dashboard was built against a transcription from Day 1, never waited on `:3002`, and runs against a throwaway local stub until Dev B's endpoints are live. That's mock-first as described.

Still needs Dev A's box before 0004 leaves `proposed`. My position stands that this should merge with two of three checked rather than sit open — #2 was about the decision being *recorded*, and blocking the record on a checkbox reproduces the problem.

The evidence correction I raised (reconciliation cost 62 insertions across 9 files, not one line — and defensive rendering rather than the `CONTRACT-Q` markers is what contained it) is **not** blocking. Amend here or I'll open it as a follow-up.

PR #5, opened 2026-08-07
feat(dashboard): Health Scan dashboard — host grid, findings, polling, detail drawer

tanifgit:
The Health Scan dashboard — tasks 1, 2 and 3 of `apps/dashboard/CLAUDE.md` §8, plus the contract reconciliation from #1/#3.

Opening this late: several commits accumulated on the branch while I was building against the transcribed contract. Reviewing them as one unit is reasonable since they're one coherent deliverable, but the branch should have been visible earlier and that's on me.

## What's here

| Task | State |
|---|---|
| 1. Dashboard against mocked schema — host grid, findings list, severity summary | ✅ |
| 2. Live polling + demo/live toggle (`?mode=live`) | ✅ |
| 3. Finding detail drawer | ✅ |
| 4. Screencast + static fallback | 🟡 `dist/index.html` builds and opens from `file://`; screencast not recorded |
| 5. Presenter cue sheet | ⬜ |

Every state in §7.3 is designed rather than defaulted: loading (skeletons, not spinners), empty, error, stale, and baseline-warming.

## Architecture worth knowing for review

**One interface, two implementations** (§4.1). `HealthScanApi` has a `mockClient` (fixture-backed, scripted progression) and a `liveClient` (fetch). `App.tsx` picks one; **no component imports a client or calls `fetch`**. That's what makes the demo/live swap a one-line change rather than a refactor.

**Guards are deliberately lenient** (`src/api/guards.ts`). Log-and-skip per entry, not reject-the-payload — one malformed host must not blank the grid mid-demo. Numeric metrics default to `0` rather than dropping the host, since a host that exists in the production belongs on the grid even with one metric missing.

That leniency has a consequence I had to address separately: **a fixture could drift from the contract and still render.** So `npm run validate:fixtures` checks all 8 fixtures strictly against Dev B's published schema. Negative-tested — reintroducing the retired `Warning` status exits 1. It skips cleanly while `contracts/` is unmerged, so it won't fail on a branch that predates #3.

**Polling** (`usePolling`): one timer, aborts in-flight on unmount and before each tick, pauses on `document.hidden`, backs off 5s → 10s → 20s → cap 30s and resets on first success. `useHealthScan` re-throws on failure specifically so the poller can back off — swallowing there would peg polling at 5s during an outage.

**Failure degrades, never blanks** (§4.2). `lastGood.ts` caches the last successful payload to `localStorage`; on failure the grid stays up, dimmed, with a "data as of HH:MM:SS" label and the banner explaining why. Live mode only — caching demo fixtures would let a stale scenario resurface as though it were real.

## Finding detail drawer (2745db8)

Task 3. Current vs. baseline side by side is the visual focus, since §7.2 calls that the reason the view exists. A right-hand drawer rather than a modal, so the grid stays visible behind it.

**It survives polls by looking its finding up from the live array rather than holding a copy** — an open drawer's numbers update as the condition develops. Stable ids (§4 Q4) are what make that safe.

**Selection is dropped when its finding disappears.** Without it the drawer would reopen by itself if the same condition recurred under a reused id — Dev B's registry is keyed by `(host, type)`, so that's plausible, and a drawer opening with no click is a ghost on stage. Guarded on `loading` so the empty first paint doesn't trigger it.

Esc closes and returns focus to the originating row (§7.3), located via `data-finding-id` rather than a per-row ref, since that row may have moved on any poll since the click.

### Two display bugs this turned up — both in the pairing the view is built around

- **`dead_host` was typed `count`, but carries throughput.** Dev B's captured sample has `current: 0, baseline: 0.4` — 0.4 being Cloud API's actual `messagesPerSec`. As a count the baseline rounded to `0`, so the comparison read **"no change" for a host that had stopped dead**. Now `rate`: `0/s vs 0.4/s → −0.4/s`.
- **`formatComparison`'s delta branches were unitless.** `0.68s` vs `0.36s` rendered as `+0.3` beside two millisecond values; now `+320 ms`. Sign is applied to the magnitude, because the unit formatters describe sizes and `formatDuration(-0.32)` has no sensible reading.

Both typechecked fine and were only visible by running real data through the real helpers.

## Contract reconciliation (ce8773f)

Addresses #1. All 13 `CONTRACT-Q` markers resolved against the published contract; 8 of 9 assumptions held.

**Q1 was the real correction, and it cost more than one line.** `Warning` doesn't exist in IRIS. But 8 fixtures used `status: "Warning"` to *mean* "degraded," so removing it made a concept I'd built on unrepresentable — those were rewritten to signal degradation through findings instead. 62 insertions across 9 files.

`StatusDot` needed real work beyond the union: its `known` check inferred recognition from "tone isn't neutral," which breaks once four statuses are legitimately neutral. Now tests set membership. Only `OK` reads healthy, only `Error` reads critical, `Retry` reads degraded, the rest neutral — so `dead_host`'s four triggering statuses deliberately don't all render critical. A stopped host isn't the same claim as a failing one, and the finding carries the alarm.

**Timestamps:** `toISOString()` is not contract-conforming — it always emits `.000Z` and the schema's pattern requires second precision. Both mock paths now emit `…52Z`. This is one of two schema defects raised on #3.

## Verification

- `npx tsc --noEmit` — clean. No `any`, no `@ts-ignore`. `strict` + `noUncheckedIndexedAccess`
- `npm run build` — clean, `dist/index.html` 190 kB single file
- `npm run validate:fixtures` — all 8 fixtures valid against Dev B's schema
- Drawer rendered via `react-dom/server` across open / closed / baseline-warming / unknown-type: dialog semantics, labels, verbatim message, metric, em-dash baseline and warm-up note all present; no `NaN`/`null`/`undefined` leaked
- Every fixture finding **and** every finding in Dev B's captured sample run through the real format helpers — all 8 types render sanely on both
- dev-stub output validated against the schema too, so `?mode=live` exercises real bytes

## Scope

Nothing from §1.1's forbidden list: no fix/remediation buttons, no root-cause narratives, no forecasts, no 0–100 score ring, no trend charts, no production selector, no chat. `docs/production-guardian-demo.html` untouched.

## Dependencies

`ajv` + `ajv-formats`, devDependencies only, for `validate:fixtures`. `ajv-formats` is not optional — without it `format: date-time` is silently ignored and the check quietly weakens. Nothing added to the runtime bundle; React + React DOM remain the only dependencies.

## Note for reviewers

`dev-stub/` is **throwaway** and sanctioned by §8 task 2 — it serves the two contract endpoints on `:3002` from the same fixtures so `?mode=live` exercises the real fetch path. Deleted once Dev B's engine is up (#3).

One known gap I'm tracking rather than hiding: my fixtures use ~18 msg/sec while Dev B's real LABDEMO captures are 0.2–1.2. Not a contract violation, but demo and live won't look continuous side by side. Filed separately.

Refs #1, #2, #3


tanifgit:
Pushed 55ff9b4 — **`dev-stub/` is gone**, live mode now points at Dev B's real engine (#8).

The stub was throwaway scaffolding from the start, but I verified the swap was invisible *before* deleting it, since "it renders" and "it renders the same" are different claims:

- captured live `/hosts` and `/findings` from the running engine → both validate against the published schema (#3)
- Q8 join key holds across **46 real findings** — every `finding.host` exactly matches a `host.host`
- server-side sort holds on every multi-finding payload (`detectedAt` desc, severity tiebreak, 0 violations)
- no duplicate `id`s, so keying React lists on `id` is safe against the real engine and not just my fixtures
- dev proxy already targeted `:3002`, so **no config change was needed** — which is the outcome I wanted from proxying in dev rather than hardcoding a base URL

`tsc --noEmit` clean, `npm run build` clean, all 8 fixtures still validate.

README's live-mode section now documents the engine's own scenario loop rather than the stub's flags, and I kept the degrade-and-recover walkthrough — killing the engine exercises the same last-good path, and that's the on-stage insurance.

One practical note now in the README: the engine's default 10 s poll needs ~14 polls to warm the baseline before anything confirms, so a full appear-and-clear cycle is over two minutes. `POLL_INTERVAL_MS=700` compresses it to ~30 s. I burned a two-minute observation window concluding "no findings ever fire" before I found that, so it's worth having written down.

## Two things I own, found by running against the real engine

Recording them here rather than fixing them in this PR, since both deserve their own change:

1. **`baselineValue: null` is overloaded.** The engine sends it for `dead_host`/`stalled_host` — rules that are absolute *by design* — but the contract defines `null` as "baseline warming up". So my drawer currently prints *"still warming up"* on a critical `dead_host` when the baseline is provably warm. Raised on #8 with a reproduction; my half is a display fix once Dev B and I settle which meaning the field carries.

2. **`baselineValue` jitters across polls** for an unchanged condition (`85.76 → 87.0 → 84.2`). Correct engine behaviour — it's a rolling mean — but an open drawer showing a static condition looks unstable. Mine to fix, probably by rounding at display precision.

Also: only 5 of 8 finding types can fire in the engine's mock loop (`elevated_error_rate`, `system_alert`, `stalled_host` are unreachable from its fixtures), so my own fixtures remain the only proof those three render. Flagged on #8.

@Ari-Glikman this is ready for your look whenever — nothing here depends on #3 or #8 merging first.

Ari-Glikman (approved):
Reviewed as the API producer, and **verified against the real engine rather than reading the diff.** Approving.

Since this branch predates `ci.yml` landing on `main`, GitHub reports no checks — so I ran what CI would, plus the integration test only I can run.

## Verification

```
$ npm run build                      # tsc --noEmit && vite build
✓ 63 modules transformed
dist/index.html  200.17 kB │ gzip: 61.19 kB      <- single self-contained file, §6
✓ built in 694ms

$ git rebase origin/main             # clean, no conflicts
$ npm run validate:fixtures
ok  scenario-{healthy,dead-host,error-storm,queue-buildup,slow-processing,system-alert,throughput-drop,baseline-warming}.json
```

Then the real thing — this branch's dashboard against the merged detection engine on `:3002`:

**It renders our live output correctly.** Host grid, findings list, severity counts, all from the engine. Confirmed the values are genuinely ours (`0.2`/`1.2`/`0.4` msg/sec, `80ms`/`50ms` processing), not fixtures, and that the finding messages appear verbatim — `"Average processing time 2.40s is 4.9x baseline"` is a string our `slow_processing` rule generated at runtime, with the `4.9x` drifting between polls as the baseline inflates.

## Q1 fully reconciled — checked, not assumed

```ts
export type HostStatus =
  | 'OK' | 'Error' | 'Inactive' | 'Retry' | 'Stopped' | 'Unconfigured' | 'Disabled';
```

No `Warning`, and `grep` finds it nowhere in `src/` or `fixtures/` except as a *severity* label, which is correct. The 83-insertion cost is real and visible in `StatusDot.tsx`.

## Your `dead_host` rendering choice works better than my rule table implied

I flagged agreement on #4 in principle; seeing it live is more convincing. In `baseline-warming`, Cloud API is `Disabled` and shows a **grey** dot — and correctly drops out of *"Hosts OK (of 4)"* → **3**, while the `Dead host` finding carries the critical badge. "A deliberately stopped host isn't a failing one" reads exactly right on screen, and the alarm still lands.

The `Retry: 'warning'` mapping matching our deliberate exclusion of `Retry` from `DEAD_STATUSES` — arrived at independently — is a good sign the contract is carrying the semantics rather than us coordinating by hand.

## One thing to consider, not blocking

**In live mode a host can be green and in "Hosts OK" while carrying 3 warnings.** `hostsOk` counts `status === 'OK'`, and IRIS reports `OK` for a host that's slow or backing up — so I saw Cloud API green with 486 queued and three warnings against it.

Defensible, and consistent with `dead_host`: the tile reflects *IRIS status*, findings carry *detection*. But it's the one place the summary row can read reassuring during a real problem, and it's a live-only effect — your fixtures set a non-OK status when they mean degraded, so it doesn't show in demo mode.

Options, all yours: leave it and rely on the card's severity border (which does show), rename to "Hosts reporting OK", or count a host as not-OK when it has findings. **I'd leave it** — "3 of 4 OK" while one is on fire is arguably the honest IRIS reading, and changing it risks double-counting what the findings list already says. Flagging so it's a recorded choice rather than something noticed on stage.

## Two documentation nits

1. **`README.md` contradicts itself on the drawer.** Line 7 says *"The finding detail drawer is not built yet"*; line 208 says **done**. It's built — 218 lines and working. Line 7 is stale.

2. `fixtures/README.md` and `CLAUDE.md` §2.3 now duplicate contract detail that's authoritative in `contracts/`. Not wrong today, but two copies of an enum drift. Consider pointing at the contract rather than restating it.

## `validate-fixture-claims.mjs` — one behaviour worth knowing

Run without the contract or engine checked out, it prints:

```
skipped — no contract at .../contracts/healthscan.schema.json
```

and **exits 0**. That's the "silently gets weaker" shape you identified in `-c ajv-formats` — though you clearly anticipated it, since it says so explicitly rather than staying quiet, and `validate:fixtures:strict` exists. Now that both are on `main` the skip won't trigger in CI. Worth making `--strict` the CI default so a missing checkout is an error rather than a note.

Confirmed `dev-stub/` is deleted, as you said it would be — the swap to our real endpoints is invisible.

## Nothing blocking

Rebase and merge whenever. Once this lands, all four CI jobs stop skipping and the demo path is on `main` end to end.

The remaining integration risk is entirely upstream of both of us: #10, where Dev A's parser has six mismatches against real IRIS output — two of which mean it finds zero hosts.

tanifgit:
`main` moved a lot — #3, #8 and #9 all merged — so I re-verified this branch against it rather than assuming it still held. **Nothing needs changing here.**

Merged `origin/main` into a scratch worktree and ran everything against the real merged artifacts:

```
merge                     CLEAN (GitHub agrees: mergeable=MERGEABLE, state=CLEAN)
tsc --noEmit              clean
npm run build             clean
validate:fixtures:strict  All fixtures match the published contract.
```

Two things about that last line are new:

- The **schema is byte-identical** to what I approved on #3 — I diffed `healthscan.schema.json`, `healthscan.d.ts`, `samples/` and `validate.mjs` against `2e13515`. Only `contracts/README.md` gained prose. So my reconciliation (ce8773f) needs no revisiting.
- The **engine-emittable half is now genuinely running**, reading Dev B's `thresholds.json` off `main`. Until #8 merged it was skipping, so this is the first time that check has actually executed against the real config rather than a temporary checkout.

## Both of my #8 findings are fixed, and I confirmed it against the running engine

Not from the commit messages — I ran the merged engine and sampled it:

```
types seen (8/8): dead_host, elevated_error_rate, growing_queue_wait, queue_buildup,
                  slow_processing, stalled_host, system_alert, throughput_drop
severities      : critical, info, warning        <- info was previously unreachable
```

`system_alert` turned out to be **structurally dead code** — it marked an alert seen on its first poll, but the registry needs 2 consecutive verdicts to confirm, so it could never fire. Every rule unit test passed because the conflict was *between* the rule and the registry. That's a better bug than the one I reported; I only observed the symptom (no `system_alert` in 46 findings).

The `info` unreachability was taken as option (1) — documented as deliberate in `thresholds.json` with a "do NOT fix this by lowering a warning band" warning, since that would require lowering the firing gate and reintroduce the false positives MVP §6 names as the top risk. Right call, and better than changing the numbers.

## The drawer fix (786a901) is correct against the real engine

The `baselineValue: null` overload fix was written against my fixtures, so I checked it against live output — 128 findings sampled:

```
comparative with null baseline (would wrongly show "warming") :  0
absolute with null baseline    (shows "no comparison applies") : 40
```

So the split lands exactly where intended: 40 real absolute findings get the corrected copy, and no comparative finding is mislabelled. Dev B hasn't had to change the contract for it.

## One item for whoever merges this

The `dashboard` CI job currently runs `npm run validate:fixtures --if-present`, which is the lenient script — it exits 0 while skipping. That was unfixable when I raised it on #9 because its dependencies weren't on `main`; **they are now**, so `validate:fixtures:strict` is safe. Raised on #11 with the one-word change and verified it passes. Not blocking this PR — the job skips on the directory guard until this merges anyway.

## Not in this branch, tracked elsewhere

The **host-name conflict** (#13): `contracts/samples/` says `EMR Source`/`Lab Router`/`FHIR Transform`/`Cloud API`, `iris/labdemo/Production.cls` defines `EMRSource`/`LabRouter`/`PIDExtractProcess`/`PatientDemographicsOperation` — **zero exact matches**, and two are different components rather than renames. `host` is the join key (§4 Q8), so this bites at integration.

My side degrades rather than breaks — verified nothing in `src/` hardcodes a host name — but my 8 fixtures use the contract's names and will need rewriting once there's a canonical list. That's ~20 minutes and it's mine; I'm waiting on the decision rather than guessing, since guessing is what created the mismatch.

PR #7, opened 2026-08-07
docs(decisions): confirm Dev C on ADR 0004, correct its evidence section

tanifgit:
Two corrections to ADR 0004, both about the record matching reality. Follows up #4, which is merged.

## 1. Dev C's confirmation box

I ticked it in a review comment on #4 — which is a *comment*, so it never edited the file. 0004 merged saying Dev C had not confirmed, which is precisely the "recorded rather than inferred" gap that line exists to close.

```diff
-- [ ] **Dev C** — dashboard (evidently followed in practice from Day 1; ...)
+- [x] **Dev C** — dashboard (confirmed: built against a transcription of §5 from Day 1,
+      never waited on `:3002`, and runs against a throwaway local stub until the real
+      endpoints exist)
```

**Dev A's box stays unticked**, now with a note that they are not yet onboarded. Filling it in on their behalf is the thing @Ari-Glikman correctly declined to do for the rest of us, and it would be no better coming from me.

0004 therefore stays `proposed`. That's the honest state, and it's why merging #4 was still right — a recorded pending decision beats an unrecorded one.

## 2. The evidence section overstated how cheap reconciliation was

It said *"roughly one line against ~1,750 lines written."* Measured against ce8773f, the Q1-attributable change is **83 insertions across 10 files**.

I raised this on #4 and said I'd either see it amended or open a follow-up — this is the follow-up.

The reason is the part worth recording, because it's the transferable bit:

- **7 fixtures used `status: "Warning"` in 8 places to *mean* "degraded".** Removing the value didn't just narrow a union — it made a concept the dashboard had built on **unrepresentable**. Every one had to be rewritten to signal degradation through findings instead.
- **`StatusDot` needed real work, not a one-liner.** Its "is this status known" test inferred recognition from *"the tone is not neutral"* — which breaks the moment four statuses are legitimately neutral, which is exactly what the real enum introduced.

So the sharper lesson: **a wrong enum *value* is cheap; a wrong enum value you've built semantics on top of is not.** Cost scales with attached meaning, not with characters.

### It also credits the practice that actually contained it

The original gave the credit to the `CONTRACT-Q` markers. Having just consumed my own markers, I don't think that's right: **defensive rendering (§2.4)** forced unknown statuses down a neutral path from the first commit, which is why the wrong value never spread past fixtures and one tone map. The markers told me *where* to look; defensive rendering is why looking was all it took. Both earn their place, they just do different jobs.

Plus one refinement to the convention Dev B proposed promoting: **the marker must state the assumption inline** (`// CONTRACT-Q1: assumed OK | Warning | Error | Inactive`). Markers naming only a question number sent me back to the question list; ones stating the assumption were self-contained.

## Why this is a PR and not a push

`docs/decisions/README.md` line 24: *"`docs/decisions/` is shared and PR-gated — a change needs review, not a direct push."*

Also worth flagging: it's an ADR convention here that a decided ADR is never rewritten in place, only superseded. 0004 is still `proposed`, not decided, so amending it is in-bounds — but @Ari-Glikman, say so if you'd rather this were a superseding record than an edit. It's your ADR and your convention.

## Verification

Numbers measured with `git diff --numstat`, not estimated. I got them wrong twice while drafting this — first "62 insertions across 9 files", then "eight fixtures" when it's 7 files and 8 occurrences — which is a decent argument for the ADR carrying measured figures rather than remembered ones.

Refs #2, #4

### DIFF
--- a/docs/decisions/0004-mock-first-contracts.md
+++ b/docs/decisions/0004-mock-first-contracts.md
@@ -50,15 +50,36 @@ This is being written after the fact, so it can be assessed rather than asserted
 **Dev C** started the dashboard against a transcription of §5 (`apps/dashboard/CLAUDE.md` §2.3)
 rather than idling for the published contract, tagging 13 assumption sites with `// CONTRACT-Q<n>`.
 When the real contract landed (PR #3), **eight of the nine assumptions were correct**; only the
-`HostStatus` enum needed changing — roughly one line against ~1,750 lines written. The bet paid
-off almost exactly as intended.
+`HostStatus` enum needed changing. The bet paid off close to as intended.
+
+**But the cost was 83 insertions across 10 files, not one line** — worth recording accurately,
+because the reason is the transferable lesson. Removing `Warning` from the enum did not merely
+narrow a union: **7 demo fixtures used `status: "Warning"` in 8 places** to *mean* "this host is
+degraded", so the correction made a concept the dashboard had built on **unrepresentable**. Each
+of those fixtures had to be rewritten to express degradation through findings instead, and
+`StatusDot`'s tone map needed a real change rather than a one-line one — its "is this status
+known" test had inferred recognition from "the tone is not neutral", which stops working the
+moment four statuses are legitimately neutral.
+
+So the sharper form of the lesson: **a wrong enum value is cheap; a wrong enum value you have
+built semantics on top of is not.** Reconciliation cost scales with how much meaning was attached
+to the wrong assumption, not with how many characters it occupied.
+
+What kept it contained was **defensive rendering** (`apps/dashboard/CLAUDE.md` §2.4) forcing
+unknown statuses down a neutral path from the very first commit, so the wrong value never spread
+beyond fixtures and one tone map. The `CONTRACT-Q<n>` markers told Dev C *where* to look;
+defensive rendering is why looking was all it took. Both practices earn their place, but they do
+different jobs and the second is the one that bounded the damage.
 
 **Dev B** published the findings contract while LABDEMO was still being built, and validated it
 against real captured metrics rather than waiting for Dev A's proxy.
 
 **The `CONTRACT-Q<n>` convention deserves promotion from accident to practice.** Tagging every
 assumption site with a greppable marker is what made reconciliation a `grep` instead of an audit.
-Recommend it explicitly for any future contract-dependent work.
+One refinement from having consumed them: **the marker must carry the assumption inline**
+(`// CONTRACT-Q1: assumed OK | Warning | Error | Inactive`). Markers that named only the question
+number sent the reader back to the question list; markers that stated the assumption were
+self-contained. Recommend it explicitly for any future contract-dependent work.
 
 ## Consequences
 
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. **The correction is right and the ADR is better for it** — merge it.

## Amend, not supersede — answering your question directly

You asked, and it's my convention so I'll be explicit: **amending is correct here.** The rule in `README.md` is that a *decided* ADR is never rewritten. 0004 is `proposed`, so it's still a draft of a decision rather than a record of one, and the whole point of its evidence section is to be accurate. Superseding a proposed ADR to fix a number in it would leave two records where the wrong figure is the one people find first.

Good instinct to ask rather than assume, though — that distinction is exactly where the convention could have been read either way.

## On the number

I verified against `ce8773f` rather than take it on trust:

```
20  3  apps/dashboard/src/components/StatusDot.tsx
```

7 fixture files, and `StatusDot` at 20 insertions. So "one line" was wrong, and **it was my figure** — I wrote it from the shape of the type change without measuring what depended on it. Your `git diff --numstat` version is the one that belongs in the record.

The specific thing I got wrong is worth naming: I measured the *edit* and not the *concept*. `Warning` wasn't a value your code passed through, it was a value your fixtures used to **mean** something, and there was no substitute for that meaning in the new enum. That's not visible from the type definition, which is precisely why the person consuming a contract can measure it and the person publishing it can't.

## The credit reassignment is the most useful part of this PR

> The markers told me *where* to look; defensive rendering is why looking was all it took.

Agreed, and I'd have kept crediting the wrong practice. The markers made reconciliation *findable*; §2.4 made it *bounded*. Those are different properties, and only the second one limits blast radius — a contract change with perfect markers and no defensive rendering would still have reached your components.

Which reframes something for our side too: the engine's equivalent of §2.4 is that unknown host statuses fall through `DEAD_STATUSES.includes(...)` to "not dead" rather than throwing, and unknown finding types can't occur because we generate them. So the property holds by construction upstream and by discipline downstream — worth knowing that asymmetry exists if the contract shifts again.

**Your marker refinement is right too** — assumption inline, not just a question number. I've followed it in the engine's `PROXY-Q` markers for Dev A (#8): each states what we assumed, with the list at the top of `src/types/proxy.ts`. If you'd landed that refinement a day earlier I'd have copied it from you rather than arriving at it separately.

## Dev A's box

Correct to leave it unticked, and correct that it'd be no better coming from you than from me. Your note explaining *why* it's empty is an improvement on my version — "not yet onboarded" is information; an empty checkbox alone reads like an oversight.

Agreed 0004 stays `proposed`. A recorded pending decision beats an unrecorded one, which was #2's whole argument.

## Your three flagged behaviours — all confirmed as intended

**Mid-condition severity escalation with a stable `id`.** Confirmed on our side, and I saw it live while verifying #8: `throughput_drop` held `f-1004` while going `warning` → `critical` as a host stopped. Your read is exactly right — re-color in place, no re-pulse. Worth noting no test of mine covered severity changing under a stable id until I watched it happen, so your flagging it is what made me check.

**Zero/null baseline never printing a ratio.** `formatComparison` refusing `Infinity×` is the right call and matches the engine: `queue_buildup` against a zero baseline emits *"Queue depth 48 with no baseline queue"* rather than a multiplier, because there is no honest ratio to state. Independently arrived at, same conclusion.

**Host names as load-bearing config.** Fair flag on `hostOverrides` keying by `"Cloud API"`. It's consistent with Q8, but you're right that a LABDEMO rename would now break config *and* the join. Recorded — and since I created those component names in the production definition, renaming them is a decision that now has three consumers.

## The cue-sheet item is yours — thanks for taking it

"Start the engine before rehearsal, not during it" matters more than it sounds: a restart loses baselines and puts the dashboard in `warming` for ~2 minutes, which on stage looks like the product not working rather than the product being honest.

---

Issue #6, opened 2026-08-07
Demo fixtures use invented throughput ~15-90x above real LABDEMO values


tanifgit:
Demo-mode fixtures were written before any real LABDEMO capture existed, so their numbers are plausible-looking rather than measured. Dev B's `contracts/samples/hosts-response.json` (#3) now gives real values, and they're **an order of magnitude apart**.

Not a contract violation — `messagesPerSec` is just a number and every fixture validates against the schema. This is about demo fidelity.

## The gap

Healthy state, side by side:

| Host | real msg/s | fixture | real avg proc | fixture |
|---|---|---|---|---|
| EMR Source | **0.2** | 18.6 | **0.01s** | 0.04s |
| Lab Router | **1.2** | 18.4 | **0.08s** | 0.08s |
| FHIR Transform | **1.2** | 18.1 | **0.08s** | 0.21s |
| Cloud API | **0.4** | 17.9 | **0.05s** | 0.36s |

Throughput is ~15–90× too high across all 8 fixtures (ranges 12.6–19.2 msg/s where reality is 0.2–1.2). `avgProcessingTime` is inflated too, though less severely — and `Lab Router` at `0.08s` happens to match exactly, which is what makes the others visibly wrong by comparison.

## Why it matters

`apps/dashboard/CLAUDE.md` §5 requires demo and live to **look continuous** — that's the whole reason fixtures use the four real LABDEMO components instead of invented hosts. Right now switching `?mode=demo` → `?mode=live` on stage makes every number on the grid drop by an order of magnitude. That reads as a broken dashboard, not a mode change.

It also undercuts the specific thing ADR 0004 (#4) argues for: samples captured from live IRIS beat invented ones *because* invented numbers encode the author's belief rather than reality. My fixtures are the counter-example sitting in the repo.

## One case that gets worse, not better

`scenario-throughput-drop` currently uses **1.7–1.8 msg/s as its "collapsed" state** — which is *above* the real healthy baseline of 0.2–1.2. Rescaled naively, "throughput has collapsed" would show numbers higher than healthy. That scenario needs its story rechecked, not just its numbers divided.

Same class of problem as the `Warning` status: a value chosen to *mean* something, where the meaning breaks when the scale changes. Worth being deliberate about rather than running a multiplier over all 8 files.

## Proposed approach

1. Anchor healthy-state values to the measured capture — for LABDEMO those four numbers are ground truth
[...]

PR #8, opened 2026-08-09
feat(detection-engine): baseline, 8 rules, thresholds, findings API

Ari-Glikman:
Dev B's four tasks from MVP §4.4. **Zero runtime dependencies** — Node 22.18+ native type stripping, `node:http`, `node:test`.

Depends on #3 (the contract) and #4 (the ADRs) for context, but branches off `main` so it can be reviewed independently.

## @tanifgit — your `:3002` stub can go

You asked to be told when the endpoints are live. They are:

```bash
cd services/detection-engine && npm install && npm start
curl localhost:3002/api/healthscan/hosts
curl localhost:3002/api/healthscan/findings
```

**It runs with nothing on `:3001`** — mock-first is the default, not a fallback (ADR 0004). It replays captured LABDEMO fixtures on a loop: healthy → degrading → dead → recovery, so findings visibly appear *and clear* without IRIS.

Two things I verified specifically because you'd built against assumptions:

- **The engine's live output validates against the schema in #3.** Not just "same author wrote both" — I captured real responses from the running service and ran them through the published schema.
- **Timestamps are second-precision.** You hit the fractional-seconds defect and fixed your side; ours emits `2026-08-09T11:11:37Z`, valid under both the old and relaxed patterns.

Your `X-Healthscan-State` header is there (`ok` / `warming` / `stale`), CORS is unconditional, and `[]` never 404s.

## What's in it

| Piece | ADR / contract |
|---|---|
| Rolling baseline, 30-min window per host+metric | ADR 0002 — warm-up returns `null`, never a seeded guess |
| The eight rules | MVP §1.3 — 2 absolute, 6 comparative |
| Finding registry: stable ids + sustained breach | contract Q4, MVP §6 |
| `thresholds.json`, hot-reloaded, validated | ADR 0003 |
| The two endpoints | contract §3 |

Each comparative rule needs **both** a baseline multiplier and an absolute floor. `1 → 5` is 5× baseline and not a problem; that pairing is the single biggest false-positive lever, and MVP §6 names false positives as a risk.

## Verification

```
tsc --noEmit --strict   clean (no any, no @ts-ignore)
tests 95  pass 95  fail 0
```

Every rule has a **does-not-fire** case alongside its fires case. A rule that fires correctly but never *stops* is broken, and a positive-only suite would not notice — the lesson from your two structural findings on #3.

Live, with nothing on `:3001`:

```
X-Healthscan-State: ok
[{"host":"Cloud API","type":"operation",...},{"host":"EMR Source","type":"service",...}]
```

Exactly the four LABDEMO hosts, `actor` normalized to `process` (Q10), framework hosts filtered.

## Two findings recorded rather than papered over

**Baseline self-inflation.** A rolling mean includes the breaching samples, so a sustained problem becomes the new normal and its comparative finding clears while the bad value persists. `queued` 0 → 486 fires `queue_buildup`, then stops once the mean has climbed enough. Inherent to ADR 0002, not a bug — pinned by a test in `test/baseline.test.ts` and documented in `CLAUDE.md` §5.1 so nobody "fixes" it unknowingly.

Absolute rules are unaffected, which is why `dead_host` stays the reliable demo headline. Options if it ever matters: exclude breaching samples, longer window, or persisted baseline — all named revisit triggers in ADR 0002, all out of MVP 1 scope.

**A correction to my own earlier claim.** I reported `queue_buildup` wasn't firing. That was a sampling artifact in my probe — polling at 0.62s against a 0.2s engine poll. It fires correctly. Separately, the `cloud-api-dead` fixture genuinely doesn't trip it: real captured depth 48 against a floor of 50, which is ADR 0003's two-condition design working. Added a `queue-buildup` fixture so the rule is actually exercised.

## @Dev A — five `PROXY-Q` markers waiting for you

`contracts/proxy-api.md` is unlanded, so `src/types/proxy.ts` is **our assumption** of your output, not an agreed contract. Following @tanifgit's convention that worked, every assumption is tagged:

```bash
grep -rn "PROXY-Q" services/detection-engine/src/
```

**PROXY-Q2 is the one that blocks a finding type.** `iris_interop_queued` has no `host` label — per-host depth isn't in the Prometheus text. It *is* in `Ens.Util.Statistics:EnumerateHostStatus` (verified on live LABDEMO: `Cloud API = 48` while disabled), so the proxy needs to read host status, not only parse `/metrics`. No extra IRIS config needed. Reasoning is in ADR 0001 (#4).

The others: array-vs-keyed shape and raw status (Q1), whether `avg*` arrive pre-aggregated across `messagetype` (Q3 — if not, aggregation moves to us and must be sample-count weighted, not a plain mean), elapsed-seconds vs timestamp (Q4), and whether you filter framework hosts (Q5 — we filter regardless).

Also: I have a real **1006-line `/api/monitor/metrics` capture** from LABDEMO for your `samples/metrics-dump.txt`. @tanifgit asked me to hand it over; say the word.

## Not in this PR

Swapping the mock for your real proxy is Day 4 and needs `contracts/proxy-*`. `PROXY_MODE=live PROXY_BASE_URL=http://localhost:3001 npm start` is wired and typechecked but **untested against a real proxy**, because there isn't one yet.

Once #3 merges this branch should rebase, so `src/types/healthscan.ts` and the real contract live in one tree.

### DIFF
--- a/services/detection-engine/CLAUDE.md
+++ b/services/detection-engine/CLAUDE.md
@@ -1,0 +1,202 @@
+# CLAUDE.md — Detection Engine & Findings API (**Developer B**)
+
+Scoped to `services/detection-engine/`. The root `CLAUDE.md` carries the shared rules — scope
+boundary, ownership, ports — and applies here too.
+
+---
+
+## 1. What you are building
+
+The **detection engine**: the brains of Health Scan. Three things:
+
+1. **Rolling baseline** — trailing 30-minute, per-host, per-metric window with warm-up handling.
+2. **Detection rules** — the eight finding types, configurable thresholds, sustained-breach logic.
+3. **Findings API** — `GET /api/healthscan/findings` and `/api/healthscan/hosts` on **`:3002`**.
+
+You consume Dev A's proxy JSON from `:3001`. You produce the contract in
+`contracts/healthscan-api.md` for Dev C.
+
+You are the *detection* layer. You read metrics, compare to baseline, emit findings. **You compute
+nothing else and act on nothing.**
+
+### 1.1 Hard scope boundary
+
+| Do NOT build | Belongs to |
+|---|---|
+| Restarting hosts, clearing queues, any remediation | **Smart Resolve** |
+| Root-cause narratives, evidence chains, confidence scores | **AI Detective** |
+| Forecasts, "will breach in N minutes", trend extrapolation | **Early Warning** |
+| A single 0–100 health score | **Health Score** |
+| Report or summary generation | **Health Summary** |
+| Natural-language endpoints | **Ask Guardian** |
+| Persisted baseline history, trend storage | Out of MVP 1 — see ADR 0002 |
+
+**A finding states what is true now, compared to what was normal. Nothing more.** If a request
+implies a row above, say so instead of building it.
+
+## 2. Architecture, and the decisions behind it
+
+Read the ADRs before changing structure — they record *why*, and each names what would reopen it:
+
+| ADR | Decision |
+|---|---|
+| `docs/decisions/0001` | Engine runs **outside IRIS**. No ObjectScript here. Consume proxy JSON over HTTP. |
+| `docs/decisions/0002` | Baseline is a **rolling in-memory window**. Nothing persisted. Warm-up is `baselineValue: null`, never a seeded guess. |
+| `docs/decisions/0003` | Thresholds live in **`thresholds.json`**, hot-reloaded. No thresholds hard-coded in rule logic. |
+| `docs/decisions/0004` | **Mock-first.** Never block on Dev A's proxy being up. |
+
+```
+services/detection-engine/
+├─ CLAUDE.md                  <- this file
+├─ package.json               <- zero runtime dependencies (§3)
+├─ tsconfig.json
+├─ thresholds.json            <- ADR 0003; the only place numbers live
+├─ fixtures/
+│  └─ proxy/                  <- mocked Dev A responses; scenario per file
+└─ src/
+   ├─ index.ts                <- wiring: poll loop + HTTP server
+   ├─ types/
+   │  ├─ proxy.ts             <- Dev A's shape. UNRATIFIED — see §4
+   │  └─ healthscan.ts        <- transcription of contracts/healthscan.d.ts
+   ├─ config/
+   │  └─ thresholds.ts        <- load, validate, hot-reload, last-good fallback
+   ├─ baseline/
+   │  └─ window.ts            <- ring buffer, per host+metric
+   ├─ detect/
+   │  ├─ engine.ts            <- runs rules, owns the finding registry
+   │  ├─ registry.ts          <- stable ids + sustained-breach state
+   │  └─ rules/               <- one file per finding type, eight total
+   ├─ proxy/
+   │  ├─ client.ts            <- fetch from :3001
+   │  └─ mockClient.ts        <- fixture-backed, same interface
+   └─ api/
+      └─ server.ts            <- the two endpoints
+```
+
+## 3. Stack
+
+**Node 20 + TypeScript. Zero runtime dependencies.** Node 20 provides everything needed: `fetch`,
+`node:http`, `node:test`, `node:fs.watch`. TypeScript and `@types/node` are the only devDeps.
+
+No Express (14 lines of `node:http` covers two GET routes), no Zod (hand-written guards are
+explicit and dependency-free), no test framework (`node:test` is built in), no Prometheus parser —
+**we consume Dev A's JSON, not Prometheus text.** Parsing is their component's job, per ADR 0001.
+
+Adding a dependency requires stating why and what it replaces.
+
+## 4. The two contracts, and their different status
+
+**Downstream — `contracts/healthscan-*` is ratified and ours.** Published in PR #3. `src/types/healthscan.ts`
+is a direct transcription. **We own it, which means we may not quietly drift from it** — a change is
+a PR to `contracts/` with a `CHANGELOG.md` entry, reviewed by all three. Dev C's UI is built
+against those exact bytes.
+
+**Upstream — Dev A's proxy contract does not exist yet.** `contracts/proxy-api.md` is unlanded.
+So `src/types/proxy.ts` is *our assumption* of their output, not an agreed contract.
+
+Follow Dev C's convention, which worked: **tag every assumption site with `// PROXY-Q<n>`** and keep
+the running list at the top of `src/types/proxy.ts`. Reconciling when Dev A lands their contract
+must be a `grep`, not an audit:
+
+```bash
+grep -rn "PROXY-Q" src/
+```
+
+Known assumptions so far, each carrying a marker:
+
+- **PROXY-Q1** — per-host objects keyed by config name, with `status` as IRIS reports it
+- **PROXY-Q2** — `queued` present per host. **Not in the Prometheus text** (`iris_interop_queued`
+  has no `host` label); the proxy must read `Ens.Util.Statistics:EnumerateHostStatus`. Raised in
+  ADR 0001 and PR #4
+- **PROXY-Q3** — `avg*` fields already aggregated across `messagetype`, weighted by sample count.
+  If Dev A passes raw per-messagetype series instead, aggregation moves here
+- **PROXY-Q4** — `lastActivityElapsedSeconds` (as IRIS gives it) vs a pre-computed timestamp.
+  We assume elapsed and convert, since that is what the metric holds
+
+## 5. Rules — the eight, and how they must behave
+
+| Finding type | Fires on |
+|---|---|
+| `dead_host` | status is `Error`, `Inactive`, `Stopped`, or `Disabled` |
+| `stalled_host` | no activity > threshold **while messages are queued** |
+| `queue_buildup` | depth over baseline multiplier **and** over the absolute floor |
+| `elevated_error_rate` | errored count rising faster than baseline |
+| `slow_processing` | avg processing time over baseline multiplier |
+| `growing_queue_wait` | avg queueing time over baseline multiplier |
+| `throughput_drop` | messages/sec below baseline fraction |
+| `system_alert` | new alert in the proxy's alerts payload |
+
+Non-negotiable behaviours:
+
+- **Sustained breach.** Per MVP §6, a rule needs **2+ consecutive breaching samples**
+  (`sustainedSamples`) before a finding is emitted. A single-sample spike emits nothing.
+- **Stable ids.** A finding's `id` is stable for the life of the condition, keyed by
+  `(host, type)`. Contract Q4 promises Dev C this — their highlight animation and detail drawer
+  depend on it. **Never regenerate an id for an ongoing condition.**
+- **Findings disappear when cleared.** No `resolvedAt`, no tombstones. Contract Q4.
+- **No baseline, no comparative finding.** Below `minBaselineSamples`, comparative rules stay
+  silent. `dead_host` and `system_alert` are absolute and still fire.
+- **Two absolute rules, six comparative.** Only `dead_host` and `system_alert` work without a
+  baseline.
+- **`message` is authoritative.** Dev C renders it verbatim and is told not to reconstruct it. It
+  must state the actual numbers — `"Queue depth 486 is 32x baseline"`, not `"Queue is high"`.
+- **Rules are pure functions** of `(sample, window, config)`. No I/O, no clock reads inside a
+  rule — pass time in. This is what makes them testable against fixtures.
+
+### 5.1 Baseline self-inflation — known, deliberate, pinned by a test
+
+A rolling mean includes the breaching samples, so **a sustained problem becomes the new normal
+and its comparative finding clears while the bad value persists.** With `queued` going 0 → 486,
+`queue_buildup` fires, then stops once the mean has climbed enough that the ratio falls below the
+multiplier.
+
+This is inherent to ADR 0002's rolling window, not a bug. `test/baseline.test.ts` pins the
+behaviour so nobody "fixes" it unknowingly. Note that the absolute rules are unaffected —
+`dead_host` keeps firing for as long as the host is down, which is why the demo's headline
+finding is a reliable one.
+
+If it needs addressing later, the options are excluding breaching samples from the window, or a
+longer window, or a persisted baseline (ADR 0002's named revisit trigger). All three are out of
+scope for MVP 1.
+
+## 6. Never invent data
+
+- **No fabricated hosts, metrics, or findings outside `fixtures/`.** Fixtures use the four LABDEMO
+  components — **EMR Source** (service), **Lab Router** (process), **FHIR Transform** (process),
+  **Cloud API** (operation) — and the eight real finding types.
+- Fixture values should be **captured from live LABDEMO**, not invented. Real numbers catch real
+  problems; `contracts/samples/` was built that way.
+- If the proxy is unreachable, serve last-known findings and report `X-Healthscan-State: stale`.
+  **Never invent a plausible reading to fill a gap**, and never emit a finding from absent data.
+
+## 7. API behaviour
+
+Per §3 of `contracts/healthscan-api.md`:
+
+- Zero findings → `200` + `[]`. **Never `404`.**
+- Warming up → `200` + `[]` + `X-Healthscan-State: warming`.
+- Proxy unreachable → `200` + last-known + `X-Healthscan-State: stale`.
+- Genuine fault → `500` + `{"error": "..."}`.
+- Always send `Access-Control-Allow-Origin: *` (contract Q9).
+- `findings` sorted `detectedAt` desc, severity tiebreak. `hosts` sorted by name.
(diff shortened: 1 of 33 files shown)


tanifgit (commented):
Reviewed as the consumer: I ran the engine, pointed the dashboard's live path at it, and validated real captured responses against the published schema myself rather than reading the diff. Your claims hold — `tsc --noEmit` clean, **95/95 tests pass**, both endpoints serve with nothing on `:3001`, and the loop genuinely shows findings appear *and clear*.

```
$ POLL_INTERVAL_MS=700 npm start          # compressed the 14-poll warm-up to watch a full cycle
[t+0.0s  state=ok] 4 findings   queue_buildup, slow_processing, growing_queue_wait, throughput_drop @ Cloud API
[t+2.8s  state=ok] 1 finding    throughput_drop escalated warning -> critical
[t+3.5s  state=ok] 2 findings   dead_host critical appears
[t+7.0s  state=ok] 0 findings   ALL CLEARED          <-- the half that matters
```

Verified independently of your verification: captured live `/hosts` and `/findings` both validate against `#/definitions/HostsResponse` / `FindingsResponse`; **Q8 join key holds across 46 findings** (every `finding.host` exactly a `host.host`); **server-side sort holds** across every multi-finding payload (`detectedAt` desc, severity tiebreak, 0 violations); no duplicate `id`s; `Ens.MonitorService` correctly filtered; `actor → process` normalized. `X-Healthscan-State` moves `warming → ok`. My `dist/` build is clean against it.

**One defect that changes what my UI tells an operator, and two coverage gaps. Not blocking the merge** — the defect is a two-line fix and I'd rather it land here than as a follow-up.

---

## 1. `dead_host` and `stalled_host` send `baselineValue: null`, which the contract defines as "baseline warming up"

`src/detect/rules/index.ts:40` and `:70` both emit `baselineValue: null` for rules that are *absolute by design* — they have no baseline because they don't need one, not because one hasn't accumulated yet.

But contract §2 defines that field as: *"`null` while the baseline is warming up — see Q3."* Q3 says the same. So `null` has exactly one documented meaning, and these two rules use it for a second, unrelated one.

**This is not theoretical — my drawer prints a sentence that is now false.** I render `baselineValue === null` as:

> "No baseline yet — still warming up"
> "The rolling baseline needs a few minutes of samples after the engine starts."

Against your engine, a `dead_host` finding shows that text *when the baseline is fully warm*. Proof from the running service — same payload, same poll:

```
X-Healthscan-State: ok                       <-- not warming
  dead_host        baselineValue=null         <-- my UI: "still warming up"
  throughput_drop  baselineValue=0.2139       <-- baseline demonstrably WARM
```

A critical finding is the one place the UI can least afford to explain itself wrongly, and "wait a few minutes for the baseline" is actively misleading next to `Cloud API is Disabled` — it suggests the finding is provisional when it's the most certain one on screen.

Worth noting this is the *same class* as the `Warning` enum problem, with the roles reversed: a value that carries meaning, reused for a second meaning, where the type system can't tell them apart. `number | null` admits both readings and neither of us would see it in a diff.

**Three ways out, my preference last:**

1. Add `absolute: true` (or `baselineApplicable: false`) to `Finding` — most explicit, but it's a contract change needing all three of us, and #3 hasn't even merged.
2. Leave the payload alone; I special-case the six-vs-two rules in the dashboard. **I don't want this** — it hardcodes your rule taxonomy into my rendering, so adding a ninth absolute rule silently breaks my copy with no signal.
3. **Document it in §2 as the second legitimate meaning of `null`, and I branch on `finding.type` being absolute using the contract's own §2.1 table as the source of truth.** No schema change, no new field, and the table already distinguishes the two rule classes — it just isn't currently load-bearing.

I'd take (3) and I can do my half in the same pass as my fixture rescale. It needs one sentence from you in `healthscan-api.md` §2, which is a smaller change than it sounds — and it makes an existing, already-agreed table authoritative rather than inventing anything.

## 2. Only 5 of the 8 finding types can fire in the mock loop

Across ~60 polls of the default scenario I observed exactly four types plus one escalation:

```
dead_host, growing_queue_wait, slow_processing, throughput_drop, queue_buildup
```

Never: **`elevated_error_rate`**, **`stalled_host`**, **`system_alert`**. And they can't, given the fixtures:

- all three fixtures have `errored` **absent on every host**, so `elevated_error_rate` has nothing to work from
- all three have `alerts: []`, so `system_alert` is unreachable
- `stalled_host` needs `requiresQueued` + 300s idle, but `lastActivityElapsedSeconds` never approaches that

Your unit tests cover all eight, so this isn't a correctness gap — it's a **demo-surface gap**, and it lands on me. `elevated_error_rate` and `system_alert` are two of the eight cards the dashboard has to show on stage; right now `PROXY_MODE=mock` can't produce either, so my only proof they render is my own fixtures. A fourth fixture with `errored` climbing and one `alerts` entry would close it. Cheap, and it makes your mock the demo's actual backstop rather than partial cover.

## 3. `baselineValue` moves every poll for an unchanged condition

Same finding id, consecutive polls: `85.76 → 87.0 → 84.2`. Correct behaviour — it's a rolling mean and §2 says `detectedAt` is first-confirmed while values are current. Flagging it as a display consequence I now own: an open drawer's baseline figure visibly jitters while the condition is static, which reads as instability rather than a moving average. Mine to solve (probably rounding at display precision); no change wanted from you.

Also confirms your **baseline self-inflation** note from the PR body is real and observable, not just theoretical — I watched `queue_buildup` clear at t+0.7s while depth stayed pinned at 486, exactly as `CLAUDE.md` §5.1 predicts. Documenting-rather-than-fixing was right for MVP 1, and `dead_host` staying absolute is what keeps the demo headline trustworthy.

---

## What I'm doing on my side

Deleting `dev-stub/` — your endpoints replace it and the swap was invisible, which was the whole test. Confirmed my dev proxy already targets `:3002`, so no config change was needed. I'll report the deletion on #5.

Not blocking on any of the above. Fix (1) here or tell me you'd rather do it as a contract follow-up and I'll take option (3)'s dashboard half either way.

Nice work on the negative tests — *"a rule that fires correctly but never stops is broken"* is the reason your loop showed findings clearing, and clearing is the thing I actually needed to see to trust it on stage.

tanifgit:
One more from rescaling my fixtures against your `thresholds.json` (#6, done in b1a5fc7). Smaller than the `baselineValue` overload and possibly deliberate — flagging because it changes what my UI can ever display.

## Severity `info` is unreachable for seven of the eight rules

Each comparative rule's **firing gate equals its warning band**, so a rule that fires can never be below `warning`:

```
queue_buildup        baselineMultiplier=5    vs severityBands.warning=5     -> info unreachable
elevated_error_rate  baselineMultiplier=3    vs severityBands.warning=3     -> info unreachable
slow_processing      baselineMultiplier=3    vs severityBands.warning=3     -> info unreachable
growing_queue_wait   baselineMultiplier=3    vs severityBands.warning=3     -> info unreachable
throughput_drop      baselineFraction=0.4    vs severityBands.warning=0.4   -> info unreachable
dead_host            severity: critical                                     -> info unreachable
stalled_host         severity: warning                                      -> info unreachable
system_alert         severity: info                                         -> REACHABLE
```

`severityFromRatio` returns `info` only when `ratio < bands.warning`, but the rule already returned `null` unless `ratio >= baselineMultiplier`. With both at `3`, the `info` branch is dead code. Same for `throughput_drop` via `severityFromFraction`, with the comparison inverted.

**Confirmed empirically, not just read off the config** — the 46 findings I captured from your running engine were exclusively `critical` and `warning`, never one `info`.

## Why it matters on my side

The severity summary row has a **Critical / Warning / Info / Hosts OK** tile set (my `CLAUDE.md` §7.2), so Info is permanently `0` against a live engine unless a `system_alert` happens to be present. Not a defect in your engine — a config consequence — but it means one of four headline tiles is effectively decorative on stage.

I've handled my half honestly: `scenario-system-alert` now carries an `info`-severity `system_alert` (alert severity `2`, above your `criticalSeverityAtOrBelow: 1`), which is the **only** honest source of an `info` finding. I deliberately did not invent an info-level `queue_buildup` to fill the tile, since that would be a fixture depicting something your engine cannot emit — the same mistake I found in my own `baseline-warming` fixture, which showed comparative findings with a null baseline that your rules structurally cannot produce.

## Three ways this could go, and I'm fine with any

1. **Deliberate for MVP 1** — false positives are named as the top risk in MVP §6, and refusing to emit `info` is a defensible way to keep the list quiet. Nothing to do; I'd just want it recorded in `thresholds.json`'s comment so the dead branch isn't read as a bug later.
2. **Drop the warning bands slightly below the gates** (e.g. gate `3`, warning band `3`, info… doesn't work — it'd need the *gate* lowered, which widens what fires). This is the one I'd be cautious about, since it trades tile completeness for exactly the false positives §6 warns about.
3. **Leave it and let `system_alert` be the info source**, which is what the fixtures now assume.

My preference is (1) — a comment in `thresholds.json` saying `info` is intentionally unreachable for comparative rules, so the next person doesn't "fix" the bands and reintroduce noise. That also makes ADR 0003's two-condition design self-documenting about what it costs.

No change needed for me to ship. My `tools/validate-fixture-claims.mjs` reproduces your bands and floors, so if you do retune, my fixtures will fail loudly rather than drift — that check caught two of my own findings that your engine would never have emitted (a `critical` `throughput_drop` that lands in the warning band, and a `slow_processing` under Cloud API's `absoluteFloorSeconds: 2.0` override I'd forgotten was host-scoped).

tanifgit:
Fixed my half of the `baselineValue: null` overload in 786a901 (on #5). **Took option (3) — no contract change needed, so nothing is blocked on you.**

The dashboard now distinguishes the two causes itself, using the contract's **§2.1 table** as the source of truth rather than a copy of your rule taxonomy:

```
dead_host / stalled_host / system_alert   -> "Detected outright — no baseline comparison applies"
                                             label reads "Baseline (n/a)"
everything else with a null baseline      -> "No baseline yet — still warming up"  (unchanged)
```

So a critical `dead_host` no longer tells the operator to wait a few minutes for something that already happened.

**Your one sentence in §2 is now documentation, not a blocker.** Still worth adding when #3 next moves — the field genuinely carries two meanings and the next consumer will hit the same thing — but nothing of mine waits on it.

Two details worth passing back:

**It was reachable in demo mode too**, which I'd initially got wrong — I described it as a live-mode problem in my review. The `stalled_host` in my own `dead-host` fixture carries `baselineValue: null`, so the false sentence appeared with no engine running at all. My fixtures were correct; only the rendering was wrong.

**Unknown types default to comparative**, deliberately. If you add a ninth rule that's absolute, my drawer will say "still warming up" until I add it to the set — which is the *safer* failure: "a baseline is expected and absent" is a milder wrong claim than asserting no baseline applies to a rule I know nothing about. It also keeps that branch reachable, which it needs to be for the warm-up window you actually documented. If a ninth rule ever lands, ping me and it's a one-line addition.

Verified by rendering the comparison line for all 20 findings across all 8 fixtures rather than trusting `tsc` — the bug was a *sentence*, so `tsc` was never going to catch it. Four absolute findings get the new copy, sixteen comparative ones unchanged.

Still open from my review and genuinely yours: the **`info` unreachability** (7 of 8 rules, firing gate == warning band) and the **5-of-8 finding types** the mock loop can't produce. Neither blocks me.

Ari-Glikman:
Both of your findings were right, and the first turned out to be **a real bug, not a config choice**. Fixed in 412f9e3.

## `system_alert` could never fire at all

I went looking to record `info` as deliberate and found something worse: the rule was **structurally dead code**.

It marked an alert seen on its first poll and returned `null` afterwards — but the registry needs `sustainedSamples` (2) consecutive verdicts to confirm. So it produced exactly one and could never reach the bar:

```
sustainedSamples=1 -> system_alert confirmed at: p0
sustainedSamples=2 -> system_alert confirmed at: NEVER      <- the default
```

An alert is a discrete **event**, not a sustained condition, so the suppression was wrong in principle as well as in effect. It now reports while the alert stays in the proxy payload and clears when it ages out.

**Your observation is what found it.** 46 findings with no `system_alert` among them was the only signal — and I'd have read it as "no alerts happened to occur" if you hadn't tied it to `info` being absent too.

Worth naming the class, because it defeated my whole test strategy: **every rule unit test passed.** The conflict was *between* the rule and the registry, not inside either. Isolation tests cannot see it. That's the same shape as your `-c ajv-formats` finding — a component that works alone and silently does nothing once composed.

## The demo loop reached 3 of 8 types

Also verified and worse than I'd assumed. No fixture carried an alert, none had a **rising** error counter, and none combined an idle host with a queue — so `stalled_host`, `elevated_error_rate` and `system_alert` were unreachable in the demo *regardless of the code*.

Added `stalled-host` and `error-storm{,-2,-3,-4}`. The storm is **four fixtures at one poll each** rather than one repeated, because the engine derives errors/min from the delta between polls — a step held for several polls has a flat counter and yields a rate of zero. That subtlety is exactly the one you hit from the other side in your `error-storm` fixture.

Verified against the running service, not just the tests:

```
dead_host|critical          elevated_error_rate|critical
stalled_host|warning        queue_buildup|warning
slow_processing|warning     growing_queue_wait|warning
throughput_drop|warning     throughput_drop|critical
system_alert|info

distinct types: 8/8
```

**`system_alert|info` is the severity you couldn't reach.** Your Info tile now has an honest live source, and you were right not to invent an info-level `queue_buildup` to fill it.

`test/scenario.test.ts` asserts the loop reaches all eight types and that `info` comes *only* from `system_alert`, so if I ever lower a gate and reintroduce it elsewhere, that fails loudly rather than drifting. 103/103 tests, tsc clean.

## `info` unreachability — took your option (1)

Recorded as deliberate in `thresholds.json` and `CLAUDE.md` §5.2, with the reasoning: reaching `info` needs the firing **gate** lowered, not the band, which widens what fires and reintroduces the false positives MVP §6 names as the top risk. Your instinct to be cautious about option (2) was right.

One existing test asserted the old buggy behaviour — *"fires once for an alert naming a host, then not again"*. It encoded the defect, so I rewrote it rather than deleting it, and added one asserting the alert's `id` stays stable across polls.

## Your fixture-claims checker

> If you'd rather it read your bands from a published artifact than a copy of your logic, say so and I'll follow whatever you expose.

Keep the copy. A duplicated constant that fails loudly when it drifts is better here than a shared artifact, for a reason specific to this project: the moment `thresholds.json` becomes a published contract, retuning it needs a three-reviewer PR — and ADR 0003's entire point is that thresholds are tunable *without* that ceremony, so the demo can be fixed by changing a number during rehearsal.

Your checker failing when I retune is the correct outcome. It means a fixture claiming something the engine won't do gets caught, which is exactly what it found in your own rescale.

If it becomes a nuisance I'd rather expose a tiny `GET /api/healthscan/thresholds` for tooling than promote the file itself — but not before it's actually a nuisance.

## On `validate-fixture-claims.mjs` more broadly

> Schema validation proves a fixture is *shaped* right. It can't prove it's *true*.

That's the sharpest thing anyone's written in this repo, and it applies to my side too. My 95 tests proved every rule behaves correctly in isolation and told me nothing about whether the demo could show them. `scenario.test.ts` is the same idea aimed at the engine: not "does this rule work" but "can anyone ever see it".

Both of the bugs in this commit lived in that gap.

Ari-Glikman:
Rebased onto `main` now that #3 and #9 are in. **This PR has real CI for the first time, and it's green.**

```
detection-engine — typecheck + tests   pass
contracts — samples vs schema          pass
dashboard — typecheck + build          pass
metrics-proxy — typecheck + tests      pass
```

From the step log rather than the conclusion:

```
# tests 110
# pass 110
# fail 0
hosts served with no proxy: 4
```

That last line is the ADR 0004 assertion running on GitHub's runners — the caveat I flagged on #9 ("never run on Actions, only locally") is now resolved for real.

## New in this push: contract drift guards (+7 tests)

The point of rebasing was that `contracts/` is on `main`, so our transcription in `src/types/healthscan.ts` can be **checked** against the published contract instead of trusted. We own that contract, which means we're the ones who can silently break it.

Seven tests compare our unions field-by-field against `contracts/healthscan.d.ts`, and our live API output against `contracts/healthscan.schema.json`:

- `HostStatus` must still **exclude `Warning`** — pinned rather than left to memory, given it cost you 83 insertions to unlearn
- every `required` field in the published schema is one we actually emit
- we emit **no** field the schema forbids (both objects are `additionalProperties: false`, so an extra field is a violation even though it'd look harmless in our own tests)
- our timestamps satisfy the published `pattern` — the defect you found where `format: date-time` accepted what `pattern` rejected

Following your rule about validators, I verified they **fail on drift** by injecting three defects into the published contract and restoring it:

```
reintroduce 'Warning'        -> "Warning must not return to HostStatus"
add required healthScore     -> "we never emit required Host field healthScore"
remove message from Finding   -> "Finding field message is not in the contract"
```

This is your *"schema validation proves a fixture is shaped right, it can't prove it's true"* idea pointed at our own types: `tsc` proves the transcription compiles, and says nothing about whether it still matches what we published.

## Also opened #11

Three CI fixes from reading the first real run on `main`. One of them is embarrassing and worth naming: **`npm test --if-present` reports green when there is no test script** — the same "silently gets weaker" failure you found in `-c ajv-formats`, reproduced in my own CI a day after I wrote the README warning about it. It now emits a warning annotation instead.

Also replaced the `@devA` CODEOWNERS placeholder with `@kskubach` — that entry was requesting review from nobody while looking correct, which is exactly the trap the file warns about.

PR #9, opened 2026-08-09
chore(ci): CODEOWNERS, PR template, and the four per-path CI jobs

Ari-Glikman:
Closes the `.github/` items from the §8 checklist in #2 — the last things on that list nobody had picked up. Shared and PR-gated, so review rather than a push.

Doing this now because it's the useful work available **while waiting on Dev A**, and it blocks nothing of theirs.

## The four per-path jobs

Per `CONTRIBUTING.md` §4, so one developer's push never goes red because another's area is mid-refactor.

| Job | Runs |
|---|---|
| `contracts` | `npm run validate` — samples vs schema |
| `detection-engine` | typecheck, tests, **and the ADR 0004 assertion** |
| `dashboard` | `npm run build` (which runs `tsc --noEmit`), plus `validate:fixtures` if present |
| `metrics-proxy` | placeholder — no-op until Dev A lands code |

**Every job guards on its own directory existing** and skips cleanly otherwise. That matters right now: two of the four components are still on unmerged branches and `services/metrics-proxy/` doesn't exist at all, so without the guards this workflow would be red on `main` from the moment it landed.

## Two jobs worth explaining

**`contracts` runs `validate.mjs`, not an `ajv-cli` one-liner.** @tanifgit found three ways the CLI form degrades (#3): validating against the root schema can't distinguish a hosts array from a findings array, Git Bash rewrites the `#/definitions/...` argument into a filesystem path, and `-c ajv-formats` silently ignores `format` when resolved from the wrong directory. A validator that quietly gets weaker is worse than one that breaks. The `README.md` records why, so nobody "simplifies" it back.

**`detection-engine` asserts ADR 0004 directly** — the engine must start and serve four hosts with **nothing on `:3001`**. If that step ever fails, mock-first is broken, and that's the promise the entire three-way parallel split rests on. Seemed worth a CI assertion rather than trusting it stays true.

## @Dev A — CODEOWNERS has a placeholder for you

`@devA` is not a real handle. GitHub **silently ignores** an unknown owner rather than erroring, so an unresolved name means reviews are quietly *not* requested for those paths — which is worse than an empty file, because it looks like it works. The file says so in a comment; swap it for the real handle when you're onboarded.

Paths currently pointing at the placeholder: `/iris/`, `/services/metrics-proxy/`, plus your third of `/contracts/`, `/docs/decisions/`, `/.github/`, and the catch-all.

## Verification

Run from clean worktree checkouts, not just eyeballed:

```
YAML valid. jobs: contracts, detection-engine, dashboard, metrics-proxy

$ # the ADR 0004 step, verbatim from ci.yml
hosts served with no proxy: 4
ADR 0004 CI STEP: PASS

$ # the contracts job, verbatim
CONTRACTS_JOB_EXIT=0
```

**Not verified:** none of this has run on GitHub Actions, only locally. The first PR after merge is the real test, and the `metrics-proxy` job is an untested placeholder that only becomes meaningful when Dev A lands code.

## On the PR template

Deliberately short. The verification section asks for **pasted output rather than a claim**, because "tests pass" isn't evidence and this project has already had one case of me reporting a green result from an `echo` rather than the tool I thought I'd run.

Refs #2

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,37 @@
+# Ownership, enforced mechanically rather than promised. CONTRIBUTING.md §4.
+#
+# Handles are placeholders where a developer is not yet onboarded. GitHub silently
+# ignores an unknown handle rather than erroring, so an unresolved name means reviews
+# are NOT requested for that path -- fix the handle, do not assume it works.
+
+# ---- Dev A -- IRIS and metrics ingestion ----
+# TODO: replace @devA with the real handle once Dev A is onboarded.
+/iris/                        @devA
+/services/metrics-proxy/      @devA
+
+# ---- Dev B -- detection engine and findings API ----
+/services/detection-engine/   @Ari-Glikman
+
+# ---- Dev C -- dashboard and demo ----
+/apps/dashboard/              @tanifgit
+/docs/demo/                   @tanifgit
+
+# ---- Shared, PR-gated ----
+# A contract change needs all three. That sounds heavy and is the point: a silent
+# contract change is the failure mode that breaks the Day-5 integration, and the
+# review takes 30 seconds.
+/contracts/                   @devA @Ari-Glikman @tanifgit
+
+# ADRs record decisions nobody should change unilaterally.
+/docs/decisions/              @devA @Ari-Glikman @tanifgit
+
+# Root config, tooling, CI.
+/tools/                       @devA @Ari-Glikman @tanifgit
+/.github/                     @devA @Ari-Glikman @tanifgit
+
+# Source material is read-only after Day 1 -- any change here wants three pairs of eyes.
+/docs/production-guardian-healthscan-mvp1.docx  @devA @Ari-Glikman @tanifgit
+/docs/production-guardian-demo.html             @devA @Ari-Glikman @tanifgit
+
+# Catch-all for anything not matched above.
+*                             @devA @Ari-Glikman @tanifgit
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,39 @@
+<!--
+Keep this short. The checklist exists to catch the three things that have actually
+cost this project time, not to be ceremony.
+-->
+
+## What and why
+
+<!-- One or two sentences. Link the issue or ADR if there is one. -->
+
+## Verification
+
+<!--
+Paste the actual output, not a claim. "tests pass" is not evidence; the summary line
+is. If something is unverified or failing, say which -- a green claim over a red build
+costs the team more than the bug did.
+-->
+
+```
+```
+
+## Checklist
+
+- [ ] Stayed inside my own area (`CONTRIBUTING.md` §2), or flagged the crossing explicitly
+- [ ] No `contracts/` file edited as part of an implementation task
+- [ ] Ran my area's typecheck / build / tests, and pasted the output above
+- [ ] No invented data outside a declared fixtures directory
+- [ ] No new dependency, or stated why and what it replaces
+
+## If this touches `contracts/`
+
+- [ ] `contracts/CHANGELOG.md` entry, dated, with the reason
+- [ ] Consumer told what changes for them — and what does **not**
+- [ ] `cd contracts && npm run validate` passes
+- [ ] All three developers on the review (CODEOWNERS asks automatically)
+
+## Scope
+
+- [ ] Adds no capability from a later module — no fixing, root-cause, forecasting,
+      single health score, report generation, or chat (root `CLAUDE.md` §2)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,144 @@
+name: CI
+
+# Per-path jobs so one developer's push never goes red because another's area is
+# mid-refactor (CONTRIBUTING.md §4). Each job only runs when its own paths change.
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  # The job worth more than the other three combined: it validates every file in
+  # contracts/samples/ against the schemas, so a contract and the shared fixtures
+  # disagreeing is caught here rather than at the Day-5 rehearsal.
+  contracts:
+    name: contracts — samples vs schema
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+      - name: Skip if contracts/ is absent
+        id: guard
+        run: |
+          if [ -d contracts ]; then echo "present=true" >> "$GITHUB_OUTPUT"
+          else echo "present=false" >> "$GITHUB_OUTPUT"; echo "contracts/ not present yet — skipping"; fi
+      # `npm ci` needs a lockfile; fall back to `npm install` before one is committed.
+      - name: Install
+        if: steps.guard.outputs.present == 'true'
+        working-directory: contracts
+        run: npm ci || npm install
+      # Deliberately not an ajv-cli one-liner. Three reasons, all found on PR #3:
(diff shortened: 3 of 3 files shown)


Ari-Glikman:
Updating the one caveat in the description: **CI has now run on GitHub Actions and passed.** All four jobs green on this PR:

```
contracts — samples vs schema:        success
detection-engine — typecheck + tests: success
dashboard — typecheck + build:        success
metrics-proxy — typecheck + tests:    success
```

Worth being precise about what that does and doesn't prove, since three of the four skipped their real work — the guards fired because `contracts/`, `services/detection-engine/` and `apps/dashboard/` aren't on `main` yet.

So this run proves the **workflow is syntactically valid, the guards work, and nothing is red on `main`**. It does *not* yet prove the jobs catch anything. Their first meaningful run is whichever of #3, #5 or #8 merges first — and on those PRs the guards will pass and the real steps will execute.

I verified the two substantive steps locally from clean worktree checkouts, which is the closest available substitute until then:

```
ADR 0004 step, verbatim:  hosts served with no proxy: 4  -> PASS
contracts job, verbatim:  all checks passed (2 samples, 4 accept, 5 reject)
```

tanifgit (commented):
Reviewed the dashboard job against my actual scripts and the CODEOWNERS file against GitHub's own resolver, rather than reading the YAML. The structure is right — per-path jobs with existence guards is the correct shape for a repo where two of four components are still on unmerged branches, and the ADR 0004 assertion is a genuinely good idea.

**Two findings, both about a check that silently doesn't happen — which is the exact failure mode this PR is written to avoid.** Neither blocks merging; the first is a two-line fix.

---

## 1. The `dashboard` job's fixture validation passes while validating nothing

`npm run validate:fixtures --if-present` runs my script, but my script **skips cleanly when `contracts/` is absent** — by design, since it had to not fail branches predating the contract. And `contracts/` is not on `main`. So today the step is green and does nothing:

```
$ npm run validate:fixtures --if-present     # exactly what ci.yml runs
skipped — no contract at .../contracts/healthscan.schema.json
EXIT=0
```

Worse, it stays green *after* #3 merges. My validator resolves `contracts/` by path but needs `ajv` — and the dashboard job only installs `apps/dashboard`. Locally that happens to work because `ajv` is a dashboard devDependency; on a clean CI checkout with `contracts/` present but uninstalled, the schema half still resolves and the second half still can't reach `thresholds.json`. Verified both scenarios:

```
Scenario A (main today, no contracts/):        skipped, EXIT=0
Scenario B (contracts/ present, uninstalled):  "All fixtures match the published contract", EXIT=0
                                               "the engine-emittable checks did NOT run"
```

So the step reports success in a state where the interesting check hasn't run. That's the `-c ajv-formats` problem again — a validator that quietly weakens — and it's mine as much as yours: my skip-silently-and-exit-0 behaviour is what makes it possible.

**Two things would fix it, and I'd rather do my half than have you work around me:**

- **Mine:** I'll add a `--strict` flag that turns both skips into failures, so CI can demand the checks actually ran while a local dashboard-only checkout still skips politely. I'll push that to #5 today.
- **Yours (one line):** drop `--if-present` from that step. My `package.json` has had `validate:fixtures` since the first dashboard commit, so `--if-present` isn't protecting against anything real, and it converts "the script is missing" — a thing worth knowing — into silence. Keep the guard on `apps/dashboard/package.json` existing; that one's doing real work.

Everything else in the dashboard job is correct. `npm run build` does run `tsc --noEmit` first (`"build": "tsc --noEmit && vite build"`), so the job name is accurate, and I verified `npm ci || npm install` works from my committed lockfile.

## 2. CODEOWNERS: the `@devA` placeholder invalidates 9 rules, not 2

Your comment says an unknown handle means reviews are silently not requested for that path. It's broader than that — **an unresolvable owner invalidates the entire rule line**, including the co-owners who *are* valid. GitHub's own resolver, run against your branch:

```
$ gh api repos/.../codeowners/errors?ref=devB/ci-and-codeowners
error lines: [9, 10, 23, 26, 29, 30, 33, 34, 37]
  "Unknown owner on line 23: make sure @devA exists and has write access"
    /contracts/   @devA @Ari-Glikman @tanifgit
```

Nine lines, not the two Dev A-only ones. The consequences that matter:

- **`/contracts/` (line 23)** — the "a contract change needs all three" gate. That's the rule the PR body calls the defence against the failure mode that breaks Day-5 integration, and it currently requests **nobody**.
- **`/.github/` (line 30)** and **`/tools/` (line 29)** — shared config, ungated.
- **`/docs/decisions/` (line 26)** — ADRs, ungated.
- **`*` (line 37)** — the catch-all, so anything unmatched has no owner at all.

Lines 13, 16, 17 (`detection-engine`, `apps/dashboard`, `docs/demo`) are **not** in the error list, so single-owner paths resolve correctly and your review request on this PR worked because of the catch-all being… well, it didn't — you requested me manually.

**Suggestion:** drop `@devA` from the shared lines for now, leaving `@Ari-Glikman @tanifgit`. Those rules then actually fire, and adding Dev A later is a one-line edit per line instead of the difference between "gated" and "silently ungated". Leave lines 9–10 as-is with the TODO — those paths are genuinely Dev A's and having no owner is honest, since `services/metrics-proxy/` doesn't exist yet either.

Worth noting `gh api .../codeowners/errors?ref=<branch>` is how to check this before merge; it 404s against `main` only because the file isn't there yet.

## On the PR template

No notes — it's the right length, and asking for pasted output rather than a claim is the correct lesson from the `echo`-instead-of-the-real-tool incident. I'd have written a longer and worse one.

## What I verified

- `contracts` job: `npm ci` inside `contracts/` then `npm run validate` → `all checks passed (2 samples, 4 accept, 5 reject)`. Correct, and the working-directory is right — this is the thing I flagged on #3, and it's handled.
- `detection-engine` job: `npm run typecheck` clean, `npm test` → **95/95**. The ADR 0004 step's logic is sound; I ran the engine with nothing on `:3001` and got 4 hosts.
- One small note on that step: `kill "$ENGINE_PID"` runs *after* `curl -fsS`, so if the curl fails the `kill` is skipped and the engine leaks until the runner tears down. Harmless on ephemeral runners, and the `test "$COUNT" -eq 4` still fails correctly — mentioning it only because a `trap 'kill $ENGINE_PID' EXIT` would be strictly better and is one line.
- `metrics-proxy` job: correctly a no-op today. Agreed it should exist now rather than later.

Not approving-or-blocking on the CI YAML since `.github/` is shared and I'd rather Dev A see the CODEOWNERS point before it lands, but nothing here needs to hold you up: fix `--if-present` and the `@devA` lines and I'm happy.

tanifgit:
Pushed my half in fb8ad73 (on #5): `npm run validate:fixtures:strict` turns both skips into failures.

So the one-line change on your side is now a two-word one — swap the step to:

```yaml
      - name: Validate fixtures
        if: steps.guard.outputs.present == 'true'
        working-directory: apps/dashboard
        run: npm run validate:fixtures:strict
```

Verified all three paths rather than just the one I wanted to work:

```
strict, no contracts/ or engine   -> FAIL  --strict, but no contract at ...   EXIT=1
lenient, same state               -> skipped politely                        EXIT=0
strict, both present              -> All fixtures match the published contract
                                     All findings self-consistent and engine-emittable
                                     EXIT=0
```

**One sequencing note before you make that swap:** strict fails until **both** #3 and #8 are on `main`, since it needs `contracts/healthscan.schema.json` and `services/detection-engine/thresholds.json`. Merging a strict step today would red the `dashboard` job immediately — the exact problem your existence guards are designed to prevent, just moved one level down.

Two reasonable orders, your call:
- leave the step as `validate:fixtures` (lenient) for now and switch to strict in the PR that merges #3 — my `fixtures/README.md` says which to use where, so it won't get lost; or
- switch now and merge #3 first, which is nearly ready anyway since I approved it.

I'd take the second if #3 lands today, the first otherwise. Either way, dropping `--if-present` is worth doing now regardless of which script name you settle on — it's the part that converts "the script is missing" into silence.

Issue #10, opened 2026-08-10
metrics-proxy parser will find no hosts against real IRIS output — 6 label/metric mismatches


Ari-Glikman:
Welcome @kskubach — good to have the proxy landed. I read `src/parser.js` against a **real 1249-line `/api/monitor/metrics` capture** from a live LABDEMO production, and the parser and IRIS disagree in six places. The first two alone mean it produces **zero hosts** against real output.

Not a criticism of the code — the parser itself is well-structured and the Prometheus line handling is right. The problem is that `fixtures/metrics.txt` is hand-written, so it encodes what IRIS was expected to emit rather than what it does. That's the exact failure ADR 0004 warns about ("invented numbers encode the author's belief rather than reality"), and it's cheap to fix now versus on Day 5.

I have the real capture and will attach it — happy for it to become `contracts/samples/metrics-dump.txt`.

## The two that break it outright

**1. The host label is `host`, not `name`.**

```
your fixture:  iris_interop_hosts{name="EMRSource",type="BS",status="Active"} 1
real IRIS:     iris_interop_hosts{id="LABDEMO",status="OK",host="EMR Source",production="LABDEMO.Production"} 1
```

`parser.js` line ~145 does `labels['name'] || labels['host'] || labels['id']`, so it falls through to `host` for `iris_interop_hosts` — but **not** for the other metric families, which carry only `host`. Worth checking, because the fallback masks it partially.

Note also `"EMR Source"` **has a space**. Config item names are display names, so `EMRSource` won't match anything.

**2. `iris_last_activity` does not exist.** The real metric is `iris_interop_last_activity`, and it is **elapsed seconds since last activity**, not a Unix timestamp:

```
iris_interop_last_activity{id="LABDEMO",host="Cloud API",production="LABDEMO.Production"} .018
iris_interop_last_activity{id="LABDEMO",host="Ens.Actor",production="LABDEMO.Production"} 85.22
```

`new Date(value * 1000).toISOString()` on `.018` gives **1970-01-01T00:00:00.018Z**. The conversion is `now − elapsed`, not `epoch × 1000`.

## The other four

**3. There is no `type` label on `iris_interop_hosts`.** `_hostType()` will always return `'unknown'`. The real labels are exactly `id`, `status`, `host`, `production`. Host type has to come from elsewhere — we get it from `hosttype` on the `avg_*` families, which carries `service` / `actor` / `operation`.

[...]

PR #11, opened 2026-08-10
chore(ci): real handle for Dev A, and stop reporting green on nothing

Ari-Glikman:
Three fixes from watching the **first real CI run** on `main` after #9 merged. I read the step logs rather than the job conclusions, which is how two of these surfaced.

## 1. CODEOWNERS was requesting no review at all for Dev A's paths

`@devA` was a placeholder and doesn't exist. GitHub **silently ignores** an unknown owner, so `/iris/` and `/services/metrics-proxy/` were configured to request nobody while looking correct.

Now `@kskubach` — and I verified it rather than assuming, since that's the entire failure mode:

```
$ gh api users/kskubach --jq '{login,type}'
{"login":"kskubach","type":"User"}

$ gh api .../collaborators --jq '.[].login'
tanifgit, kskubach, Ari-Glikman, InterSystems-Israel
```

## 2. A wall of npm errors inside a passing job

`npm ci || npm install` fails loudly first when there's no lockfile, and `services/metrics-proxy` has none:

```
metrics-proxy  Install  npm error code EUSAGE
metrics-proxy  Install  npm error The `npm ci` command can only install with an existing package-lock.json
```

The job passed via the fallback, so this was pure noise — and noise in a green log trains people to stop reading it. It now checks for the lockfile and says which path it took.

@kskubach — committing a `package-lock.json` for `services/metrics-proxy` would let CI use `npm ci` and get reproducible installs. Not urgent.

## 3. `npm test --if-present` reports green when there is no test script

This is the one worth flagging: it's **the same failure Dev C found in `-c ajv-formats`** (#3) — a check that silently gets weaker instead of failing — reproduced in my own CI a day after I wrote the README warning about it.

The step now emits a GitHub warning annotation when a component defines no test script, so *"nothing was verified"* is visible rather than indistinguishable from success.

Kept `--if-present` for **typecheck**, because `metrics-proxy` is plain JS and legitimately has no such script — a missing one there is correct, not suspicious.

## What the first run actually proved

Worth recording, since three of four jobs skipped their real work (their directories aren't all on `main` yet):

| Job | What really happened |
|---|---|
| `contracts` | **genuinely validated** — 11/11 checks, the job Dev C called most valuable |
| `metrics-proxy` | **genuinely ran Dev A's 123 lines of parser tests**, all passing |
| `detection-engine` | skipped — not on `main` until #8 |
| `dashboard` | skipped — not on `main` until #5 |

So CI works. Verified: YAML parses, all four jobs present.

Refs #9

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,13 +1,12 @@
 # Ownership, enforced mechanically rather than promised. CONTRIBUTING.md §4.
 #
-# Handles are placeholders where a developer is not yet onboarded. GitHub silently
-# ignores an unknown handle rather than erroring, so an unresolved name means reviews
-# are NOT requested for that path -- fix the handle, do not assume it works.
+# GitHub silently ignores an unknown handle rather than erroring, so an unresolved name
+# means reviews are NOT requested for that path. Verify a handle resolves before
+# trusting it -- a typo here looks identical to working config.
 
 # ---- Dev A -- IRIS and metrics ingestion ----
-# TODO: replace @devA with the real handle once Dev A is onboarded.
-/iris/                        @devA
-/services/metrics-proxy/      @devA
+/iris/                        @kskubach
+/services/metrics-proxy/      @kskubach
 
 # ---- Dev B -- detection engine and findings API ----
 /services/detection-engine/   @Ari-Glikman
@@ -20,18 +19,18 @@
 # A contract change needs all three. That sounds heavy and is the point: a silent
 # contract change is the failure mode that breaks the Day-5 integration, and the
 # review takes 30 seconds.
-/contracts/                   @devA @Ari-Glikman @tanifgit
+/contracts/                   @kskubach @Ari-Glikman @tanifgit
 
 # ADRs record decisions nobody should change unilaterally.
-/docs/decisions/              @devA @Ari-Glikman @tanifgit
+/docs/decisions/              @kskubach @Ari-Glikman @tanifgit
 
 # Root config, tooling, CI.
-/tools/                       @devA @Ari-Glikman @tanifgit
-/.github/                     @devA @Ari-Glikman @tanifgit
+/tools/                       @kskubach @Ari-Glikman @tanifgit
+/.github/                     @kskubach @Ari-Glikman @tanifgit
 
 # Source material is read-only after Day 1 -- any change here wants three pairs of eyes.
-/docs/production-guardian-healthscan-mvp1.docx  @devA @Ari-Glikman @tanifgit
-/docs/production-guardian-demo.html             @devA @Ari-Glikman @tanifgit
+/docs/production-guardian-healthscan-mvp1.docx  @kskubach @Ari-Glikman @tanifgit
+/docs/production-guardian-demo.html             @kskubach @Ari-Glikman @tanifgit
 
 # Catch-all for anything not matched above.
-*                             @devA @Ari-Glikman @tanifgit
+*                             @kskubach @Ari-Glikman @tanifgit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,18 +122,23 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: '22'
-      # Dev A's component does not exist yet, so this job is a no-op placeholder
-      # rather than a failure. It exists so the fourth per-path job from
-      # CONTRIBUTING.md §4 is present, and starts working the moment they land code.
       - name: Skip if the service is absent
         id: guard
         run: |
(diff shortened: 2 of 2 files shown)


Ari-Glikman:
Requested review from @tanifgit and @kskubach — I'd opened this without asking anyone, which was an oversight. `.github/` is in the all-three bucket per `CONTRIBUTING.md` §4.

Worth noting why it wasn't auto-requested: **CODEOWNERS on `main` still says `@devA`, and GitHub silently ignores an unresolved handle.** So this PR couldn't benefit from the fix it contains — the `/.github/` line requested nobody. It's a neat demonstration of the actual problem: the config looked correct and did nothing.

@kskubach — two of the three fixes touch your component, so specifically for you:

- The `metrics-proxy` CI job was printing a wall of `npm error code EUSAGE` inside a **passing** job, because there's no `package-lock.json` for `services/metrics-proxy`. Committing one would let CI use `npm ci` and get reproducible installs.
- Your 123 lines of parser tests **do** genuinely run and pass in CI — I checked the step log, not just the green tick. The `--if-present` change is about components that have *no* test script reporting green anyway, which yours isn't.

Neither is urgent, and neither blocks you. Separately, #10 is the one that does need your attention — six label mismatches between the parser and real IRIS output, two of which mean zero hosts are found. My offer there stands: I can point your proxy at the live IRIS instance and post the actual `/proxy/metrics` JSON, since that instance is local to me and you can't reach it.

kskubach (approved):
Approving — all three fixes are correct, and #1 and #3 both fix checks that were reporting green while verifying nothing.

**On the CODEOWNERS handle:** `@kskubach` is right, and thank you for verifying it resolves instead of assuming. That was exactly the failure mode — my paths were configured to request nobody while looking correct. Confirmed from my side: `gh api user` returns `kskubach`, and I have WRITE on this repo.

**On the `npm ci` fallback (#2):** taking the action. I'll commit a `package-lock.json` for `services/metrics-proxy` — it has exactly one direct dependency (express), so the lockfile is cheap and gets CI reproducible installs. Tracking it alongside my parser fixes rather than leaving it on your list.

**On `npm test --if-present` (#3):** this is the one I'd have missed, and the framing is what makes it land — the same silent-weakening class as `-c ajv-formats`, a day after writing the warning about it. Worth noting your fix is already load-bearing for me: `metrics-proxy` *does* define a test script, so the warning path won't trigger there, but it means the day someone renames it the log says so instead of going quietly green.

Keeping `--if-present` on typecheck is the right call — plain CommonJS, no `tsc`, so a missing script there is correct rather than suspicious.

tanifgit (approved):
Approving. All three fixes are right, and the CODEOWNERS one I can confirm independently — GitHub's resolver now returns **zero errors** on your branch, against the nine it reported on #9:

```
$ gh api repos/.../codeowners/errors?ref=devB/ci-followups
(no errors)
```

That's the check that matters, since the whole failure mode was a file that looks correct and silently isn't.

Reading step logs rather than job conclusions is what found #2 and #3, and that's the transferable part — a green job with a wall of `npm error` in it is exactly how people learn to stop reading logs.

## One thing this PR leaves unfixed, and it's now safe to fix

The `dashboard` job still runs:

```yaml
run: npm run validate:fixtures --if-present
```

That's the green-on-nothing case from my #9 review. When I raised it, the fix wasn't safe — strict mode would have failed because `contracts/` and `services/detection-engine/` weren't on `main`. **Both are now**, so the blocker is gone. Verified against a merge of my branch with current `main`:

```
EXIT=0
```

Note the second line — the engine-emittable half is now genuinely running, reading your `thresholds.json` off `main`. Before #8 merged it was skipping.

So the swap is one word:

```yaml
run: npm run validate:fixtures:strict
```

It'll take effect when #5 merges; until then the job skips on the directory guard either way. Worth doing in this PR since it's the same class as your fix #3 and you're already in the file — but it's your call whether it rides along or waits for #5.

Also: dropping `--if-present` matters independently of the script name. It converts "the script is missing" into silence, and that's a thing worth failing on.

## On fix #3 being the same failure as `-c ajv-formats`

> reproduced in my own CI a day after I wrote the README warning about it

Worth noting the warning still did its job — it's why you recognised the shape in your own logs instead of accepting a green run. The failure mode isn't "we didn't know", it's that `--if-present` and `|| npm install` and `-c` all *read* as defensive while removing the check. My `--strict` flag exists because my own exit-0 skip was the same trap: written to be helpful locally, silently wrong in CI.

## Small note on the annotation

Emitting a GitHub warning for a missing test script is the right call over failing — `metrics-proxy` legitimately has none yet, and failing would just get the job disabled. Once Keren lands tests (PR #13 has 22 passing), that annotation should stop appearing on its own, which is a good property: the warning retires itself rather than needing a follow-up.

Issue #12, opened 2026-08-10
PROXY-Q2: per-host queue depth is not in /api/monitor/metrics — queue_buildup cannot fire


kskubach:
## Summary

`/api/monitor/metrics` does not expose per-host queue depth. `iris_interop_queued` is emitted **once per production**, with labels `id` (namespace) and `production` only:

```
iris_interop_queued{id="LABDEMO",production="LABDEMO.Production"} 0
```

There is no `host` label on that family. So the proxy cannot populate `queued` per host from the Prometheus text at all, and **`queue_buildup` cannot fire** — it is one of the eight finding types in the MVP 1 spec.

This is PROXY-Q2 in `services/detection-engine/src/types/proxy.ts`. Answering it: **the proxy currently publishes `queued: 0` on every host and the real per-production total as `_meta.productionQueued`.** That is honest but not sufficient for the rule.

## Why it is not just a naming question

Per-host queue depth exists in IRIS, but in a different place: `Ens.Util.Statistics:EnumerateHostStatus` returns a result set including the queue size per config item. The poller does not read it — `src/poller.js` only does two HTTP GETs against `/api/monitor/`. Closing this gap means adding a third data source, which is a real change to the proxy's shape, not a parser fix.

Recorded as an accepted cost in `docs/decisions/0001-detection-engine-location.md`.

## Options

1. **Add a small REST endpoint in LABDEMO** that wraps `EnumerateHostStatus` and returns per-host queue depth as JSON; the poller GETs it on the metrics interval and merges by host name. Cost: one new `%CSP.REST` route in Dev A's area, one more poll. Keeps the join key (`host`) identical because `EnumerateHostStatus` returns the config item name.
2. **Leave `queued` per-production only** for MVP 1 and have `queue_buildup` fire at the production level rather than per host. Cost: the finding cannot name which host is backing up, which is most of its value.
3. **Drop `queue_buildup` from MVP 1.** Cost: seven of eight finding types instead of eight; the demo loses its most visually obvious failure mode.

I lean to (1) — it is a contained amount of work in `iris/**`, and it is the only option that keeps the finding useful.

## Blocking / not blocking

Not blocking Dev B: the field exists in the contract and is typed `number`; it is just always `0` today. Worth deciding before `contracts/proxy-api.md` is published, because whether `queued` is per-host or per-production is a contract statement, not an implementation detail.

[...]

PR #13, opened 2026-08-10
fix(metrics-proxy): parse the real /api/monitor/metrics label shape

### LINKED ISSUE #10: metrics-proxy parser will find no hosts against real IRIS output — 6 label/metric mismatches

kskubach:
Fixes #10.

Ari's capture was right about the important thing: the parser was written against what IRIS was expected to emit, and real IRIS disagrees. Six mismatches, all fixed here, plus one the issue didn't catch.

## The one that wasn't in the issue

`id` is the **namespace**, not a host. The old code fell back to it for a host name, so every per-production line — `iris_interop_queued{id="LABDEMO",production="LABDEMO.Production"} 0` — manufactured a host called `LABDEMO`. It had no activity metrics, so it would have produced a spurious `dead_host` finding **on every poll**, and it isn't filtered by `isFrameworkHost()` in the detection engine. There's now a regression test asserting neither `LABDEMO` nor `LABDEMO.Production` can appear in `hosts`.

## One correction to the issue

Claim 1, "produces zero hosts," is overstated — the `|| labels['host']` fallback did resolve all four hosts with correct spaced names. The output was wrong, not empty, which is worse: it looked plausible.

## The six

| | Was | Is |
|---|---|---|
| host label | `name` | `host` — display name, contains spaces |
| status | `value === 1 ? 'Active' : 'Inactive'` | the `status` label, passed through (`OK`\|`Error`\|`Inactive`\|`Retry`\|`Stopped`\|`Unconfigured`\|`Disabled`) |
| type | `type` label on `iris_interop_hosts` | `hosttype`, which rides on `avg_*` only; `actor` normalized to `process` |
| last activity | `iris_last_activity`, epoch | `iris_interop_last_activity`, **elapsed seconds** |
| `avg_*` | last-write-wins across message types | weighted by `iris_interop_sample_count` |
| `queued` | per host | per production — see below |

## avg_* weighting (Q12)

`iris_interop_sample_count` turned out to be its own metric family keyed by (host, messagetype), not a label on the `avg_*` line, so it needs a pre-pass to index before aggregation. Lab Router in the fixture handles two types: `.08` s over 15 samples and `.04` s over 5. Weighted → `.07`. A plain mean says `.06`; last-write-wins said `.04`. Test asserts `.07`.

A host with no sample-count lines falls back to an unweighted mean rather than dropping out — also tested.

## PROXY-Q2 / queued — filed as #12

`iris_interop_queued` genuinely has no host label. `queued` is `0` on every host and the real per-production total is published as `_meta.productionQueued`. Per-host depth needs `Ens.Util.Statistics:EnumerateHostStatus`, which the poller doesn't read. **`queue_buildup` cannot fire today.** #12 has three options and a recommendation; it's a contract-shape decision, so worth settling before `contracts/proxy-api.md`.

## 503 → 200 + `warming: true`

Agreed on the issue. During the ~10 s startup window both endpoints now answer `200` with an empty list and `warming: true`. A 503 was indistinguishable from the proxy being down.

## `lastActivity` and `lastActivityElapsedSeconds`

Both published, per PROXY-Q4. `lastActivity` is ISO 8601 and now computed correctly (`polledAt - elapsed`), rather than passing an elapsed-seconds figure through a `Date` constructor. A host with no activity line keeps both `null` instead of reading as "active just now."

## Fixture

`fixtures/metrics.txt` rewritten to the real label shape. It's a **representative excerpt**, and its header says so — the full 1249-line capture should land as `contracts/samples/metrics-dump.txt`, since only Ari's instance can produce it. @Ari-Glikman, that's still the offer from the issue and it'd let all three of us test against identical bytes.

## `package-lock.json`

Added, as committed to in the review on #11 — `npm ci` needs it.

## Verification

```
$ node --test src/parser.test.js
# tests 22
# pass 22
# fail 0
```

Was 13 tests, all asserting the guessed shape (`EMRSource`, `status 'Active'`, epoch `lastActivity`). They passed against a parser that was wrong, which is the case for not writing tests from the same guess as the code.

## Still open, not in this PR

`contracts/proxy-api.md` + `proxy.schema.json` answering PROXY-Q1–Q5. Separate PR, needs all three reviewers. Two things to settle there: whose field names win (`errored` vs `messagesErrored`), and the host-name conflict — `contracts/samples/` says `EMR Source`/`Lab Router`/`FHIR Transform`/`Cloud API`, but the current `iris/labdemo/Production.cls` items are unspaced and only partly overlapping. `host` is the documented join key, so that one bites at integration if we leave it.


tanifgit:
Welcome @kskubach. Not my area, so no verdict on the parser — but I'm the end consumer of whatever `host` string comes out of it, so two notes from that seat.

## The host-name conflict is worse than "spacing", and it lands on my grid

You flagged it as still-open. Checking it concretely against `main`, there are **zero exact matches**:

```
contracts/samples/hosts-response.json :  EMR Source, Lab Router, FHIR Transform, Cloud API
iris/labdemo/Production.cls           :  EMRSource, LabRouter, PIDExtractProcess,
                                         PatientDemographicsOperation, ActivityReporter
exact matches                         :  NONE
```

Two of them aren't a naming difference at all — they're **different components**:

| Contract / dashboard | `iris/labdemo` |
|---|---|
| `FHIR Transform` | `PIDExtractProcess` |
| `Cloud API` | `PatientDemographicsOperation` |

Why it matters downstream: `host` is the documented join key (contract §4 Q8 — *"`finding.host` is always exactly a `host.host` value, same string, same case"*). My grid joins findings to host cards on that string, and Dev B's `thresholds.json` has a `hostOverrides` entry keyed on `"Cloud API"` exactly. Against the current production definition that override silently never applies, and any finding would attach to a card that isn't there.

**My side degrades rather than breaks**, which I verified rather than assumed: nothing in `src/` hardcodes a host name, the grid renders whatever it's given, and an unmatched `finding.host` just means the finding shows in the list without tinting a card. So this isn't a dashboard bug — but it does mean the demo would show five unfamiliar component names and no host↔finding correlation, which is most of what the view is for.

**Not asking you to rename anything.** `iris/**` is your area and the production may well be right — `PIDExtractProcess` is arguably a more honest name than `FHIR Transform`. What I'd ask is only that the decision be explicit and land in `contracts/proxy-api.md`, because whichever way it goes, three things move together: your production items, `contracts/samples/`, and my 8 demo fixtures. I'll rewrite the fixtures the moment there's a canonical list — that's ~20 minutes and it's mine to do.

Ari also noted two competing LABDEMO definitions (his live in the namespace, yours in `iris/`). Same decision, so probably worth settling once in one place rather than twice.

## #12 / `queue_buildup` — agreed on option (1), with one thing to weigh

For what it's worth from the consumer end, option (1) is the only one that keeps the finding worth rendering. Option (2) — production-level `queue_buildup` — is the one I'd push back on: my grid and detail drawer both key on a host, so a finding with no host either can't be placed or has to be attributed to an arbitrary one. Dev B already hit that shape with `system_alert` and resolved it by attributing to the named host; a production-wide queue has no name to attribute to.

Option (3) costs one of eight finding types, and it's the **most visually obvious failure mode** in the demo — a queue climbing into the hundreds is the thing an audience reads instantly. Ari induced a real depth of 48 by disabling Cloud API, so it's a genuine behaviour, not a synthetic one.

One thing worth knowing before you build (1): Dev B's `queue_buildup` has `absoluteFloor: 50`, so a real captured depth of **48 will not trip the rule**. That's deliberate on his side, but it means "wired up and returning real numbers" and "the finding fires" are two different milestones — worth inducing a depth above 50 when you test the new source, or it'll look broken when it's correct.

**Demo mode is unaffected either way.** My fixtures produce `queue_buildup` at 486 and 212 with `baselineValue: 0`, matching the engine's no-baseline-queue path, so all eight types render in demo regardless of how #12 lands. That's the fallback holding up, which is what it's for.

## The correction you made to Ari's issue is the useful kind

> Claim 1, "produces zero hosts," is overstated — the fallback did resolve all four hosts with correct spaced names. The output was wrong, not empty, which is worse: it looked plausible.

Agreed, and "wrong but plausible" is the recurring theme of this sprint — the schema that accepted `[]` ambiguously, the fixture set that validated while misstating the product, my own drawer printing a true-sounding sentence about a warm baseline. Empty fails loudly on day one; plausible survives to the rehearsal. Worth having said out loud.

And the `id="LABDEMO"` catch — a manufactured host that isn't caught by `isFrameworkHost()` and would have produced a spurious `dead_host` **every poll** — is exactly the kind of thing that only shows up when you read real bytes. That one would have reached my grid as a fifth card with a critical finding on it.

kskubach:
## Second commit: 2e164ac — real capture, and the alerts endpoint

I got a full 313-line `/api/monitor/metrics` capture off my own IRIS 2024.1 running the post-`1801a50` production, and committed it verbatim as `services/metrics-proxy/fixtures/metrics-live-capture.txt`. That closes the offer I made above — @Ari-Glikman, no need to export yours unless your instance disagrees with mine, in which case that disagreement is the interesting artifact.

Running the parser against it found four more defects. The pattern is the same one as the phantom `LABDEMO` host: **the code produced plausible output rather than empty output**, so nothing failed loudly.

### 1. `iris_interop_messages` was unmapped

Not in `METRIC_MAP` at all. It's the only message-count family this IRIS emits, present for 12 of 13 hosts — silently dropped.

### 2. Absent is not zero

This instance emits **no `iris_interop_messages_errored` lines and no `iris_interop_last_activity` lines whatsoever**. Not zero-valued lines — the families are absent. The old code initialised every numeric field to `0`, so:

- `elevated_error_rate` saw `errored: 0` on all 13 hosts and was structurally unable to fire, while looking like a measurement
- `stalled_host` had `lastActivity: null` everywhere and couldn't evaluate
- `dead_host` would have fired on 0 hosts

Every numeric field now defaults to `null`, and `_meta.absentFamilies` names what IRIS didn't send:

```json
"_meta": {
  "production": "ProductionGuardian.LabDemo.Production",
  "productionQueued": 0,
  "absentFamilies": ["iris_interop_messages_errored", "iris_interop_last_activity"],
  "hostCount": 13,
  "applicationHostCount": 4
}
```

@DevB — this is the field that matters to you. `null` means "IRIS didn't tell me", `0` means "I measured zero". A real measured zero still reports as `0`: `messagesPerSec` is emitted with value `0` in this capture and stays `0`. There's a test for exactly that, because a null-default that swallows legitimate zeros is the same bug in the other direction.

### 3. Framework hosts weren't distinguishable — and one of them owns the only latency data

9 of the 13 hosts are `Ens.`/`EnsLib.` plumbing (`Ens.Alarm`, `Ens.Actor`, `Ens.ScheduleHandler`, `EnsLib.Background.*`, …). Only 4 are application items: `EMRSource`, `LabRouter`, `PIDExtractProcess`, `PatientDemographicsOperation`.

The sharp edge: **`Ens.MonitorService` is the only host in the entire capture with an `avg_*` series** (`.051727`). Every application host has none. So any aggregate over all hosts reports framework timing as the production's processing time.

Hosts are **flagged `isFramework: true`, not dropped**. Dropping them would hide hosts from the detection engine with no way to ask why one is missing.

`ActivityReporter` is matched by an explicit name entry rather than the prefix rule. Its class is `Ens.Activity.Operation.Local`, but the metric label carries the **item** name, which has no prefix — same shape of bug as the phantom `LABDEMO` host, where the filter looked at the wrong string. It's handled under both spellings so it keeps working across #14's rename.

### 4. `/proxy/alerts` published an empty list for any non-array payload

The poller did `if (!Array.isArray(alerts)) alerts = []`. So an instance returning `{"alerts":[…]}` — or an HTML 401 page — yielded zero alerts, and `system_alert` could never fire with nothing anywhere indicating why.

New `src/alerts.js` reports how the payload was read via `_meta.shape`: `array`, `wrapped:alerts`, `single-object`, `empty`, `unparseable`, `unrecognized-object`. `empty` is a legitimate zero and carries no error; `unparseable`/`unrecognized-object` keep the body (truncated to 500 chars) so the cause is diagnosable from the endpoint alone. It never throws, and it does **not** promote an arbitrary object into a fabricated alert — `{"namespace":"LABDEMO","count":0}` is not an alert.

`index.js` cross-checks against `iris_system_alerts_new` from the metrics side and sets `_meta.suspectShapeMismatch` when metrics say there are alerts and this list is empty. The intervals differ (10 s vs 30 s) so it's a hint, not a verdict — but that combination is precisely the symptom of a wrong shape mapping, and cheaper to surface here than to debug from a blank dashboard. `iris_system_alerts_new` is `1` in this capture, so the check is exercised against real data.

The alerts payload shape is still **not pinned by a capture** — I haven't been able to get `/api/monitor/alerts` to return a non-empty body. That's why this is a normalizer with diagnostics rather than a mapping. If anyone's instance has a pending alert, a copy of that body is worth more than any of the above.

### What I deliberately did *not* do

`type` stays `"unknown"` for the 12 hosts without an `avg_*` series. `hosttype` rides only on `avg_*`, and it's tempting to read "PIDExtractProcess" as a process — but an item name is arbitrary and says nothing about the item's class. Guessing would be fabricated data that looks like a measurement, which is the same failure as #2.

### Mock

`mock-server.js` now serves the live capture by default, so the mock reproduces framework hosts and absent families instead of the tidy excerpt. `MOCK_FIXTURE=metrics.txt npm run mock` gets the old behaviour. Alerts go through the same normalizer, so the mock can't publish a shape the real proxy would reject (ADR 0004).

### Verification

```
$ npm test
# tests 50
# pass 50
# fail 0
```

Was 22. The new ones are mostly negative assertions — no host invented from any of the ~200 `id="IRISSYS"`/`id="all"`/`id="Lock_Table"` lines, Windows `dir="c:\intersystems\..."` paths surviving label parsing intact, `errored` null rather than 0 on every host.

Smoke-tested over HTTP on both fixtures: `/proxy/metrics` returns the 4 application hosts + 9 framework-flagged + `_meta.absentFamilies`, `/proxy/alerts` returns `shape: "array"` with 1 alert, `/proxy/health` ok. CI green on all four jobs.

### Still blocking @DevB

`contracts/proxy-api.md` + `proxy.schema.json` answering PROXY-Q1–Q5, separate PR with all three reviewers. It now has three more things to pin down beyond what I listed above: `isFramework`, `null`-vs-`0` semantics with `_meta.absentFamilies`, and the `_meta.shape` contract on `/proxy/alerts`. That's next on my list.

Ari-Glikman (approved):
Approving — and I ran your fixed parser against the **real 1236-line capture**, not just read the diff. All six mismatches resolved, plus the seventh you found that I missed.

```
$ node -e "buildSnapshot(parsePrometheusText(realCapture))"
hosts found: 15
  Cloud API          type=operation  status=OK  mps=0.4  proc=0.05  last=2026-08-10T07:59:55Z
  EMR Source         type=service    status=OK  mps=0.2  proc=0.01
  Lab Router         type=process    status=OK  mps=1.2  proc=0.08
  FHIR Transform     type=process    status=OK  mps=1.2  proc=0.08
  ...11 framework hosts...
phantom LABDEMO host present? false
_meta: {"productionQueued":0,"hostCount":15,"applicationHostCount":4}
```

Spaced display names, `actor` → `process`, seconds not ms, `status` passed through, correct `lastActivity`. `npm test` → **71/71**.

## One thing that will break our engine today — a 1-line fix on either side

`queued` comes through as **`null`**, and our boundary guard requires a finite number:

```
our guard accepts Dev A host with queued:null?  ->  false
...with queued:0 instead?                       ->  true
```

Consequence: `isProxyHost()` rejects the entry, our client log-and-skips it, and **the engine reports zero hosts** — the dashboard goes empty rather than degraded. Worse than a wrong number, because it looks like the production is gone.

`Host.queued` in the published contract is `{"type": "integer", "minimum": 0}` and **required** — no `null` — so `0` is the contract-conformant value for "we cannot measure this yet". Your PR body says the proxy publishes `queued: 0` with the real total in `_meta.productionQueued`, which is exactly right; it's just that the *parser* emits `null` when the family has no host label.

Two ways to close it, and **I'm happy to take either**:

1. **You coerce to `0`** in `buildSnapshot` — matches your stated intent and the contract, one line.
2. **We accept `number | null`** and treat `null` as unknown — one line in our guard, and arguably more honest since `0` claims a measurement we don't have.

My preference is **(1) for now, (2) once #12 settles.** Reason: `0` is a real reading once #12's option (1) lands, so nullability would become permanent contract surface for a temporary gap. But if you'd rather not encode a value you can't measure, say so and I'll widen our guard instead — this is your data and I don't want to push you into asserting `0` if you consider it a lie.

## Your correction to my issue is right and better-stated than mine

> "produces zero hosts" is overstated — the output was wrong, not empty, which is worse: it looked plausible.

Owning that: I traced the `|| labels['host']` fallback and wrote "masks it partially" in the same issue, then led with a headline contradicting my own paragraph. Wrong-but-plausible is the worse failure mode and you characterised it more precisely than I did.

## The `id="LABDEMO"` phantom host is the find of this PR

I read your parser against my capture and still missed it. It's nastier than any of the six:

- **a spurious finding every poll** — a host with no activity metrics is exactly what our `dead_host`/`stalled_host` paths latch onto
- **our framework filter would not have caught it.** `isFrameworkHost()` matches `Ens.` / `EnsLib.` prefixes; `LABDEMO` matches neither, so it would have reached the dashboard as a fifth host card

Your regression test asserting neither `LABDEMO` nor `LABDEMO.Production` can appear is the right guard, and it belongs upstream where you put it. I've noted on our side that our filter is prefix-based and therefore not a safety net for this class.

## Q12 — you found the part I described wrongly

I told you to weight `avg_*` by `iris_interop_sample_count` as though the count rode on the same line. It doesn't: it's **its own family keyed by (host, messagetype)**, needing a pre-pass to index. Materially different work than my description implied.

The `.08`/15 + `.04`/5 → `.07` test is the right shape, and asserting against both the plain mean (`.06`) and last-write-wins (`.04`) is what makes it meaningful — all three look plausible and only one is right. Unweighted fallback when counts are missing, rather than dropping the host, matches how we handle absent data.

## `503` → `200` + `warming: true`

This matters more than it reads. Our engine responds differently to "proxy is down" (serve last-known, report `X-Healthscan-State: stale`) versus "proxy has nothing yet" — and a `503` made those indistinguishable. It also aligns us on one convention: `contracts/healthscan-api.md` §3 has us returning `200` + `[]` rather than an error for the same reason.

Publishing **both** `lastActivity` and `lastActivityElapsedSeconds` settles PROXY-Q4 without either of us guessing. We'll consume the elapsed form and ignore the ISO one rather than double-convert. Keeping both `null` when there's no activity line — rather than reading as "active just now" — is the same instinct as our `baselineValue: null`.

tanifgit:
Not a review request, just a flag: this is **approved by @Ari-Glikman, all four CI jobs green, and MERGEABLE** as of now, and it has been sitting since this morning. Nothing appears to be waiting on it.

Mentioning it because it is upstream of the last Day-1 gate item — `contracts/proxy-api.md` and `proxy.schema.json` are the only pieces still missing, and the label-shape answers this PR establishes (`host` not `name`, `hosttype` on `avg_*` only, elapsed-seconds rather than epoch, sample-count weighting) are most of what that contract has to state. Easier to write it against merged code than against a branch.

I have not merged it myself — `services/metrics-proxy/` is yours and you may be holding it deliberately. Say if you would rather someone else landed it.

No action needed from me on this one. Issue #10 is already closed, and nothing of mine is blocked: the dashboard consumes Dev B's API rather than the proxy directly, and demo mode consumes neither.

PR #14, opened 2026-08-10
feat(labdemo): route ADT^A01 straight to the operation, ADT-only generator

kskubach:
LABDEMO pipeline as it actually runs on the instance. All in `iris/**`.

## Pipeline is one hop shorter

```
before:  EMRSource → LabRouter → PIDExtractProcess → PatientDemographicsOperation
after:   EMRSource → LabRouter → PatientDemographicsOperation
```

`LabRouter` now applies the `HL7ToPID` DTL inline on the send and targets the operation directly. `PIDExtractProcess` was a pass-through: transform, then `SendRequestAsync`. The routing engine does both. One fewer queue and one fewer host on the way to the same record.

`PIDExtractProcess.cls` stays in the repo but is no longer a production item.

## ORU^R01 dropped

Removed from the routing rule and from `HL7Generator`. ORU carries a PID too, so it upserted the same `PatientRecord` by a second path — two message types, one visible outcome. The demo is patient demographics; ADT^A01 is the message that means "a patient arrived."

## Compile fixes

These were the reason the classes wouldn't load:

- **`</Storage>`** in `PatientRecord.cls` and `PIDExtractProcess.cls` — UDL closes a Storage block with `}`. The XML end-tag form only appears in the XML export format, and it fails to parse in `.cls`.
- **`PatientDemographics.cls` had a Storage block at all** — it extends `Ens.Request`, which isn't stored by that map. Removed.
- **`$$$LOGWARNING` / `$$$LOGINFO` with a line-wrapped argument.** These are macros; the preprocessor doesn't see the continuation, so the expansion broke. Concatenation now builds into a variable first, then the macro takes the variable.
- **`do {"error": (message)}.%ToJSON()`** doesn't parse — a JSON literal isn't a valid method-call target. Bind to `err`, then call on it.
- **`%ToJSON(%response.Output)` wrote nothing.** No-argument `%ToJSON()` writes to the current device, which during a `%CSP.REST` dispatch is the response.
- **`RunContinuous` wrote `filename` as the file body** instead of `msg`, so every message generated in continuous mode contained its own path. `Run()` was correct; only the continuous loop had it.

## DTL uses the brace property syntax

`source.{PID:3(1).1}` instead of `source.GetValueAt("PID:3(1).1")`, throughout. The brace form is resolved at compile time against the 2.5 schema, so a wrong path is a compile error. `GetValueAt` returns an empty string for a bad path and the record silently lands with a blank field — the failure mode is a patient with no name and no error anywhere.

## Paths and the web-app prefix

Drop directory is `C:\Practice\IN\HL7`, archive `C:\Practice\ARCHIVE`, set in both `Production.cls` and `HL7Generator` (they have to agree or the service never sees the file). They were pointing at `/tmp/labdemo/...`, which doesn't exist on a Windows instance.

`EnableMetrics.Verify()` now prefixes both `/api/monitor/` GETs with `^ProductionGuardian.Setup("WebAppPrefix")`. On an instance fronted by a web server the API lives under an instance prefix, and `Verify()` was reporting a bare 404 against a working endpoint — which reads as "metrics aren't enabled" when they are. Default is `""`, so an instance on the private web server is unaffected.

## Operation returns Ens.StringResponse

Carrying the log line, so the outcome shows in the message trace instead of only the event log. Makes a failed upsert visible where you'd look for it.

## Note for the proxy contract

This changes the production item names, which are the **join key** between hosts and findings (`healthscan-api.md` Q8). Items are now `EMRSource`, `LabRouter`, `PatientDemographicsOperation`, `ActivityReporter` — `PIDExtractProcess` is gone.

`contracts/samples/hosts-response.json` and `findings-response.json` say `EMR Source`, `Lab Router`, `FHIR Transform`, `Cloud API`. Neither the spacing nor the last two match. `FHIRTransform` and `CloudAPI` don't exist in this production at all.

Not fixed here — `contracts/` needs its own PR with all three reviewers, and Dev C built the UI against those bytes. Flagging it because it surfaces at Day-5 integration as findings that join to no host, and it's the reason to settle names when `contracts/proxy-api.md` gets published. @Ari-Glikman @tanifgit

## Verification

Not verified against a live instance — I don't have one attached. Each compile fix above is a specific ObjectScript parse rule, not a guess, but the honest status is "reviewed, not compiled." Worth a `LoadDir` before merge.

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -12,7 +12,7 @@ Developer A owns:
 | Path | Purpose |
 |---|---|
 | `iris/setup/` | One-time ObjectScript setup scripts for the demo namespace |
-| `iris/labdemo/` | LABDEMO production class, HL7 generator, trigger toggles |
+| `iris/labdemo/` | LABDEMO production, HL7 generator, patient demographics pipeline |
 | `services/metrics-proxy/` | Node.js proxy: polls `/api/monitor/metrics` + `/api/monitor/alerts`, exposes per-host JSON |
 
 Do **not** touch `services/detection-engine/`, `apps/dashboard/`, `contracts/` (read only), or any root config shared with other devs without a PR.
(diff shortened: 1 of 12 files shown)


tanifgit:
@kskubach — reviewing this as the dashboard consumer, not the `iris/**` reviewer. **The rename in here resolves most of the host-name conflict I raised on #13**, and I want to confirm that rather than leave you guessing what it did downstream.

## Three of four now match the contract exactly

```
contract / dashboard :  EMR Source, Lab Router, FHIR Transform, Cloud API
this PR              :  EMR Source, Lab Router, Cloud API, Ens.ActivityReporter
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ exact match on all three
```

Before this PR there were **zero** exact matches (`EMRSource`, `LabRouter`, `PIDExtractProcess`, `PatientDemographicsOperation`). `host` is the documented join key — contract §4 Q8, *same string, same case* — so that mattered: findings wouldn't have tinted their host cards, and Dev B's `hostOverrides` keyed on `"Cloud API"` would silently never have applied. Now it does.

`Ens.ActivityReporter` is correctly invisible to me — the engine filters on the `Ens.`/`EnsLib.` prefixes (`isFrameworkHost`, `src/types/proxy.ts:91`), so renaming it into the `Ens.` namespace means it drops out before it reaches the dashboard. Deliberate or not, it's the right side of that filter.

## One open question, and it's genuinely yours

**`FHIR Transform` no longer exists**, because `PIDExtractProcess` was a pass-through and you removed it. I think that's the correct engineering call — a host that transforms and forwards is a queue for no reason, and the routing engine does both. But it leaves the demo with **three hosts where every artifact says four**:

- `contracts/samples/hosts-response.json` — four hosts, Dev B's area
- my 8 demo fixtures — four hosts each
- my host grid renders 4 loading skeletons, so the layout shifts when 3 arrive
- `apps/dashboard/CLAUDE.md` §5 names the four LABDEMO components

Two ways to go, and **it's your call because it's a production-design question, not a display one**:

1. **Accept three hosts.** I update fixtures + skeletons, Dev B updates `contracts/samples/`. Honest, and simpler is usually better on stage. Cost: ~20 min mine, a contract-samples PR for Dev B.
2. **Keep a fourth host that does real work.** If any step deserves its own host — an actual FHIR transform rather than a PID pass-through — a four-host pipeline reads better in a demo about pipeline health, and nothing downstream changes. Cost: yours.

I lean (1) unless a fourth host earns its place. Don't reinstate a pass-through purely to match my fixtures — that's the fixtures following reality, which is the direction it should go.

**Either way I'm not blocked**, and nothing here needs to hold up this PR. Demo mode is fixture-driven and renders four hosts regardless; only live mode reflects the real production. Tell me which way you land and I'll do my half.

## Two things I verified so you don't have to take them on faith

- **The dashboard degrades rather than breaks** on a host-count or name change. Nothing in `apps/dashboard/src/` hardcodes a host name — the grid renders whatever it's given, and an unmatched `finding.host` still shows in the findings list, just without tinting a card. So a mismatch is a cosmetic loss, not a crash.
- **`Cloud API` keeps the `hostOverrides` entry working.** Dev B's `thresholds.json` raises `slow_processing.absoluteFloorSeconds` to `2.0` for `"Cloud API"` specifically. That's keyed on the exact string, so the rename switched it from dead config to live config.

## Unrelated, but worth flagging while you're in here

The compile-fix list is the useful part of this PR for the rest of us — particularly `GetValueAt` returning empty for a bad path versus the brace syntax failing at compile time. *"A patient with no name and no error anywhere"* is the same class as the two bugs that have cost this project the most: the schema that accepted `[]` ambiguously, and my drawer printing a true-sounding sentence about a warm baseline. Wrong-but-plausible survives to the rehearsal; wrong-and-loud doesn't. Worth a line in `iris/CLAUDE.md` so the next person reaches for the brace form by default.

kskubach:
## Closing this — superseded on `main`, but three fixes in here did NOT land

The work this PR proposed has been done directly on `main` (`83ffe1f`…`55970e4`), on a live instance, which is strictly better than this branch: this PR's own verification section said *"reviewed, not compiled."* `main` is now the compiled truth.

Merging it at this point would actively regress things — it **re-adds two files `main` deleted** (`Process/PIDExtractProcess.cls`, `FHIRRoutingRule.cls`, +111 lines) and reverts the item renames to a different scheme than the one that shipped.

### What landed, and better than proposed

| This PR proposed | `main` did |
|---|---|
| Drop `PIDExtractProcess` from the production, keep the class | **Deleted the class** (`2ca6764`) |
| Rename items to `EMRSource` / `LabRouter` / `PatientDemographicsOperation` | Renamed to **`EMR Source` / `Lab Router` / `Cloud API`** — the *spaced* contract names |
| Rename the activity reporter to `Ens.ActivityReporter` | It is `Ens.Activity.Operation.Local` |
| ORU^R01 removed | Done (`ce77c4b`) |

The naming outcome is the important difference. I proposed unspaced names and argued `contracts/` should change to match; `main` went the other way and made the production match the contract. **That is the right call** — `host` is the join key, `contracts/` is the frozen artifact two other developers built against, and the production is the cheaper thing to change. My proposal would have forced a contract PR that is now unnecessary.

Verified against the live instance today: 12 hosts, 3 application — `Cloud API`, `EMR Source`, `Lab Router` — and the proxy's mock fixture now carries those same bytes (#13, `f949472`).

### Three fixes in here are NOT on `main`

Not blockers, but they don't exist anywhere else, so I'm recording them rather than letting them die with the branch:

**1. `</Storage>` in `iris/labdemo/Data/PatientRecord.cls:158`** — still there on `main`. UDL closes a Storage block with `}`; the XML end-tag form belongs to the XML export format. If the class currently compiles on your instance then my read of this was wrong and it should just be dropped — but it's worth one `LoadDir` to confirm, because the symptom is a class that silently won't load.

**2. `Message/PatientDemographics.cls` has a Storage block on `main` (lines 46–98)** — it extends `Ens.Request`, which isn't stored by that map.

**3. `EnableMetrics.Verify()` hardcodes `/api/monitor/metrics` with no web-app prefix.** This one is now *confirmed real*, not theoretical. Working through #13 against your instance, the monitor API is at:

```
http://localhost/iris4health_2024_1/api/monitor/metrics
```

`Verify()` requests `/api/monitor/metrics` on port 80 and gets the web server's 404, so **it reports "metrics aren't enabled" against an instance where they are working fine.** I hit exactly this in the proxy and had to add an `IRIS_BASE_PATH` setting to fix it (`694e902`). The same one-line prefix from this branch fixes `Verify()`.

**4. Stale docs.** `main`'s `Production.cls:4`, `RoutingRule.cls:2-4` and 6 places in `iris/labdemo/README.md` still describe the `PIDExtractProcess` hop and ORU^R01 routing. `README.md:138-140` tells you to suspend `PIDExtractProcess` to demo queue buildup and add a `hang 5` to it for slow processing — both now impossible, and those are two of the eight finding triggers.

I'll pick these up as a small follow-up PR against current `main` rather than reviving this branch, so it's a clean diff against what actually ships.

### @tanifgit

Answering your review comment here since this is closing: **yes, the host-name conflict is resolved, and in the direction you'd want.** The production now emits `EMR Source`, `Lab Router`, `Cloud API` — exactly the spelling in `contracts/samples/hosts-response.json`, so nothing changes on the dashboard side and no contract PR is needed for naming.

One residual gap: the production has **three** application hosts; the sample lists **four**, including `FHIR Transform`, which no longer exists. #15 removes it and is still awaiting your review — merging that closes the last difference.

kskubach:
Closed as superseded by the direct work on `main` (`83ffe1f`…`55970e4`). See the comment above for what landed, what `main` did better, and the three fixes from this branch that still need picking up separately.

Not deleting `devA/labdemo-pipeline-simplify` yet — the three unlanded fixes are only on it. I'll delete the branch once the follow-up PR is open.

kskubach:
Recorded the four unlanded items as #17 so they survive this branch being deleted. Nothing else from here needs carrying forward.

PR #15, opened 2026-08-10
contracts: drop FHIR Transform from the samples

kskubach:
## What

`FHIR Transform` is removed from `contracts/samples/`. Per §4 of the root `CLAUDE.md` this is its own PR, with a `contracts/CHANGELOG.md` entry, and needs all three of us.

**No schema change, no field change.** `host` is `{"type": "string", "minLength": 1}` with no `enum`, so nothing in `healthscan.schema.json` or `healthscan.d.ts` moves. If you read `host` as an opaque string — and both of you do — there is nothing to change on your side.

## Why

The LABDEMO production no longer has a `FHIR Transform` host, so IRIS cannot emit that `host` label. §2 of `healthscan-api.md` says only application config items appear — a host the production cannot produce does not belong in the bytes you two mock against.

Verified against [Production.cls](iris/labdemo/Production.cls): the items are `EMR Source`, `Lab Router`, `Cloud API`, plus `Ens.ActivityReporter`, which the framework filter drops.

### It was real — this is a removal, not a correction of a fabrication

Worth stating plainly, because @Ari-Glikman's live capture in #10 **does** show `host="FHIR Transform"`, and my first version of this PR body wrongly claimed the production never had it. Git history disagrees with me: it was an `<Item>` in `Production.cls` from the first commit until `1801a50`.

It was an `EnsLib.MsgRouter.RoutingEngine` whose rule forwarded everything straight to `CloudAPI` — no DTL, and nothing FHIR, as its own class comment said. It existed to generate metric activity. Keren removed it in `1801a50` when the pipeline became HL7→PID, and #14 settled on the three hosts above.

**So the samples aren't wrong, they're older than the pipeline change.** Restoring the host instead would mean re-adding a pass-through item to the production for no reason beyond making a fixture true. Nothing in the pipeline produces FHIR; that name was aspirational from the start.

### One consequence for @Ari-Glikman

Your instance still runs the older production definition, so a fresh capture there will show four hosts until you reload from `iris/labdemo/`. Expected, and it breaks nothing — the parser reads whatever `host` labels arrive, and an extra host is over-coverage rather than a failure.

## The one thing worth reviewing carefully

I did **not** delete the `FHIR Transform` finding. The samples carry **exactly one finding per type**, so deleting `f-1038` would have zeroed `slow_processing` coverage for anyone mocking against these bytes.

It is reassigned to `Lab Router` instead, and that costs nothing: the finding cites `baselineValue: 0.08`, which is *already* Lab Router's measured `avgProcessingTime` in `hosts-response.json`. Its `host` now names a host that exists and its baseline still agrees with that host's row — it agrees better than it did before.

## Changes

| File | Change |
|---|---|
| `samples/hosts-response.json` | `FHIR Transform` entry removed; three hosts remain, still alphabetical per §1 |
| `samples/findings-response.json` | `f-1038` `host`: `FHIR Transform` → `Lab Router` |
| `healthscan-api.md` §2 | "exactly four: …, FHIR Transform, Cloud API" → "exactly three: EMR Source, Lab Router, Cloud API" |
| `CHANGELOG.md` | entry with the reasoning above |

## Verification

```
$ node validate.mjs
...
```

## Follow-up outside `contracts/` — deliberately not in this PR

`FHIR Transform` also appears in each of our own areas, and I am not editing anyone else's:

- `services/detection-engine/fixtures/proxy/*.json` (8 files) and the four-component sentence in `services/detection-engine/CLAUDE.md` — @tanifgit
- the four-component sentence in `apps/dashboard/CLAUDE.md` — @Ari-Glikman
- the four-component sentence in the root `CLAUDE.md` — shared, needs its own PR; happy to open it if you both agree with this one
- `services/metrics-proxy/fixtures/metrics.txt` — mine, **already done** on the #13 branch

Nothing breaks in the meantime.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -10,16 +10,26 @@ Every contract change, dated, with the reason. Newest first.
 `enum`, so this is samples plus one prose sentence. Consumers that read `host` as an opaque string
 need no edit.
 
-**Why:** the LABDEMO production has no FHIR Transform host. It never did. The HL7→PID transform is
-a DTL applied *inside* Lab Router's process, so it produces no config item and therefore no
-`host` label in `/api/monitor/metrics`. §2 of `healthscan-api.md` says only application config
-items appear; a host that IRIS cannot emit does not belong in a sample that Dev B and Dev C mock
-against. Verified against `iris/labdemo/Production.cls`: the items are `EMR Source`, `Lab Router`,
-`Cloud API`, plus the framework-filtered `Ens.ActivityReporter`.
+**Why:** the LABDEMO production no longer has a FHIR Transform host, so IRIS cannot emit that
+`host` label. §2 of `healthscan-api.md` says only application config items appear; a host the
+production cannot produce does not belong in a sample Dev B and Dev C mock against. Verified
+against `iris/labdemo/Production.cls`: the items are `EMR Source`, `Lab Router`, `Cloud API`, plus
+the framework-filtered `Ens.ActivityReporter`.
 
-The alternative — adding a fourth component so the fixture becomes true — would be inventing a
-production host to satisfy a fixture. Nothing in the pipeline produces FHIR, and the root
-`CLAUDE.md` forbids fabricated demo data.
+**It was real when the samples were written — this is a removal, not a correction of a fabrication.**
+Worth stating plainly, because Dev B's live capture in issue #10 does show
+`host="FHIR Transform"`. It was an `EnsLib.MsgRouter.RoutingEngine` whose rule forwarded everything
+straight to CloudAPI: no DTL, and nothing FHIR, as its own class comment said. It existed to
+generate metric activity. Keren removed it in `1801a50` when the pipeline became HL7→PID, and
+PR #14 settled on the three hosts above. The samples are simply older than that change.
+
+Restoring it instead would mean re-adding a pass-through host to the production for no reason
+other than making a fixture true. Nothing in the pipeline produces FHIR — that name was aspirational
+from the start.
+
+**Dev B's instance still runs the older production definition,** so a fresh capture there will show
+four hosts until it is reloaded from `iris/labdemo/`. That is expected and breaks nothing: an extra
+host is over-coverage, and the parser reads whatever `host` labels arrive.
 
 **Changes:**
 
@@ -38,11 +48,11 @@ Verified: `node validate.mjs` passes; both samples still valid.
 
 ### Follow-up outside `contracts/`, not part of this PR
 
-`FHIR Transform` also appears in `services/detection-engine/fixtures/proxy/*.json` (8 files),
-`services/metrics-proxy/fixtures/metrics.txt`, and the four-component sentence in the root
-`CLAUDE.md`, `apps/dashboard/CLAUDE.md`, `services/detection-engine/CLAUDE.md`. Those are each in
-their owner's area. Dev A will fix the metrics-proxy fixture; the rest is on the owning developer.
-Nothing breaks meanwhile — an extra host in a fixture is over-coverage, not a failure.
+`FHIR Transform` also appears in `services/detection-engine/fixtures/proxy/*.json` (8 files) and the
+four-component sentence in the root `CLAUDE.md`, `apps/dashboard/CLAUDE.md`,
+`services/detection-engine/CLAUDE.md`. Those are each in their owner's area, so they are not touched
+here. `services/metrics-proxy/fixtures/metrics.txt` is Dev A's and is already done on the PR #13
+branch. Nothing breaks meanwhile — an extra host in a fixture is over-coverage, not a failure.
 
 ## 2026-08-09 — Schema fixes from Dev C's review (Dev B)
 
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,46 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-08-10 — `FHIR Transform` removed from the samples (Dev A)
+
+**No schema change and no field change.** `host` is `{"type": "string", "minLength": 1}` with no
+`enum`, so this is samples plus one prose sentence. Consumers that read `host` as an opaque string
+need no edit.
+
+**Why:** the LABDEMO production has no FHIR Transform host. It never did. The HL7→PID transform is
+a DTL applied *inside* Lab Router's process, so it produces no config item and therefore no
+`host` label in `/api/monitor/metrics`. §2 of `healthscan-api.md` says only application config
+items appear; a host that IRIS cannot emit does not belong in a sample that Dev B and Dev C mock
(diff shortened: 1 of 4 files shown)


tanifgit (approved):
Approving. You answered the question from my #14 comment with option (1) and did it properly — own PR, CHANGELOG entry, all three reviewers. **Right call, and for the reason that matters: the fixture was wrong, not the production.**

> The alternative was adding a fourth component so the fixture becomes true. Nothing in the pipeline produces FHIR, so that would be inventing a production host to satisfy a fixture.

That's exactly it, and it's the direction this should go. I explicitly did not want you reinstating a pass-through to match my fixtures.

## Verified every claim rather than reading the diff

```
$ node validate.mjs

hosts:            Cloud API (0.05), EMR Source (0.01), Lab Router (0.08)   3, alphabetical
finding types:    8/8 still covered
orphan findings:  none — every finding.host exists in hosts
f-1038 ratio:     claims 5.1x;  0.41/0.08 = 5.1x                          exact
f-1038 baseline:  0.08  vs Lab Router avgProcessingTime 0.08              AGREES
schema/.d.ts:     zero diff against main                                   as you said
```

**Your careful bit is the right bit.** Deleting `f-1038` would have zeroed `slow_processing` coverage for both consumers, since the samples carry exactly one finding per type. Reassigning to `Lab Router` costs nothing *and* improves it — the finding's `baselineValue: 0.08` now matches the host row it names, where before it cited a baseline belonging to a host that doesn't exist. It agrees better than it did. Good instinct to flag that as the thing to review closely.

`host` being an open string with no `enum` is what makes this a zero-code change for me — I read it as opaque and join on it, nothing more. Confirmed: no host name is hardcoded anywhere in `apps/dashboard/src/`.

## My follow-up, quantified

Two of the attributions in your list are swapped — `services/detection-engine/**` is Ari's and `apps/dashboard/**` is mine. No matter, the work is clear:

**Mine:** 8 demo fixtures, **19 occurrences** of `FHIR Transform`, plus the four-component sentence in `apps/dashboard/CLAUDE.md` §5 and my host grid rendering 4 loading skeletons. I'll do it once this merges — not before, since the samples are the source of truth and I'd rather follow them than race them.

Worth noting it's a little more than a find-and-replace, which is why I'm not rushing it: `scenario-dead-host` and `scenario-throughput-drop` use FHIR Transform as the *downstream* host that stalls behind Cloud API, so removing it means re-telling those two scenarios with three hosts rather than deleting a row. Same class as the `throughput-drop` rescale in #6 — the numbers move, but the story has to still make sense. My `validate-fixture-claims.mjs` will catch me if it doesn't.

**Agreed on the root `CLAUDE.md` sentence needing its own PR** — please open it. Root `CLAUDE.md` says every edit there is a merge conflict for the other two, so one deliberate PR beats three of us touching it.

## One thing for whoever merges

This lands *after* my #5, so `main` will briefly have contract samples with three hosts and dashboard fixtures with four. Harmless — an extra host in a demo fixture is over-coverage, exactly as you say, and demo mode never consults `contracts/samples/`. The `dashboard` CI job validates fixture *shape* against the schema, which is unchanged, so it stays green. Flagging only so nobody sees the mismatch on `main` and thinks something broke.

Ari-Glikman (approved):
Approving. Verified the premise on `main` rather than taking it from the PR body — #14's commits are merged and `Production.cls` now has exactly three application items:

```
<Item Name="EMR Source"    <Item Name="Lab Router"    <Item Name="Cloud API"
<Item Name="Ens.Activity.Operation.Local"    (framework filter drops this)
```

So the samples are stale, not wrong, exactly as you say. `node validate.mjs` → 11/11 with your changes.

## Correcting your ownership list — those 8 fixtures are mine, not Dev C's

Your follow-up table has this the wrong way round:

> - `services/detection-engine/fixtures/proxy/*.json` (8 files) and the four-component sentence in `services/detection-engine/CLAUDE.md` — @tanifgit
> - the four-component sentence in `apps/dashboard/CLAUDE.md` — @Ari-Glikman

Per `CONTRIBUTING.md` §2, `services/detection-engine/**` is **Dev B (me)** and `apps/dashboard/**` is **Dev C**. Easy swap to make and it doesn't change your PR — just flagging so nobody waits on the wrong person. **I'll take the detection-engine ones**; @tanifgit has the dashboard sentence.

Confirmed the scope on my side: all 8 proxy fixtures carry a `FHIR Transform` host, plus one sentence in my `CLAUDE.md`. I'll do it as its own PR after this merges, so the contract moves first and my fixtures follow it rather than the reverse.

## Reassigning `f-1038` rather than deleting it was the right call

This is the part worth praising. Deleting it would have zeroed `slow_processing` coverage for both consumers, and the samples carry exactly one finding per type — so a reader would have had nothing to mock that rule against.

And your reasoning holds arithmetically: `baselineValue: 0.08` **is** Lab Router's measured `avgProcessingTime` in `hosts-response.json`. I checked, because a reassigned host with a mismatched baseline would be a fixture asserting something the engine couldn't produce — the exact class of bug Dev C's `validate-fixture-claims.mjs` was built to catch. It agrees better than it did before, as you said.

## On correcting your own PR body

> my first version of this PR body wrongly claimed the production never had it. Git history disagrees with me

Worth saying: that's the right instinct and it's the second time this week someone here has corrected their own claim against evidence rather than defending it. My "zero hosts" headline on #10 needed the same treatment and you gave it to me on #13. It's the reason this integration is going to work.

## One consequence for me, and it's fine

You're right that my instance still runs the older production definition, so a fresh capture there shows four hosts. Two notes:

1. I won't reload from `iris/labdemo/` yet — my engine's fixtures still reference the four-host shape, and I'd rather change the fixtures and the instance in one deliberate step than have them disagree mid-review.
2. The capture in #10 remains valid as *evidence about label shapes*, which is what it was for. The host list in it is now historical, and I'll say so if anyone reads it as current.

Nothing blocking. Merge whenever — and once it's in, I'll follow with the detection-engine fixture PR.

Ari-Glikman:
My follow-up is **ready and pushed** on `devB/drop-fhir-transform` — deliberately not opened as a PR yet, so the contract lands first and my fixtures track it rather than the other way round. I'll open it the moment this merges.

What it does, for visibility:

- `FHIR Transform` removed from all **8** `services/detection-engine/fixtures/proxy/*.json`
- `CLAUDE.md` §6 stops restating the host list and points at `iris/labdemo/Production.cls` instead — a host list duplicated across three areas is what went stale here, so naming the authoritative source seemed better than correcting the copy

The thing I actually checked, because it was the real risk: **the demo loop still reaches all eight finding types with three hosts.** `stalled_host` uses Lab Router and `dead_host` uses Cloud API, so neither depended on the removed host — but that was luck rather than design, and worth confirming rather than assuming:

```
OK   dead_host             critical      OK   slow_processing       warning
OK   stalled_host          warning       OK   growing_queue_wait    warning
OK   queue_buildup         warning       OK   throughput_drop       warning
OK   elevated_error_rate   critical      OK   system_alert          info

serves with no proxy (ADR 0004): hosts: 3 -> Cloud API, EMR Source, Lab Router
110/110 tests, tsc clean
```

Three hosts, matching your `Production.cls` items exactly.

PR #18, opened 2026-08-11
fix(detection-engine): drop FHIR Transform from the proxy fixtures

Ari-Glikman:
The detection-engine half of #15, which is now merged. Follows it deliberately rather than racing it, so the contract moved first and these fixtures track it.

`FHIR Transform` is no longer an `<Item>` in `iris/labdemo/Production.cls` — @kskubach removed it in #14 when the pipeline became `EMR Source → Lab Router → Cloud API`. IRIS cannot emit that host label any more, so a fixture carrying it asserts a host the production cannot produce.

## Changes

- Removed from all **8** `fixtures/proxy/*.json`
- `CLAUDE.md` §6 stops restating the host list and points at `Production.cls` instead

That second one is the more interesting change. A host list duplicated across `contracts/samples/`, my fixtures, the dashboard's docs and the root `CLAUDE.md` is precisely what went stale here — four copies, one source of truth, and the copies drifted. Naming the authoritative source beats correcting the copy.

## The thing actually worth checking

Not "did the host go away" but **can the demo still show all eight finding types with three hosts**:

```
OK   dead_host             critical      OK   slow_processing       warning
OK   stalled_host          warning       OK   growing_queue_wait    warning
OK   queue_buildup         warning       OK   throughput_drop       warning
OK   elevated_error_rate   critical      OK   system_alert          info
```

It can — but that was **luck, not design**. `stalled_host` happens to use Lab Router and `dead_host` Cloud API, so neither depended on the removed host. Had the fixtures been laid out slightly differently, removing a host would have silently dropped a finding type from the demo, and `test/scenario.test.ts` would have caught it only because it exists. Worth stating plainly: the coverage assertion is what made this a five-minute change instead of a bug found during the screencast.

## Verification

```
tsc --noEmit strict          clean
tests                        110/110 pass
serves with no proxy         hosts: 3 -> Cloud API, EMR Source, Lab Router   (ADR 0004)
```

Three hosts, matching `Production.cls`'s application items exactly. Rebased on `main` post-#15 and re-verified there — the contract samples and these fixtures now name the same three hosts, which is the point of `contracts/samples/` being the shared bytes.

`Ens.MonitorService` remains in `healthy.json` on purpose: it's the framework host that proves `isFrameworkHost()` filters it, and the engine reports 3 rather than 4 as a result.

## Also includes one incidental fix

`package-lock.json` still carried `"node": ">=20.6.0"` while `package.json` requires `>=22.18.0` since the switch to native type stripping. One line, regenerated by `npm install` rather than hand-edited. Unrelated to the rest — happy to split it out if you'd rather.

Refs #15, #14

### DIFF
--- a/services/detection-engine/CLAUDE.md
+++ b/services/detection-engine/CLAUDE.md
@@ -188,9 +188,11 @@ all eight types.
 
 ## 6. Never invent data
 
-- **No fabricated hosts, metrics, or findings outside `fixtures/`.** Fixtures use the four LABDEMO
-  components — **EMR Source** (service), **Lab Router** (process), **FHIR Transform** (process),
-  **Cloud API** (operation) — and the eight real finding types.
+- **No fabricated hosts, metrics, or findings outside `fixtures/`.** Fixtures use the LABDEMO
+  application components — **EMR Source** (service), **Lab Router** (process), **Cloud API**
+  (operation) — and the eight real finding types. The authoritative list is the `<Item>` set in
+  `iris/labdemo/Production.cls`; do not restate it here beyond this note, because a host list
+  duplicated across areas is exactly what went stale when `FHIR Transform` was removed.
 - Fixture values should be **captured from live LABDEMO**, not invented. Real numbers catch real
   problems; `contracts/samples/` was built that way.
 - If the proxy is unreachable, serve last-known findings and report `X-Healthscan-State: stale`.
--- a/services/detection-engine/fixtures/proxy/cloud-api-dead.json
+++ b/services/detection-engine/fixtures/proxy/cloud-api-dead.json
@@ -30,19 +30,6 @@
       "sampleCount": 24
     },
     {
-      "name": "FHIR Transform",
-      "type": "actor",
-      "status": "OK",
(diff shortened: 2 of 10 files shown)


Ari-Glikman:
CI is red here on one step, and **it's my check that's wrong, not this change.**

The ADR 0004 step in `ci.yml` hardcoded `test "$COUNT" -eq 4`. Removing FHIR Transform means the engine now correctly serves **3**, so my own assertion caught a correct change and called it a failure:

```
hosts served with no proxy: 3
##[error]Process completed with exit code 1
```

Fixed in **#19** — changed to `-ge 1`, since the exact number was never the point of that step. It asserts the engine starts and serves with no proxy; a host count belongs to `Production.cls`, which that job doesn't own.

Everything in this PR is green: **110/110 tests, tsc clean**, and the three other CI jobs pass. #18 will go green once #19 merges.

Mildly embarrassing but a fair catch by the harness — and it's the third instance of the same root cause this week: a fact copied into a place that doesn't own it, then drifting. The host list was in four files, and the count was in CI too.

PR #20, opened 2026-08-11
fix(detection-engine): tune floors to measured LABDEMO baselines

Ari-Glikman:
ADR 0003 committed to replacing the guessed defaults *"once each of the eight triggers has been induced against LABDEMO."* IRIS is up, so this is that pass — and the guesses were wrong in a way that mattered.

## The absolute floors dominated on every host

Measured steady state, stable across four samples 10s apart:

| | avg processing | avg queue wait |
|---|---|---|
| EMR Source | `0` | `0` |
| Lab Router | `0.08s` | `0` |
| Cloud API | `0.05s` | `0.03s` |

Against a `0.08s` baseline, a `1.0s` floor is **12.5×** — so `baselineMultiplier: 3.0` never bound. Proven against the rule itself rather than reasoned about:

```
BEFORE (floor 1.0s, baseline 0.08s)
  0.24s ( 3.0x) -> NO FINDING
  0.80s (10.0x) -> NO FINDING       <- a 10x regression, silent
  1.00s (12.5x) -> critical         <- and it skips warning entirely

AFTER (floor 0.2s)
  0.16s ( 2.0x) -> no finding
  0.24s ( 3.0x) -> warning
  0.80s (10.0x) -> critical
```

**Two defects in one.** `slow_processing` could not report a real 10× slowdown, and its `warning` band was unreachable.

@tanifgit — that second half is the same class as the `info`-unreachable finding you caught on #8, and I missed it then because I was reading the severity bands rather than asking what the *firing gate* actually was. The gate was the floor, not the multiplier, on every single host.

## Changes

```
slow_processing     absoluteFloorSeconds  1.0 -> 0.2
growing_queue_wait  absoluteFloorSeconds  1.0 -> 0.15
Cloud API override                        2.0 -> 0.3
```

Cloud API keeps a higher floor deliberately: its latency includes a simulated remote call, so it tolerates more than the in-process hosts. `0.3s` is ~6× its measured `0.05s` rather than 40×.

## What I deliberately did not change

**`queue_buildup`'s `absoluteFloor: 50`.** LABDEMO idles at `0` queued, and when I disabled Cloud API a real backup hit **70 within three minutes**. So 50 sits correctly between noise and a genuine problem — the one guessed number that happened to be right.

**The multipliers and severity bands.** They're the right shape and were never the binding constraint. Changing them would have masked the real problem.

## Verification

```
tsc --noEmit strict     clean
tests                   110/110
demo loop               all 8 finding types, all 3 severities
```

One thing worth recording: my first attempt put `hostOverrides` *inside* `rules` instead of beside it, and the config validator rejected it with `unknown rule "hostOverrides"`. That's the "never fall back to zeros, refuse the file loudly" guard from ADR 0003 earning its place — a permissive loader would have silently dropped the override and I'd have shipped Cloud API on the wrong floor.

Entirely within `services/detection-engine/`. Refs ADR 0003, #8

### DIFF
--- a/services/detection-engine/src/config/thresholds.ts
+++ b/services/detection-engine/src/config/thresholds.ts
@@ -42,6 +42,13 @@ export interface DurationRuleConfig {
   enabled: boolean;
   baselineMultiplier: number;
   absoluteFloorSeconds: number;
+  /**
+   * Zero-baseline hosts only. An infinite ratio cannot be graded by the bands, so
+   * severity comes from absolute magnitude: this many times the floor earns `critical`,
+   * anything above the floor alone is a `warning`. Without it the floor becomes a
+   * critical trigger and there is no warning tier at all.
+   */
+  criticalFloorMultiple: number;
   severityBands: { warning: number; critical: number };
 }
 
@@ -111,12 +118,14 @@ export const DEFAULT_CONFIG: ThresholdConfig = {
       enabled: true,
       baselineMultiplier: 3.0,
       absoluteFloorSeconds: 1.0,
+      criticalFloorMultiple: 4.0,
       severityBands: { warning: 3.0, critical: 10.0 },
     },
     growing_queue_wait: {
       enabled: true,
       baselineMultiplier: 3.0,
       absoluteFloorSeconds: 1.0,
+      criticalFloorMultiple: 4.0,
       severityBands: { warning: 3.0, critical: 10.0 },
     },
     throughput_drop: {
--- a/services/detection-engine/src/detect/rules/index.ts
+++ b/services/detection-engine/src/detect/rules/index.ts
@@ -168,9 +168,21 @@ function durationRule(
       const ratio = baseline > 0 ? current / baseline : Number.POSITIVE_INFINITY;
       if (ratio < rule.baselineMultiplier) return null;
 
+      // A zero baseline makes every ratio infinite, so the multiplier cannot bind and
+      // the floor is the entire gate. This branch previously hardcoded `critical`,
+      // which meant a host whose normal queue wait is 0 went from silent to critical
+      // at the floor with no warning tier -- and lowering the floor to 0.15s made that
+      // trigger 150ms of queue wait. MVP §6 names false positives as the top risk.
+      //
+      // Instead, judge an infinite ratio on absolute magnitude against the floor:
+      // `criticalFloorMultiple` x the floor earns `critical`, anything above the floor
+      // is a `warning`. That restores a two-tier response for zero-baseline hosts
+      // without making the floor itself a critical trigger. Found by Dev C on #20.
       const severity = Number.isFinite(ratio)
         ? severityFromRatio(ratio, rule.severityBands)
-        : 'critical';
+        : current >= rule.absoluteFloorSeconds * rule.criticalFloorMultiple
+          ? 'critical'
+          : 'warning';
       const comparison = Number.isFinite(ratio) ? ` is ${formatRatio(ratio)} baseline` : '';
 
       return {
--- a/services/detection-engine/test/rules.test.ts
+++ b/services/detection-engine/test/rules.test.ts
@@ -259,6 +259,80 @@ describe('growing_queue_wait', () => {
     assert.match(verdict.message, /Average queue wait 1\.84s/);
   });
 
+  describe('zero baseline — the infinite-ratio path', () => {
+    // A zero baseline makes every ratio infinite, so the bands cannot grade it and the
+    // floor is the whole gate. This used to hardcode `critical`, meaning a host whose
+    // normal wait is 0 went silent -> critical at the floor with no warning tier. Found
+    // by Dev C on #20. LABDEMO measures 0 here on two of three hosts, so this is the
+    // common case rather than an edge one.
+    const floor = DEFAULT_CONFIG.rules.growing_queue_wait.absoluteFloorSeconds;
+    const criticalAt = floor * DEFAULT_CONFIG.rules.growing_queue_wait.criticalFloorMultiple;
+
+    it('does NOT fire below the floor', () => {
+      const verdict = evaluate(
+        'growing_queue_wait',
+        host({ avgQueueingTime: floor * 0.5 }),
+        warmBaseline('avgQueueingTime', 0),
+      );
+      assert.equal(verdict, null);
+    });
+
+    it('warns at the floor rather than going straight to critical', () => {
+      const verdict = evaluate(
+        'growing_queue_wait',
+        host({ avgQueueingTime: floor }),
+        warmBaseline('avgQueueingTime', 0),
+      );
+      assert.ok(verdict !== null);
+      assert.equal(verdict.severity, 'warning', 'the floor must not be a critical trigger');
+    });
+
+    it('still warns just below the critical multiple', () => {
+      const verdict = evaluate(
+        'growing_queue_wait',
+        host({ avgQueueingTime: criticalAt * 0.99 }),
+        warmBaseline('avgQueueingTime', 0),
+      );
+      assert.equal(verdict?.severity, 'warning');
+    });
+
+    it('escalates to critical at the critical multiple', () => {
+      const verdict = evaluate(
+        'growing_queue_wait',
+        host({ avgQueueingTime: criticalAt }),
+        warmBaseline('avgQueueingTime', 0),
+      );
+      assert.equal(verdict?.severity, 'critical');
+    });
+
+    it('omits the ratio from the message, since there is no honest one to state', () => {
+      const verdict = evaluate(
+        'growing_queue_wait',
+        host({ avgQueueingTime: 1.0 }),
+        warmBaseline('avgQueueingTime', 0),
+      );
+      assert.ok(verdict !== null);
+      assert.doesNotMatch(verdict.message, /baseline/, 'must not claim an Infinity ratio');
+    });
+
+    it('applies the same two tiers to slow_processing', () => {
+      const procFloor = DEFAULT_CONFIG.rules.slow_processing.absoluteFloorSeconds;
+      const procCritical = procFloor * DEFAULT_CONFIG.rules.slow_processing.criticalFloorMultiple;
+      const atFloor = evaluate(
+        'slow_processing',
+        host({ avgProcessingTime: procFloor }),
+        warmBaseline('avgProcessingTime', 0),
+      );
+      const atCritical = evaluate(
+        'slow_processing',
+        host({ avgProcessingTime: procCritical }),
+        warmBaseline('avgProcessingTime', 0),
+      );
+      assert.equal(atFloor?.severity, 'warning');
+      assert.equal(atCritical?.severity, 'critical');
+    });
+  });
+
   it('does NOT fire on a healthy sub-second wait', () => {
     const verdict = evaluate(
       'growing_queue_wait',
--- a/services/detection-engine/thresholds.json
+++ b/services/detection-engine/thresholds.json
@@ -1,12 +1,9 @@
 {
-  "_comment": "ADR 0003. The only place detection numbers live. Hot-reloaded on change. Defaults are conservative starting points, to be corrected once each of the eight triggers has been induced against LABDEMO.",
-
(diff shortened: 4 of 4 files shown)


tanifgit (approved):
Approving. **The change is right, the diagnosis is right, and the evidence is real** — I verified it against the rule rather than reading the diff.

## Verified rather than taken on trust

Your BEFORE/AFTER table reproduces exactly. I traced it through `durationRule` in `src/detect/rules/index.ts:159-183` by hand, then ran your suite against your actual thresholds file:

```
BEFORE (floor 1.0s, baseline 0.08s)
  0.24s  0.24 < 1.0            -> null            NO FINDING     as you say
  0.80s  0.80 < 1.0            -> null            NO FINDING     as you say
  1.00s  ratio 12.5, >= 10     -> critical        skips warning   as you say

AFTER (floor 0.2s)
  0.16s  0.16 < 0.2            -> null            no finding      as you say
  0.24s  ratio 3.0,  >= 3      -> warning         as you say
  0.80s  ratio 10.0, >= 10     -> critical        as you say

$ npx tsc --noEmit          clean
$ npm test                  tests 110  pass 110  fail 0   (suites 32)
```

Your scenario suite still passes intact, which is the part I cared about most: `produces all eight finding types`, `reaches critical, warning and info`, and `sources info only from system_alert` are all green under the new floors. The retune didn't buy reachability in one rule by breaking coverage in another.

**Only three things actually changed.** The diff is bigger than the change — `severityBands` got expanded to multiline and the blank lines between sections went — so I diffed the *parsed* objects to be sure nothing rode along in the reformat:

```
rules.slow_processing.absoluteFloorSeconds              1 -> 0.2
rules.growing_queue_wait.absoluteFloorSeconds           1 -> 0.15
hostOverrides.Cloud API.slow_processing.absoluteFloor…  2 -> 0.3
```

Nothing else. And the `_comment_info_severity` text is identical after parsing, despite the reformat rewriting its `§` as a `\u`-escape — I checked, because a reformat that silently altered the comment warning people off a change would be an unfortunate thing to lose.

## On the diagnosis

> I missed it then because I was reading the severity bands rather than asking what the *firing gate* actually was. The gate was the floor, not the multiplier, on every single host.

That's the right lesson and it's the more useful half of this PR. `baselineMultiplier: 3.0` sitting next to `absoluteFloorSeconds: 1.0` reads like two independent conditions, and only one of them was ever load-bearing. Worth noting the guessed value wasn't unreasonable in isolation — 1.0s "feels" like a slow message. It was only wrong *relative to a baseline nobody had measured yet*, which is precisely the cost ADR 0004 predicted and the reason it committed to this pass.

## What I'd add: the same question, asked of the other branch

You asked "what is the firing gate?" for the finite-ratio path and found the floor. Asking it of the **infinite-ratio** path gives a different answer, and it's the one case your fix doesn't reach.

`index.ts:168-173` sets `ratio = Infinity` when the baseline is `0`, and then severity is **hardcoded `'critical'`** rather than going through `severityFromRatio`. So for a zero-baseline host the multiplier can never bind, there is no warning tier at all, and the floor is the *entire* gate — for `critical` specifically.

LABDEMO's measured `avgQueueingTime` is `0` on EMR Source and Lab Router. Working the reachability out per host under your new numbers:

```
slow_processing (avgProcessingTime)
  EMR Source   baseline 0      ratio Infinity -> hardcoded CRITICAL at >= 0.2s    NO warning tier
  Lab Router   baseline 0.08   multiplier binds (3.0x)    warning reachable 0.24-0.80s
  Cloud API    baseline 0.05   FLOOR binds (6.0x)         warning reachable 0.30-0.50s

growing_queue_wait (avgQueueingTime)
  EMR Source   baseline 0      ratio Infinity -> hardcoded CRITICAL at >= 0.15s   NO warning tier
  Lab Router   baseline 0      ratio Infinity -> hardcoded CRITICAL at >= 0.15s   NO warning tier
  Cloud API    baseline 0.03   FLOOR binds (5.0x)         warning reachable 0.15-0.30s
```

Two observations from that:

**1. On four of the six (host, metric) pairs the first severity anyone can ever observe is still `critical`.** For `growing_queue_wait` that's two of the three hosts. Your PR fixes an unreachable warning band caused by the floor; this is the same band, still unreachable, caused by the zero baseline instead.

**2. This PR lowers exactly where that `critical` fires, by 6.7×.** On Lab Router, a sustained `avgQueueingTime` of `0.15s` now produces a `critical` finding where `1.0s` was needed before. `sustainedSamples: 2` is a real mitigation — at the production 10s poll that's 20s of continuous queueing, not a spike — so I don't think this is alarming. But "150ms of queue wait on a host that normally has none" reaching straight for `critical`, with no warning step in between, is a strong claim, and MVP §6 names false positives as the top risk.

I'm deliberately not proposing a number, because it's your rule and the measurement is yours. The shape of the options, as I see it: treat an infinite ratio as `warning` and require a separate absolute bound for `critical`; or give the zero-baseline case its own explicit threshold; or decide it's correct as-is and record *why*, the way you did for `queue_buildup`'s floor. **Any of those is fine by me and none of them blocks this PR** — the current behaviour is pre-existing, not introduced here. I'd just rather it were a decision with a comment on it than an emergent property of `Infinity`.

## What this costs me, already handled

Four host rows in my demo fixtures would be flagged by your retuned engine and carry no finding:

```
NEW after #20   error-storm     Cloud API   avgQueueingTime 0.84s vs baseline 0.02
NEW after #20   queue-buildup   Lab Router  avgQueueingTime 0.62s vs baseline 0
NEW after #20   system-alert    Cloud API   avgQueueingTime 0.21s vs baseline 0.02
PRE-EXISTING    dead-host       Lab Router  avgQueueingTime 41.6s vs baseline 0
```

**No action for you, and it does not gate merging in either direction.** My findings all still *validate* against your file — lowering a floor can't invalidate a finding that already cleared a higher one, and my strict validator passes 8/8 either way. I confirmed `scenario-healthy` stays clean under the new floors too, which was the case I was actually worried about; a healthy fixture that starts producing findings would be the damaging one. `baseline-warming` is safe because comparative rules can't fire with a null baseline.

The interesting part is that **my own validator is structurally blind to this.** `validate-fixture-claims.mjs` walks the findings and asks "would the engine emit this?" It never walks the hosts and asks "is there a finding the engine *would* emit that's missing?" That's the same failure mode you've now named twice this week — a check that gets quietly weaker instead of failing — sitting in the tool I wrote to catch it. The `dead-host` row is the proof: `41.6s` was already miles above the old `1.0s` floor and slipped through anyway, so this was never really about your retune.

Both are mine: a reverse check in the validator first, then the four rows once the check can prove the fix. Written up in #21 (dashboard side of the FHIR Transform removal, now open and green) so it isn't only in a review comment.

## Taken on trust

`queue_buildup`'s `absoluteFloor: 50` staying put, on the strength of a real backup hitting **70 within three minutes** with Cloud API disabled. I have no LABDEMO access so I can't check that one — but I'd rather say so than let it read as verified. Agreed on the reasoning that it's the one guessed number that happened to land between noise and a genuine problem, and agreed with not touching the multipliers or bands when they were never the binding constraint.

## Nit, ignore if you like

The reformat turned the literal `§` in `_comment_info_severity` into a six-character `\u`-escape. It parses back to the same string, so this is cosmetic — but that comment exists to be *read in the raw file* by whoever is about to lower a warning band, and an escape sequence mid-sentence is slightly worse at that job than the character itself.

tanifgit:
Correcting one thing I told you above, because it now affects when you can merge this.

I said:

> **No action for you, and it does not gate merging in either direction.** [...] Merge order between us does not matter.

**That was true of the check that existed when I wrote it, and is not true any more.** I've since built the reverse sweep I promised — walk the hosts, fail on a finding the engine *would* emit that the fixture omits — and it's in #22. With that in place there is now a real ordering constraint, and it points at your PR:

**Merging #20 on its own turns the `dashboard` CI job red on `main`.** That job runs `validate:fixtures`, which now includes the sweep, and your new floors surface three fixture omissions:

```
error-storm     Cloud API   growing_queue_wait  0.84s vs 0.02  ratio 42
queue-buildup   Lab Router  growing_queue_wait  0.62s, no baseline
system-alert    Cloud API   growing_queue_wait  0.21s vs 0.02  ratio 11
```

**And I cannot pre-fix them**, which is the part that creates the constraint rather than just the work. Under today's `1.0s` floor the emittable check correctly rejects the very findings the sweep will later demand:

```
FAIL  scenario-queue-buildup.json — f-1044 growing_queue_wait @ Lab Router
        0.62s is below absoluteFloorSeconds 1 — would not fire
```

So across your boundary the two directions genuinely conflict: today those three findings must **not** exist, and the moment your floors land they **must**. There is no state of my fixtures that satisfies both.

Nothing here is an argument against #20 — the approval stands, the retune is right, and this is the drift detector doing precisely what it was built for. It just shouldn't land unannounced and leave `main` red while someone wonders why.

**Suggested sequence: #21 → #22 → #20**, and I'll have the three-row follow-up ready to push the moment #20 lands, so the red window is minutes. Alternatively merge #20 first and I'll fix forward immediately — I don't mind which, I'd just rather you knew there was a sequence than discover it from a red build.

One more thing that fell out of doing this properly, and it's yours to judge — non-blocking, no action needed for #20.

`queueBuildup` appends `with no baseline queue` for the infinite-ratio case (`index.ts:99-101`), but `durationRule` appends an empty string (`index.ts:174`). Same situation, two presentations. In `dead-host` they now land two rows apart on screen:

```
Critical  Queue buildup       · Lab Router   Queue depth 212 with no baseline queue
Critical  Growing queue wait  · Lab Router   Average queue wait 41.60s
```

The second doesn't say why there's no ratio to quote. I used your real string rather than the nicer one, since a fixture may not invent wording the engine doesn't produce — so if you'd rather `durationRule` said something equivalent, my fixtures follow you rather than the reverse.

Ari-Glikman:
You found the half of this I missed, and it's a real defect — **fixed in fbb34da**, so this PR now does both.

## Confirmed exactly as you described

`src/detect/rules/index.ts:171-173` hardcoded `: 'critical'` on the infinite branch. Reproduced on a zero-baseline Lab Router with my new floor:

```
0.14s -> no finding
0.15s -> CRITICAL          <- no warning tier at all
41.6s -> CRITICAL
```

And you're right that **my retune made it worse specifically**, lowering that trigger 6.7×. "150ms of queue wait reaches for critical" is not a claim I want the product making, and MVP §6 names false positives as the top risk.

The part that decided it for me: **this isn't an edge case.** LABDEMO measures `avgQueueingTime` of `0` on two of three hosts, so the zero-baseline path is the *common* one for `growing_queue_wait`, not the exception. I'd been reasoning about the finite-ratio path as the normal case and it isn't.

## The fix — your first option, made configurable

An infinite ratio can't be graded by a ratio band, so severity now comes from absolute magnitude:

```
criticalFloorMultiple x floor  -> critical
above the floor alone          -> warning

AFTER (floor 0.15s, multiple 4.0)
  0.14s -> no finding      0.59s -> warning
  0.15s -> warning         0.60s -> critical
  0.30s -> warning         41.6s -> critical
```

Configurable per ADR 0003 rather than hardcoded, default `4.0`. The message still omits the ratio, since there's no honest one to state.

You said you'd rather it were "a decision with a comment on it than an emergent property of `Infinity`" — that's exactly what it was, and the comment is now in the code at the branch plus in the config type.

**116/116 tests** (+6), covering both duration rules and asserting explicitly that *the floor must not be a critical trigger*.

## Your `§` nit — fixed, and it was a real regression

Not cosmetic in effect: `ensure_ascii` in my reformat had escaped it. Restored to the literal character. Your reasoning is right — that comment exists to be read in the raw file by whoever is about to lower a warning band, and my own tooling had degraded it.

## On your validator being blind to this

> It never walks the hosts and asks "is there a finding the engine *would* emit that's missing?"

That's the sharpest observation in the review, and the `dead-host` row at `41.6s` proves it was never about my retune — it cleared the old `1.0s` floor by 40× and slipped through anyway.

Worth noting the symmetry: my `scenario.test.ts` has the same blind spot in reverse. It asserts every finding type *can* appear, and never asks whether a fixture host *should* be producing a finding it isn't. Both tools check the direction their author was worried about. I'll add the reverse check on our side too — it belongs next to the coverage assertion rather than in your validator alone.

## And on `queue_buildup`'s floor of 50

Thank you for saying you couldn't verify it rather than letting it read as checked. For the record so it's not just my word: production running, Cloud API disabled, ~3 minutes elapsed —

```
Cloud API | status=Disabled | queue='70' | elapsed=176
```

That's also the capture behind #12, where it turned out the metrics text reports `status="OK"` and `queued 0` for that same host at that same instant. Happy to run any specific state you want measured — the instance is local to me, so it's cheap.

PR #21, opened 2026-08-11
fix(dashboard): drop FHIR Transform from the demo fixtures

tanifgit:
## What and why

Follows @kskubach's #15, which removed `FHIR Transform` from `contracts/samples/`. This is the dashboard's side of it — the third and last of the three areas, after #15 and @Ari-Glikman's #18.

Refs #15. No `contracts/` file touched.

The host left the LABDEMO production when the pipeline became HL7→PID, so as #15's CHANGELOG entry puts it, **the fixtures were older than the production rather than wrong.** I kept that distinction in `fixtures/README.md` and `CLAUDE.md` §5 rather than just deleting the name, because "we removed a fabricated host" and "a real host left the production" lead a future reader to different conclusions.

## The two that needed re-telling, not a deleted row

I flagged in my #15 review that this would not be a find-and-replace. Two scenarios used `FHIR Transform` as a narrative participant:

**`dead-host`** — it was the host that stalled behind a `Disabled` Cloud API, which is the demo headline. **Lab Router takes that role**: keeps its own queue depth of 212, inherits the stall characteristics (`messagesPerSec` 0, `avgQueueingTime` 41.6, idle 402 s). All four findings survive — `dead_host` on Cloud API, and `throughput_drop` + `stalled_host` + `queue_buildup` on Lab Router. The `stalled_host` message moves from "338 message(s) are queued" to 212, so it agrees with the row it describes.

**`throughput-drop`** — the drop propagated through three hosts, and dropping one would have left a two-stage story. Cloud API's row **already** asserted `messagesPerSec: 0.1` against its measured `0.4` baseline, so I added the finding that row was already implying rather than inventing a value. Full-length propagation again: 0.2→0, 1.2→0.4, 0.4→0.1.

Note the severities there: only EMR Source is `critical`. The two downstream fractions are 0.33 and 0.25, and `throughput_drop`'s critical band is at or below 0.1. That is in the scenario `note` now so nobody "fixes" it later.

## The thing worth reviewing carefully

**Both `stalled_host` findings in the entire set were on `FHIR Transform`** — `dead-host` and `baseline-warming`. A straight row-and-finding delete would have zeroed that type's coverage for anyone reading these fixtures, the same trap @kskubach avoided with `f-1038` and `slow_processing` in #15. Both are reassigned to Lab Router instead.

Same shape in `error-storm`: its Lab Router finding (was FHIR Transform) is the **only `elevated_error_rate` in the set with a non-zero baseline**, so it is the only one exercising the ratio path rather than the clean-baseline path. `4.2 / 1.1 = 3.8x`, unchanged.

Straight row deletions only in `healthy`, `queue-buildup` and `system-alert`, where the host carried no findings.

## One judgement call I did not silently make

`slow-processing` keeps a `growing_queue_wait` **baseline of 0.02** on Lab Router, where the measured steady-state table reads `0`. That number came from the finding this scenario used to hang on `FHIR Transform`, and it did not match FHIR Transform's measured `0` either — so it is inherited, not introduced.

I kept it rather than zeroing it, and documented why in `fixtures/README.md`: the engine's baseline is a rolling 30-minute mean (ADR 0002), so a host that idles at zero still carries a small non-zero mean if anything queued during the window. Zeroing it would change the engine's own message to the "no baseline" wording, and a fixture may not invent that string. Happy to be told this should be a `0` and the finding dropped instead.

## Also in here

- `HostGrid` loading skeletons **4 → 3**, with the comment updated. This was the only code change; nothing in `src/` hardcodes a host name, as I confirmed on #15.
- The four-component sentences in `apps/dashboard/CLAUDE.md` §5, §8 and §9, and in `fixtures/README.md`.
- **Three rows of the README scenario table that were stale for unrelated reasons.** It claimed `queue-buildup` shows two warnings (it is two criticals), `dead-host` shows an `Inactive` dot and two criticals (it is `Disabled`, three criticals and a warning), and `throughput-drop` collapses to `1.8 msg/sec` — left over from the issue #6 rescale, where the real value is now `0`. Adjacent to my edits and false as written, so I corrected them. Flagging because it is outside the stated scope of this PR.

## Verification

```
ok    scenario-baseline-warming.json
ok    scenario-dead-host.json
ok    scenario-error-storm.json
ok    scenario-healthy.json
ok    scenario-queue-buildup.json
ok    scenario-slow-processing.json
ok    scenario-system-alert.json
ok    scenario-throughput-drop.json

$ npx tsc --noEmit
(clean)

$ npm run build
dist/index.html  199.53 kB | gzip: 61.33 kB   built in 459ms

hosts across all fixtures:  Cloud API | EMR Source | Lab Router
all 8 finding types still covered  (scripted check, every fixture)
```

**Rendered and checked in a browser, not just built** — a fixture set can validate perfectly and still look wrong:

- `?scenario=dead-host` → 3 critical / 1 warning / 0 info, "Hosts OK (of 3)" = 2, Lab Router showing 212 queued / 0 msg/sec / 41.6 s / "7 minutes ago" with **3 active findings**, Cloud API a grey `Disabled` dot with 1.
- `?scenario=throughput-drop` → three-stage propagation, one finding per host, EMR Source red-bordered and the two downstream cards amber.
- `?mode=live` against the engine on `:3002` → renders real findings, including an `info` from `system_alert`, which is the only rule that can produce one.

## One consequence for @Ari-Glikman, and it is not blocking this PR

Your #20 lowers `growing_queue_wait`'s floor from `1.0s` to `0.15s`. I checked this branch against your actual thresholds file and **my findings all still validate** — lowering a floor cannot invalidate a finding that already cleared a higher one, so 8/8 passes either way. Merge order between us does not matter.

But the other direction bites. **Four host rows in these fixtures would be flagged by your retuned engine and carry no finding:**

```
NEW after #20   error-storm     Cloud API   avgQueueingTime 0.84s vs baseline 0.02
NEW after #20   queue-buildup   Lab Router  avgQueueingTime 0.62s vs baseline 0
NEW after #20   system-alert    Cloud API   avgQueueingTime 0.21s vs baseline 0.02
PRE-EXISTING    dead-host       Lab Router  avgQueueingTime 41.6s vs baseline 0
```

`scenario-healthy` stays clean under the new floors — I checked that specifically, since a healthy fixture that starts producing findings would be the damaging case. `baseline-warming`'s `0.31s` is safe because comparative rules cannot fire with a null baseline.

**And my own validator is structurally blind to this.** `validate-fixture-claims.mjs` walks the findings and asks "would the engine emit this?" It never walks the hosts and asks "is there a finding the engine *would* emit that is missing?" That is the same failure mode you named twice this week — a check that gets quietly weaker instead of failing — sitting in the tool I wrote to catch exactly that. The `dead-host` row proves it, since that one was already above the old `1.0s` floor and slipped through anyway.

Both are mine and both are follow-ups, not this PR: a reverse check in `validate-fixture-claims.mjs`, then the four rows once the check can prove the fix. Your #20 is right on the merits and I will say so on it properly.

## Checklist

- [x] Stayed inside my own area (`CONTRIBUTING.md` §2) — `apps/dashboard/**` only
- [x] No `contracts/` file edited as part of an implementation task
- [x] Ran my area's typecheck / build / tests, and pasted the output above
- [x] No invented data outside a declared fixtures directory — the one added finding uses a value the fixture's own host row already asserted
- [x] No new dependency

## Scope

- [x] Adds no capability from a later module — detection and surfacing only

### DIFF
--- a/apps/dashboard/CLAUDE.md
+++ b/apps/dashboard/CLAUDE.md
@@ -305,7 +305,7 @@ Going through the dev proxy means **CORS is never your problem** and Dev B can b
 
 Demo mode is a **first-class deliverable**, not a stub — it is the Day-5 fallback and the screencast source.
 
-- Fixtures use the LABDEMO production's four real components, so demo and live look continuous: **EMR Source** (service), **Lab Router** (process), **FHIR Transform** (process), **Cloud API** (operation).
+- Fixtures use the LABDEMO production's three real components, so demo and live look continuous: **EMR Source** (service), **Lab Router** (process), **Cloud API** (operation). There was a fourth, `FHIR Transform`, until `contracts/` PR #15 dropped it — a real pass-through routing item that left the production when the pipeline became HL7→PID.
 - Fixture files are plain JSON matching the contract exactly — the same guards run over them. If a fixture fails validation, the contract transcription is wrong.
 - Each fixture is a scenario file containing `{ hosts: Host[], findings: Finding[] }`. `scenario-healthy.json` has zero findings — that exercises the empty state, which is easy to forget and looks bad if broken on stage.
 - `mockClient` supports a **scripted progression**: healthy → degrading → findings appear, advancing on each poll so the dashboard visibly comes alive during the demo without anyone touching IRIS. Make it loop, and make it restartable from the UI.
@@ -415,7 +415,7 @@ From §4.5 of the MVP plan, in order:
 
 | # | Task | Est. | Done when |
 |---|---|---|---|
-| 1 | Dashboard against mocked schema — host grid, findings list, severity summary | 1.5 d | Renders all four LABDEMO hosts and all fixture findings; every state designed; `tsc` clean |
+| 1 | Dashboard against mocked schema — host grid, findings list, severity summary | 1.5 d | Renders all three LABDEMO hosts and all fixture findings; every state designed; `tsc` clean |
 | 2 | Live polling + demo/live toggle (`?mode=live`) | 0.75 d | Visible change reflected within 10 s; API failure degrades to last-good + banner, never blanks |
 | 3 | Finding detail view | 0.5 d | Click a finding → current vs. baseline, metric, timestamp; keyboard accessible; drawer survives polls |
 | 4 | Screencast + static fallback | 0.5 d | `dist/index.html` opens from `file://` in demo mode; screencast recorded |
@@ -431,7 +431,7 @@ Sequencing note: tasks 1–3 need only the contract, so they proceed regardless 
 
 - **Ask before widening scope.** If a request implies any module from §1.1, say so instead of building it.
 - **The contract is upstream of you.** Never edit `contracts/`. Never add a field to a type to make a component work — surface it as a contract question.
-- **Never invent data.** No placeholder hosts, no made-up metrics, no fabricated findings outside `fixtures/`. Fixtures use the four LABDEMO components and the eight real finding types.
+- **Never invent data.** No placeholder hosts, no made-up metrics, no fabricated findings outside `fixtures/`. Fixtures use the three LABDEMO components and the eight real finding types.
 - **No new dependencies** without stating why and what it replaces.
 - **No `fetch` outside `src/api/`.** No component talks to the network.
 - **No hard-coded colors, spacing, or radii.** Use the tokens in §7.1. If a token is missing, add it to `tokens.css` rather than inlining a value.
--- a/apps/dashboard/README.md
+++ b/apps/dashboard/README.md
@@ -29,12 +29,12 @@ Every scenario is reachable by URL, which is also how you grab screenshots.
 | URL | What you should see |
 |---|---|
 | `/` | Scripted progression — advances one fixture per refresh, then loops |
-| `/?scenario=healthy` | All four hosts OK, zero counts, the designed empty state |
-| `/?scenario=queue-buildup` | Lab Router queue at 486, two warnings |
-| `/?scenario=dead-host` | Cloud API grey `Inactive` dot, two criticals, red left borders |
+| `/?scenario=healthy` | All three hosts OK, zero counts, the designed empty state |
+| `/?scenario=queue-buildup` | Lab Router queue at 486, two criticals |
+| `/?scenario=dead-host` | Cloud API grey `Disabled` dot, three criticals and a warning, red left borders |
 | `/?scenario=error-storm` | Cloud API in `Error`, 218 errored messages |
-| `/?scenario=slow-processing` | FHIR Transform at 1.86 s per message |
-| `/?scenario=throughput-drop` | EMR Source intake collapsed to 1.8 msg/sec |
+| `/?scenario=slow-processing` | Lab Router at 1.86 s per message |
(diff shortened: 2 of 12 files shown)


Ari-Glikman (approved):
Approving. The third and last of the three areas, and the most careful of them — verified rather than read.

```
$ npm run build                          -> dist/index.html 199.53 kB, single file
```

## Your `0.02` judgement call — you're right, and here's the proof

You flagged this at me rather than deciding silently, so here's the reproduction against the actual engine:

```
174 samples at 0 + 6 samples at 0.6s  ->  rolling mean = 0.02
```

Exactly `0.02`. Your reasoning is precisely correct: a 30-minute window at 10s polls holds 180 samples (ADR 0002), so a host that idles at zero **does** carry a small non-zero mean if anything queued during the window. `0.02` is a mean of 174 zeros and a brief burst — it is a *legitimate reading for an idle host*, not a leftover.

**Keep it, and don't drop the finding.** Two reasons beyond the arithmetic:

1. Zeroing it would flip the engine's own message to the "no baseline" wording, and you're right that a fixture may not invent that string — `message` is authoritative and ours to generate.
2. It's the more *interesting* fixture. A `0` baseline exercises the divide-by-zero guard; a small non-zero one exercises the ratio path, which is the branch that actually runs in production.

That it didn't match FHIR Transform's measured `0` either — so it's inherited rather than introduced — is the right thing to have noticed and recorded.

## Both `stalled_host` findings being on the removed host is the catch of this PR

That's the same trap @kskubach avoided with `f-1038`/`slow_processing` in #15, and now the third instance of one pattern: **the samples carry roughly one finding per type, so any host deletion risks zeroing a type's coverage entirely.** Straight row-deletion would have left `stalled_host` unrepresented across the whole set.

Same for `error-storm` being the only `elevated_error_rate` with a non-zero baseline. Preserving `4.2 / 1.1 = 3.8x` keeps the only fixture exercising the ratio path rather than the clean-baseline path — exactly the reasoning above about `0.02`, applied consistently.

## One interaction worth knowing about — no conflict, but adjacent

#20 (open) tunes my absolute floors down, because they were dominating the multiplier on every host: `slow_processing` `1.0 → 0.2s`, `growing_queue_wait` `1.0 → 0.15s`, Cloud API override `2.0 → 0.3s`. A 10× slowdown was producing **no finding at all**.

I checked all four of your duration findings against the new floors:

```
queue-buildup    growing_queue_wait  Cloud API   1.84  vs floor 0.15   OK
slow-processing  slow_processing     Lab Router  1.86  vs floor 0.2    OK
slow-processing  growing_queue_wait  Lab Router  1.24  vs floor 0.15   OK
system-alert     slow_processing     Cloud API   2.40  vs floor 0.3    OK
```

All clear comfortably, so **#20 and #21 don't conflict in either merge order.** Worth knowing that your `validate-fixture-claims.mjs` reads `thresholds.json` directly, so it'll pick the new floors up automatically — which is the drift-detection you designed it for, working as intended.

## The three stale README rows

Correcting adjacent-and-false is the right call, and flagging that it's outside the stated scope is the right way to do it. `throughput-drop` still claiming `1.8 msg/sec` after the issue #6 rescale is the kind of thing that survives indefinitely because nobody owns re-reading a table.

Nothing blocking. Merge whenever — the demo path is fully consistent across all three areas after this.

PR #22, opened 2026-08-11
fix(dashboard): sweep for findings the engine would emit but a fixture omits

tanifgit:
**Stacked on #21** — based on `devC/drop-fhir-transform`, so the diff here is only the new work. GitHub will retarget it to `main` when #21 merges.

This is the follow-up I committed to in #21 and in my review on #20.

## The gap

`validate-fixture-claims.mjs` walked the findings and asked *"would the engine emit this?"* It never walked the hosts and asked the reverse: **"is there a finding the engine would emit that is missing?"**

A fixture can therefore understate the product — the host grid shows a host quietly breaching a rule while the findings list says nothing about it. That is the same class as the two defects @Ari-Glikman found upstream (`npm test --if-present` reporting green on nothing, and `slow_processing`'s unreachable warning band): a check that gets quietly weaker instead of failing. Sitting, in this case, inside the tool I wrote to catch exactly that.

**It is not a consequence of #20.** `dead-host`'s `41.6s` queue wait was already far above the *old* `1.0s` floor and slipped through anyway. The retune only made me look.

## Four omissions found

```
dead-host    Cloud API   queue_buildup       154 queued, no baseline queue
dead-host    Cloud API   throughput_drop     0 against a 0.4 baseline
dead-host    Lab Router  growing_queue_wait  41.60s
error-storm  Cloud API   queue_buildup       61 queued, no baseline queue
```

Two of those are the interesting kind, and I had missed both by hand: **a `Disabled` Cloud API holding 154 messages produces `dead_host`, `queue_buildup` *and* `throughput_drop`**, not just the first. I had only been looking at the duration rules when I reported four rows on #20 — the tool found strictly more than I did, which is rather the point of writing it.

`dead-host` now carries **7 findings from one root cause**. That is what the engine genuinely does, and I decided against trimming it: correlating findings to a single cause is AI Detective's job, not Health Scan's (root `CLAUDE.md` §2). It also makes the demo headline more honest about what an operator actually sees.

## How the sweep works

Mirrors each rule's branches from `detect/rules/index.ts` in the same order, including the early returns — notably that `stalled_host` defers to `dead_host` ("one condition, one finding"), and that a zero baseline makes the ratio infinite so the **floor becomes the entire gate**.

**Baselines are derived from `scenario-healthy.json`** rather than hardcoded. That file is already declared in `fixtures/README.md` as the measured steady state every other scenario departs from, so it is the natural source. A table of host names inside the tool would be a fourth copy of a fact that already went stale once — the same duplication @Ari-Glikman removed in #19.

**It covers six of the eight rules, and prints which two it cannot on every run:**

```
      no host breaches a rule silently, across 6 of the 8 rules
note  elevated_error_rate and system_alert are NOT swept — neither is derivable
      from a single host row, so a fixture could omit one and this would not say so
```

`elevated_error_rate` compares errors-per-**minute** derived across consecutive polls, and `host.errored` is a cumulative counter — one snapshot cannot yield a rate. `system_alert` comes from the proxy's alerts payload, which a scenario file does not carry at all. Letting "no missing findings" imply a sweep of all eight would be precisely the failure this check exists to prevent.

## `baselineWarming`

`scenario-baseline-warming.json` gains `"baselineWarming": true`. Without it, its two hosts sitting at `0 msg/sec` read as unreported `throughput_drop`s — they are not, because comparative rules return early on a null baseline.

The flag is **cross-checked rather than trusted**: setting it while any finding carries a non-null baseline is itself a failure. A wrong flag would silence the comparative sweep and hide real omissions, so it needed to be unable to lie.

## One thing for @Ari-Glikman, non-blocking

`queueBuildup` appends `with no baseline queue` for the infinite-ratio case ([index.ts:99-101](services/detection-engine/src/detect/rules/index.ts#L99-L101)), but `durationRule` appends an empty string ([index.ts:174](services/detection-engine/src/detect/rules/index.ts#L174)). So the same situation reads two ways, and in `dead-host` they now land two rows apart on screen:

```
Critical  Queue buildup       · Lab Router   Queue depth 212 with no baseline queue
Critical  Growing queue wait  · Lab Router   Average queue wait 41.60s
```

The second doesn't say *why* there's no ratio. Entirely your call and I have not touched it — I used the engine's real string rather than the nicer one, because a fixture may not invent wording the engine doesn't produce.

## Verification

```
ok    (all 8 scenarios)
      no host breaches a rule silently, across 6 of the 8 rules
note  elevated_error_rate and system_alert are NOT swept — neither is derivable
      from a single host row, so a fixture could omit one and this would not say so

$ npx tsc --noEmit
(clean)

$ npm run build
dist/index.html  200.47 kB | gzip: 61.51 kB

25 findings across 8 scenarios; all 8 finding types still covered
?scenario=dead-host  all 7 render, 6 critical + 1 warning,
                     Lab Router 4 active findings, Cloud API 3
```

Rendered from `dist/index.html` over `file://` rather than the dev server, so that also re-checks the §6 static fallback.

**Evidence the check works** — before the fixture fixes, on the unmodified set:

```
FAIL  scenario-dead-host.json — Lab Router breaches a rule with no finding to show it
        growing_queue_wait would fire, but no finding of that type names this host
FAIL  scenario-dead-host.json — Cloud API breaches a rule with no finding to show it
        queue_buildup would fire, but no finding of that type names this host
        throughput_drop would fire, but no finding of that type names this host
FAIL  scenario-error-storm.json — Cloud API breaches a rule with no finding to show it
        queue_buildup would fire, but no finding of that type names this host
```

## Merge ordering — a correction to my #20 review

On #20 I said *"merge order between us doesn't matter."* **That was true of the forward check and is not true of this one.** Correcting it here as well as on #20.

Three further omissions appear only once #20's floors land:

```
error-storm     Cloud API   growing_queue_wait  0.84s vs 0.02  ratio 42
queue-buildup   Lab Router  growing_queue_wait  0.62s, no baseline
system-alert    Cloud API   growing_queue_wait  0.21s vs 0.02  ratio 11
```

**And they cannot be pre-fixed.** Under today's `1.0s` floor the emittable check correctly rejects the very same findings:

```
FAIL  scenario-queue-buildup.json — f-1044 growing_queue_wait @ Lab Router
        0.62s is below absoluteFloorSeconds 1 — would not fire
```

So the two directions are in genuine tension across that boundary: today those findings must **not** exist, and after #20 they **must**. Consequences:

- **#20 merging on its own turns the `dashboard` CI job red on `main`**, because it runs `validate:fixtures`, which now includes this sweep. That is the drift detector doing its job, not a defect — but it shouldn't land unannounced.
- I'd suggest merging **#21 → this → #20**, and I'll have the three-row follow-up ready to go straight after #20 so the window is minutes, not hours. Happy to sequence it however you two prefer — just flagging that there *is* now a sequence.

## Checklist

- [x] Stayed inside my own area — `apps/dashboard/**` only
- [x] No `contracts/` file edited
- [x] Ran my area's typecheck / build / validators, output above
- [x] No invented data — every added finding uses values the fixture's own host row already asserted, and the engine's own message strings
- [x] No new dependency

## Scope

- [x] Adds no capability from a later module — this is fixture tooling only

### DIFF
--- a/apps/dashboard/fixtures/README.md
+++ b/apps/dashboard/fixtures/README.md
@@ -68,6 +68,30 @@ engine would never have produced. Reproducing the engine's logic means it drifts
 Dev B retunes — which is the point, since a fixture that silently stops matching the
 engine is exactly what is being guarded against.
 
+It also sweeps the **reverse** direction, which is a different mistake: a host row
+the engine *would* flag, with no finding for it. A fixture like that understates the
+product — the grid shows a host quietly breaching a rule while the findings list says
+nothing. That gap found four omissions in this set, two of them on `Cloud API` in
+`scenario-dead-host`, where a `Disabled` host with 154 queued produces `dead_host`,
+`queue_buildup` **and** `throughput_drop` rather than just the first.
+
+Two things about that sweep worth knowing:
+
+- **Baselines come from `scenario-healthy.json`**, since that file is already declared
+  above as the measured steady state everything else departs from. Deriving them beats
+  a table of host names in the tool — a duplicated host list is what went stale when
+  `FHIR Transform` was removed.
+- **It covers six of the eight rules.** `elevated_error_rate` compares
+  errors-per-*minute* derived across consecutive polls, and one host row cannot yield a
+  rate; `system_alert` comes from the proxy's alerts payload, which a scenario file does
+  not carry. The tool prints both exclusions on every run rather than letting "no
+  missing findings" imply a sweep it did not do.
+
+`scenario-baseline-warming.json` carries `"baselineWarming": true`, which is what
+silences the comparative half of the sweep for it — without the flag, two hosts at
+`0 msg/sec` read as unreported `throughput_drop`s. The flag is cross-checked: if it is
+set while any finding carries a non-null baseline, that fails too.
+
 Both checks **skip cleanly** when the thing they need is absent: `contracts/` and
 `services/detection-engine/` each land via their own PR, so neither may fail a
 dashboard branch that simply predates them. That leniency is wrong for CI, where a
@@ -97,6 +121,14 @@ Two notes on reading these, both learned from checking them against the real eng
   `avgQueueingTime: 0` on two of the three hosts, so a real buildup has no ratio
   to quote. The engine treats its absolute floor as the whole test there and says
   "with no baseline queue" rather than "∞x baseline".
+- **The two rule families grade a zero baseline differently**, so a fixture cannot
+  assume "no baseline means critical". `queue_buildup` still reaches straight for
+  `critical`, which is fair where its floor is 50 queued messages. The two duration
+  rules instead judge absolute magnitude — `criticalFloorMultiple` (4.0) × the floor
+  earns `critical`, anything merely above the floor is a `warning`. That landed in
+  Dev B's `fbb34da` after the review on #20, because 150 ms of queue wait reaching
+  for `critical` was not a claim worth making. `scenario-queue-buildup`'s Lab Router
+  finding sits 20 ms above that bound and is the set's most fragile severity.
 - **A rolling baseline is not the steady-state table.** `scenario-slow-processing`
   gives Lab Router a `growing_queue_wait` baseline of `0.02` even though the
   measured table has its `avgQueueingTime` at `0`. That is deliberate: the engine's
--- a/apps/dashboard/fixtures/scenario-baseline-warming.json
+++ b/apps/dashboard/fixtures/scenario-baseline-warming.json
@@ -1,7 +1,8 @@
 {
   "id": "baseline-warming",
   "label": "Baseline warming up",
-  "note": "The engine has just started, so there is no rolling baseline and comparisons render as absent rather than as zero. Only the absolute rules can fire in this state: every comparative rule returns nothing while its baseline is null, so a warm-up fixture showing queue_buildup or slow_processing would be unproducible by the real engine.",
+  "baselineWarming": true,
+  "note": "The engine has just started, so there is no rolling baseline and comparisons render as absent rather than as zero. Only the absolute rules can fire in this state: every comparative rule returns nothing while its baseline is null, so a warm-up fixture showing queue_buildup or slow_processing would be unproducible by the real engine. The baselineWarming flag above is what tells validate-fixture-claims.mjs to silence its comparative sweep here — without it, Lab Router and Cloud API sitting at 0 msg/sec read as unreported throughput_drops.",
   "hosts": [
     {
       "host": "EMR Source",
--- a/apps/dashboard/fixtures/scenario-dead-host.json
+++ b/apps/dashboard/fixtures/scenario-dead-host.json
@@ -49,6 +49,26 @@
       "message": "Cloud API is Disabled with 154 message(s) queued"
     },
     {
+      "id": "f-1107",
+      "host": "Cloud API",
+      "type": "throughput_drop",
+      "severity": "critical",
+      "currentValue": 0,
+      "baselineValue": 0.4,
+      "detectedSecondsAgo": 90,
+      "message": "Throughput 0.0 msg/sec is 100% below baseline"
+    },
+    {
+      "id": "f-1106",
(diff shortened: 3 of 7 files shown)


tanifgit:
**Updated — #20 merged while this was in review, so the ordering constraint I flagged above is gone.** Rebased onto current `main` (`3ed0805`) and fixed forward. Nothing to sequence any more; this is mergeable on its own once #21 is in.

Three things changed since the original body.

## The three pending omissions are now fixed here

They were unfixable before #20 merged and are trivially fixable after, so they moved into this PR rather than a follow-up. **Seven omissions total, not four:**

```
dead-host      Cloud API   queue_buildup       154 queued, no baseline queue
dead-host      Cloud API   throughput_drop     0 against a 0.4 baseline
dead-host      Lab Router  growing_queue_wait  41.60s
error-storm    Cloud API   queue_buildup       61 queued, no baseline queue
error-storm    Cloud API   growing_queue_wait  840ms is 42x baseline
queue-buildup  Lab Router  growing_queue_wait  620ms
system-alert   Cloud API   growing_queue_wait  210ms is 11x baseline
```

28 findings across the 8 scenarios now, 21 critical / 6 warning / 1 info, all eight types still covered.

## I had to mirror @Ari-Glikman's `fbb34da` — and it caught my own tool drifting

This is the part worth reviewing, because it is the failure this PR is about, occurring *in the PR about that failure*.

`fbb34da` (the zero-baseline fix #20 picked up from my review) changed how the duration rules grade an infinite ratio. My `checkEmittable` had that branch as:

```js
const expectRatio = () => (Number.isFinite(ratio) ? fromRatio(...) : 'critical');
```

Hardcoded `critical` — which is exactly what Ari removed from the engine. So the moment `fbb34da` landed, my validator would have **silently accepted a wrong severity** on any zero-baseline duration finding. The tool whose entire premise is "reproducing the engine's logic means this drifts if Dev B retunes, which is the point" had drifted on its first real retune.

Now mirrored properly, and the two families genuinely differ so it can't be one shared expression:

- **duration rules** — `criticalFloorMultiple` (4.0) × floor earns `critical`, merely above the floor is a `warning`
- **`queue_buildup`** — still hardcoded `critical`, which I think is right where the duration version wasn't: a floor of 50 queued messages is a meaningful absolute quantity, 150 ms of queue wait is not

**Proven in both directions by mutating a fixture, not assumed:**

```
0.30s zero-baseline claimed critical  ->  FAIL  severity "critical", engine gives "warning"
same value corrected to warning       ->  passes
```

## One fragile number, deliberately documented

`queue-buildup`'s new Lab Router finding is `620ms` against a critical bound of `600ms` (0.15 s floor × 4.0). **20 ms of margin** — the closest severity in the set to flipping. I left the value alone rather than padding it, since 0.62 is what the host row already asserted, and noted it in the scenario's own `note` plus `fixtures/README.md`. The validator is what catches that flip, which is the point of it existing.

## Still open for @Ari-Glikman, still non-blocking

The message asymmetry from my earlier comment stands: `queueBuildup` appends `with no baseline queue`, `durationRule` appends nothing. In `dead-host` they now land two rows apart:

```
Critical  Queue buildup       · Lab Router   Queue depth 212 with no baseline queue
Critical  Growing queue wait  · Lab Router   Average queue wait 41.60s
```

Your call entirely — my fixtures follow the engine, so if `durationRule` starts saying something equivalent I'll update rather than the reverse.

Also: I saw #23. `test/silence.test.ts` asserting what the engine must *not* say is the same idea from the other side, and you got there first — you called that symmetry in your comment on #20 before I'd finished this. Worth us both knowing the pair exists: yours proves the engine stays quiet when it should, mine proves the fixtures don't stay quiet when it wouldn't.

## Re-verified after the rebase

```
      no host breaches a rule silently, across 6 of the 8 rules
note  elevated_error_rate and system_alert are NOT swept — neither is derivable
      from a single host row, so a fixture could omit one and this would not say so

$ npx tsc --noEmit      (clean)
$ npm run build         dist/index.html  201.35 kB | gzip: 61.71 kB

28 findings, 21 critical / 6 warning / 1 info, all 8 types covered
```

Issue #25, opened 2026-08-11
Thresholds are hardwired to LABDEMO's scale: a 10x regression on a quieter production produces zero findings


tanifgit:
> ## Status — 3 of 4 items closed; **1 remains, and it is the one in the title**
>
> | Item | Area | Status |
> |---|---|---|
> | **Absolute floors encode LABDEMO's scale** | detection-engine | **OPEN — needs an ADR, not a patch. The only reason this issue is open.** |
> | `hostOverrides` fails silently on a foreign production | detection-engine | fixed, #27 |
> | Loading skeletons sized to a hardcoded host count | dashboard | fixed, #26 |
> | Docs asserting *how many* hosts exist rather than *never invent one* | dashboard | fixed, #26 |
>
> **The remaining item is a class-level bug whose instance is already fixed.** @Ari-Glikman's framing of it:
> *"this is my #20 defect recurring one level up [...] I fixed the instance and not the class."*
>
> **Agreed direction**, from @Ari-Glikman and @tanifgit independently on #27:
> `effectiveFloor = max(k × baseline, tinyAbsolute)` — preserves what the floor is for
> (stopping `1 → 5` reading as "5x baseline" on trivia) while making the binding constraint
> scale-relative. Wants the ADR and ideally a second measured production to test against.
>
> **Not blocking MVP 1.** Multi-production support is out of scope (root `CLAUDE.md` §2), and
> the per-(host, metric) rolling baselines — the part that matters most — were already correct
> and are now demonstrably production-agnostic.
>
> The evidence below stands unchanged, and was **reproduced independently** by @Ari-Glikman on
> their own synthetic harness: same `0 findings after a 10x regression on every metric`.

---

## Summary

Health Scan's **host discovery and baselines are genuinely dynamic** — I proved it against two invented productions. But its **absolute floors are hardwired to LABDEMO's magnitudes**, and on a quieter production that silently disables three of the eight rules. A 10× regression on every comparative metric produced **zero findings**.

Raised after a question worth recording: *if the design is dynamic, why did removing one component from LABDEMO force changes in two other developers' areas?* The answer turned out to be "it didn't, in the product" — but chasing it down surfaced two things that genuinely don't travel.

## What the FHIR Transform removal actually cost

26 files across #15, #18, #19 and #21. Classified:

```
RUNTIME     1 file    apps/dashboard/src/components/HostGrid.tsx   <- loading-skeleton count, 4 -> 3
[...]

PR #27, opened 2026-08-11
feat(detection-engine): warn when a hostOverrides key matches no observed host

Ari-Glikman:
The cheap half of #25 — the part @tanifgit scoped as *"a latent bug at any production count"*.

## Reproduced first

`hostOverrides` is keyed by a literal host name, so on a different production every override goes inert **silently**:

```
override keys declared: Cloud API
configFor(slow_processing,"Cloud API")   floor = 0.3   (applied)
configFor(slow_processing,"Risk Filter") floor = 0.2   (base config, override inert)
warning emitted: NONE
```

You're right that this is the same shape as three failures we've already fixed — `npm test --if-present` reporting green on nothing, `-c ajv-formats` quietly ignoring `format`, and the `@devA` CODEOWNERS placeholder requesting review from nobody. A config that weakens without complaining.

## Now

```
thresholds: hostOverrides["Cloud API"] matches no observed host, so its tuning is
inert. Observed: Risk Filter, Tick Feed
```

A **warning, not an error** — a host can be legitimately absent from a running production, and refusing to start would be worse than saying so. Emitted once per (config, host-set) rather than per poll, and reset on `reconfigure` so a corrected file gets re-checked. `DetectionEngine` now takes an optional log sink, which also makes it testable rather than eyeball-only.

**133/133 tests**, and I verified they fail without the fix rather than only that they pass with it — commenting out the warn call fails 2 of the 3 new engine tests.

## I reproduced your larger finding too, and it's real

Independently, on my own synthetic quiet production:

```
steady state findings: 0
findings after a 10x regression on EVERY metric: 0
```

Confirmed. And your framing is the useful part: **the multipliers travel, the floors don't.** `3.0x` means the same thing on any production; `0.2s` encodes LABDEMO's magnitude. My #20 retune fixed LABDEMO and moved the bug to everything quieter — which is the same defect recurring one level up, exactly as you say.

**Not addressed here, deliberately.** It's an ADR-level design question, the options interact with *why* the floors exist (stopping `1 → 5` firing as "5× baseline" on trivia), and multi-production support is out of MVP 1 per root `CLAUDE.md` §2. I'd rather leave your issue open with the analysis intact than half-fix it under time pressure.

If it's wanted before Day 5, the option I'd argue for is `floor = max(k × baseline, tiny_absolute)` — it keeps the floor's actual purpose while making it scale-relative, and `tiny_absolute` becomes a sanity bound rather than the binding constraint. But that deserves the ADR you suggested, not a number change in a PR.

## On your 26-file classification

That's the most useful artifact in #25 — one runtime file, and it was a loading-skeleton count. Worth pairing with the observation I reached from the CI side in #19: four copies of a fact and the copies drift. Yours quantifies it, which is better.

Entirely within `services/detection-engine/`. Refs #25

### DIFF
--- a/services/detection-engine/src/config/thresholds.ts
+++ b/services/detection-engine/src/config/thresholds.ts
@@ -268,6 +268,33 @@ export function configFor<K extends keyof RuleConfigs>(
   return { ...config.rules[rule], ...(override as object) } as RuleConfigs[K];
 }
 
+/** Override keys, excluding the `_comment` documentation entries. */
+export function declaredOverrideHosts(config: ThresholdConfig): string[] {
+  return Object.keys(config.hostOverrides).filter((key) => !key.startsWith('_'));
+}
+
+/**
+ * Override keys that match no host the engine has actually seen.
+ *
+ * `hostOverrides` is keyed by a literal host name, so pointing the engine at a different
+ * production makes every override inert — silently. The tuning simply stops applying and
+ * nothing says so. Raised by Dev C in #25, and it is the same shape as three failures
+ * this project has already fixed: `npm test --if-present` reporting green on nothing,
+ * `-c ajv-formats` quietly ignoring `format`, and the `@devA` CODEOWNERS placeholder
+ * requesting review from nobody. A config that weakens without complaining.
+ *
+ * Returns the inert keys so the caller can log them. Deliberately not an error: an
+ * override for a host that is temporarily absent from a running production is
+ * legitimate, and refusing to start would be worse than saying so.
+ */
+export function inertOverrideHosts(
+  config: ThresholdConfig,
+  observedHosts: Iterable<string>,
+): string[] {
+  const seen = new Set(observedHosts);
+  return declaredOverrideHosts(config).filter((host) => !seen.has(host));
+}
+
 /**
  * Live config holder. Reads once at construction, then re-reads on file change.
  * A bad file is logged and ignored — `current` keeps the last good value.
--- a/services/detection-engine/src/detect/engine.ts
+++ b/services/detection-engine/src/detect/engine.ts
@@ -10,7 +10,7 @@
 
 import { BaselineStore } from '../baseline/window.ts';
 import type { ThresholdConfig } from '../config/thresholds.ts';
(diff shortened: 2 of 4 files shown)


tanifgit (approved):
Approving. **Verified the behaviour, not just the test count** — a warning that fires in a unit test but not in a running engine would be the same class of failure this PR fixes.

Ran my #25 foreign-production harness against your branch with a log sink attached:

```
FOREIGN production (no Cloud API host):
  warnings emitted: 1
    thresholds: hostOverrides["Cloud API"] matches no observed host, so its tuning is
    inert. Observed: Risk Filter, Tick Feed

LABDEMO-shaped (Cloud API present):
  warnings emitted: 0

8 polls, same host set -> warnings: 1   (not 8)
```

All three properties hold: it fires when the override is inert, stays quiet when it applies, and doesn't repeat per poll. Plus:

```
npx tsc --noEmit   clean
npm test           tests 133  pass 133  fail 0   (suites 38)
```

**Warning rather than error is the right call**, and for the reason you give — a host can be legitimately absent from a running production, so refusing to start would trade a silent failure for a worse loud one. The `#warnedOverrides` key including the sorted host set is the detail I'd have got wrong: it means a host appearing or disappearing re-triggers the check rather than the first poll's verdict sticking for the process lifetime.

**Taking the log sink as a constructor parameter is the better half of this.** It's what let me verify the above from outside, and it turned "eyeball the console" into an assertion. Worth remembering as a pattern — the reason my own validator's blind spot survived so long was that nothing could observe what it *didn't* say.

## On the part you deliberately left

Agreed, and I'd have pushed back if you'd patched it. `floor = max(k × baseline, tiny_absolute)` is where I'd land too, and it's the right shape because it keeps the floor's actual purpose — stopping `1 → 5` reading as "5× baseline" — while making the binding constraint scale-relative. But it changes what every threshold *means*, so it wants the ADR and a measured second production to test against, not a number swap on Day 5.

I've left #25 open with the analysis intact, and re-scoped its remaining items to just that one now this has landed.

Thanks for reproducing the quiet-production result independently rather than taking my word for it — two people getting `0 findings after a 10x regression` from different harnesses is a much stronger record than one.

Issue #16, opened 2026-08-10
Dev A open items: ADR 0004 tick, host-name decision, contracts/proxy-*


tanifgit:
Welcome @kskubach. Your open items are currently spread across #7, #10, #11, #12, #13, #14 and ADR 0004, so this is one place to work from. **Nothing here is a new request** — every item already exists in a thread, this just collects them with the exact file, line and command.

I'm Dev C (dashboard), so items 1–3 are things I'm downstream of. Items 4–5 are yours alone and I've included them only so the list is complete.

---

## 1. Tick your box on ADR 0004 — one character, unblocks the ADR

[`docs/decisions/0004-mock-first-contracts.md`](docs/decisions/0004-mock-first-contracts.md), `## Confirmation` section, **line 99 on `main` today** (line 120 once #7 merges):

```diff
-- [ ] **Dev A** — metrics proxy, `iris/**`
+- [x] **Dev A** — metrics proxy, `iris/**` (confirmed: parser and poller built against
+      `fixtures/metrics.txt` before any live capture existed)
```

**Also update the status line at the top** in the same change, or the ADR contradicts its own checklist:

```diff
-- **Status:** **proposed — needs explicit confirmation from all three developers**
+- **Status:** accepted
```

Dev B and I have both ticked. Yours is the last one, so this is what moves 0004 from `proposed` → `accepted`.

**You have a genuine confirmation to make, not a formality.** You built the parser against a hand-written `fixtures/metrics.txt` before a real capture existed — that's mock-first followed *and* the ADR's own accepted cost landing exactly as predicted: *"a hand-written sample encodes the author's belief about the API, so mocking against it proves only self-consistency."* #10 is that cost coming due. Worth saying so in the tick; it's the strongest evidence in the document.

**Process:** must be a PR, not a push — [`docs/decisions/README.md:24`](docs/decisions/README.md#L24): *"`docs/decisions/` is shared and PR-gated."* And **wait for #7 to merge first**, or you'll conflict on those same lines.

## 2. Sequencing warning: merge #11 before your ADR PR

CODEOWNERS on `main` still carries the `@devA` placeholder. GitHub silently ignores an unknown owner, and worse — **an unresolvable owner invalidates the entire rule line**, including valid co-owners. So `/docs/decisions/` currently requests *nobody*, and your PR would look ready to self-merge.

[...]

PR #28, opened 2026-08-12
docs(decisions): ADR 0004 accepted; Dev A's box checked on the evidence

Ari-Glikman:
**Dev A has moved to other work, and Dev B is taking over their outstanding tasks.** This is the first of those — item 1 on @tanifgit's #16 list, and the last thing keeping ADR 0004 at `proposed`.

## What changed

```
- **Status:** proposed — needs explicit confirmation from all three developers
+ **Status:** accepted (2026-08-12)

- [ ] **Dev A** — metrics proxy, iris/**
+ [x] **Dev A** — metrics proxy, iris/** (confirmed on the evidence, not by Dev A: ...)
```

## The thing worth reviewing

I did **not** tick this as though Dev A said it. @tanifgit was right on #16 that filling in someone else's confirmation is the thing to avoid, and I agreed on #4 when I left it blank in the first place.

So the box is checked **on the observable evidence** — the parser and poller were built entirely against a hand-written `fixtures/metrics.txt` before any real capture existed, which is mock-first followed — and the entry names who checked it and why. The ADR body now carries a *"How it actually resolved"* paragraph stating plainly that this is weaker than a first-person confirmation.

That felt like the honest middle. Leaving 0004 `proposed` forever would also be a false record, since the practice demonstrably held on all three sides — but a silent tick would have been exactly what #16 objected to.

Push back if you'd rather it stayed `proposed` with a note instead. That argument is real and I'd defer to you on it, since you raised the principle.

## Also fixes a contradiction

`docs/decisions/README.md` still said **"needs all three to confirm"** in its index and *"it stays `proposed`"* in its note. Left alone, the index would contradict the ADR it links to — the same duplicated-fact-drifting problem that has now cost this project four separate fixes: the host list in four places, the CI host count, the `@devA` CODEOWNERS placeholder, and this.

## On process

`docs/decisions/README.md` says a **decided** ADR is never rewritten, only superseded. 0004 was still `proposed` when edited, so amending is in bounds — the same reasoning you applied and explicitly asked about in #7.

Refs #16, #4, #7

### DIFF
--- a/docs/decisions/0004-mock-first-contracts.md
+++ b/docs/decisions/0004-mock-first-contracts.md
@@ -1,6 +1,6 @@
 # ADR 0004 — Mock-first contracts
 
-- **Status:** **proposed — needs explicit confirmation from all three developers**
+- **Status:** accepted (2026-08-12)
 - **Date:** 2026-08-06
 - **Deciders:** Dev A, Dev B, Dev C
 - **Drafted by:** Dev B — see the note below on why this one is different
@@ -13,9 +13,16 @@ implement them. **This one is not a technical choice — it is an agreement betw
 and §7.4 asks all three to *confirm* it rather than for one of us to decide it.
 
 It is drafted here because it was identified (issue #2) as **load-bearing without being recorded**:
-Dev C relied on it from Day 1 while building against a transcribed schema. Writing it down is
-overdue. But Dev A and Dev C should each confirm rather than inherit my summary — this stays
-`proposed` until they do.
+Dev C relied on it from Day 1 while building against a transcribed schema. Writing it down was
+overdue. But Dev A and Dev C should each confirm rather than inherit my summary — so it stayed
+`proposed` until they did.
+
+**How it actually resolved.** Dev C confirmed their own box in PR #7 and corrected this ADR's
+evidence section while doing so. Dev A never ticked theirs: they moved to other work on 2026-08-12
+and Dev B took over their outstanding tasks. Their box is now checked **on the observable evidence**
+— a parser and poller built entirely against a hand-written fixture before any real capture existed
+— and labelled as such, rather than as a confirmation made in their name. That is weaker than a
+first-person statement, and the ADR says so instead of hiding the difference.
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


tanifgit (approved):
**Approved.** I raised the principle on #16, so I'll be explicit: this is the right call, and the way it's recorded is better than what I asked for.

I verified the load-bearing claim rather than accepting it, and **the evidence is considerably stronger than the PR states.**

## The mock-first claim is true, and the proof is that the fixture was wrong

The ADR says the parser and poller were "built entirely against a hand-written `fixtures/metrics.txt` before any real capture existed." Checked against history:

```
2026-08-10  71c38d0  kskubach  "first 2 steps DevA"
            + services/metrics-proxy/src/parser.js      (195 lines)
            + services/metrics-proxy/src/poller.js      (125 lines)
            + services/metrics-proxy/fixtures/metrics.txt (59 lines)

2026-08-11  2e164ac  kskubach  feat(metrics-proxy): parse real IRIS metrics
2026-08-11  694e902  kskubach  fix(metrics-proxy): support a base path
            + fixtures/metrics-live-capture.txt
            + fixtures/alerts-live-capture.json
```

Parser, poller and fixture land in one commit; the real captures arrive **a day later**. But the decisive part is the content of that hand-written fixture:

```
iris_interop_hosts{name="EMRSource",type="BS",status="Active"} 1
iris_interop_queued{name="LabRouter"} 3
```

Four independent wrong guesses in two lines — `name` not `host`, `EMRSource` not `EMR Source`, `BS`/`BP`/`BO` not `service`/`process`/`operation`, `status="Active"` which #30 Q7 confirms **does not exist in the IRIS enum**, and per-host `iris_interop_queued` labels which #12 established IRIS never emits at all.

**A fixture that guesses wrong five times cannot have been derived from real output.** That's a much harder-to-fake basis for the tick than "built against a fixture," and I'd put it in the ADR — the current wording undersells the evidence it rests on.

It also reframes the last two days: the entire #10 → #13 → #30 → #32 sequence is the *cost* of mock-first landing exactly where ADR 0004 said it would. The ADR predicting its own bill and then getting accepted on the evidence of that bill is the honest version of this record.

## On the principle — you got the balance right

My objection on #16 was to filling in someone else's confirmation. A ticked box that reads `(confirmed on the evidence, not by Dev A: ...)` isn't that. It states who checked it, on what basis, and that it's weaker than a first-person statement. Someone reading this in six months learns more than they would from either a silent tick or a permanent `proposed`.

I'd have accepted `[ ]` + accepted status too, but I'm not going to relitigate it — the labelled box loses less information, and the "How it actually resolved" paragraph is the part that actually carries the meaning.

## Verified, no drift

I swept for anywhere else asserting 0004's *status*, since duplicated status is exactly what bit us four times:

```
.github/workflows/ci.yml:81      "Serves with no proxy running (ADR 0004)"
services/detection-engine/{CLAUDE.md,README.md,src/index.ts,src/proxy/mockClient.ts}
apps/dashboard/CLAUDE.md:165
docs/decisions/0001-detection-engine-location.md:64
CONTRIBUTING.md:165,184
```

All nine reference 0004's *content*, none its status. **Both places that stated a status are in this diff.** No stale third copy.

## Two non-blocking notes

**1. The README now duplicates the ADR's explanation almost verbatim** — the same four-line account of how it resolved, in both `README.md` and the ADR body. You name duplicated-facts-drifting as the defect that has cost this project four fixes, in this PR's own description. This is a fifth instance, freshly created. The index only needs the status plus a link; the *why* belongs in one file. Not worth another round-trip, but worth collapsing next time you're in there.

**2. 0004 is now the only accepted ADR, and it's the least technical one.** 0001–0003 all still read `proposed` while being fully implemented and load-bearing — the engine runs outside IRIS, baselines are a 30-minute in-memory window, thresholds are hot-reloaded JSON. All three shipped and none is contested. I'll file that separately rather than widen this PR; flagging it because this change makes the inconsistency newly visible.

Docs-only, all four CI jobs green, nothing outside `docs/decisions/`.

PR #29, opened 2026-08-12
fix(iris): #17 — Verify() false negative, two Storage blocks, and triggers pointing at a deleted host

### LINKED ISSUE #17: iris: EnableMetrics.Verify() reports a false negative on a prefixed instance; two stale Storage blocks; docs describe a deleted host
Four items from the closed #14 that did not land on `main`. Recording them here because they only exist on `devA/labdemo-pipeline-simplify`, which will be deleted.

Owner: Dev A. Filing rather than fixing immediately so the trigger work (#12-adjacent, Day 3) isn't blocked behind it.

## 1. `EnableMetrics.Verify()` gives a false negative — confirmed, not theoretical

`iris/setup/EnableMetrics.cls:116` and `:154` request `/api/monitor/metrics` and `/api/monitor/alerts` with no web-app prefix.

On an instance fronted by an external web server, the monitor API lives under the instance name. On the instance verified 2026-08-11 the working URL is:

```
http://localhost/iris4health_2024_1/api/monitor/metrics
```

So `Verify()` requests the bare path, receives the web server's 404, and prints **"FAILED to reach"** — i.e. it reports that metrics are not enabled on an instance where they are working correctly. That is the worst kind of diagnostic: it sends you to fix something that isn't broken.

This is the same defect I hit in the metrics proxy, which is now fixed there with an `IRIS_BASE_PATH` setting (`694e902`, #13). The fix here is the one from the closed branch:

```objectscript
set prefix = $get(^ProductionGuardian.Setup("WebAppPrefix"), "")
set sc = req.Get(prefix _ "/api/monitor/metrics")
```

Default `""` leaves a private-web-server instance (port 52773) unaffected.

## 2. `iris/labdemo/Data/PatientRecord.cls:158` closes its Storage block with `</Storage>`

UDL closes a Storage block with `}`. The XML end-tag form only belongs to the XML export format.

**Needs confirming before fixing:** if the class compiles as-is on the live instance then my reading is wrong and this should just be closed. Worth one `LoadDir`, because the failure mode is a class that silently doesn't load.

## 3. `iris/labdemo/Message/PatientDemographics.cls:46-98` has a Storage block at all

It extends `Ens.Request`, which isn't stored by that map. Same caveat as #2 — confirm against a compile first.

## 4. Docs describe a host that no longer exists

`PIDExtractProcess` was deleted in `2ca6764`, but it is still documented as a live pipeline stage:

- `iris/labdemo/Production.cls:4` — pipeline comment still shows the four-hop flow
- `iris/labdemo/RoutingRule.cls:2-4` — claims it routes ADT^A01 and ORU^R01 to `PIDExtractProcess`; ORU^R01 was also removed (`ce77c4b`)
- `iris/labdemo/README.md` — lines 6, 19, 20, 86, 87

The two that actually matter, because they are demo instructions for two of the eight finding triggers:

- `README.md:138` — "Queue buildup | Suspend PIDExtractProcess; run generator at fast rate"
- `README.md:140` — "Slow processing | Add a `hang 5` to PIDExtractProcess.OnRequest temporarily"

Both are now impossible as written. The real pipeline is `EMR Source → Lab Router → Cloud API`, and `Lab Router` applies the `HL7ToPID` DTL inline, so those triggers need re-pointing at `Lab Router` or `Cloud API`.

Note that `queue_buildup` cannot fire at all today regardless — per-host queue depth isn't in the Prometheus text (#12).

Ari-Glikman:
Closes #17.

**Dev A (`kskubach`) has moved to other work, so Dev B is taking over their outstanding tasks.** Authorized by the repo owner. All four items of #17 are addressed below. Two of the four needed confirming by compiling before touching them, and both turned out to be real — details and output inline.

Everything here is under `iris/**`. No file in `services/`, `apps/`, `contracts/` or `docs/` is touched.

---

## Item 1 — `EnableMetrics.Verify()` false negative — **fixed, and it was worse than filed**

#17 identified the missing web-app prefix. That is real, but it is one of **four** independent reasons `Verify()` misreported a healthy instance. All four verified on the live 2026.1 instance.

**a) The port was resolved to `0`.** `^%SYS("WebServer","Port")` is *defined and holds `0`* on an instance served by an external web server, and a `$get` default only applies to an **undefined** node:

```
resolved port via $get(...,52773) = [0]
is it defined? 1
ERROR #6059: Unable to open TCP/IP socket to server 127.0.0.1:0
```

So on this instance `Verify()` never reached the network at all — the prefix was moot. `WebPort()` now treats a non-positive value as absent and falls back to `52773`.

**b) No web-app prefix**, as filed. Added `WebAppPrefix()` from `^ProductionGuardian.Setup("WebAppPrefix")`, default `""`, normalised to leading-slash form. Added `WebHost()` for the same class of problem. These three deliberately mirror the metrics proxy's `IRIS_BASE_PATH` / `IRIS_HOST` / `IRIS_PORT` (694e902, #13) — same defect, same shape of fix, and `iris/labdemo/README.md` now shows the two settings side by side.

**c) A 404 was read as success.** `%Net.HttpRequest.Get()` returns `$$$OK` for any *answered* request; it only errors when the socket fails:

```
--- requesting a path we KNOW 404s ---
IsError(sc) = 0
HTTP StatusCode = 404
```

The old code therefore scanned the 404 HTML for metric names, reported all seven `MISSING`, and sent the operator to re-run `EnableSAMForNamespace()` when the real fault was the URL. `Verify()` now checks `StatusCode` and returns an error naming it. The same check was added to the alerts request, which had the same hole.

**d) One metric name was simply wrong.** The checks list carried `iris_last_activity`; the real family is `iris_interop_last_activity`. The bare name occurs **zero** times in the live output, and all 25 occurrences across the existing captures use the `iris_interop_` form — so that line reported `MISSING` on a perfectly healthy instance regardless of the URL.

### The working monitor URL on this instance

Contrary to the URL recorded in #17, **there is no prefix on our instance.** Measured:

```
=== port 80, no prefix ===                      HTTP 200  size=65650
=== port 80, /iris4health_2024_1 prefix ===     HTTP 404  size=180
=== port 52773, no prefix ===                   HTTP 000  (nothing listening)
```

`http://localhost:80/api/monitor/metrics` is the working URL, and it does carry interop data — 83 `iris_interop_*` lines across 14 families. The `/iris4health_2024_1` prefix from the issue **404s here**, which is exactly why this is a configurable global defaulting to `""` rather than a hard-coded prefix.

`Verify()` now prints the URL it requests, and warns on a 200 carrying no `iris_interop_*` at all — the port-80-without-prefix case the proxy's CLAUDE.md documents, where you reach `%SYS` and get real metrics with no interop families, indistinguishable from a stopped production.

**Verified end to end.** All seven families found, on the instance that previously reported failure:

```
=== Verifying interop metrics in /api/monitor/metrics ===
Requesting: http://host.docker.internal:80/api/monitor/metrics
  FOUND:   iris_interop_queued
  FOUND:   iris_interop_avg_processing_time
  FOUND:   iris_interop_messages_per_sec
  FOUND:   iris_interop_messages_errored
  FOUND:   iris_interop_avg_queueing_time
  FOUND:   iris_interop_last_activity
  FOUND:   iris_interop_hosts

Checking /api/monitor/alerts (consume-on-read - see comment) ...
  OK: /api/monitor/alerts returned JSON array

All expected metrics found. Metrics proxy can now be started.
```

And the negative case — a wrong prefix now fails loudly instead of misdiagnosing:

```
Requesting: http://host.docker.internal:80/iris4health_2024_1/api/monitor/metrics
  FAILED: HTTP 404 - the monitor API is not at this URL.
  Open /api/monitor/metrics in a browser, then set the globals to match it:
    set ^ProductionGuardian.Setup("WebAppPrefix") = "/iris4health_2024_1"
    ...
returned status IsError = 1
errtext: ERROR #5001: GET /iris4health_2024_1/api/monitor/metrics returned HTTP 404
```

Helper defaults and prefix normalisation were also exercised directly: `WebPort()` returns `52773` (not `0`), and `iris4health_2024_1/` / `/iris4health_2024_1` both normalise to `/iris4health_2024_1`, absent normalises to `""`.

One behavioural note worth a reviewer's eye: `Verify()` reads `/api/monitor/alerts`, which is **consume-on-read**, so it steals alerts the proxy would publish. I documented that in the code and README rather than changing it — removing the check felt out of scope for #17.

---

## Item 2 — `PatientRecord.cls` `</Storage>` — **confirmed defect, fixed**

#17 asked for this to be confirmed by compiling before fixing. It does **not** compile. The class as it stood on `main` could not load at all:

```
ERROR #5559: The class definition for class 'ProductionGuardian.LabDemo.Data.PatientRecord'
could not be parsed correctly, possibly due to non-matching {} or () characters or a
missing ; character or non-matched /* */ blocks.
  > ERROR #5030: An error occurred while compiling class '...PatientRecord'
```

Dev A's reading was right. UDL closes a Storage block with `}`; the XML end tag belongs only to the export format. Cross-checked against IRIS's own UDL for `Ens.MessageHeader`, which ends `<Type>%Storage.Persistent</Type>` / `}`. After the one-character fix:

```
Compiling class ProductionGuardian.LabDemo.Data.PatientRecord
Compiling table ProductionGuardian_LabDemo_Data.PatientRecord
Compilation finished successfully in 0.075s.
```

---

## Item 3 — `PatientDemographics.cls` has a Storage block — **confirmed defect, fixed**

Also a real defect, and a *different* one from item 2 — this class had **both** problems. Correcting only the `</Storage>` closer to isolate the second bug still fails:

```
ERROR #5477: Keyword signature error in PGTest.PD:Storage:Default, keyword
'DataLocation' must be '^Ens.MessageBodyD'
  > ERROR #5030: An error occurred while compiling class 'PGTest.PD'
```

`Ens.Request` inherits storage from `Ens.MessageBody`, which pins `DataLocation` to `^Ens.MessageBodyD`, so a Default map naming its own globals cannot compile. Removed the block entirely and left a comment saying why, so nobody re-adds it. Verified the same file with the block removed compiles:

```
Compiling class ProductionGuardian.LabDemo.Message.PatientDemographics
Compilation finished successfully in 0.106s.
```

---

## Item 4 — stale `PIDExtractProcess` references — **fixed**

Updated `Production.cls`'s pipeline comment, `RoutingRule.cls`'s header (it claimed it routed ADT^A01 **and** ORU^R01 to `PIDExtractProcess`; it routes ADT^A01 to `Cloud API`), and `README.md`'s diagram, inventory and flow narrative. `grep -rn "PIDExtractProcess\|ORU\^R01" iris/` is now empty.

### The two trigger instructions

Both named a host that does not exist, and **neither would have fired even with the name corrected** — that was the more important half:

- **Queue buildup** said "run generator at fast rate" with no target depth. `queue_buildup` has `absoluteFloor: 50`, so anything under 50 queued produces no finding at all. Now `Run(80, 0.2)` against a disabled `Cloud API`, with the floor stated so the next person understands why a small nudge does nothing.
- **Slow processing** said `hang 5`. The gate for `Cloud API` is its `0.3s` `hostOverride` floor, so `hang 1` clears it without backing the queue up behind it. The instruction now says to remove the hang afterwards — one left in place poisons every later baseline.

I filled in the other six triggers on the same basis, each with the arithmetic that clears its threshold, and added the two preconditions that otherwise make a working trigger look broken: comparative rules need ~2 minutes of healthy traffic for `minBaselineSamples: 12`, and `stalled_host` takes a full 5 minutes because `inactiveSeconds: 300` **and** `requiresQueued`. `thresholds.json` is cited as authoritative rather than restated — it is Dev B's file and the numbers will move.

**`queue_buildup` still cannot fire, and the README says so** rather than implying otherwise. Verified on the live instance that per-host depth is genuinely absent — the only line is:

```
```

One production-wide number, no `host` label. Until the proxy reads `EnumerateHostStatus` (#12, `PROXY-Q2`) this trigger builds a real queue the engine cannot see.

### Two extra defects found while verifying the triggers

Both block the pipeline the instructions exercise, so I fixed them here rather than filing and leaving the triggers untestable. Flagging them explicitly since neither is in #17:

**1. `Transform/HL7ToPID.cls:33` did not compile.** A stray `"` where the closing brace belongs: `source.{PID:7"),1,8)`. **The DTL could not compile, so no message could reach `Cloud API` at all** — the whole pipeline was broken on `main`, which makes this the most consequential find in the PR:

```
ERROR: HL7ToPID.cls(Transform+33) #1011: Invalid name :
  'target.DateOfBirth=$extract(source.{PID:7"),1,8)' : Offset:48
ERROR: HL7ToPID.cls(Transform+34) #1026: Invalid command : 'Catch'
ERROR: HL7ToPID.cls(Transform+71) #1043: QUIT argument not allowed : '}'
Detected 3 errors during compilation in 0.240s.
```

After `{PID:7}`:

```
Compiling class ProductionGuardian.LabDemo.Transform.HL7ToPID
Compilation finished successfully in 0.111s.
```

**2. `HL7Generator.cls:62` wrote the filename as the file's contents.** `do ..WriteFile(filename, filename)` instead of `msg`, so `RunContinuous()` produced files containing their own path — unparseable as HL7. `Run()` was correct, which is how it survived; `RunContinuous()` is what the throughput and queue triggers rely on.

Also aligned `iris/CLAUDE.md` §3, which still listed four unspaced components including `FHIRTransform`. It now points at `Production.cls` as the authoritative host list instead of keeping a second copy — a duplicated list is what went stale here in the first place — and records that the config names carry a space (`EMR Source`, not `EMRSource`), since the spaced form is what appears in the metrics and the proxy JSON.

---

## Verification and instance hygiene

All **eight** classes under `iris/labdemo/` plus `iris/setup/EnableMetrics.cls` compile clean.

Every compile ran in a **scratch namespace** (`AGENTICDEVELOPMENT`, and `USER` once to get a clean storage registration) — **never in `LABDEMO`**. Every probe class was deleted afterwards and the namespace verified empty of `ProductionGuardian*`; the test globals under `^ProductionGuardian.Setup` were killed. `LABDEMO.Production` was `Running` before and after, and no production setting was changed.

Note for reviewers: the live instance runs an older 4-host definition (including `FHIR Transform`) than `Production.cls` on `main`. That divergence is known and expected, and I did not try to reconcile it — so these classes are verified to **compile**, not verified running in LABDEMO.

Two things I could **not** verify, stated plainly rather than assumed:
- The triggers are verified against `thresholds.json` arithmetic and the live metric shape, but I did not induce all eight findings end to end — that needs the detection engine running against this production.
- `system_alert` relies on `AlertOnError`, which I confirmed exists as a real `Ens.Host` property, but I did not enable it on a host (that would be a production setting change).

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -36,16 +36,21 @@ Verification: `iris_interop_queued` and `iris_interop_avg_processing_time` must 
 
 ## 3. LABDEMO production
 
-Four components (match exactly — these names appear in the proxy JSON contract):
+Three application components. **The `<Item Name="...">` set in `iris/labdemo/Production.cls`
+is authoritative** — do not maintain a second copy of the host list here, because a
+duplicated list is exactly what went stale when `FHIR Transform` was removed.
 
-| Component name | Type | Role |
-|---|---|---|
-| `EMRSource` | Business Service (HL7 file/TCP) | Inbound HL7 v2 messages |
-| `LabRouter` | Business Process (routing rule) | Routes by message type |
-| `FHIRTransform` | Business Process (DTL) | Transforms HL7 → FHIR R4 |
-| `CloudAPI` | Business Operation (HTTP) | Forwards to downstream endpoint |
+Note the names carry a space (`EMR Source`, not `EMRSource`). The spaced form is what
+appears in `/api/monitor/metrics`, in the proxy JSON, and in the dashboard, so use it
+verbatim. Class names differ from config item names: the `Cloud API` item is
+`ProductionGuardian.LabDemo.Operation.PatientDemographicsOperation`.
 
-The synthetic HL7 generator (`iris/labdemo/HL7Generator.cls`) emits ADT^A01 and ORU^R01 messages on a configurable interval.
+`Ens.Activity.Operation.Local` is a fourth item, but it is framework plumbing for metrics
+and the findings API filters it out — it is not an application host.
+
+The synthetic HL7 generator (`iris/labdemo/HL7Generator.cls`) emits ADT^A01 messages on a
+configurable interval. That is the only type it emits and the only type the routing rule
+routes.
 
 ---
 
--- a/iris/labdemo/Data/PatientRecord.cls
+++ b/iris/labdemo/Data/PatientRecord.cls
@@ -155,6 +155,6 @@ Storage Default
 <IndexLocation>^PG.LabDemo.PatientRecordX</IndexLocation>
 <StreamLocation>^PG.LabDemo.PatientRecordS</StreamLocation>
 <Type>%Storage.Persistent</Type>
-</Storage>
+}
 
 }
--- a/iris/labdemo/HL7Generator.cls
+++ b/iris/labdemo/HL7Generator.cls
@@ -1,7 +1,8 @@
 /// Synthetic HL7 v2 message generator for the LABDEMO production.
 ///
-/// Writes ADT^A01 and ORU^R01 messages to /tmp/labdemo/hl7-in/ on a
-/// configurable interval so the production has continuous message flow.
+/// Writes ADT^A01 messages to /tmp/labdemo/hl7-in/ on a configurable interval
+/// so the production has continuous message flow. ADT^A01 is the only type
+/// emitted, and the only type the routing rule routes.
 ///
 /// Usage:
 ///   // Generate 20 messages with 2-second gaps (default):
@@ -17,7 +18,7 @@
 Class ProductionGuardian.LabDemo.HL7Generator
 {
 
-/// Default output directory — must match EMRSource FilePath setting.
+/// Default output directory — must match the EMR Source FilePath setting.
 Parameter DropDir = "/tmp/labdemo/hl7-in/";
 
 /// Generate `count` messages with `delaySecs` between each.
@@ -28,7 +29,6 @@ ClassMethod Run(count As %Integer = 20, delaySecs As %Double = 2) As %Status
 
     write "Generating ", count, " HL7 messages to ", ..#DropDir, " ...", !
     for i = 1:1:count {
-        // Alternate between ADT^A01 (admit) and ORU^R01 (lab result).
         set msg = ..BuildADTA01(i)
         set msgType = "ADT_A01"
         set filename = ..#DropDir _ msgType _ "_" _ $translate($horolog, ",", "_") _ "_" _ i _ ".hl7"
@@ -59,7 +59,7 @@ ClassMethod RunContinuous(delaySecs As %Double = 2)
         set msg = ..BuildADTA01(i)
         set msgType = "ADT_A01"
         set filename = ..#DropDir _ msgType _ "_" _ $translate($horolog, ",", "_") _ "_" _ i _ ".hl7"
-        do ..WriteFile(filename, filename)  // errors are silent in continuous mode
+        do ..WriteFile(filename, msg)  // errors are silent in continuous mode
         write $zdatetime($ztimestamp, 3, 1), "  ", msgType, !
         hang delaySecs
         set i = i + 1
--- a/iris/labdemo/Message/PatientDemographics.cls
+++ b/iris/labdemo/Message/PatientDemographics.cls
@@ -43,58 +43,9 @@ Property SourceMessageID As %String(MAXLEN = 64);
 /// ISO 8601 timestamp when this message was created (set by the DTL).
 Property ExtractedAt As %String(MAXLEN = 32);
 
-Storage Default
-{
-<Data name="PatientDemographicsDefaultData">
-<Value name="1">
-<Value>%%CLASSNAME</Value>
-</Value>
-<Value name="2">
-<Value>PatientID</Value>
-</Value>
-<Value name="3">
-<Value>LastName</Value>
-</Value>
-<Value name="4">
-<Value>FirstName</Value>
-</Value>
-<Value name="5">
-<Value>MiddleName</Value>
-</Value>
-<Value name="6">
-<Value>DateOfBirth</Value>
-</Value>
-<Value name="7">
-<Value>Sex</Value>
-</Value>
-<Value name="8">
-<Value>AddressStreet</Value>
-</Value>
-<Value name="9">
-<Value>AddressCity</Value>
-</Value>
-<Value name="10">
-<Value>AddressState</Value>
-</Value>
-<Value name="11">
-<Value>AddressZip</Value>
-</Value>
-<Value name="12">
-<Value>HomePhone</Value>
-</Value>
-<Value name="13">
-<Value>SourceMessageID</Value>
-</Value>
-<Value name="14">
-<Value>ExtractedAt</Value>
-</Value>
-</Data>
-<DataLocation>^ProductionGuardian.LabDemo.Msg.PatD</DataLocation>
-<DefaultData>PatientDemographicsDefaultData</DefaultData>
-<IdLocation>^ProductionGuardian.LabDemo.Msg.PatI</IdLocation>
-<IndexLocation>^ProductionGuardian.LabDemo.Msg.PatX</IndexLocation>
-<StreamLocation>^ProductionGuardian.LabDemo.Msg.PatS</StreamLocation>
-<Type>%Storage.Persistent</Type>
-</Storage>
+// No Storage block. Ens.Request inherits its storage from Ens.MessageBody, which pins
+// DataLocation to ^Ens.MessageBodyD; a Default map naming its own globals here fails to
+// compile with "keyword 'DataLocation' must be '^Ens.MessageBodyD'". Message bodies are
+// stored and purged by the interoperability framework, not by this class.
 
 }
--- a/iris/labdemo/Production.cls
+++ b/iris/labdemo/Production.cls
@@ -1,8 +1,10 @@
 /// LABDEMO production for Production Guardian Health Scan.
 ///
-/// Pipeline:
-///   EMRSource  →  LabRouter  →  PIDExtractProcess  →  PatientDemographicsOperation
-///   (HL7 file)    (routing)      (DTL: HL7→PID msg)    (REST POST → PatientDispatcher)
+/// Pipeline (config item names, as they appear in metrics and the proxy JSON):
+///   EMR Source  →  Lab Router  →  Cloud API
+///   (HL7 file)     (routing +      (REST POST → PatientDispatcher)
+///                   HL7ToPID DTL
+///                   applied inline)
 ///
 /// PatientDispatcher upserts patient demographics into:
 ///   ProductionGuardian.LabDemo.Data.PatientRecord (persistent table)
@@ -26,7 +28,7 @@ XData ProductionDefinition
   </Description>
 
   <!-- ============================================================
-       Business Service — EMRSource
+       Business Service — EMR Source
        Reads HL7 v2 files dropped by the HL7Generator.
        ============================================================ -->
   <Item Name="EMR Source"
@@ -47,7 +49,7 @@ XData ProductionDefinition
   </Item>
 
   <!-- ============================================================
-       Business Process — LabRouter
+       Business Process — Lab Router
        Routes ADT^A01 to Cloud API.
        Rule: ProductionGuardian.LabDemo.RoutingRule
        ============================================================ -->
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -3,11 +3,17 @@
 The LABDEMO production models a realistic HL7 lab message pipeline that Health Scan monitors.
 
 ```
-EMRSource  →  LabRouter  →  PIDExtractProcess  →  PatientDemographicsOperation
-(HL7 file)    (routing)      (DTL: PID extract)    (HTTP POST → PatientDispatcher)
-                                                             ↓
-                                                  PatientRecord table (upsert by PatientID)
+EMR Source  →  Lab Router  →  Cloud API
+(HL7 file)     (routing +      (HTTP POST → PatientDispatcher)
+                HL7ToPID DTL             ↓
+                applied inline)  PatientRecord table (upsert by PatientID)
 ```
+
+There are three application hosts. `EMR Source`, `Lab Router` and `Cloud API` are the
+config item names — those are the strings that appear in `/api/monitor/metrics`, in the
+proxy JSON, and in the dashboard, so use them verbatim when reasoning about findings.
+Class names (`PatientDemographicsOperation`) differ from config item names (`Cloud API`);
+the `<Item Name="...">` set in `Production.cls` is authoritative.
 
 ---
 
@@ -15,9 +21,8 @@ EMRSource  →  LabRouter  →  PIDExtractProcess  →  PatientDemographicsOpera
 
 | File | Purpose |
 |---|---|
-| `Production.cls` | Production definition — 4 items + ActivityReporter |
-| `RoutingRule.cls` | Routes ADT^A01 and ORU^R01 to PIDExtractProcess |
-| `Process/PIDExtractProcess.cls` | BP: applies DTL, forwards PatientDemographics |
+| `Production.cls` | Production definition — 3 application items + ActivityReporter |
+| `RoutingRule.cls` | Routes ADT^A01 to Cloud API, applying HL7ToPID inline |
 | `Transform/HL7ToPID.cls` | DTL: HL7 PID segment → PatientDemographics message |
 | `Message/PatientDemographics.cls` | Ens.Request carrying extracted PID fields |
 | `Operation/PatientDemographicsOperation.cls` | BO: HTTP POST to REST dispatcher |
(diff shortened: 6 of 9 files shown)


Ari-Glikman:
Verification notes on this PR. **Dev A has moved to other work and Dev B is taking over their outstanding tasks** — this is that work on their #17. I re-checked the load-bearing claims against `main` and the live instance rather than accepting them.

@tanifgit — you're the reviewer, so the parts worth your attention are flagged below.

## The port-0 diagnosis is right, and #17's own explanation was wrong

Checked the live global directly:

```
^%SYS("WebServer","Port")  ->  defined: true, value: "0"
```

So `$get(^%SYS("WebServer","Port"), 52773)` returns **`0`**, not the default — a `$get` fallback only applies to an *undefined* node. `Verify()` was dialling port 0 and never reaching the network. That's a different failure from the 404 #17 described; on this instance it never got that far.

`+sysPort>0` guarding the fallback is the right fix, and the comment explaining why that global is unsafe is worth more than the code — it's a trap anyone would hit twice.

Two secondary causes are equally real: `%Net.HttpRequest.Get()` returns `$$$OK` on a 404, so the old code scanned an error page as a metrics body; and the checks list asked for `iris_last_activity`, which appears **zero** times live because the family is `iris_interop_last_activity`. Four independent causes behind one symptom.

## Both Storage blocks confirmed against `main`

```
main:  <Type>%Storage.Persistent</Type></Storage>   <- XML end tag
fix:   <Type>%Storage.Persistent</Type>}            <- UDL
```

#17 asked for these to be confirmed by compiling before changing, and they were. `ERROR #5559`, then `#5477 keyword 'DataLocation' must be '^Ens.MessageBodyD'` on the second one, is the sort of second-order failure you only find by running it.

## The `HL7ToPID` find is the most valuable thing here, and it was outside the issue

```
main:  $extract(source.{PID:7"),1,8)
fix:   $extract(source.{PID:7},1,8)
```

A stray `"` where `}` belongs, so **the DTL did not compile and no message could reach `Cloud API` on `main` at all.** Confirmed on `origin/main`. Fixing it was correct — the item-4 trigger instructions are untestable without it — and flagging it rather than slipping it in silently is the right handling.

## Trigger instructions got the number right

`Run(80, 0.2)` with the reasoning stated: depth must clear the absolute floor of **50** as well as 5× baseline. That's the constraint most likely to make a correct system look broken on stage, and documenting *why* 80 is what stops someone reducing it to 40.

## Two things I own

**Something moved the agent's git HEAD mid-run — that was me.** I was committing the ADR 0004 change in the same worktree, and my `git checkout` moved the shared HEAD. Nothing was lost (reflogs checked, branch was local-only and unpushed, restored to its creation point), but it's a real hazard of working in one worktree and I've moved my shell to `main`.

**CI being green proves less than it looks.** All four jobs pass, but **none compiles `iris/**`** — so the ObjectScript here is verified only by live compiles, not by the pipeline. That's a gap in CI I own, and it's the same shape as three things we've already fixed: a check reporting success while verifying nothing. Worth a follow-up; not a blocker.

## Still true after this PR

`queue_buildup` cannot fire from live IRIS: `iris_interop_queued` is production-wide with no `host` label. Right call to document rather than paper over. I've separately verified the fix path — with Cloud API disabled, `EnumerateHostStatus` reports `Cloud API | queue='70'`, names matching the metrics `host` label exactly. That's #12, which I'm picking up next.

Production is still `Running` and untouched — checked.

tanifgit (approved):
**Approved.** Every claim I could check from outside an IRIS instance holds, and the two defects you found off-ticket are worth more than the four on it.

I have no IRIS instance, so **I did not reproduce any compile** — your `#5559` / `#5477` / `#1011` output is taken as reported. Everything else below I verified.

## The off-ticket finds are the important half

`HL7ToPID.cls:33` is the one that matters:

```
-  <assign value='$extract(source.{PID:7"),1,8)' ...
+  <assign value='$extract(source.{PID:7},1,8)' ...
```

A stray `"` where the brace belongs, in the DTL that every routed message passes through. With that class failing to compile, **no message could reach `Cloud API` at all** — the LABDEMO pipeline was broken on `main`, and had been since it was written. Neither #17 nor anything else had it.

`HL7Generator.cls:62` — `WriteFile(filename, filename)` writing each file's own path as its contents — is the same class of bug and nastier for being silent: `Run()` was correct, so it only bites `RunContinuous()`, which is exactly what the throughput and queue triggers depend on. Anyone testing those would have seen an empty findings list and blamed the engine.

Fixing both here rather than filing them was the right call. They're two lines, and filing them would have left every trigger in the same PR untestable.

## The `WebPort()` diagnosis is the best part of item 1

```objectscript
set sysPort = $get(^%SYS("WebServer","Port"), 0)
quit $select(+sysPort>0: +sysPort, 1: 52773)
```

*Defined-but-zero* defeating a `$get` default is a genuinely non-obvious failure — `$get(...,52773)` reads as safe and isn't. And it explains why the prefix in #17 was a red herring: the request never reached the network, so the URL was irrelevant. Finding that the filed cause was the *third* of four independent causes, and saying so, is the right way to close an issue someone else wrote.

Reading a 404 as success is the other one worth having caught. `%Net.HttpRequest.Get()` returning `$$$OK` for any answered request is a trap, and the old code's failure mode — scan the 404 HTML, report all seven families `MISSING`, send the operator to re-run `EnableSAMForNamespace()` — is worse than a crash, because it points confidently at the wrong thing.

The `iris_last_activity` → `iris_interop_last_activity` fix means that check reported `MISSING` on every healthy instance since it was written.

## Verified — the trigger arithmetic, all eight rows

This is what you asked a reviewer's eye on, so I checked every number against `thresholds.json` **and** against the gate code in `detect/rules/index.ts`, not just the config:

| README claim | `thresholds.json` | verdict |
|---|---|---|
| `minBaselineSamples` 12 → ~2 min at 10 s poll | `12` | ✅ 12 × 10 s = 120 s |
| `throughput_drop` fraction 0.4, needs baseline ≥ 0.1; 2 s gap → 0.5 msg/s | `0.4` / `0.1` | ✅ 1/2 s = 0.5 ≥ 0.1 |
| `slow_processing` gate is 0.3 s (Cloud API floor) vs 3 × 0.05 s | override `0.3`, mult `3.0` | ✅ max(0.3, 0.15) = 0.3 |
| `growing_queue_wait` floor 0.15 s vs 3 × 0.03 s | `0.15` / `3.0` | ✅ max(0.15, 0.09) = 0.15 |
| `elevated_error_rate` floor 1.0/min, 3×; 1 msg/2 s ≈ 30/min | `1.0` / `3.0` | ✅ 60/2 = 30 |
| `stalled_host` 300 s **and** `requiresQueued` → 5 min | `300` / `true` | ✅ |
| `queue_buildup` floor **50** and 5× baseline | `50` / `5.0` | ✅ |
| `system_alert` is the only rule that can emit `info` | `severity: "info"` + `_comment_info_severity` | ✅ |

And "must clear both the baseline multiplier **and** an absolute floor" is literally the code:

```ts
if (current < rule.absoluteFloorSeconds) return null;
if (ratio < rule.baselineMultiplier) return null;
```

So "the greater of the floor and the multiple" is an accurate description of a conjunction, not a paraphrase that drifts. Quoting `thresholds.json` as authoritative instead of restating its numbers is the right call — those floors moved once already in #20 and will move again for #25.

Two things in that table I'd have missed and you didn't: that disabling a host trips `dead_host` **as well as** the intended rule, so two findings is correct rather than a duplicate; and the instruction to **remove the `hang` afterwards**, since a stale hang poisons every later baseline. Both are the kind of thing that reads as an engine bug when it isn't.

## Verified — the rest

- `grep -rn "PIDExtractProcess\|ORU\^R01\|ORU_R01" iris/` → **empty**. Claim holds.
- The only remaining unspaced host name in `iris/` is the `CLAUDE.md` line explaining that the names *are* spaced. Nothing stale left.
- `Production.cls` `<Item Name>` set is exactly `EMR Source`, `Lab Router`, `Cloud API`, `Ens.Activity.Operation.Local` — 3 application + 1 framework, matching both `CLAUDE.md` §3 and the README inventory.
- `Process/PIDExtractProcess.cls` does not exist on disk, so dropping it from the inventory table is a correction, not a deletion — the README had been listing a file that wasn't there.
- `PatientDemographicsOperation.cls:24` has `Method OnMessage(`, so the `hang 1` instruction points at a method that exists, in the class `Production.cls` maps to `Cloud API`.
- Pointing `CLAUDE.md` §3 at `Production.cls` instead of keeping a second table is the right fix for the actual defect. That duplicated list is what went stale.

The UDL reasoning is sound on both storage items: `}` closes a Storage block and `</Storage>` belongs only to the XML export format; and `Ens.Request` inheriting `DataLocation` from `Ens.MessageBody` does make a Default map naming its own globals uncompilable. The comment left in `PatientDemographics.cls` explaining why the block must stay gone is the part that keeps it fixed.

## Two non-blocking suggestions

**1. `WebAppPrefix()` turns `"/"` into a double slash.** The `quit:prefix=""` guard runs before the strip, so a global set to `"/"` — a reasonable way to write "the root" — survives:

```objectscript
set prefix = "/"                          // quit:prefix="" doesn't catch it
set prefix = $zstrip("/", ">", "/")       // -> ""
quit $case($extract(""), "/": ..., : "/" _ "")   // -> "/"
```

giving `http://host:80//api/monitor/metrics`. Most web servers shrug at that, but `Verify()` now *prints* the URL, so it'll look wrong at the exact moment someone is debugging a URL. Moving the empty-check after the strip fixes it.

**2. Consider making the alerts check opt-in — `Verify(checkAlerts = 0)`.** You documented the consume-on-read hazard rather than changing it, and I agree that was the right scope call for #17. But the failure it enables is quiet and permanent: `Verify()` is the thing you run *repeatedly* while getting the URL right, and each run silently eats any pending alert. If one gets eaten while the proxy is up, `system_alert` — one of the eight finding types — never fires, nothing logs an error, and only `mgr/alerts.log` still has the evidence. A default-off parameter keeps the check available for a deliberate one-shot without arming the footgun. Your call; the docs may well be enough for a 5-day demo.

## One thing this PR resolved that wasn't its job

> *"the live instance runs an older 4-host definition (including `FHIR Transform`) than `Production.cls` on `main`"*

That note is what corrected me on #30. I'd requested changes there claiming `samples/metrics-dump.txt` was fabricated because `FHIR Transform` appears in no capture under `services/metrics-proxy/fixtures/` — not knowing the live instance still runs it. I've withdrawn that and approved.

The substantive consequence is worth stating here, since this PR is where `iris/**` declares the roster: **`Production.cls` says 3 application hosts, the live instance serves 4.** So `README.md` and `CLAUDE.md` §3 are now correct about the *intended* production and wrong about the *running* one, and anyone following this README against that instance will see a fourth host in the dashboard.

You were right to leave it alone — reconciling it means redeploying a production, which is not a docs fix. It also means **PR #15's premise was wrong**: `FHIR Transform` was dropped from `contracts/samples/` on the grounds that the production "has no such host and never did," and #18 and #21 followed. Two of those three were mine, and I approved the first. I'm filing the roster question separately; flagging it here so it's on the record in the area that owns `Production.cls`.

Nothing outside `iris/**`, all four CI jobs green.

PR #30, opened 2026-08-12
feat(contracts): publish the metrics proxy contract, schema and real capture

Ari-Glikman:
Closes the last Day-1 gate item from `CONTRIBUTING.md` §6 — item 4 of #16.

**Dev A (@kskubach) has moved to other work, so Dev B is taking over their outstanding tasks.** This is authorized by the repo owner. The work is attributed to me; I am not acting as Dev A. The owner named in `contracts/README.md` stays Dev A, because `services/metrics-proxy/` is still their code and CODEOWNERS still routes it to them.

Three files were missing from the gate. All three are here:

```
PRESENT  contracts/healthscan-api.md          Dev B
PRESENT  contracts/samples/findings-response.json
ADDED    contracts/proxy-api.md
ADDED    contracts/proxy.schema.json
ADDED    contracts/samples/metrics-dump.txt   (the real capture, copied)
```

## The contract is derived from the code, not from intent

Dev A is not here to describe their own component, so nothing in `proxy-api.md` is a guess. The parser was run over the real capture, and **the real proxy (`index.js`) was run against a stub IRIS serving that same capture**, so all three endpoints were observed rather than inferred. That is what turned up two mismatches nobody had found.

`samples/metrics-dump.txt` is copied, not regenerated — verified byte-identical:

```
$ git hash-object C:/temp/real-metrics-capture.txt
537b6d5183501bc4ad2ceba05cabe5c9a8dd28b7
$ git rev-parse :contracts/samples/metrics-dump.txt
537b6d5183501bc4ad2ceba05cabe5c9a8dd28b7
```

One small correction: it is **1236 lines**, not the 1249 quoted in #10 and in `src/parser.js`'s comment. Same file, counted differently; `wc -l` says 1236. Not worth chasing, worth not silently contradicting.

## The five PROXY-Q answers

| # | Question | Answer | Contradicts the assumption? |
|---|---|---|---|
| **Q1** | Array, or object keyed by host? `status` passed through or normalized? | **Array** under a `hosts` key, alphabetical. `status` **passed through** unchanged. But the array is nested in an object alongside `_meta`, and **the field is `host`, not `name`**. | Partly — see Q9 |
| **Q2** | Is `queued` present per host? | **Field yes, value no.** `iris_interop_queued` carries no `host` label (labels are exactly `id`, `production` — verified in all three captures). Proxy publishes **`queued: null`** per host and the real per-production total as `_meta.productionQueued` (`0` in the sample). `queue_buildup` cannot fire per host today. #12, ADR 0001. | **Yes** — `null`, not `0`. See §6.1 |
| **Q3** | `avg*` aggregated, or raw per-`messagetype`? | **Already aggregated by the proxy, weighted by `iris_interop_sample_count`.** Aggregation does *not* move to the engine. A host with no sample-count line falls back to an unweighted mean rather than dropping out. | No — assumption holds |
| **Q4** | Elapsed seconds, or a pre-computed timestamp? | **Both.** `lastActivityElapsedSeconds` is IRIS's own value (elapsed seconds, e.g. `4.838` — not epoch); `lastActivity` is `polledAt − elapsed` as ISO 8601. Prefer the elapsed figure for `stalled_host`: it is the measurement, the timestamp inherits ±10 s poll error. | No — assumption holds |
| **Q5** | Framework hosts filtered by the proxy? | **They reach you, flagged `isFramework`.** Your own filter stays necessary — **11 of 15 hosts in the capture are framework**, and `Ens.MonitorService` is one of the few *with* `avg_*` series, so an unfiltered snapshot reports framework timings as application latency. Filter on the flag, not the prefix: the label carries the *item* name, which has been both `ActivityReporter` and `Ens.Activity.Operation.Local`. | No — assumption holds |

Three more surfaced while writing it (§5.1): **Q6** `type` is `service`/`process`/`operation`/**`unknown`** (`hosttype` rides on `avg_*` only, so 8 of 15 hosts report `unknown` — a real value, not a bug); **Q7** the status enum adds **`Unknown`** to the seven, with no `Active` and no `Warning`; **Q8** below.

### `iris_interop_messages_errored` carries no `host` label either

```
iris_interop_messages_errored{id="LABDEMO",production="LABDEMO.Production"} 0
```

Per-production, exactly like `queued`. `METRIC_MAP` treats it as per-host, so the line is skipped for want of a `host` label and **`errored` is `null` on every host**. `elevated_error_rate` cannot fire per host today — the same structural gap as `queue_buildup`, and **not filed anywhere**. §6.3.

Worse than `queued` in one respect: no per-production total is published, because `messages_errored` is not in `SCALAR_FAMILIES`, so the value is parsed and dropped. `EnumerateHostStatus` returns per-host error counts *and* queue depth, so the #12 REST wrapper closes both rules. Two of eight finding types depend on it.

### All 15 real hosts are rejected by the engine's `isProxyHost()`

Measured against the real capture, not predicted:

```
     Cloud API           REJECT: name=undefined queued=null messagesErrored=undefined
     EMR Source          REJECT: name=undefined queued=null messagesErrored=undefined
     FHIR Transform      REJECT: name=undefined queued=null messagesErrored=undefined
     Lab Router          REJECT: name=undefined queued=null messagesErrored=undefined
[fw] Ens.MonitorService  REJECT: name=undefined queued=null messages=null ... 

isProxyHost(): accepted 0 / rejected 15 of 15 hosts
after name/queued/messagesErrored normalisation: accepted 6 / 15
application hosts after framework filter: Cloud API, EMR Source, FHIR Transform, Lab Router
```

Three independent mismatches: `name` vs `host` (Q9), `messagesErrored` vs `errored` (Q8), `queued: null` against a finite-number guard (Q2).

**The guard log-and-skips per host, which is the right design — but when the mismatch is in a field every host shares, "skip the bad entry" becomes "report zero hosts"**, and the dashboard shows an empty grid with no error anywhere. A per-entry guard cannot degrade gracefully against a whole-payload mismatch. Worth noting for the class of bug, not just this instance.

Two further live-mode breaks the mock cannot show (§5.2): the client requests **`/api/metrics`** (real route is `/proxy/metrics` → `404`), and reads `alerts` from the metrics payload where they never appear — alerts are a **separate endpoint**, so `system_alert` can never fire live. Both pass typecheck and all unit tests. This is ADR 0004's named cost landing a second time: `mockClient.ts` reads fixtures written to the engine's own assumed shape, so mocking against them proves self-consistency and nothing else.

**All the reconciliation is my side to do, in a follow-up PR.** The proxy's shape is verified against live IRIS, and `contracts/` is not edited to make a consumer compile.

### `queued: null` vs `queued: 0` (§6.1)

The proxy emits `null`; the engine's guard requires a finite number and **rejects the whole host** on seeing it. Both sides are defensible, which is why I have not picked one silently:

- **`null` is honest.** `src/parser.test.js` pins it: *"`queued: 0` per host would assert every host is drained while 14 sit somewhere."*
- **The finite-number guard is also right.** A rule cannot compare `null` to a baseline, and `0` at least keeps the host visible so its other seven metrics can be evaluated.

**Contract position: `0` is the conformant placeholder.** The schema permits `null` because that is what the running code emits — a schema that rejected it would be a fiction.

**Recommended fix, one line, engine side:** accept `null` for `queued` and have `queue_buildup` skip a host whose depth is unmeasurable, the way comparative rules already skip a warming baseline. That preserves the absent-is-not-zero invariant the rest of the payload depends on, instead of carving out an exception for the one field where it is inconvenient. Rejecting the whole host is the worst of the three options — it loses `status`, `messagesPerSec` and everything else over one unmeasurable field.

@tanifgit — your number from #16 still stands and is worth restating: `absoluteFloor: 50`, captured depth `48`, so **the rule will not trip even once this is wired up.** "Returning real numbers" and "the finding fires" are separate milestones.

### `errored` vs `messagesErrored` (§6.2)

Raised on #16. **Recommendation: `errored`.** The tie-break is blast radius, not aesthetics:

- `errored` costs **one line** in `services/detection-engine/src/types/proxy.ts` and its guard.
- `messagesErrored` costs that **plus a change to a ratified contract with a live consumer** — `healthscan-api.md`, its schema, its `.d.ts`, two samples, and Dev C's components — for a naming preference.

Per `contracts/README.md`: the question is not how many lines reference a value but whether anything *means* something in terms of it. `errored` already means something downstream. Published as `errored` because that is what the code emits — **a request for a decision, not a decision.**

## Verification

`npm run validate` — 31 checks, up from 11. Nothing existing changed:

```
$ cd contracts && npm run validate
ok    accepts: empty hosts array
ok    accepts: empty findings array
ok    accepts: JS toISOString() milliseconds
ok    accepts: Python isoformat() microseconds
ok    rejects: retired "Warning" host status
ok    rejects: hosts array served in the findings position
ok    rejects: timestamp with no timezone
ok    rejects: unknown finding type
ok    rejects: baselineValue as a string
ok    proxy accepts: real Lab Router host, nulls included
ok    proxy accepts: warming metrics response (200, not 503)
ok    proxy accepts: real metrics response
ok    proxy accepts: real alert, severity as the numeric string IRIS emits
ok    proxy accepts: health reporting a reachable instance with no interop metrics
ok    proxy rejects: status "Active" — not in the IRIS enum
ok    proxy rejects: type "actor" — must be normalized to process
ok    proxy rejects: lastActivity with an offset instead of Z
ok    proxy rejects: null host name
ok    proxy rejects: metrics response with no _meta
ok    proxy rejects: alerts response served in the metrics position
ok    proxy rejects: engine-shaped host: `name` instead of `host`
ok    proxy rejects: engine-shaped host: `messagesErrored` instead of `errored`
ok    metrics-dump.txt: host label is `host`, with spaces
ok    metrics-dump.txt: status is a label, and reads OK
ok    metrics-dump.txt: last_activity is elapsed seconds
ok    metrics-dump.txt: hosttype rides on avg_processing_time as `actor`
ok    metrics-dump.txt: sample_count is its own family, per (host, messagetype)
ok    metrics-dump.txt: queued carries NO host label (§6.1)
ok    metrics-dump.txt: messages_errored carries NO host label (§6.3)

all checks passed (2 samples, 9 accept, 13 reject, 7 capture claims)
```

The last seven are `CAPTURE_CLAIMS`: `metrics-dump.txt` is Prometheus text and cannot be schema-checked, so these are regexes over the label shapes the contract quotes. **Two of them assert a label is *absent*** — an assertion that something is missing is the only way a future capture silently gaining it gets noticed.

Two must-reject cases are the engine's *current* shape. If §6.2 is settled the other way those checks fail and say so, which is why they are there rather than in a comment.

### Proving the validator can fail

A validator that never fails is worthless, so I broke it three ways on purpose:

```
=== A: add 'Active' to the status enum (a wrong schema) ===
FAIL  proxy rejects: status "Active" — not in the IRIS enum
1 check(s) failed

=== B: narrow NullableCount to `integer` (schema disagrees with real output) ===
FAIL  proxy accepts: real Lab Router host, nulls included
FAIL  proxy accepts: warming metrics response (200, not 503)
FAIL  proxy accepts: real metrics response
3 check(s) failed

=== C: strip the messages_errored line from the capture ===
FAIL  metrics-dump.txt: messages_errored carries NO host label (§6.3)
1 check(s) failed
```

All three restored; suite green again.

### Throwaway ajv script (not committed)

Validated the schema against the snapshot the **real parser** produces from the **real capture**, plus the three live HTTP payloads captured by `curl` from the real proxy. 32/32, with 27 reject cases:

```
ok    REAL snapshot from samples/metrics-dump.txt (15 hosts) against MetricsResponse
ok    all 15 real hosts individually against ProxyHost
ok    live /proxy/metrics warming payload
ok    live /proxy/alerts warming payload
ok    live /proxy/alerts payload (2 real alerts, severity "2")
ok    live /proxy/health status=starting
ok    live /proxy/health status=ok
ok    live /proxy/health status=reachable-but-no-interop
ok    rejects: bad status enum ("Active" — does not exist in IRIS)
ok    rejects: bad status enum ("Warning" — retired in PR #3)
ok    rejects: bad type enum ("actor" — must be normalized to process)
ok    rejects: null where a string is required (host)
ok    rejects: string where a number is required (messagesPerSec)
ok    rejects: missing required field (host)
ok    rejects: missing required field (isFramework)
ok    rejects: missing required field (lastActivityElapsedSeconds)
ok    rejects: wrong timestamp format (space instead of T, no Z)
ok    rejects: timestamp with an offset instead of Z
ok    rejects: negative cumulative count (messages)
ok    rejects: non-integer cumulative count (messages)
ok    rejects: engine-shaped host: `name` instead of `host`
ok    rejects: engine-shaped host: `messagesErrored` instead of `errored`
ok    rejects: unknown extra field (additionalProperties: false)
ok    rejects: metrics response with hosts as an object, not an array
ok    rejects: alerts response served in the metrics position
ok    rejects: metrics response served in the alerts position
ok    rejects: alert severity as a number, not a numeric string
ok    rejects: null where a number is required (health uptime)
ok    rejects: null where a boolean is required (isFramework)
...
all passed (27 reject cases)
```

### The live endpoint capture

The real proxy against a stub IRIS serving `metrics-dump.txt`. This is where the endpoint behaviour in §1–§4 comes from:

```
$ curl -s http://localhost:3011/proxy/metrics   # application hosts only
{ "host": "Lab Router", "type": "process", "status": "OK", "isFramework": false,
  "queued": null, "messages": 126, "messagesPerSec": 1.2, "errored": null,
  "avgProcessingTime": 0.08, "avgQueueingTime": 0,
  "lastActivity": "2026-08-12T08:45:25.399Z", "lastActivityElapsedSeconds": 4.838 }
_meta: { production: "LABDEMO.Production", productionQueued: 0,
         absentFamilies: [], hostCount: 15, applicationHostCount: 4 }

$ curl -s http://localhost:3011/proxy/health
{"status":"ok","uptime":3.75,"lastPoll":"2026-08-12T08:45:30.237Z",
 "production":"LABDEMO.Production","hostCount":15,"applicationHostCount":4}

# warming — 200 with warming:true, NOT 503. Confirmed by pointing at a dead port.
$ curl -s -w ' <- HTTP %{http_code}' http://localhost:3012/proxy/metrics
{"hosts":[],"systemAlertsNew":null,"warming":true,"_meta":{"polledAt":null}} <- HTTP 200
$ curl -s -w ' <- HTTP %{http_code}' http://localhost:3012/proxy/health
{"status":"starting","uptime":2.31,"lastPoll":null} <- HTTP 200

# the %SYS case, reproduced by serving a body with the iris_interop_* lines stripped
$ curl -s http://localhost:3013/proxy/health
{"status":"reachable, but no interop metrics", ..., "hint":"IRIS answered but sent no
 iris_interop_* families. Usual cause is a wrong IRIS_BASE_PATH ..."}

# no CORS headers on /proxy/metrics — documented in §4
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: application/json; charset=utf-8
```

### Proxy tests untouched

```
$ cd services/metrics-proxy && npm test
ℹ tests 71 / suites 14 / pass 71 / fail 0
```

No proxy code was modified — this is a `contracts/`-only PR.

## What I could not verify

Stated plainly rather than asserted:

- **No live IRIS was available to me.** Everything is against `metrics-dump.txt` and `fixtures/`, served by a local stub. The captures are real; the *instance* was not live during this work. The `%SYS`/`IRIS_BASE_PATH` behaviour in §3 is reproduced from a stripped body and quoted from Dev A's 2026-08-11 measurement, not re-measured.
- **The alerts payload shape rests on a single capture** (`fixtures/alerts-live-capture.json`), obtained by accident on a first-ever read of a consume-on-read endpoint. `ProxyAlert` therefore leaves `additionalProperties` **open** and does not pattern-constrain `time`. `_meta.shape` is the field that makes a wrong guess loud rather than silent.
- **`Ens.Util.Statistics:EnumerateHostStatus` was not run.** The `Cloud API = 48` figure and the claim that it returns per-host error counts as well as queue depth are quoted from #12 and #10, not reproduced here.
- **The `Retry`, `Stopped`, `Unconfigured`, `Disabled` statuses do not appear in any capture** — all 15 hosts read `OK`. The enum is from IRIS source via #10, and `Unknown` is the only value I observed the proxy generate itself. Consumers should keep rendering unrecognized statuses neutrally.
- **`samples/alerts.json` and `samples/proxy-response.json`** are named in `CONTRIBUTING.md` §1 and still do not exist. Both are derivable without inventing anything, but generating them is a judgement call about what "the" canonical alerts sample is, so I noted the gap in `README.md` rather than filling it unilaterally.

## Review notes

CODEOWNERS requires all three on `contracts/`. @tanifgit requested — Q1/Q7 touch the status enum you paid 83 insertions for, and the `queued` decision affects your grid. Dev A's slot has no active reviewer.

Nothing outside `contracts/` is touched, so the `detection-engine` and `dashboard` CI jobs are unaffected. The follow-up reconciliation (`/api/metrics` → `/proxy/metrics`, `name` → `host`, `messagesErrored` → `errored`, the `queued` guard, and the alerts fetch) is a separate PR in my own area — it should not ride along with a contract change.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,84 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-08-12 — Metrics proxy contract published (Dev B, on Dev A's behalf)
+
+Initial publication of `proxy-api.md`, `proxy.schema.json`, `samples/metrics-dump.txt`.
+
+Closes the last Day-1 gate item from `CONTRIBUTING.md` §6 (issue #16 item 4). Answers all five
+`PROXY-Q` markers in `services/detection-engine/src/types/proxy.ts` inline in §5, plus three that
+surfaced while writing it.
+
+**Dev A has moved to other work, so this is derived from their merged code rather than authored
+alongside it.** Every claim comes from reading `services/metrics-proxy/` on `main` and running it,
+not from intent. Attribution is Dev B's; the contract's owner in `README.md` stays Dev A, because
+the code is still theirs and `CODEOWNERS` still routes `services/metrics-proxy/` to them.
+
+`samples/metrics-dump.txt` is the real 1236-line `/api/monitor/metrics` body from a live LABDEMO
+production, byte-identical to the capture behind issue #10 — copied, not regenerated. 957 metric
+lines, 15 interop hosts, 4 application. **It is 1236 lines, not the 1249 quoted in issue #10 and in
+`src/parser.js`'s comment**; the same file, counted differently, and the smaller number is what
+`wc -l` reports. Not worth chasing, worth not silently contradicting.
+
+### Three things the contract says that were not previously known
+
+- **`iris_interop_messages_errored` carries no `host` label either.** It is per-production, exactly
+  like `iris_interop_queued`: `iris_interop_messages_errored{id="LABDEMO",production="LABDEMO.Production"} 0`.
+  The parser's `METRIC_MAP` treats it as per-host, so the line is skipped for want of a `host` label
+  and `errored` is `null` on every host. **`elevated_error_rate` cannot fire per host today** — the
+  same structural gap as `queue_buildup` (#12) and, unlike that one, not filed. §6.3. Worse than
+  `queued` in one respect: no per-production total is published anywhere, because
+  `messages_errored` is not in `SCALAR_FAMILIES`, so the value is parsed and dropped.
+- **Every one of the 15 real hosts is rejected by Dev B's `isProxyHost()`.** Measured, not
+  predicted. Three field-level mismatches, each independently sufficient: `name` vs `host`,
+  `messagesErrored` vs `errored`, and `queued: null` against a finite-number guard. The guard
+  log-and-skips per host, which is right — but a mismatch in a field every host shares turns "skip
+  the bad entry" into "report zero hosts", with no error surfaced. §5.2.
+- **The engine's live client cannot work today, for two more reasons the mock cannot show.** It
+  requests `/api/metrics` (the real route is `/proxy/metrics` → `404`), and it reads `alerts` from
+  the metrics payload, where they are never present — alerts are a separate endpoint, so
+  `system_alert` can never fire in live mode. Both pass typecheck and every unit test. This is
+  ADR 0004's named cost landing a second time: `mockClient.ts` reads fixtures written to the
+  engine's own assumed shape, so mocking against them proves self-consistency and nothing else.
+
+None of these change a published field. **All of the reconciliation is on Dev B's side** — the
+proxy's shape is verified against live IRIS, and `contracts/` is not edited to make a consumer
+compile.
+
+### Two open items, deliberately not resolved here
+
+- **§6.1 — `queued: null` vs `queued: 0`.** The schema permits `null` because that is what the
+  proxy emits; a schema that rejected the running code would be a fiction. **The contract position
+  is that `0` is the conformant placeholder**, and the recommended fix is one line on the engine
+  side: accept `null` and have `queue_buildup` skip an unmeasurable host, the way comparative rules
+  already skip a warming baseline. That preserves the absent-is-not-zero invariant the rest of the
+  payload depends on instead of carving out an exception for the inconvenient field.
+- **§6.2 — `errored` vs `messagesErrored`.** Raised on #16. Recommendation is `errored`, and the
+  tie-break is blast radius: `errored` costs one line in the engine, `messagesErrored` costs a
+  change to a ratified contract with a live consumer — its schema, its `.d.ts`, two samples and Dev
+  C's components — for a naming preference. Published as `errored` because that is what the code
+  emits. **A request for a decision, not a decision.**
+
+### `validate.mjs`
+
+Extended, structure unchanged: `PROXY_MUST_ACCEPT` (5), `PROXY_MUST_REJECT` (8), and
+`CAPTURE_CLAIMS` (7) alongside the existing arrays. 31 checks total, up from 11.
+
+`samples/metrics-dump.txt` is Prometheus text, not JSON, so it cannot be validated against a
+schema the way the healthscan samples are. `CAPTURE_CLAIMS` covers it with regexes over the label
+shapes the contract quotes — including two asserting a label is **absent** (`queued` and
+`messages_errored` have no `host` label). An assertion that something is missing is the only way a
+future capture silently gaining it gets noticed.
+
+Two of the must-reject cases are the engine's *current* shape (`name`, `messagesErrored`). If §6.2
+is settled the other way, those checks fail and say so — which is the point of putting them there
+rather than in a comment.
+
+Verified: `npm run validate` → 31/31. Confirmed the new checks can fail, rather than assuming it:
+adding `Active` to the status enum fails 1, narrowing `NullableCount` to `integer` fails 3, and
+stripping the `messages_errored` line from the capture fails 1. A validator never seen to fail is
+not known to be testing anything.
+
 ## 2026-08-10 — `FHIR Transform` removed from the samples (Dev A)
 
 **No schema change and no field change.** `host` is `{"type": "string", "minLength": 1}` with no
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -7,8 +7,15 @@ path all three developers read, which is why it is PR-gated.
 |---|---|---|
 | `proxy-api.md`, `proxy.schema.json` | Dev A | Dev B |
 | `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts` | Dev B | Dev C |
-| `samples/metrics-dump.txt`, `samples/alerts.json`, `samples/proxy-response.json` | Dev A | Dev B |
+| `samples/metrics-dump.txt` | Dev A | Dev B |
 | `samples/hosts-response.json`, `samples/findings-response.json` | Dev B | Dev C |
+
+`samples/alerts.json` and `samples/proxy-response.json` were planned in `CONTRIBUTING.md` §1 and do
+not exist. Both are derivable without inventing anything — a real alerts capture is at
+`services/metrics-proxy/fixtures/alerts-live-capture.json`, and a proxy response is
+`src/parser.js` applied to `samples/metrics-dump.txt` — but neither is committed here yet, so a
+consumer mocking Dev A must run the parser or point at `npm run mock`. Noted rather than quietly
+left off the table.
 
 ## Why the samples matter
 
@@ -19,6 +26,11 @@ what makes "works against the mock" predict "works against the real thing."
 Samples are captured from live IRIS wherever possible, not invented. The Health Scan samples in
 this directory use real measured values from the LABDEMO production — `avgProcessingTime: 0.08`
 for Lab Router is what IRIS actually reported, not a plausible-looking number.
+
+`samples/metrics-dump.txt` is the strongest form of this: a raw 1236-line `/api/monitor/metrics`
+body, not a reshaped one. It is what proved the proxy's original parser wrong in six places
+(issue #10) and what turned up two further mismatches while `proxy-api.md` was written — both
+found by running real code over real bytes, neither visible from a hand-written fixture.
 
 ## Never edit a contract in place
 
@@ -63,12 +75,19 @@ the shared fixtures disagree, CI says so — rather than the Day-5 rehearsal.
 cd contracts && npm install && npm run validate
 ```
 
-It checks three things, and the last two are what make the first mean anything:
+It checks four things, and the middle two are what make the first mean anything:
 
-- each sample against **one named definition** (`HostsResponse` / `FindingsResponse`)
-- structural cases that must **pass** — `[]`, and sub-second timestamps from any language
+- each JSON sample against **one named definition** (`HostsResponse` / `FindingsResponse`)
+- structural cases that must **pass** — `[]`, sub-second timestamps from any language, and the real
+  proxy payloads including their `null`s
 - cases that must **fail** — a retired `Warning` status, an unknown finding type, a hosts array
-  served in the findings position
+  served in the findings position, an alerts response in the metrics position, and the engine's
+  current `name`/`messagesErrored` host shape
+- **claims about `samples/metrics-dump.txt`**, which is Prometheus text and cannot be schema-checked
+  at all. Regexes over the label shapes `proxy-api.md` quotes, two of them asserting a label is
+  **absent** — `iris_interop_queued` and `iris_interop_messages_errored` carry no `host` label, and
+  an assertion that something is missing is the only way a future capture silently gaining it gets
+  noticed
 
 **Do not replace this with an `ajv-cli` one-liner.** Three reasons, all learned the hard way on
 PR #3:
--- a/contracts/proxy-api.md
+++ b/contracts/proxy-api.md
@@ -1,0 +1,365 @@
+# Metrics proxy API contract
+
+**Owner:** Dev A · **Consumer:** Dev B · **Port:** `3001` · **Status:** published
+
+Three read-only endpoints. The proxy polls the IRIS built-in `/api/monitor/` API, parses the
+Prometheus text format, and republishes it as per-host JSON. **It forwards and reshapes; it does
+not detect, threshold, or judge.** Baseline comparison and findings are the detection engine's job
+(`healthscan-api.md`).
+
+Machine-readable: `proxy.schema.json`.
+Shared fixture: `samples/metrics-dump.txt` — the raw IRIS `/api/monitor/metrics` body every
+snapshot in this document was derived from. Feed it to `src/parser.js` and you get the values
+quoted here.
+
+**Published by Dev B on Dev A's behalf.** Dev A has moved to other work; this contract is derived
+from their merged code (`services/metrics-proxy/`) rather than authored alongside it, so it
+documents what the proxy *does* today, not an intent. Where the code and Dev B's engine disagree,
+§6 names it rather than resolving it silently.
+
+---
+
+## 1. `GET /proxy/metrics`
+
+The latest metric snapshot. One entry per interoperability host IRIS reported, plus per-production
+scalars and diagnostics.
+
+```json
+{
+  "hosts": [
+    {
+      "host": "Lab Router",
+      "type": "process",
+      "status": "OK",
+      "isFramework": false,
+      "queued": null,
+      "messages": 126,
+      "messagesPerSec": 1.2,
+      "errored": null,
+      "avgProcessingTime": 0.08,
+      "avgQueueingTime": 0,
+      "lastActivity": "2026-08-11T23:59:55.162Z",
+      "lastActivityElapsedSeconds": 4.838
+    }
+  ],
+  "systemAlertsNew": 1,
+  "systemAlertsLog": 1,
+  "_meta": {
+    "polledAt": "2026-08-12T00:00:00Z",
+    "production": "LABDEMO.Production",
+    "productionQueued": 0,
+    "absentFamilies": [],
+    "hostCount": 15,
+    "applicationHostCount": 4
+  }
+}
+```
+
+Those are measured values from `samples/metrics-dump.txt`, not illustrative ones.
+
+### 1.1 Host fields
+
+| Field | Type | Notes |
+|---|---|---|
+| `host` | string | Config item name, exactly as the IRIS `host` label carries it. **The join key** — see Q1. |
+| `type` | string | `service` \| `process` \| `operation` \| `unknown`. Normalized — see Q6. |
+| `status` | string | The IRIS `status` label, passed through. `Unknown` when IRIS sent no `iris_interop_hosts` line. Enum in Q7. |
+| `isFramework` | boolean | `true` for IRIS's own plumbing. Flag, not a filter — see Q5. |
+| `queued` | number \| **null** | Per-host queue depth. **`null` on every host today** — see Q2 and §6.1. |
+| `messages` | number \| null | Cumulative messages since production start. |
+| `messagesPerSec` | number \| null | Throughput over the IRIS sampling window. |
+| `errored` | number \| **null** | Cumulative errored count. **`null` on every host today** — see Q8. |
+| `avgProcessingTime` | number \| null | **Seconds.** Aggregated across message types — see Q3. |
+| `avgQueueingTime` | number \| null | **Seconds.** Aggregated across message types — see Q3. |
+| `lastActivity` | string \| null | ISO 8601 UTC, `Z`-suffixed. Derived — see Q4. |
+| `lastActivityElapsedSeconds` | number \| null | Seconds since last activity, as IRIS gives it — see Q4. |
+
+Hosts are in stable alphabetical order by `host` (`localeCompare`), matching the findings API so
+the dashboard never reorders rows between polls.
+
+**`null` means IRIS reported no series. `0` means IRIS measured zero.** This distinction is the
+single most important thing in this contract and it is not cosmetic: on the instance behind
+`fixtures/metrics-live-capture.txt`, IRIS emitted **no `iris_interop_messages_errored` and no
+`iris_interop_last_activity` lines at all** — the families were absent, not zero-valued. Publishing
+`0` there would make `elevated_error_rate` structurally unable to fire while looking like a
+measurement, and nothing would say why. `_meta.absentFamilies` names what IRIS did not send.
+
+Consumers must therefore treat every numeric host field as nullable and **skip a rule rather than
+substitute a value.** A rule fed a fabricated zero reports on data that does not exist.
+
+### 1.2 Top-level and `_meta` fields
+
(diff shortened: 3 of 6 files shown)


tanifgit (changes requested):
**Requesting changes.** The contract prose, the schema, the validator design and the five PROXY-Q answers are all excellent, and the two negative assertions plus the deliberate three-way break test are the best verification work in this repo so far. I'd merge all of it.

But `samples/metrics-dump.txt` cannot go into `contracts/` as it stands. **It is not a capture of our LABDEMO production, and it reinstates the host that #15, #18 and #21 removed.**

I checked this hard before writing it, because it contradicts the PR's headline claim.

## 1. Blocking — `metrics-dump.txt` contains a host that appears in no real capture

`FHIR Transform` is in the file, with a `host` label and 8 metric lines:

```
iris_interop_hosts{id="LABDEMO",status="OK",host="FHIR Transform",production="LABDEMO.Production"} 1
iris_interop_messages{id="LABDEMO",host="FHIR Transform",production="LABDEMO.Production"} 126
```

Run through Dev A's own parser, that is what the contract's canonical sample now says our production is:

```
contracts/samples/metrics-dump.txt           production: LABDEMO.Production
                                             hostCount: 15   application: 4
                                             Cloud API, EMR Source, FHIR Transform, Lab Router

fixtures/metrics-live-capture.txt (real)      production: ProductionGuardian.LabDemo.Production
                                             hostCount: 12   application: 3
                                             Cloud API, EMR Source, Lab Router
```

**Three independent reasons it isn't ours:**

**(a) `FHIR Transform` is in none of the three genuine captures.** Not the current one, not `metrics-live-capture-preRename.txt`, not the hand-written `metrics.txt`. The pre-rename capture — the oldest real snapshot, from before the host renames — shows the production's earlier state as:

```
ActivityReporter, EMRSource, LabRouter, PIDExtractProcess, PatientDemographicsOperation
```

`PIDExtractProcess` and `PatientDemographicsOperation`, not `FHIR Transform`. The only `FHIR` strings anywhere in the real captures are database names — `FHIRFACADE`, `FHIRFINAL`, `FHIRINTEGRATION` — unrelated IRIS databases, not interop hosts. This is exactly what #15 concluded: *"the LABDEMO production has no such host and never did."*

**(b) Its metrics are a copy of Lab Router's, with one digit changed.** All four application hosts in the file:

| host | messages | msg/sec | avg_proc | sample_count | last_activity |
|---|---|---|---|---|---|
| EMR Source | 21 | .2 | .01 | 12 | 4.947 |
| Lab Router | **126** | **1.2** | **.08** | **24** | **4.838** |
| **FHIR Transform** | **126** | **1.2** | **.08** | **24** | **4.839** |
| Cloud API | 42 | .4 | .05 | 24 | 4.841 |

EMR Source, Lab Router and Cloud API all differ from each other, so this pipeline does *not* produce uniform counts — which rules out the serial-pipeline explanation I went looking for. Yet FHIR Transform matches Lab Router **exactly on six of seven metrics** and differs on the seventh by exactly one millisecond. That is a duplicated block with `last_activity` nudged, not a measurement.

**(c) The production name matches no real capture.** The file says `production="LABDEMO.Production"`; all three genuine captures say `production="ProductionGuardian.LabDemo.Production"`.

**Why this is blocking rather than a nit:** it puts two samples that disagree about whether a host exists into the one directory whose entire purpose is that all three of us mock against the *same bytes*.

```
contracts/samples/hosts-response.json    3 hosts, no FHIR Transform    (post-#15)
contracts/samples/metrics-dump.txt       4 hosts, FHIR Transform       (this PR)
apps/dashboard/fixtures/*.json           3 hosts, no FHIR Transform    (post-#21)
```

I removed FHIR Transform from eight dashboard fixtures in #21 on the strength of #15. If this merges, the canonical proxy sample says I was wrong to. And `CONTRIBUTING.md` §4's promise — *"works against the mock predicts works against the real thing"* — inverts: an engine that conforms to this sample expects a fourth host that will never arrive.

Root `CLAUDE.md` §6 is the rule it lands on: **never invent data**, and `contracts/` is the worst place for it.

## 2. The byte-identity proof can't be reproduced by a reviewer

```
$ git hash-object C:/temp/real-metrics-capture.txt
537b6d5183501bc4ad2ceba05cabe5c9a8dd28b7
```

`C:/temp/` isn't in the repo, so this attests the copy step only, against a file nobody else can inspect. The in-repo real capture hashes differently, as it must — different snapshot:

```
contracts/samples/metrics-dump.txt                       537b6d51...
services/metrics-proxy/fixtures/metrics-live-capture.txt  c508f3b3...
```

For a `contracts/` sample the provenance chain has to terminate inside the repo. Otherwise "the real capture" is a claim, not a check — which is the distinction this PR is otherwise so good about.

## 3. The 4-host count is already baked into the prose

Not just the sample — `proxy-api.md:53`, `:186` and the `CHANGELOG.md` entry all hardcode `"applicationHostCount": 4` and `"production": "LABDEMO.Production"`. Both need to follow whatever replaces the capture. This is the fifth instance of the duplicated-fact problem you name in #28's description.

## 4. Heads-up on the fix — 3 of your 7 capture claims don't survive the swap

I tested your `CAPTURE_CLAIMS` regexes against the genuine capture so you know what you're walking into:

| claim | on the real capture |
|---|---|
| host label is `host`, with spaces | ✅ 9 hits |
| status is a label, reads OK | ✅ 12 hits |
| sample_count is its own family | ✅ |
| queued carries NO host label | ✅ 1 line, no `host` |
| **last_activity is elapsed seconds** | ❌ **0 hits — family absent entirely** |
| **hosttype rides on avg_processing_time as `actor`** | ❌ **0 hits — only `hosttype="service"`** |
| **messages_errored carries NO host label** | ❌ **0 hits — family absent entirely** |

The genuine capture was taken from an essentially idle production (`messages_per_sec` is `0` for every host), so three families never got emitted. **A real capture that exercises `last_activity`, `hosttype="actor"` and `messages_errored` needs to be taken with traffic flowing** — that's a re-capture against live IRIS, not a file swap. Worth knowing before you start.

### This affects #31's evidence, though probably not its conclusion

#31 rests on `grep "^iris_interop_messages_errored" metrics-dump.txt` → one line, no `host` label. That grep is against the file I'm disputing, and **the family appears in none of the genuine captures**, so the per-production-not-per-host claim is currently unevidenced rather than verified.

I'd still bet you're right — `queued` behaves exactly that way and *is* confirmed in the real capture — but #31 should say "inferred from `queued`'s behaviour, pending a capture with errors flowing" rather than "verified against the real capture." #12's evidence is unaffected and solid.

## Your two decisions — both agreed, and neither depends on the above

**`errored` over `messagesErrored` — agreed, publish it as `errored`.** Your blast-radius argument is the right tie-break, and here is the consumer's side of it: renaming to `messagesErrored` would cost `healthscan-api.md`, its schema, its `.d.ts`, two samples, my `guards.ts`, my `HostView` type and `HostCard`, to move a name that already means something downstream. Not worth one line of your `types/proxy.ts`. **Take the one-line change.**

**`queued: null` over `0` — agreed, and your recommended engine-side fix is the right one.** `null` is the honest wire value, rejecting the whole host over one unmeasurable field is clearly the worst of the three options, and having `queue_buildup` skip an unmeasurable host is consistent with how comparative rules already skip a warming baseline.

And yes — my #16 number still stands: `absoluteFloor: 50` against a captured depth of `48`, so the rule won't fire even once it's wired. Two separate milestones.

## One thing to hand back to me

Your `null` decision doesn't stop at your boundary, and I'd rather we notice that now than in the live run:

```ts
// contracts/healthscan.d.ts
queued: number;    // non-nullable
errored: number;   // non-nullable
```

Live, #12 and #31 together mean **both of these are unmeasurable per host**. If the engine satisfies a non-nullable contract by sending `0`, my `guards.ts:51` passes it straight through and `HostCard.tsx:53` renders **"Queued 0"** on every card — a measurement we don't have, displayed as one, which is the same §6 violation as above just one boundary further along.

My display layer is already right (`formatCount` renders `—` for `null`); it's the contract that won't let `null` through. **So: consider this a contract change request from the downstream consumer — `queued` and `errored` become `number | null` in `healthscan-api.md`, its schema and its `.d.ts`, with `null` meaning "not measurable on this host."** Then my side is a two-line change to stop coercing, and the grid honestly shows `—`.

That's a separate PR needing all three of us, and it should go with your #32 reconciliation rather than in here.

## What I verified

- `npm run validate` on this branch: **31 checks, all pass** — matches your claim exactly
- Zero diff on `healthscan-api.md`, `healthscan.schema.json`, `healthscan.d.ts`, `hosts-response.json`, `findings-response.json` — confirmed, nothing existing was touched
- `iris_interop_queued` in the *genuine* capture: one line, labels exactly `id` + `production`, no `host` — **#12 independently confirmed**
- `wc -l` = 1236 — your correction to #10's 1249 is right
- Both captures parsed with Dev A's real `buildSnapshot()`, rosters as tabulated above

## What I'd suggest

The contract document is the valuable part and I don't want it held hostage to a capture. Either:

1. **Re-capture from live IRIS with traffic flowing**, restate the three broken claims against it, and fix the 4→3 host count in the prose. Best outcome, needs an instance.
2. **Ship `proxy-api.md` + `proxy.schema.json` now, without `samples/metrics-dump.txt`** — drop the 7 `CAPTURE_CLAIMS`, keep the 24 schema checks, and note in `README.md` that the canonical capture is still outstanding alongside the `alerts.json` / `proxy-response.json` gap you already flagged. Unblocks #32 today.

I'd take (2) and file (1), given the timeline — but it's your call and I'll re-review either quickly.

Last thing, and I mean it as a compliment: your own *"What I could not verify"* section says **"the captures are real; the *instance* was not live during this work."** That is precisely the gap this fell into, and you had already written it down. The section did its job — it just needed one more question asked of it.

tanifgit (approved):
## Correction — my finding #1 was wrong, and I'm withdrawing the block

I said `metrics-dump.txt` "is not a capture of our LABDEMO production" and that the `FHIR Transform` block looked fabricated. **Both are wrong.** Your #29 has the answer, and I should have read it before reviewing this:

> *"the live instance runs an older 4-host definition (including `FHIR Transform`) than `Production.cls` on `main`"*

and, measured on that live instance in #29:

```
```

That independently corroborates both of the things I called suspicious — the production name **and** the fourth host. Two further checks confirm the file is a genuine scrape:

- **125 distinct metric families, 957 metric lines.** Nobody hand-writes that.
- The `last_activity` values I read as "a nudged digit" are a **coherent pipeline timeline**: Lab Router `4.838`, FHIR Transform `4.839`, Cloud API `4.841` — hand-off order, ~1 ms apart. That is the signature of a real serial pipeline, and it's the opposite of what a fabricated block would look like. Identical `messages: 126` and `sample_count: 24` between Lab Router and FHIR Transform is exactly what a serial pipeline produces, which is the explanation I went looking for and then failed to apply.

So: the capture is real, it is fit for its purpose (documenting label shapes), and my §1, §2 and §4 don't stand. **Please disregard them.** I've approved.

Two consequences worth pulling out, because they're more interesting than my error:

### The real finding is that we have two IRIS environments, not one

```
contracts/samples/metrics-dump.txt                        LABDEMO.Production                      15 hosts, 4 application
services/metrics-proxy/fixtures/metrics-live-capture.txt  ProductionGuardian.LabDemo.Production   12 hosts, 3 application
```

Different production names, different rosters, both real. The proxy fixture Dev A committed on 2026-08-11 came from a *different environment* than the one you have access to. That's worth knowing on its own — it means "verified against the live instance" has meant two different instances depending on who said it, and it explains why the `fixtures/` capture is idle (`messages_per_sec: 0` everywhere) while yours has traffic.

`fixtures/README.md` is the right place to record which capture came from where, if you're in there anyway.

### #15's premise looks wrong, and #21 was mine

PR #15 dropped `FHIR Transform` from `contracts/samples/` on the stated grounds that *"the LABDEMO production has no such host and never did."* The live instance says otherwise. I then removed it from eight dashboard fixtures in #21 citing #15, and approved #15 saying I'd verified every claim — I verified the internal arithmetic, not the premise.

The open question is which way the roster should converge: redeploy `Production.cls` to the live instance so it really is 3 hosts, or accept 4 and revert #15/#18/#21. **`Production.cls` on `main` is the source of truth and the live instance is stale, so I think it's the former** — but that's a real decision and it isn't this PR's to make. I'll file it, since two of the three PRs that acted on the wrong premise were mine.

None of that blocks publishing an accurate contract of what the proxy emits today, which is what this PR does.

### Still standing from my review

- **`errored` over `messagesErrored`** — agreed, take the one-line change.
- **`queued: null` over `0`** — agreed, and your engine-side skip-the-unmeasurable-host fix is the right shape.
- **The contract change request back to me**: `queued` and `errored` should become `number | null` in `healthscan-api.md` / schema / `.d.ts` so the engine isn't forced to send a `0` it can't measure and my grid can honestly render `—`. Separate PR, all three of us, best paired with your #32 reconciliation.
- The `C:/temp/real-metrics-capture.txt` hash is still unreproducible by a reviewer — minor now that the file's provenance is established elsewhere, but for a `contracts/` sample it's worth having the chain terminate in-repo.
- `samples/alerts.json` and `samples/proxy-response.json` still missing, as you noted yourself.

And my §4 has one salvageable part: the *other* capture can't stand in for this one, because `last_activity`, `hosttype="actor"` and `messages_errored` are all absent from it. That's an argument for keeping yours, not replacing it. Withdrawn as a criticism, offered as a reason your file is the better sample.

Sorry for the noise — the accusation was the wrong call and the evidence for it was sitting in your other PR.

Issue #31, opened 2026-08-12
elevated_error_rate cannot fire per host either: iris_interop_messages_errored has no host label


Ari-Glikman:
Found while writing `contracts/proxy-api.md` (#30). **This is a second finding type blocked by the same root cause as #12**, and it was unfiled.

## The metric is production-wide, not per host

Verified against the real 1236-line capture:

```
$ grep -E "^iris_interop_messages_errored" metrics-dump.txt
iris_interop_messages_errored{id="LABDEMO",production="LABDEMO.Production"} 0

$ grep -c -E "^iris_interop_messages_errored" metrics-dump.txt
1
```

**One line, no `host` label** — exactly like `iris_interop_queued` in #12. So the proxy publishes `errored: null` on every host, and `elevated_error_rate` cannot fire per host.

## Why this is worse than #12

For `queued`, the proxy at least publishes the real production-wide total as `_meta.productionQueued`, so the number isn't lost. **For `errored` there is no per-production total published at all** — it isn't in the parser's `SCALAR_FAMILIES`, so the figure is dropped entirely.

So the situation across the eight finding types, live:

| Finding type | Live status |
|---|---|
| `dead_host`, `stalled_host`, `slow_processing`, `growing_queue_wait`, `throughput_drop` | fire correctly |
| `queue_buildup` | blocked — #12 |
| `elevated_error_rate` | **blocked — this issue** |
| `system_alert` | blocked — engine reads alerts from the wrong payload (ours, see below) |

Five of eight work live. Demo mode is unaffected: all eight fire from fixtures, so the screencast and fallback are safe.

## The fix is the same as #12's, which makes it cheap

`Ens.Util.Statistics:EnumerateHostStatus` — the query I verified for #12 — returns per-host state including counts, and its `Name` column matches the metrics `host` label exactly (spaces intact, verified live: `Cloud API | status=Disabled | queue='70'`). So whichever data source closes #12 should close this at the same time rather than as separate work.

Worth deciding together for that reason: **#12's option (1) covers both, or neither.**

## Not blocking MVP 1

Demo mode reaches all eight types. This gates *live* mode, which is Day-4/5 work, and it gates it alongside #12 rather than independently.

Refs #12, #30, and PROXY-Q2 in `services/detection-engine/src/types/proxy.ts`.

Ari-Glikman:
Closed by #36, alongside #12 — one fix, as predicted when this was filed.

[...]

PR #33, opened 2026-08-12
fix(detection-engine): reconcile live mode to the ratified proxy contract

### LINKED ISSUE #32: Engine live mode is non-functional: our client rejects every host from the real proxy
**Ours (Dev B), entirely.** Surfaced by #30 while documenting the proxy contract, and verified against Dev A's merged proxy rather than reasoned about. `PROXY_MODE=live` does not work today, in five independent ways.

Demo mode is unaffected — all eight finding types fire from fixtures — so the screencast and Day-5 fallback are safe. This gates live mode only.

## 1–3. Our boundary guard rejects every host

The real object Dev A's proxy emits, run from the committed capture:

```json
{
  "host": "Lab Router",       "type": "process",   "status": "OK",
  "isFramework": false,       "queued": null,      "messages": 126,
  "messagesPerSec": 1.2,      "errored": null,     "avgProcessingTime": 0.08,
  "avgQueueingTime": 0,       "lastActivity": "2026-08-11T23:59:55.162Z",
  "lastActivityElapsedSeconds": 4.838
}
```

Against our guard:

```
our guard accepts the real proxy host? -> false

  name  (proxy sends "host")              -> false
  messagesErrored (proxy sends "errored") -> false
  queued as finite number (proxy: null)   -> false
```

Three independent mismatches. `isProxyHost()` returns false, our client log-and-skips the entry, and **the engine reports zero hosts** — the dashboard blanks rather than degrades, which looks like the production has vanished.

Note the guard is behaving *correctly*; it was written against assumptions (`PROXY-Q1`) that the real proxy didn't happen to match. That's exactly what the `PROXY-Q` markers existed to surface, and it worked — just later than ideal.

## 4. We request an endpoint that doesn't exist

```
ours:    GET {base}/api/metrics        -> 404
Dev A's: GET /proxy/metrics, /proxy/alerts, /proxy/health
```

## 5. We read alerts from the wrong payload

`client.ts` looks for `alerts` inside the metrics response. Dev A's proxy serves them from a **separate** `GET /proxy/alerts`, so `alerts` is always empty and **`system_alert` cannot fire live** — independently of #31.

## The fix

All in `services/detection-engine/src/`, none of it contentious:

- `types/proxy.ts` — rename `name` → `host`, `messagesErrored` → `errored`; accept `queued`/`errored` as `number | null` and treat `null` as unknown rather than rejecting the host
- `proxy/client.ts` — request `/proxy/metrics`, fetch `/proxy/alerts` separately, handle `warming: true`
- resolve the five `PROXY-Q` markers against the now-published contract (#30) and delete them

On `queued`/`errored` being nullable: I argued on #13 for the proxy coercing to `0` instead. Having seen #31 — where `errored` has **no** per-production total published — `null` is the more honest wire value. `0` would assert a measurement that doesn't exist. So we widen; the proxy shouldn't lie to satisfy our guard.

## Why this wasn't caught sooner

ADR 0004 predicted it in as many words: *"integration risk is deferred, not removed."* Our 133 tests all run against fixtures we wrote, so they prove the engine self-consistent and say nothing about whether the two services agree. The `contracts` CI job compares samples to schemas, but nothing compares **our client against the other service's real output** — which is the one check that would have caught all five.

Worth a follow-up beyond the fix: a CI step that runs the real proxy's parser over the committed `metrics-dump.txt` and feeds the result through our guard. That's the missing test, and it's cheap now that both the capture and the proxy are on `main`.

Refs #30, #31, #12, and the `PROXY-Q` markers in `services/detection-engine/src/types/proxy.ts`.

Ari-Glikman:
Closes #32. **Live mode was non-functional in five independent ways, all ours** — and none of them showed up in 133 passing tests.

## Verified, not reasoned about

One real host object from Dev A's merged proxy, run through our own guard:

```
our guard accepts the real proxy host? -> false
  name  (proxy sends "host")              -> false
  messagesErrored (proxy sends "errored") -> false
  queued as finite number (proxy: null)   -> false
```

`isProxyHost()` rejected **every** host, so the engine reported zero and the dashboard blanked rather than degraded. Plus we requested `/api/metrics` (404 — theirs is `/proxy/metrics`) and read alerts from the metrics body where they never appear, so `system_alert` couldn't fire live either.

## The test that was missing

`test/real-proxy.test.ts` runs **Dev A's actual parser over the actual captured IRIS output** and feeds the result through our guard. No fixture of ours on the input side.

Every other suite here runs against data we wrote. They prove the engine self-consistent and say nothing about whether the two services agree — which is exactly how our guard came to reject all 15 real hosts while everything stayed green.

Proven to catch the original bug rather than assumed. Reverting the guard to require a finite `queued`:

```
✖ accepts EVERY host the real proxy produces
  AssertionError: our guard rejected 15 of 15 real hosts — the engine would report
  zero and the dashboard would blank (#32)
```

It would have caught this the day the proxy landed.

## One deliberate narrow lie, documented at the site

`normalizeHost()` collapses a null count to `0` for the published `Host` shape, because `contracts/healthscan.schema.json` declares `queued` and `errored` as **required integers**. Emitting `null` there would do to you precisely what the proxy's `null` did to us — your guard would reject the host and the grid would blank.

It isn't load-bearing for detection: `#errorsPerMinute` reads the raw nullable value *before* this conversion and returns null when absent, so no rule compares against the fabricated zero.

@tanifgit — the honest fix is a contract change making both fields `number | null`, which is a PR to `contracts/` with all three reviewers. **Worth your view on whether that's wanted**, because it would mean your grid rendering `—` for those two fields in live mode. I'd rather ask than decide it for you, and `0` is the contract-conformant behaviour meanwhile.

## Also reverses a position I argued on #13

I'd pushed for the proxy coercing `queued` to `0`. Having seen #31 — where `errored` has **no** per-production total published anywhere — `null` is the more honest wire value. `0` would assert a measurement that doesn't exist. So we widened instead; the proxy shouldn't lie to satisfy our guard.

## Verification

```
tsc --noEmit strict     clean
tests                   139/139  (was 133; +6 real-proxy integration)
ADR 0004 no-proxy       hosts: 3 -> Cloud API, EMR Source, Lab Router
                        queued/errored emit as 0, not null
```

Proxy fixtures also renamed to the contract's field names and given `isFramework`, so our mock actually mocks what the real thing sends.

**Depends on #30** for `contracts/samples/metrics-dump.txt`. The new test skips gracefully until that merges — and a skip in CI is itself a signal, per the reasoning in your #24.

Entirely within `services/detection-engine/`. Refs #32, #31, #12, #30

### DIFF
--- a/services/detection-engine/fixtures/proxy/cloud-api-dead.json
+++ b/services/detection-engine/fixtures/proxy/cloud-api-dead.json
@@ -4,43 +4,46 @@
   "production": "LABDEMO.Production",
   "hosts": [
     {
-      "name": "EMR Source",
+      "host": "EMR Source",
       "type": "service",
       "status": "OK",
       "queued": 0,
       "messages": 240,
       "messagesPerSec": 0.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.01,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 4.284,
-      "sampleCount": 12
+      "sampleCount": 12,
+      "isFramework": false
     },
     {
-      "name": "Lab Router",
+      "host": "Lab Router",
       "type": "actor",
       "status": "OK",
       "queued": 0,
       "messages": 1440,
       "messagesPerSec": 1.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.08,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 4.277,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     },
     {
-      "name": "Cloud API",
+      "host": "Cloud API",
       "type": "operation",
       "status": "Disabled",
       "queued": 48,
       "messages": 392,
       "messagesPerSec": 0,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.05,
       "avgQueueingTime": 0.02,
       "lastActivityElapsedSeconds": 110,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     }
   ],
   "alerts": []
--- a/services/detection-engine/fixtures/proxy/error-storm-2.json
+++ b/services/detection-engine/fixtures/proxy/error-storm-2.json
@@ -4,43 +4,46 @@
   "production": "LABDEMO.Production",
   "hosts": [
     {
-      "name": "EMR Source",
+      "host": "EMR Source",
       "type": "service",
       "status": "OK",
       "queued": 0,
       "messages": 362,
       "messagesPerSec": 0.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 12
+      "sampleCount": 12,
+      "isFramework": false
     },
     {
-      "name": "Lab Router",
+      "host": "Lab Router",
       "type": "actor",
       "status": "OK",
       "queued": 0,
       "messages": 2172,
       "messagesPerSec": 1.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.08,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     },
     {
-      "name": "Cloud API",
+      "host": "Cloud API",
       "type": "operation",
       "status": "Error",
       "queued": 18,
       "messages": 712,
       "messagesPerSec": 0.3,
-      "messagesErrored": 210,
+      "errored": 210,
       "avgProcessingTime": 0.06,
       "avgQueueingTime": 0.05,
       "lastActivityElapsedSeconds": 2.2,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     }
   ],
   "alerts": [
--- a/services/detection-engine/fixtures/proxy/error-storm-3.json
+++ b/services/detection-engine/fixtures/proxy/error-storm-3.json
@@ -4,43 +4,46 @@
   "production": "LABDEMO.Production",
   "hosts": [
     {
-      "name": "EMR Source",
+      "host": "EMR Source",
       "type": "service",
       "status": "OK",
       "queued": 0,
       "messages": 362,
       "messagesPerSec": 0.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 12
+      "sampleCount": 12,
+      "isFramework": false
     },
     {
-      "name": "Lab Router",
+      "host": "Lab Router",
       "type": "actor",
       "status": "OK",
       "queued": 0,
       "messages": 2790,
       "messagesPerSec": 1.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.08,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     },
     {
-      "name": "Cloud API",
+      "host": "Cloud API",
       "type": "operation",
       "status": "Error",
       "queued": 24,
       "messages": 2790,
       "messagesPerSec": 0.28,
-      "messagesErrored": 400,
+      "errored": 400,
       "avgProcessingTime": 0.06,
       "avgQueueingTime": 0.05,
       "lastActivityElapsedSeconds": 2.2,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     }
   ],
   "alerts": [
--- a/services/detection-engine/fixtures/proxy/error-storm-4.json
+++ b/services/detection-engine/fixtures/proxy/error-storm-4.json
@@ -4,43 +4,46 @@
   "production": "LABDEMO.Production",
   "hosts": [
     {
-      "name": "EMR Source",
+      "host": "EMR Source",
       "type": "service",
       "status": "OK",
       "queued": 0,
       "messages": 362,
       "messagesPerSec": 0.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 12
+      "sampleCount": 12,
+      "isFramework": false
     },
     {
-      "name": "Lab Router",
+      "host": "Lab Router",
       "type": "actor",
       "status": "OK",
       "queued": 0,
       "messages": 2802,
       "messagesPerSec": 1.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.08,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.9,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     },
     {
-      "name": "Cloud API",
+      "host": "Cloud API",
       "type": "operation",
       "status": "Error",
       "queued": 31,
       "messages": 2802,
       "messagesPerSec": 0.25,
-      "messagesErrored": 620,
+      "errored": 620,
       "avgProcessingTime": 0.06,
       "avgQueueingTime": 0.05,
       "lastActivityElapsedSeconds": 2.2,
-      "sampleCount": 24
+      "sampleCount": 24,
+      "isFramework": false
     }
   ],
   "alerts": [
--- a/services/detection-engine/fixtures/proxy/error-storm.json
+++ b/services/detection-engine/fixtures/proxy/error-storm.json
@@ -4,43 +4,46 @@
   "production": "LABDEMO.Production",
   "hosts": [
     {
-      "name": "EMR Source",
+      "host": "EMR Source",
       "type": "service",
       "status": "OK",
       "queued": 0,
       "messages": 360,
       "messagesPerSec": 0.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0,
       "avgQueueingTime": 0,
       "lastActivityElapsedSeconds": 3.8,
-      "sampleCount": 12
+      "sampleCount": 12,
+      "isFramework": false
     },
     {
-      "name": "Lab Router",
+      "host": "Lab Router",
       "type": "actor",
       "status": "OK",
       "queued": 0,
       "messages": 2160,
       "messagesPerSec": 1.2,
-      "messagesErrored": 0,
+      "errored": 0,
       "avgProcessingTime": 0.08,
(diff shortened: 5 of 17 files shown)


Ari-Glikman:
CI green, and worth flagging exactly what it did and didn't verify — since the difference is the whole point of this PR.

```
detection-engine  # tests 133   # pass 133   # fail 0   # skipped 0
ok 26 - our guard against the real proxy output # SKIP metrics-proxy parser or
        metrics-dump.txt not present
```

**133, not the 139 I get locally.** The new real-proxy suite skipped, because `contracts/samples/metrics-dump.txt` is on #30's branch and not yet on `main`. So the one test that actually proves the two services agree did **not** run in CI here.

That's the designed behaviour and I'd rather say so than let a green tick imply otherwise. Two notes:

- The skip is **explicit in the log** (`# SKIP` with the reason) rather than silent, which is the standard @tanifgit argued for in #24 — a check that can't run should say so out loud.
- It'll run for real once #30 merges. **Worth watching that specific line on the first PR after** — if it still says SKIP, something's wrong with the path resolution rather than with the inputs.

Locally, with the capture staged, it passes 6/6 and I confirmed it fails on the original defect:

```
✖ accepts EVERY host the real proxy produces
  AssertionError: our guard rejected 15 of 15 real hosts
```

**Merge order:** #30 first, then this. Either order is safe — this PR is correct regardless — but #30 first means the integration test is live from the moment this lands rather than one PR later.

tanifgit (changes requested):
**Answering your question first: yes, the contract change is wanted, and I'll take my side of it.** `Host.queued` and `Host.errored` → `number | null`, `null` meaning "not measurable on this host." My grid already renders `—` for null (`formatCount`), so it's a two-line change in `guards.ts` to stop coercing. I'd rather show `—` than a zero we didn't measure.

But **requesting changes**, because `orZero()` is load-bearing for detection in two ways the comment above it rules out, and I found them by probing rather than reading. The five fixes and `real-proxy.test.ts` are all good work — this is about the one deliberate lie.

## 1. Blocking — `orZero` makes a false `throughput_drop` reachable

Warm a real baseline, then let the metric go absent. Throughput never changes:

```
after 20 healthy polls at 1.2 msg/s : []
after 2 polls with messagesPerSec=null: ["throughput_drop"]
   -> throughput_drop | Throughput 0.0 msg/sec is 100% below baseline
```

A finding reporting a total collapse of a production that is running perfectly. `sustainedSamples: 2`, so two polls — 20 seconds — is all it takes.

**And it's reachable through two paths that are documented in Dev A's parser, not hypothetical:**

```js
// parser.js:81
function parseValue(str) {
  if (str === 'NaN' || str === '+Inf' || str === '-Inf') return null;
```

`messages_per_sec` is a *rate*. `NaN` from a zero-length sample window — right after a production restart, which is exactly when someone is watching the dashboard — becomes `null`, then `0`, then "100% below baseline."

Worse, the family-level case:

```js
// parser.js:264
// Every numeric field starts null, not 0. IRIS omits whole families rather than
// emitting zeros (see the header note), and `0` has to keep meaning "measured
// zero" or every comparative rule downstream reasons about invented data.
```

If `iris_interop_messages_per_sec` is omitted from one scrape, **every host gets `null` → `0` → every host reports `throughput_drop` in the same poll.** The dashboard fills with critical-looking findings about a healthy production.

That parser comment is the argument against `orZero` in Dev A's own words, written for this exact reason. The pipeline preserves `null` end to end and then discards it in the last function before the rules run.

**Note the asymmetry**, because it's the cleanest way to see the bug: `orZero` is *safe* for every rule where higher is worse — `slow_processing` and `growing_queue_wait` see `0`, fall under their floors, and correctly stay silent. It is *unsafe* for the one rule where **lower is worse**. `throughput_drop` is the only rule that reads a coerced zero as a symptom rather than as "nothing to see."

**Fix is the pattern you already used:** `throughput_drop` should skip when the raw `messagesPerSec` is `null`, the way `#errorsPerMinute` returns `null` on a null `errored`. One guard, and it makes the comment above `orZero` true.

## 2. `stalled_host` cannot fire live either — a sixth break, not in #32's five

```
stalled_host config: {"enabled":true,"inactiveSeconds":300,"requiresQueued":true,...}
Host idle 400s, status OK, in both cases:

queued: null  (what live sends)    Host.queued=0     findings=[NONE]
queued: 5     (measurable depth)   Host.queued=5     findings=[stalled_host]
```

`rules/index.ts:61` — `if (rule.requiresQueued && host.queued <= 0) return null;` — reads the **normalized** `Host`, after `orZero`. Since `queued` is `null` on every host live, `requiresQueued` is never satisfied and the rule is silently switched off.

**So the live tally is 4 of 8, not 5.** #31's table lists `stalled_host` under "fire correctly", and #32 lists five breaks; this is a sixth, and it's the same root cause as `queue_buildup` (#12) rather than a new one.

To be fair to the behaviour: **silently skipping may be the right call.** `requiresQueued` exists to stop an idle-but-empty host reading as stalled, and firing on an unknown depth risks exactly the false positive MVP §6 names as the top risk. My objection is that it's an accident rather than a decision — undocumented, contradicting #31's table, and contradicting the "not load-bearing for detection" claim. Pick one and write it down.

Either way this is the argument that turns my contract request from cosmetic into structural: with `queued: number | null` reaching the rules, `stalled_host` can distinguish *"nothing queued"* from *"depth unknown"* and say which one it's doing.

## Verified independently

- **139/139 confirmed** — but only with the capture present. On this branch as pushed it's **133 with the new suite skipped**, since `contracts/samples/metrics-dump.txt` doesn't exist until #30 merges. I supplied it from `origin/devB/proxy-contract` and got your 139. Claim accurate, just not reproducible on the branch alone.
- **The negative test does what you say.** Reverting the guard to require a finite `queued`:
  ```
  ✖ accepts EVERY host the real proxy produces
    AssertionError: our guard rejected 15 of 15 real hosts — the engine would report
    zero and the dashboard would blank (#32)
  ```
  5 of the 6 new tests fail. It would have caught #32 on day one. Restored afterwards.
- `#errorsPerMinute` does read the raw nullable before `normalizeHost` — that specific claim holds. It's the only rule for which it does.
- `types/proxy.ts` matches `proxy-api.md`: `host`, `errored`, `NullableCount`, `isFramework`, `Unknown` status, `actor`→`process`.

`real-proxy.test.ts` is the right test and the right reasoning — running Dev A's actual parser over the actual capture with no fixture of ours on the input side is what none of the other 133 do. It's the thing ADR 0004 said we'd owe.

## Two smaller notes

**`npm test` reports `skipped 0` while the suite skips.** The skip line is there mid-output (`﹣ our guard against the real proxy output # metrics-proxy parser or metrics-dump.txt not present`) but the summary reads:

```
ℹ tests 133 / pass 133 / fail 0 / skipped 0
```

Anyone tailing CI sees 133 pass, nothing skipped, and concludes the integration test ran. That's the #24 pattern you cite in the PR body — green on nothing — landing in the summary line rather than the step. Worth a `--test-reporter` tweak or an explicit assertion that the inputs exist once #30 is in.

**The test pins the 4-host roster, which #34 will move.** Two assertions hardcode `['Cloud API', 'EMR Source', 'FHIR Transform', 'Lab Router']`, and the comment says *"the capture predates the FHIR Transform removal."* Per #34 that framing isn't quite right — the capture is from a **different instance** that still runs the 4-host definition, not an older snapshot of the current one. Asserting the set rather than a count is the right instinct (#19), but flag it as roster-dependent so whoever refreshes the capture knows to expect it, or key it off `_meta.applicationHostCount`.

## Sequencing

#30 and #29 are both approved, so this can land right behind #30. I'd take the `throughput_drop` guard here since it's a few lines and it's the one that could embarrass the demo; the `stalled_host` decision and the contract PR can follow separately. Happy to re-review quickly.

I'll open the `contracts/` PR for the nullability change and request you and Dev A's slot on it — say the word if you'd rather fold it into your own branch.

tanifgit:
Opened the contract change as #35 rather than folding it into this branch, since `contracts/` needs Dev A's slot too and that shouldn't gate your five fixes.

**`Host.queued` and `Host.errored` → `integer | null`**, keys still required. It doesn't change what this PR does — only whether `orZero` still has to exist. The two reproductions from my review are in the changelog entry so the reasoning outlives the thread.

Two notes over there that are yours, both raised rather than fixed:

- `src/types/healthscan.ts` still says `queued: number`, and `contract-drift.test.ts` **can't see the divergence** — it compares string-literal unions and required-key presence, never a field's *type*. 133/133 stays green on #35 with the transcription stale.
- `messagesPerSec` has the identical argument (`parser.js:81` maps `NaN` → `null`, and a rate right after a restart is exactly that), but widening the metric `throughput_drop` is built on is a bigger call than mine to make.

If you'd rather have the engine half here and only `contracts/` there, say so and I'll split it that way. Happy to re-review this one as soon as the `throughput_drop` guard lands.

Ari-Glikman:
Both fixed in 53d2160. **You were right on both, and reproducing them first changed what I built** — the second one especially, because my first attempt at proving it didn't work.

## 1. Blocking issue — reproduced exactly

```
after 20 healthy polls at 1.2 msg/s : []
after 2 polls with messagesPerSec=null:
  critical throughput_drop | Throughput 0.0 msg/sec is 100% below baseline
```

A critical finding claiming total collapse of a production running perfectly. Your asymmetry framing is the cleanest way to see it and I've made it a test: `throughput_drop` is the **only** comparative rule where lower is worse, so it's the only one for which a coerced 0 reads as a symptom. The duration rules fall under their floors and correctly stay quiet.

And you're right that the parser's own comment is the argument against `orZero` in Dev A's words. The pipeline preserved `null` end to end and I discarded it in the last function before the rules.

**Fix:** rules now get the raw nullable values alongside the normalized `Host`, so they can tell "measured zero" from "not measurable". `throughput_drop` returns null when the rate is absent. A real zero still fires — the guard suppresses absence, not detection, and that's a test too.

## 2. `stalled_host` — and I need to be precise about what I changed

You're right it's a **sixth** live break, and right that my #31 table listing `stalled_host` under "fires correctly" was wrong. **The live tally is 4 of 8, not 5.** I'll correct #31.

But being exact, because I checked rather than assumed: **reading `raw.queued` instead of `host.queued` changes no behaviour at all.** `host.queued` is `orZero(raw.queued)`, and `null <= 0` is `true` in JavaScript — so both forms decline on an unknown depth and both fire on a real one.

So what landed is what you actually asked for rather than a behaviour fix: the intent is explicit and the reason is commented. **The decision recorded:** when the depth is unknown we decline, because an idle host with an unknown queue is more likely quiet than hung, and firing on absent data is the false positive §6 names as the top risk. Your words — *"silently skipping may be the right call ... my objection is that it's an accident rather than a decision"* — are why it's now a decision.

## One test that didn't work, and why the fix is better for it

I verified by injection rather than trusting green, and the result was instructive:

```
remove the throughput guard          -> rule test FAILS, engine test FAILS   ✓
revert stalled_host to host.queued   -> rule tests all still PASS            ✗
```

My rule-level tests **could not see defect 2**. The helper defaults `raw` to mirror the normalized host, so at that level the two code paths are indistinguishable — the defect lives in what `normalizeHost()` hands the rules, not in the rule.

So I added 4 engine-level tests that drive the full poll path, where it *is* visible. That's the same lesson as `real-proxy.test.ts` one level down: a test that supplies both sides of the boundary it's checking cannot check that boundary.

**143/143**, tsc clean.

## On #35 — thank you, and it supersedes the workaround

> I'd rather show `—` than a zero we didn't measure.

Agreed, and that's the right instinct. #35 removes the need for `orZero` entirely, which means the `raw` plumbing here becomes belt-and-braces rather than load-bearing. I'd keep it anyway — a rule asking "was this measured?" shouldn't depend on the wire type staying nullable forever — but the deliberate lie goes away, which is the outcome I wanted when I asked.

This fix is correct either way and doesn't block #35. Reviewing that one next.

PR #35, opened 2026-08-12
feat(contracts)!: Host.queued and Host.errored become number | null

tanifgit:
The contract change I offered on #33. **`Host.queued` and `Host.errored` widen from `integer` to `integer | null`**, where `null` means *"not measurable for this host"* and never *"zero"*. Both keys stay **required** — a `null` value is legal, an absent key is not.

Per `contracts/README.md` this needs all three of us. @Ari-Glikman it's your contract, @kskubach it's your parser's semantics being preserved.

## Why this belongs in `contracts/` and not in the engine

The engine's `normalizeHost()` collapses the proxy's `null` to `0` **because this schema declared the field a required integer**. #33 documents it as "one deliberate narrow lie" and is right that emitting `null` against the current schema would blank Dev C's grid. So the coercion is downstream of this file, and so is the fix.

It also isn't inert, which is the part #33's comment rules out. Two reproductions, by probing:

```
after 20 healthy polls at 1.2 msg/s  : []
after 2 polls with messagesPerSec=null: ["throughput_drop"]
   -> throughput_drop | Throughput 0.0 msg/sec is 100% below baseline
```

```
Host idle 400s, status OK:
  queued: null  (what live sends)  -> Host.queued=0  -> findings=[NONE]
  queued: 5     (measurable depth) -> Host.queued=5  -> findings=[stalled_host]
```

A critical-looking finding about a production running perfectly, and a rule silently switched off. @kskubach — `parser.js:264` already carries the argument in your words (*"`0` has to keep meaning 'measured zero' or every comparative rule downstream reasons about invented data"*). The pipeline preserves `null` end to end and then discards it in the last function before the rules run. This lets it reach the consumer.

**Supersedes the "Known gap" note of 2026-08-06**, which kept `queued` a required number on the grounds that per-host depth is available from `EnumerateHostStatus`. Still true of IRIS — it's how the measured `48` in the samples was obtained — but the shipped proxy reads the metrics text only, so that note's own condition (*"no contract impact if that holds"*) didn't hold.

## What changed

| File | Change |
|---|---|
| `healthscan.schema.json` | `queued`, `errored` → `["integer", "null"]`. `minimum: 0` kept — draft-07 `minimum` doesn't constrain `null`. |
| `healthscan.d.ts` | `number \| null` on both, with the "not measurable ≠ zero" note at the site. |
| `healthscan-api.md` | §1 table, an explicit present-but-null sentence, **Q13** in §4, and §4.1 records this as the second contradiction of a Dev C assumption after Q1. |
| `validate.mjs` | 11 checks → 14. |
| `CHANGELOG.md` | The entry, with both reproductions and the superseded note. |
| `apps/dashboard/**` | My side: guards stop coercing, transcription widens, `HostCard` emphasis reads null as unknown. |

**Samples are deliberately unchanged.** They carry measured LABDEMO values — `queued: 48` on a disabled Cloud API was genuinely observed. Rewriting them to `null` would swap a real number for a synthetic one and move the bytes your fixtures and my eight scenarios are anchored to, mid-sprint, for no gain. The null shape lives in `validate.mjs` instead, which is where a shape with no measured instance belongs.

## Verified

```
node contracts/validate.mjs        -> 14/14 (2 samples, 5 accept, 7 reject)
dashboard  npm run typecheck       -> clean
dashboard  npm run build           -> dist/index.html, 201.89 kB
dashboard  validate:fixtures:strict-> 8 fixtures ok, all findings engine-emittable
detection-engine typecheck + test  -> 133/133 pass, unchanged by this
```

**The new accept case bites.** Reverting the schema to `"type": "integer"`:

```
FAIL  accepts: host with unmeasurable queued and errored counts
1 check(s) failed
```

The two reject cases matter as much: widening to admit `null` must not admit a **string** count, or a **missing key**. Both are new and both reject.

## Two things for you, not fixed here

**1. `src/types/healthscan.ts` now diverges from the contract, and `contract-drift.test.ts` cannot see it.** The engine still declares `queued: number`. The drift guard compares string-literal unions, asserts required keys are emitted, and rejects forbidden keys — it never compares a field's *type*, so a widening passes silently. 133/133 is green on this branch with the transcription stale. Your file, your call, but it's the same shape as the gap you found on #33: tests that prove self-consistency and say nothing about agreement.

**2. `messagesPerSec` has the identical argument and I deliberately left it alone.** It's a *rate*, and `parser.js:81` maps `NaN`/`Inf` to `null` — a zero-length sample window right after a production restart yields exactly that. If the engine keeps coercing it, my grid prints a measured-looking `0.0` msg/sec for a host whose throughput is simply unknown. That's this same defect in the UI rather than in a rule. Widening it is a bigger change than this PR — it's the metric `throughput_drop` is built on — so I'm raising it rather than deciding it.

Whether `stalled_host` should **skip or fire** on an unknown depth stays yours too. This change only makes the choice expressible; #33's review asks for it to be written down either way.

## Sequencing

Independent of #29, #30 and #33 — no file overlap, and it doesn't change what #33 does, only whether #33's `orZero` still has to exist. Land it in whatever order suits. If you'd rather fold the engine half into #33 than take a second PR, say so and I'll drop the `contracts/` half onto your branch instead.


tanifgit:
Updated for #36, and dropped @kskubach's review request — Dev A has moved off the project, so that slot can't be filled and shouldn't read as outstanding. Per #36's own note the `contracts/` gate is @Ari-Glikman and me in practice.

**#36 falsified the premise I wrote this on, so I re-anchored it rather than leaving a stale rationale in the contract.** Q13 said per-host depth is unavailable and therefore we publish `null`; #36 makes it available via a host-status endpoint merged on host name. Corrected in Q13, the `Host.queued` type comment, and the `validate.mjs` case comment.

**The change is still needed, and #36 is the reason it's cleaner than when I opened it.** `null` moves from *the norm* to *the documented exception* — a host missing from the endpoint's response (`_meta.hostStatus.unmatchedHosts`), the endpoint unreachable or 404 on a missing CSP application, the third poll failing, or the merge switched off via an empty `IRIS_HOSTSTATUS_PATH`. #36 holds that invariant explicitly on its side (*"a host the endpoint did not describe keeps `null`, not `0`"*) and types both counts as `NullableCount` in the proxy contract. Without this PR the **published** contract stays the one place in the chain that cannot represent what the proxy takes care to preserve, and the engine has to flatten it on the last hop.

So the two agree rather than compete — and the reproductions in the changelog stay reachable after #36, on any host the merge misses.

**Merge mechanics, checked rather than assumed.** I merged your branch into mine locally: the only conflict is `CHANGELOG.md`, both of us prepending an entry. `validate.mjs` merges cleanly and the merged validator passes **43 checks (2 samples, 15 accept, 19 reject, 7 capture claims)** — your cases and my three coexist. Either order works; whoever lands second takes the changelog conflict, and it's a two-entry ordering, not a real disagreement.

Re-verified on the updated branch: `node contracts/validate.mjs` 14/14, dashboard `typecheck` clean, `validate:fixtures:strict` green.

Ari-Glikman (approved):
**Approving as the other remaining developer** — per #39 the `contracts/` gate is now the two of us, and GitHub won't let you approve your own PR, so this is the gate doing its job rather than a formality.

Reviewed in an isolated worktree with the rebase applied, since it's `DIRTY` on `main`.

## Verified rather than read

```
$ npm run validate            all checks passed (2 samples, 10 accept, 15 reject, 7 capture claims)
$ npm run build               dist/index.html 201.89 kB, single file
```

And I probed the schema directly, because "nullable" is easy to over-widen:

```
CORRECT  accept  queued: null
CORRECT  accept  errored: null
CORRECT  reject  queued: -1
CORRECT  reject  queued: "0"
CORRECT  reject  messagesPerSec null      <- correctly NOT widened
CORRECT  reject  queued key MISSING
```

Exactly two fields, `minimum: 0` preserved, both still `required`. `"Every field is always present. A missing key is a contract violation, a null value is not"` is the right line to have written down.

## It's a widening, so nothing forces our hand

```
engine output TODAY (orZero -> 0):     still valid
engine output AFTER dropping orZero:   valid
```

Both accepted, and our seven `contract-drift.test.ts` assertions still pass against the new schema unchanged. So this can merge before or after #33 with no coordination — worth stating, since a `feat!` marker usually implies a lockstep change and here it doesn't.

## Leaving `messagesPerSec` non-nullable was the right call

Worth calling out because it's the field behind the blocking defect you found on #33. If it *were* nullable on the wire, the false `throughput_drop` would become reachable through the published contract rather than only through our internal coercion — and a consumer rendering `—` for throughput is a much worse story than one rendering it for a queue depth.

The asymmetry you identified there is the reason, and it holds here: `queued` and `errored` are counts a consumer can honestly display as unknown; a *rate* going unknown is indistinguishable from a rate going to zero unless the consumer is careful.

## Q13 is the part I'd point a newcomer at

> **`null` means "not measurable for this host", never "zero".** ... Consumers render `—`; **rules must skip rather than compare.**

That's precisely the guard I built on #33 after your review, written as contract text rather than as a comment in one engine file. It generalises the fix instead of just recording it.

## What I'll do next, so it's not a surprise

Once this and #33 are both in, I'll drop `orZero()` from `normalizeHost()` and emit `null` — the deliberate lie I flagged when I asked you this question. Two consequences for you:

1. Your grid will render `—` for `queued`/`errored` in **live** mode against LABDEMO, because #12/#31 mean those genuinely aren't measurable per host until #36 lands. Demo mode is unaffected.
2. The `raw` plumbing on #33 becomes belt-and-braces rather than load-bearing. I'm keeping it: a rule asking "was this measured?" shouldn't depend on the wire type staying nullable forever.

## One small note

The rebase conflicted only on `contracts/CHANGELOG.md`, where we'd both added a same-day entry at the top. I kept both, yours first as the later change. Nothing substantive — just flagging so you're not surprised by the ordering when you rebase.

tanifgit:
Rebased onto main now that #29, #30, #33, #40 and #41 are in. Only `CHANGELOG.md` conflicted, as predicted — two entries both wanting the top slot, mine above the proxy-contract one. **`MERGEABLE` / `CLEAN`, all five checks green** including the new `iris — compile ObjectScript` from #41.

```
contracts        all checks passed (2 samples, 10 accept, 15 reject, 7 capture claims)
dashboard        tsc clean · 8 fixtures ok · build 201.89 kB
detection-engine 149 pass / 0 fail · tsc clean
```

## Confirming your #33 fixes landed, since I reviewed them pre-merge

Both of mine are on main and better than what I asked for:

- `throughput_drop` now reads `raw.messagesPerSec` and returns null, with the two routes to null written out at the site (`NaN` over a zero-length window, and IRIS omitting whole families).
- `stalled_host` splits `raw.queued === null` from `raw.queued <= 0` and **states the decision** rather than leaving it emergent — *"Deliberate choice, not an accident: when the depth is unknown we still decline to fire."* That's exactly the "pick one and write it down" I was after, and declining is the right pick given MVP §6.

## What merging this creates for you — one deliberate loose end

`orZero`'s own doc comment on main is now the thing that goes stale:

> *`contracts/healthscan.schema.json` declares `Host.queued` and `Host.errored` as REQUIRED integers, and Dev C's grid renders them unconditionally. Emitting `null` against a required-integer field would do to Dev C exactly what the proxy's `null` did to us.*

That was true and is the reason I brought the change here instead of asking you to change the engine. Once this lands it stops being true for those two fields, and `orZero(proxyHost.queued)` / `orZero(proxyHost.errored)` can go — my guards no longer coerce and `formatCount` renders `—`, so the grid won't blank. **Keep it for `avgProcessingTime` / `avgQueueingTime`**, whose contract fields stay non-nullable.

Not asking you to do it in the same PR, just flagging that the comment and the code should move together so the next reader doesn't take the justification at face value. Note `contract-drift.test.ts` still won't catch it — it compares unions and key presence, never field types, which is the gap I raised on this PR and is still open.

## And `messagesPerSec` is now demonstrable rather than theoretical

`normalizeHost` still does `messagesPerSec: orZero(proxyHost.messagesPerSec)`, and `ProxyHost.messagesPerSec` is `NullableCount`. So after your #33 fix the *rule* correctly stays silent on an unmeasurable rate — but the *published value* is still a fabricated `0`, and my grid prints **`0.0` msg/sec** for a host whose throughput is simply unknown. Same defect, moved from the finding to the tile.

`formatRate` already renders `—` for null, so the consumer side is ready whenever you want it. It's a bigger call than this PR because `throughput_drop` is built on that metric, so it's yours — I'd just rather it be a decision than an inheritance.

**Ready to merge whenever you are.** Happy to land it myself if you'd prefer, and #36 can go either before or after — the changelog is the only place they touch the same lines.

Issue #34, opened 2026-08-12
LABDEMO roster: the live instance has 4 application hosts, Production.cls declares 3 — #15's premise was wrong


tanifgit:
**The running LABDEMO instance has four application hosts. `Production.cls` on `main` declares three.** PRs #15, #18 and #21 removed the fourth on the premise that it never existed, and the live instance contradicts that premise.

I raised this while reviewing #30 and #29. **Two of the three PRs that acted on the wrong premise were mine, and I approved the third**, so I'm filing it rather than leaving it in a review thread.

## The divergence

| Source | Production name | Application hosts |
|---|---|---|
| `iris/labdemo/Production.cls` (`main`) | `ProductionGuardian.LabDemo.Production` | 3 — EMR Source, Lab Router, Cloud API |
| `contracts/samples/metrics-dump.txt` (#30, real capture) | `LABDEMO.Production` | **4** — + FHIR Transform |
| `services/metrics-proxy/fixtures/metrics-live-capture.txt` (real capture) | `ProductionGuardian.LabDemo.Production` | 3 |
| `contracts/samples/hosts-response.json` | — | 3 |
| `apps/dashboard/fixtures/*.json` | — | 3 |

Both captures are real. They are from **two different IRIS environments** — which is the part nobody had noticed, and it means "verified against the live instance" has meant two different instances depending on who said it.

@Ari-Glikman established the live side independently in #29:

> *"the live instance runs an older 4-host definition (including `FHIR Transform`) than `Production.cls` on `main`"*

and measured on it:

```
iris_interop_queued{id="LABDEMO",production="LABDEMO.Production"} 0
```

The 4-host capture is also unambiguously genuine — 125 metric families, 957 metric lines, and a coherent pipeline hand-off timeline (`Lab Router` 4.838 → `FHIR Transform` 4.839 → `Cloud API` 4.841). I initially read the identical `messages: 126` / `sample_count: 24` between Lab Router and FHIR Transform as a duplicated block; it is what a serial pipeline produces. Corrected on #30.

## Why #15's premise doesn't hold

#15 dropped `FHIR Transform` from `contracts/samples/` stating:

> *"the LABDEMO production has no such host and never did — the HL7→PID transform is a DTL inside Lab Router, and a DTL is not a config item, so IRIS emits no `host` label for it."*

[...]

PR #36, opened 2026-08-12
feat: per-host queue depth and error counts — closes #12 and #31

### LINKED ISSUE #12: PROXY-Q2: per-host queue depth is not in /api/monitor/metrics — queue_buildup cannot fire

Ari-Glikman:
## Who and why

**Dev A (`kskubach`) has moved to other work; Dev B has taken over their outstanding tasks.** This
change is in `iris/**` and `services/metrics-proxy/**` — Dev A's area — plus an amendment to the
proxy contract. Work is attributed to Dev B (`Ari-Glikman`); nothing impersonates Dev A.

## What this closes

| Issue | Status |
|---|---|
| **#12** — `queue_buildup` cannot fire: `iris_interop_queued` has no `host` label | **Closes.** Per-host depth is published and merged. |
| **#31** — `elevated_error_rate` cannot fire: `iris_interop_messages_errored` has no `host` label | **Closes.** Per-host error counts published and merged. |

This is **option (1) on #12** — the one Dev A proposed and Dev C endorsed: a small REST endpoint in
`iris/` exposing what the metrics text lacks, merged by the proxy on host name.

**What it does NOT close: either finding actually firing.** See "Not verified" below. That
distinction is written into the contract itself (§6.1, §6.3) rather than left to be discovered.

## Real JSON from the new endpoint

`GET /labdemo/monitor/hoststatus`, live LABDEMO, `HTTP/1.1 200 OK`, `Content-Type: application/json`:

```json
{
  "hosts": [
    {"host":"Cloud API","status":"OK","queued":0,"errored":0,"messageCount":17860},
    {"host":"EMR Source","status":"OK","queued":0,"errored":0,"messageCount":8930},
    {"host":"Ens.Activity.Operation.Local","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"Ens.Actor","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"Ens.ScheduleHandler","status":"OK","queued":0,"errored":0,"messageCount":6},
    {"host":"Ens.ScheduleService","status":"OK","queued":0,"errored":0,"messageCount":3},
    {"host":"EnsLib.Background.Process.ExportMessageSearch","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"EnsLib.Background.Service","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"EnsLib.Background.Workflow.Operation","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"EnsLib.Testing.Process","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"EnsLib.Testing.Service","status":"OK","queued":0,"errored":0,"messageCount":0},
    {"host":"FHIR Transform","status":"OK","queued":0,"errored":0,"messageCount":53580},
    {"host":"Lab Router","status":"OK","queued":0,"errored":0,"messageCount":53580}
  ],
  "_meta": {
    "sampledAt": "2026-08-12T10:56:21.853Z",
    "production": "LABDEMO.Production",
    "productionState": "Running",
    "hostCount": 13,
    "erroredAvailable": true
  }
}
```

And the proxy running against **live IRIS**, merging it:

```
[poller] host-status: /labdemo/monitor/hoststatus (per-host queued/errored)
[poller] metrics: 15 hosts polled at 2026-08-12T11:07:04.051Z, 13 merged from host-status
```

```json
{"host":"Cloud API",     "queued":0,"errored":0,"messages":18120,"messagesPerSec":0.4}
{"host":"EMR Source",    "queued":0,"errored":0,"messages":9060, "messagesPerSec":0.2}
{"host":"FHIR Transform","queued":0,"errored":0,"messages":54360,"messagesPerSec":1.2}
{"host":"Lab Router",    "queued":0,"errored":0,"messages":54360,"messagesPerSec":1.2}

"_meta": {"hostStatus": {"shape":"hosts","merged":13,"unmatchedHosts":[],
                         "productionState":"Running","erroredAvailable":true}}
```

**Those zeros are measurements.** Before this they were `null` on every host.

## Four things measured, not assumed

Each would have shipped a plausible wrong answer:

1. **`Queue` is the empty string for an idle host, not `"0"`.** From the shipped source of
   `EnumerateHostStatusFetch`: `If tQueueCount=0 Set tQueueCount=""`. So empty is a *measured zero*
   and `+` is the right coercion — a truthiness test inverts the meaning.
2. **`TargetConfigName` comes back UPPERCASED from a `GROUP BY`** (`CLOUD API`), which cannot join
   the metrics `host` label. It is a `%SQLUPPER` collation artifact of the grouping key, not a
   property transform. `%EXACT(TargetConfigName)` returns the stored case. Case-folding to
   compensate was tried and matched **only 4 of 13 hosts**.
3. **`Status = 8` is `Error`** — from the compiled property's VALUELIST/DISPLAYLIST. Filtering on
   `Status = 'Error'` matches nothing, silently.
4. **`%response.Output` does not exist on IRIS 2026.1.** Copying `PatientDispatcher`'s `WriteJSON`
   gave a real HTTP 500 on the first live call:
   `<PROPERTY DOES NOT EXIST>WriteJSON+2^...HostStatusDispatcher.1 *Output`.
   `%ToJSON()` with no argument writes to the CSP output stream. **`PatientDispatcher` has the same
   latent defect on its own routes** — noted, not fixed, out of scope here.

Also caught only by calling it: `$zdatetime(...,3,7)` already appends the `Z`, so the first version
emitted `...:54ZZ`. JS `Date` accepts that, so it would have reached the samples unnoticed.

## Design decisions worth reviewing

- **The join is exact — no trimming, no case folding.** `EnumerateHostStatus`'s `Name` and the
  metrics `host` label are the same string, spaces intact. Silently mapping `CloudAPI` onto
  `Cloud API` would attribute one host's queue depth to another and hide a rename. Divergence is
  reported as `_meta.hostStatus.unmatchedHosts`.
- **A host the endpoint did not describe keeps `null`, not `0`** — the same "absent is not zero"
  invariant the parser already holds. `merged: 0` with `shape: "hosts"` is the case worth alerting
  on: answered correctly, matched nothing.
- **`IRIS_HOSTSTATUS_PATH` is absolute; `IRIS_BASE_PATH` is deliberately not applied.** They locate
  different web apps. Prefixing 404s, and the failure is quiet. There is a test asserting the base
  path is *not* applied.
- **`Promise.allSettled`, not `all`** — a failed third poll must not take the metrics poll with it.
  Setting `IRIS_HOSTSTATUS_PATH=` empty restores the old behaviour exactly.
- **Read-only throughout.** The endpoint reads host state and counts rows. No writes, no production
  changes.

## Test output

```
services/metrics-proxy    96 pass / 0 fail     (was 71)
contracts                 40/40 checks pass    (2 samples, 14 accept, 17 reject, 7 capture claims)
detection-engine         133 pass / 0 fail     tsc --noEmit clean — UNTOUCHED
smoke (mock)              15/15
smoke (live IRIS)         15/15
```

**Proven to fail without the fix, not merely to pass with it:**

- Stubbing `mergeHostStatus` to a no-op: **7 of 21 fail** in `hoststatus.test.js`.
- Reverting `pollMetrics` to its single-GET form: **3 of 4** new poller cases fail.
- Removing `statusFromMetrics` + `HostStatusMeta` from the schema in memory: the new contract cases
  fail. *Honest caveat:* `_meta.hostStatus` alone would have validated either way, since
  `MetricsMeta` is not `additionalProperties: false` — before `HostStatusMeta` existed it was
  accepted **unvalidated** rather than rejected.

**Found while doing this:** `npm test` enumerated test files explicitly, so `src/hoststatus.test.js`
was collected by `node --test` but **not** by `npm test` — 21 tests would have passed locally and
never run in CI. Added to the script; that is why the count is 96 and not 75.

## Not verified — please read before treating #12/#31 as demo-ready

**No non-zero queue depth or error count was observed on live IRIS.** The production is healthy: it
drains immediately, and all **163,392** rows in `Ens.MessageHeader` are `Status = 9` (Completed).
400 samples of `Ens.Queue.GetCount` and 40 of `EnumerateHostStatus` read `0`/empty throughout.
Inducing either state requires disabling or misconfiguring a host — a **production change**, which I
treated as out of bounds on the shared instance.

So the plumbing is verified end to end and **the two findings actually firing is not.** The non-zero
path rests on the depth of `70` measured earlier on this instance with `Cloud API` disabled (#12),
and is covered by schema cases plus a unit test explicitly labelled synthetic. Note also
`queue_buildup`'s `absoluteFloor: 50` (#16): real numbers arriving and the rule tripping are
separate milestones, and `48` would not trip it.

Someone with authority to change a production setting should induce a backlog once and confirm.

## contracts/ changed — needs all three reviewers

Per `CONTRIBUTING.md` §4 and CODEOWNERS, any `contracts/` change requires **@devA @devB @devC**.
With Dev A gone that is Dev B and Dev C in practice; flagging it explicitly rather than letting the
CODEOWNERS prompt be the only signal.

**Both fields keep their `number | null` type — no consumer is required to change anything.** What
changed is which case is normal. `CHANGELOG.md` has the full entry.

**This branch also carries PR #30**, which is what *introduces* `contracts/proxy-api.md` and
`proxy.schema.json` — they are not on `main` yet. This change makes two of their documented values
wrong, so the amendment had to sit on top of them rather than beside them. **Review #30 first, or
close it in favour of this branch.**

**The engine needs no change, and I verified that rather than assuming it:** on
`devB/live-mode-reconcile` (PR #33), `ProxyHost.queued`/`.errored` are `NullableCount` and
`isProxyHost` gates them with `isNullableCount`, which accepts `null` or any finite number.
`services/detection-engine/**` and `apps/dashboard/**` are untouched by this PR.

## New setup step

The dispatcher needs its **own** CSP web application — one path maps to one dispatch class, so it
cannot share `/labdemo`. `iris/labdemo/README.md` §1b documents it. Nothing said so before, and the
failure mode is a 404 that surfaces only as `queued`/`errored` staying `null`.


Ari-Glikman:
Verification notes. **Dev A moved to other work and Dev B has taken over their outstanding tasks** — this is that work on #12/#31. I re-checked the significant claims against `main` and the live instance rather than accepting the report.

@tanifgit — the parts worth your attention are the merge order and the "not verified" section.

## Two corrections to my own brief, both right

I told the agent `contracts/proxy-api.md` and `proxy.schema.json` were on `main`. **They aren't** — they're in my own open #30. And I implied #33 had merged; it's open. Confirmed both:

```
git ls-tree origin/main contracts/ | grep -i proxy   ->  (nothing)
gh pr view 33 --json state                            ->  OPEN
```

Handling it by merging #30's branch in was the right call — an amendment should sit on top of the files it amends. Which creates a merge-order question, below.

## Merge order: #30 first, then rebase this

This PR contains all six of #30's files, but as a **genuine superset**, not a copy — it amends them by +428 lines:

```
contracts/proxy-api.md      | 225 +++++++++-----------
contracts/proxy.schema.json |  66 ++++++-
contracts/validate.mjs      | 117 +++++++++++
contracts/CHANGELOG.md      |  80 ++++++++
```

**#30 is already approved**, so the clean sequence is merge #30, rebase this onto `main`, then review this. Merging this first would strand #30 as a no-op and lose the separate reasoning in its history. I'd rather not close #30 in favour of this — the proxy contract landing is the Day-1 gate item and deserves its own commit.

## The latent defect find is real, and now filed as #37

Verified in existing code:

```
iris/labdemo/REST/PatientDispatcher.cls:177    do dyn.%ToJSON(%response.Output)
```

`%CSP.Response` has no `Output` property on IRIS 2026.1, and `WriteJSON` is the sole response writer for **four** routes (lines 89, 109, 120, 146) — so every JSON route on that dispatcher 500s. It compiles fine; the property access only fails at runtime, which is why nobody caught it.

Right call not to fix it here. Filed as **#37** so it isn't lost. Same shape as three things we've already fixed: code that compiles and looks correct while doing nothing.

## On "not verified", which is the important section

> the plumbing is verified end to end and **the two findings actually firing is not**

This is the right way to report it and I want to reinforce rather than soften it: 400 samples of `Ens.Queue.GetCount`, 40 of `EnumerateHostStatus`, all zero; and all 163,392 `Ens.MessageHeader` rows at `Status = 9` (Completed), so there were no errors to count.

Declining to induce either — both need a production change on a shared instance — was correct. I made the same call earlier today when I stopped short of setting `FailurePercent: 40` to watch errored counts climb.

So the honest state: **#12 and #31 are plumbed, not demonstrated.** `queue_buildup`'s `absoluteFloor: 50` sits between plumbed and firing, which means "wired up and returning real numbers" and "the finding appears" remain two separate milestones — exactly the caution @tanifgit raised on #16.

The non-zero path rests on my earlier measurement (`Queue='70'` with Cloud API disabled) and is covered by a test labelled synthetic. That's weaker than a live observation and the PR says so.

## The three technical corrections are all more accurate than what I gave the agent

- **`TargetConfigName` uppercasing is a `GROUP BY` collation artifact**, not a property transform — and my suggested `$ZCONVERT` mapping matched only 4 of 13 hosts when tested. `%EXACT()` is correct and I was wrong.
- **`$zdatetime(...,3,7)` already appends `Z`**, so the first version emitted `...ZZ` — which JS `Date` accepts, meaning it would have reached the samples silently. Good catch; that's the class of bug that survives review.
- **`npm test` enumerated test files explicitly**, so 21 new tests ran under `node --test` but never in CI. That's the fourth instance this week of a check that passes while verifying less than it appears to.

## The disclosed weak spot

> `_meta.hostStatus` would validate with or without my schema addition, since `MetricsMeta` isn't `additionalProperties: false`

Worth stating what that means: the field was previously accepted **unvalidated** rather than rejected, so the schema addition documents it without yet enforcing it. Tightening `MetricsMeta` is a contract change of its own. Fine to leave, better to have said.

## One process note

The agent's git HEAD collided with mine again — my `#33` work briefly appeared reverted in the working tree. Verified nothing was lost: `origin/devB/live-mode-reconcile` still has the rewritten client. Second occurrence today; the lesson is that parallel agents in one worktree need isolation, which this agent solved for itself by moving to a separate worktree.

tanifgit (changes requested):
This is the strongest PR on the repo so far, and the four "measured, not assumed" items are the reason — `%SQLUPPER` on the `GROUP BY` key, `Queue` being `""` for a measured zero, `Status = 8`, and `%response.Output` not existing are each a wrong answer that would have shipped looking right. The `+rs.Get("Queue")` reasoning in particular is the sort of thing that only comes from reading the shipped source.

Requesting changes for **one small thing** with a silent failure mode. The other two are non-blocking.

## 1. Blocking, and a two-line fix — the class doc contradicts §1b added in this same PR

`HostStatusDispatcher.cls`'s own header says:

```
///   GET /labdemo/hoststatus     — per-host queue depth, status and error counts
///
/// Register this dispatcher under a CSP web application, following the same convention
/// as PatientDispatcher (see ../README.md §1):
///   Path: /labdemo  →  Dispatch class: ProductionGuardian.LabDemo.REST.HostStatusDispatcher
```

Everything else in the PR says `/labdemo/monitor` and says it emphatically — README **§1b** (*"It needs its own entry because a CSP application maps one path to one dispatch class"*), `.env.example` (`IRIS_HOSTSTATUS_PATH=/labdemo/monitor/hoststatus`), the poller log, and the live `curl`. The class header points at **§1**, which is PatientDispatcher's.

So the file most likely to be open when someone registers this instructs them to do the exact thing the PR body says is impossible. Following it means either the save collides, or `/labdemo` gets repointed off PatientDispatcher — **breaking the patients API** — and the endpoint then lives at `/labdemo/hoststatus` while the proxy polls `/labdemo/monitor/hoststatus`. That 404s, and per your own note the failure "surfaces only as `queued`/`errored` staying `null`."

Fix the header to `/labdemo/monitor` and point at §1b. Re-approving on sight.

## 2. Non-blocking — the join is reported in one direction, and it's the other one that reaches consumers

`unmatchedHosts` covers *endpoint → metrics*. The direction that leaves `queued: null` in front of a consumer is *metrics → endpoint*, and nothing reports it. I ran your `mergeHostStatus` over four application hosts:

```
ALL FOUR described by the endpoint:
  _meta.hostStatus: merged=4 hostCount=4 shape=hosts unmatchedHosts=[]
  published queued: Cloud API=0  EMR Source=0  FHIR Transform=0  Lab Router=0

Lab Router MISSING from the endpoint (rename, or the query missed it):
  _meta.hostStatus: merged=3 hostCount=3 shape=hosts unmatchedHosts=[]
  published queued: Cloud API=0  EMR Source=0  FHIR Transform=0  Lab Router=null
```

**`merged === hostCount` in both.** That's the check `HostStatusMeta` explicitly recommends (*"Compare against hostCount"*), and in this failure it says everything is fine. The only signal is `queued: null` on the host — which is precisely the state the description says this meta exists to disambiguate: *"how a consumer tells 'every queued is null because the endpoint is down' from 'every queued is genuinely 0'."* For one host dropping out, it can't.

The arithmetic workaround doesn't rescue it either. In your own live run `snapshot.hosts.length - merged` is `15 - 13 = 2` **while everything is healthy** — the two are `Ens.Alarm` and `Ens.MonitorService`, which the endpoint doesn't enumerate and which get framework-filtered anyway. I diffed the captures:

```
metrics-dump.txt        15 host labels
hoststatus-live-capture 13 hosts    -> absent: Ens.Alarm, Ens.MonitorService
```

So a consumer cannot distinguish "framework hosts absent, normal" from "an application host dropped". Benign today because all four application hosts merge — worth saying that plainly, since it means this is a missing diagnostic and not a wrong number.

Suggestion: report the reverse set too, e.g. `unmergedHosts` for metrics hosts that got nothing, or make it application-hosts-only so the framework noise doesn't mask the real case. Your `isFramework` flag is already on the host.

## 3. Non-blocking — the caveat you disclosed has a one-line fix

You noted that `_meta.hostStatus` "would have validated either way, since `MetricsMeta` is not `additionalProperties: false`". Confirmed — neither `MetricsMeta` nor `HostStatusMeta` sets it. Worth closing, because these fields exist specifically to be *trusted*: if the proxy ever emitted `unmatched` for `unmatchedHosts`, the schema passes, and a consumer reads `undefined` and concludes there were none. A silently-misspelled diagnostic is worse than a missing one. `additionalProperties: false` on both, and the negative test proves what it appears to.

## Verified independently

- **`$get(errored(name), 0)` defaulting to 0 is right**, and I checked it isn't the invented-zero case: the aggregate ran successfully (`erroredAvailable` true), so a host absent from it genuinely has no error rows. The `errCount = ""` / `erroredAvailable` split keeps the unmeasurable case null. Consistent with the parser's invariant.
- **`mergeHostStatus` degrades as advertised.** Null/empty status yields the snapshot unchanged with `available: false`, and `shape: 'hosts'` + `merged: 0` is distinguishable from it — the "answered correctly, matched nothing" case your schema calls out.
- **`shape` enum covers what the code can emit.** A JSON `null` body lands on `unrecognized-object` via `typeof null`, which is in the enum; it collides with the wrong-web-app case, but `keys` separates them, so not worth a change.
- The `status` overwrite is the right call and I want it — `Disabled` is in my `HostStatus` union and renders today, and keeping `statusFromMetrics` alongside instead of discarding it is what makes the disagreement auditable. It doesn't reach my contract (`Host` is `additionalProperties: false`), so nothing on my side needs to change.

## Two notes beyond this PR

**Your capture is more evidence for #34.** `FHIR Transform` with `messageCount: 53580` — equal to Lab Router, consistent with the serial pipeline — so the live instance is unambiguously 4 application hosts, and this PR commits a 4-host fixture. My dashboard fixtures are 3-host per `Production.cls`. That divergence is now in two more files and still has no decision on #34; with Dev A gone, `iris/` is yours, so the redeploy-vs-revert call has effectively become yours to make.

**#30 vs this branch:** you asked for a decision. Close #30 in favour of this one. It's approved but this branch carries its commits and corrects two of its documented values, so merging #30 separately just means landing something we already know is wrong and then fixing it. One review, one merge.

**#35 and this don't conflict except in `CHANGELOG.md`** — I merged them locally, `validate.mjs` merges clean and the merged validator passes 43 checks. Whoever lands second takes a two-entry ordering conflict. Also filed **#38**: CODEOWNERS still names `@kskubach` on 10 lines including `/contracts/`, and when their access goes, per #11 every one of those lines un-gates silently. Your call on `/iris/` and `/services/metrics-proxy/` since you've taken the area.

tanifgit (approved):
Approving. Both findings fixed, and I re-ran my own probe against your branch rather than reading the diff.

**1. Class header** — now `/labdemo/monitor`, points at §1b, and states the consequence inline. That last part is the bit that makes it stick.

**2. `undescribedHosts`** — verified it reports exactly the case that was invisible, and stays quiet in the healthy shape:

```
healthy live shape (framework absent by design)
  merged=4 hostCount=4  merged===hostCount? true
  undescribedHosts=[]

Lab Router dropped — the case that was undetectable
  merged=3 hostCount=3  merged===hostCount? true
  undescribedHosts=["Lab Router"]
```

Excluding framework hosts is the right call and I'd have got it wrong — I was thinking of it as "metrics hosts minus merged", which reports 2 on a healthy live instance. Filtering on `isFramework` is what makes an empty list actually mean "nothing is wrong". Softening `merged === hostCount` to "a coarse check, but see undescribedHosts" in the schema matters too, since that description was the thing pointing consumers at the check that couldn't see it.

## Nit, not worth another round

`undescribedHosts` is absent in the one case where *every* application host is undescribed:

```
endpoint down entirely
  shape=hosts merged=0 hostCount=0
  undescribedHosts=undefined
  queued: Cloud API=null  EMR Source=null  FHIR Transform=null  Lab Router=null
```

`mergeHostStatus` returns early when `byHost.size === 0`, so the field only exists on the success path. It's schema-valid (not `required`) and the state *is* signalled — `merged: 0` with `shape: 'hosts'` is documented as the interesting failure — so nothing is hidden. It's just an odd shape: present when one host is affected, absent when all of them are. A consumer rendering "these hosts have unknown depth" shows nothing in the worst case. `undescribedHosts: []` on that branch would be consistent, or leave it and it's fine.

## Still open from my review, and I'll take your answer either way

**Finding 3 wasn't addressed and the commit doesn't mention it** — `MetricsMeta`, `HostStatusMeta` and `AlertsMeta` are still open objects:

```
CLOSED  ProxyHost          CLOSED  MetricsResponse
OPEN    MetricsMeta        CLOSED  AlertsResponse
OPEN    HostStatusMeta     CLOSED  HealthResponse
OPEN    AlertsMeta
OPEN    ProxyAlert         <- deliberate, and documented
```

I flagged it non-blocking so I'm not re-blocking on it. But there's a decent argument *for* leaving them open that you may already have had: `_meta` is diagnostics, and closing it means adding a future diagnostic field fails validation for anyone pinned to today's schema — the opposite of what you want from a channel whose whole job is carrying new information about failures. `ProxyAlert` carries exactly that reasoning already.

If that's the call, say so at the site the way you did for `ProxyAlert` and it's settled — the reason a thing is open is what stops someone "fixing" it later. If not, `additionalProperties: false` on the two meta objects closes the caveat you disclosed in the PR body.

## The rest, verified

- `$get(errored(name), 0)` — checked again against `erroredAvailable`; a host absent from a successful aggregate genuinely has zero error rows, so that default is a measurement, not an invention.
- 99 tests, +3 covering the healthy case, the previously-undetectable failure, and framework exclusion. The middle one is the one that would have caught this.
- Rebased on main cleanly now that #29 and #30 are in.

Also: nice catch filing and fixing #37 off your own footnote — `%response.Output` breaking every `PatientDispatcher` route was mentioned as out of scope here and landed as #40 within the hour.

PR #39, opened 2026-08-12
chore: reassign Dev A's areas to Dev B and drop the stale three-way gate

### LINKED ISSUE #38: CODEOWNERS still names Dev A on 10 rule lines — every all-three gate un-gates silently the moment their access is revoked
Dev A has moved off the project (#36). `.github/CODEOWNERS` still names `@kskubach` on **10 rule lines**, including every all-three gate and the `*` catch-all.

**Nothing is broken right now.** They still hold push access, so the handle resolves:

```
$ gh api repos/InterSystems-Israel/ProductionGuardian/collaborators
tanifgit / kskubach / Ari-Glikman / InterSystems-Israel   (all push: true)

$ gh api repos/.../codeowners/errors?ref=main
{"errors":[]}
```

**The problem is what happens when that access is revoked** — the normal next step after someone leaves. This is PR #11's bug in reverse, and the warning is already written at the top of the file we'd be breaking:

> *GitHub silently ignores an unknown handle rather than erroring, so an unresolved name means reviews are NOT requested for that path.*

Per #11, an unresolvable owner invalidates the **entire rule line, including the valid co-owners**. So the moment `@kskubach` stops resolving, these stop requesting anyone at all:

| Path | Currently | After revocation |
|---|---|---|
| `/contracts/` | @kskubach @Ari-Glikman @tanifgit | **nobody** |
| `/docs/decisions/` | all three | **nobody** |
| `/tools/` | all three | **nobody** |
| `/.github/` | all three | **nobody** |
| `*` (catch-all) | all three | **nobody** |
| `/iris/`, `/services/metrics-proxy/` | @kskubach | **nobody** |

`/contracts/` losing its gate is the one that matters: root `CLAUDE.md` §4 makes all-three review the mechanism that stops a silent contract change, and #11 exists precisely because that gate was once requesting nobody without anyone noticing. We'd reintroduce it, and `codeowners/errors` would report it only if someone thought to look.

It also has a visible cost today: every PR requests a review that will never arrive. #35 sat with a pending `@kskubach` request until I removed it by hand.

## Proposed change

Dev B has taken over Dev A's area (#36), so:

```diff
-# ---- Dev A -- IRIS and metrics ingestion ----
-/iris/                        @kskubach
-/services/metrics-proxy/      @kskubach
+# ---- IRIS and metrics ingestion (Dev A's former area, Dev B since 2026-08-12) ----
+/iris/                        @Ari-Glikman
+/services/metrics-proxy/      @Ari-Glikman
```

and `@kskubach` comes off the five shared lines, leaving `@Ari-Glikman @tanifgit`.

**The judgement call is what "all three" means with two developers**, and it isn't mine alone to make — @Ari-Glikman, this is your new area. Two readings:

1. **Two-way gate.** Both remaining developers review. Same guarantee as before, one fewer name. My preference — it keeps the property that no `contracts/` change lands on one person's say-so.
2. **Acknowledge the gate is now just "the other person"** and say so in `CONTRIBUTING.md` §4, since "all three" will read as stale to anyone joining later.

These aren't exclusive — (1) is the CODEOWNERS edit, (2) is the doc sentence, and I think we want both.

Happy to open the PR; `.github/**` is shared-with-review per root `CLAUDE.md` §3, so it needs your sign-off either way. Flagging rather than just doing it because reassigning `/iris/` and `/services/metrics-proxy/` to you should be your call to confirm, not mine to assume from a PR description.

Ari-Glikman:
Closes #38.

## Confirming, since you asked rather than assumed

> reassigning `/iris/` and `/services/metrics-proxy/` to you should be your call to confirm, not mine to assume from a PR description

Right call, and: **confirmed.** `/iris/`, `/services/metrics-proxy/` and `contracts/proxy-*` are Dev B's. Dev A moved off on 2026-08-12 and we took over their outstanding tasks — #29, #30, #36 and the ADR 0004 tick were all that work.

Taking **both** your options, since they aren't exclusive.

## 1. CODEOWNERS — `@kskubach` off all 10 lines

Five shared lines become `@Ari-Glikman @tanifgit`; the two former Dev A areas become `@Ari-Glikman`.

Your framing of the urgency is what made this worth doing now rather than later: per #11, **an unresolvable owner invalidates the entire rule line including valid co-owners.** So the moment their push access is revoked, `/contracts/`, `/docs/decisions/`, `/tools/`, `/.github/` and the `*` catch-all would all request nobody — #11's bug in reverse, with the warning already written at the top of the file we'd be breaking.

`/contracts/` losing its gate is the one that matters, and #11 exists precisely because that gate was once requesting nobody without anyone noticing.

```
$ gh api repos/.../codeowners/errors?ref=main --jq '.errors|length'
0
```

## 2. The doc sentence — and I went slightly further

"all three developers" → **"every remaining developer"** in root `CLAUDE.md` §4 and `CONTRIBUTING.md` §2, with a dated team-change note in the ownership map.

Your point that "all three" reads stale to anyone joining later is right, and I've recorded the thing I think a newcomer would actually wonder — whether the gate still means anything with two people:

> GitHub will not let an author approve their own PR, so with two people the gate is exactly "the other person looked at it".

That's the honest version. The count was never the point; requiring someone other than the author is.

**Historical prose describing the original three-way split is left as written** — `CONTRIBUTING.md` §4's rationale, the §1 tree comments, ADR 0004's evidence. It records how the project was structured, and rewriting it would erase the reason the `contracts/` mechanism exists at all.

## One thing found while editing

Root `CLAUDE.md` §6 still named "the four LABDEMO components (EMR Source, Lab Router, FHIR Transform, Cloud API)" — `FHIR Transform` left the production in #14. It now points at `iris/labdemo/Production.cls` as authoritative instead of restating the list.

That's the **fifth** time a copied host list has gone stale (contract samples, engine fixtures, dashboard fixtures, the CI count, and now this), which rather makes your #25 point about frozen copies for the last time.

Refs #38, #11, #36, #25

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,9 +4,11 @@
 # means reviews are NOT requested for that path. Verify a handle resolves before
 # trusting it -- a typo here looks identical to working config.
 
-# ---- Dev A -- IRIS and metrics ingestion ----
-/iris/                        @kskubach
-/services/metrics-proxy/      @kskubach
+# ---- IRIS and metrics ingestion ----
+# Dev A's former area. Dev A moved off the project 2026-08-12 and Dev B took over their
+# outstanding tasks (#36, #38), so these are Dev B's now.
+/iris/                        @Ari-Glikman
+/services/metrics-proxy/      @Ari-Glikman
 
 # ---- Dev B -- detection engine and findings API ----
 /services/detection-engine/   @Ari-Glikman
@@ -16,21 +18,26 @@
 /docs/demo/                   @tanifgit
 
 # ---- Shared, PR-gated ----
-# A contract change needs all three. That sounds heavy and is the point: a silent
-# contract change is the failure mode that breaks the Day-5 integration, and the
-# review takes 30 seconds.
-/contracts/                   @kskubach @Ari-Glikman @tanifgit
+# A contract change needs EVERY remaining developer -- two since 2026-08-12, three before
+# that. The count is not the point; requiring someone other than the author is. A silent
+# contract change is the failure mode that breaks the Day-5 integration, and the review
+# takes 30 seconds.
+#
+# Note this now means a contract PR cannot be self-approved: GitHub refuses to let an
+# author approve their own PR, so with two developers the gate is exactly "the other
+# person looked at it".
+/contracts/                   @Ari-Glikman @tanifgit
 
 # ADRs record decisions nobody should change unilaterally.
-/docs/decisions/              @kskubach @Ari-Glikman @tanifgit
+/docs/decisions/              @Ari-Glikman @tanifgit
 
 # Root config, tooling, CI.
-/tools/                       @kskubach @Ari-Glikman @tanifgit
-/.github/                     @kskubach @Ari-Glikman @tanifgit
+/tools/                       @Ari-Glikman @tanifgit
+/.github/                     @Ari-Glikman @tanifgit
 
 # Source material is read-only after Day 1 -- any change here wants three pairs of eyes.
-/docs/production-guardian-healthscan-mvp1.docx  @kskubach @Ari-Glikman @tanifgit
-/docs/production-guardian-demo.html             @kskubach @Ari-Glikman @tanifgit
+/docs/production-guardian-healthscan-mvp1.docx  @Ari-Glikman @tanifgit
+/docs/production-guardian-demo.html             @Ari-Glikman @tanifgit
 
 # Catch-all for anything not matched above.
-*                             @kskubach @Ari-Glikman @tanifgit
+*                             @Ari-Glikman @tanifgit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,15 +1,15 @@
 # CLAUDE.md — Production Guardian (shared rules)
 
-Rules that apply to **all three developers**. Deep, area-specific instructions live in the nearest `CLAUDE.md` to the files being edited:
+Rules that apply to **every developer**. Deep, area-specific instructions live in the nearest `CLAUDE.md` to the files being edited:
 
 | Area | File |
 |---|---|
 | Dashboard / UI (Dev C) | `apps/dashboard/CLAUDE.md` |
 | Detection engine + findings API (Dev B) | `services/detection-engine/CLAUDE.md` |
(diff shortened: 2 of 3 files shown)


tanifgit (approved):
Approving. Taking both options was right, and the sentence I'd have left out is the one that does the most work:

> GitHub will not let an author approve their own PR, so with two people the gate is exactly "the other person looked at it".

That reframes the rule from a headcount into an invariant, which is what makes it survive the next team change instead of needing another PR. Leaving the historical prose as written is also the right instinct — the `contracts/` mechanism only makes sense if you can still read why it was built.

Verified the thing the issue was actually about, on your branch:

```
$ gh api repos/.../codeowners/errors?ref=devB/codeowners-reassign
{"errors":[]}
```

All 10 lines resolve, and the two former Dev A areas now point at you.

## One line missed, and it's the same class this PR exists to fix

`.github/CODEOWNERS`, just above the source-material rules:

```
# Source material is read-only after Day 1 -- any change here wants three pairs of eyes.
/docs/production-guardian-healthscan-mvp1.docx  @Ari-Glikman @tanifgit
/docs/production-guardian-demo.html             @Ari-Glikman @tanifgit
```

"three pairs of eyes" over a two-name list. This isn't the historical prose you deliberately preserved — it's a live statement about a current rule, in the file whose stale names you just fixed. Worth the one-word change while you're here.

## On the fifth stale host list

You're right, and I'll take the sixth: `apps/dashboard/CLAUDE.md` §5 still restates "EMR Source (service), Lab Router (process), Cloud API (operation)". It's the least dangerous of the six — it already says *"Do not treat that list as a constant... the rule being enforced here is never invent a host, not there are three hosts"* — but it's still a frozen copy, and #34 means it may be describing the wrong number right now. I'll align it with your root `CLAUDE.md` §6 wording and point at `Production.cls` instead of restating. My file, my change.

Which leaves #34 as the last one, and now unambiguously yours: `iris/` is your area as of this PR, `Production.cls` is the authority per `iris/CLAUDE.md` §3, and #36's capture is the third independent confirmation that the deployed instance runs four. My fixtures follow whichever way you take it — redeploying to three is one line in my CLAUDE.md and no fixture change; keeping four means re-adding a host to eight fixtures, which I'd rather know before I touch them.

PR #42, opened 2026-08-12
docs(dashboard): stop restating the LABDEMO host list in the rule file

tanifgit:
Taking the sixth instance of the stale host list you counted on #39. My area, my copy.

`apps/dashboard/CLAUDE.md` §5 named the three hosts inline *and* told the reader not to treat the list as a constant, in the same sentence. That contradiction is exactly what let the other five drift: a standing instruction that restates data will always eventually disagree with the data.

Now it points at the `<Item>` set in `iris/labdemo/Production.cls` as the authority and doesn't restate, matching your root `CLAUDE.md` §6 wording. It also names **#34** as the live case where the repo and the deployed instance disagree about the count — which means this file stops being wrong whichever way that lands, rather than needing another edit after the decision.

**`fixtures/README.md` deliberately keeps its list.** That one documents what eight frozen files actually contain, including the measured baseline (`EMR Source 0.2`, `Lab Router 1.2`, `Cloud API 0.4`), and it already draws the distinction the rule file was missing:

> *A fixture is a frozen snapshot of one production and is expected to need editing when that production changes; nothing in `src/` is.*

Stripping the hosts out of there would lose the measured numbers issue #6 exists to protect, and a snapshot description is supposed to be specific. The rule file was the wrong place; that one is the right place.

Docs only — no code, no fixture, no behaviour change. Nothing in `src/` referenced a host name or count before this (enforced by #26) and nothing does now.


tanifgit:
Pushed a second commit, and it makes this PR **depend on #35** — worth stating rather than letting the ordering be discovered.

The same frozen-copy problem sits one section up from the host list. §2.3 of this file duplicates the `Host` interface, and §2.5 says "all twelve" Day-1 questions. #35 widens `queued`/`errored` to `number | null` and adds **Q13**, so the moment it lands, §2.3 contradicts `src/types/healthscan.ts` — in the very section whose rule is *"keep `src/types/healthscan.ts` as a direct transcription of the contract"*.

So §2.3 now shows the nullable fields with the "not measurable, NOT zero" note, and §2.5 gains a Q13 row and says thirteen.

**Merge order: #35, then this.** Either order leaves main consistent within a minute, but this-then-#35 means the doc briefly describes a type the contract hasn't published yet, which is the wrong way round for a file that calls the contract "the law".

Not folding it into #35 because that PR is approved and `CLEAN`, and amending it would drop your approval for a docs edit in a different area.

Ari-Glikman (approved):
Approving. The distinction you drew is the part I'd have got wrong.

## `fixtures/README.md` keeping its list is right, and I nearly argued the opposite

> A fixture is a frozen snapshot of one production and is expected to need editing when that production changes; nothing in `src/` is.

That's the line. I'd been treating "don't restate the host list" as a flat rule after counting five stale copies, and it isn't — a **rule** file that restates data will drift, but a **snapshot description** is supposed to be specific, and stripping the measured baseline (`EMR Source 0.2`, `Lab Router 1.2`, `Cloud API 0.4`) out of it would lose exactly what issue #6 exists to protect.

So the sixth instance was real and the seventh would have been me removing something load-bearing. Good catch on both halves.

Naming **#34** rather than a count is the right move too — it means this file stays correct whichever way that lands, instead of needing another edit after the decision.

## One sequencing note, not a change request

The diff does more than the title says — it also updates the `Host` type to `queued: number | null` / `errored: number | null` and adds Q13. All correct, but **that contract isn't on `main` yet**:

```
$ contracts/healthscan.schema.json on main
  queued    integer
  errored   integer
```

So merged as-is, `apps/dashboard/CLAUDE.md` would describe a nullable contract for however long #35 sits open — a rule file disagreeing with the schema, which is the same class of drift this PR is fixing.

**Merge #35 first and it's moot.** It's approved and green on all five jobs, including the new `iris-compile`. Either you merge it or I will; nothing is blocking it.

## Q13's wording is better than mine

> `null` is **"not measurable for this host"**, never zero ... Render `—`; never compare against it.

That's the same rule I ended up enforcing in the engine after your #33 review, and yours states it in one line. Worth noting we arrived at it from opposite ends: you from "I'd rather show `—` than a zero we didn't measure", me from a false `throughput_drop` firing on a coerced zero. Same conclusion, and the contract text now carries it rather than two separate code comments.

Docs-only, `apps/dashboard/` is yours, no code or fixture touched. Merge whenever #35 is in.

Issue #43, opened 2026-08-12
queue_buildup cannot fire live from a cold baseline: the queue inflates its own mean faster than 5x


Ari-Glikman:
Found in the **first end-to-end run of the full chain** (`IRIS → proxy → engine → dashboard`, MVP §4.6 Day 5). Everything else worked; this did not, and it matters for the demo.

## What happened

Disabled `Cloud API` on the live production and watched. `dead_host` and `throughput_drop` fired **immediately** — genuinely from real IRIS, and `dead_host`'s message carried the per-host depth that #12 landed:

```
critical  dead_host        Cloud API is Disabled with 6 message(s) queued
critical  throughput_drop  Throughput 0.0 msg/sec is 100% below baseline
```

Then I watched the queue climb, expecting `queue_buildup` at 50:

```
queued=54   findings: dead_host,throughput_drop
queued=62   findings: dead_host,throughput_drop
queued=70   findings: dead_host,throughput_drop
queued=94   findings: dead_host,throughput_drop
queued=122  findings: dead_host,throughput_drop
```

**It never fired.** The `absoluteFloor: 50` was cleared at 54 and stayed cleared all the way to 122.

## Why

Self-inflation (ADR 0002, `CLAUDE.md` §5.1) — but worse than the documented case. There, a *sustained* value eventually becomes the new normal and an existing finding clears. Here the finding **never appears at all**, because the queue is *monotonically growing* while the baseline is being built from those same samples. The rolling mean chases the value up, so `current / mean` never reaches `5.0`.

Concretely: a queue growing ~8 per poll produces a mean that is always a large fraction of the current value. `5×` is unreachable for a linear ramp from zero.

This is distinct from the case pinned in `test/silence.test.ly` — that one fires and then clears. This one is silent throughout.

## Why the fixture tests didn't catch it

`fixtures/proxy/queue-buildup.json` jumps **0 → 486 in one poll**, so the first breaching sample sees a near-zero mean and a huge ratio. Real IRIS ramps. The fixture is honest about the *value* and wrong about the *shape*, and shape is what this rule is sensitive to.

That's the same lesson as the last five in this repo, one level up: our mock was self-consistent and didn't predict the real thing. ADR 0004 said integration risk is deferred, not removed — this is it arriving.

## What this means for the demo

[...]

PR #45, opened 2026-08-12
test(detection-engine): pin that queue_buildup is blind to a ramp (#43)

Ari-Glikman:
Test-only. Pins the defect found in the first live end-to-end run (#43) so it can't be silently "fixed" or forgotten while we decide on the real change.

## What it pins

A disabled `Cloud API` on live LABDEMO queued 6 → 122 messages. `queue_buildup` **never fired**, despite clearing its `absoluteFloor: 50` at sample one.

Same mechanism as the self-inflation property already in this file, one stage earlier:

| Shape | Behaviour |
|---|---|
| Step change (existing test) | fires, then clears once the mean catches up |
| **Linear ramp (new test)** | **silent throughout — never fires at all** |

## The part worth knowing

The ratio ceiling is **~2.18× and scale-invariant**. `+8/poll` and `+30/poll` peak at the same value, because scaling a ramp scales its mean identically. The test asserts that invariance directly:

```
assert.ok(Math.abs(slow - fast) < 0.01, 'the ceiling is scale-invariant, so tuning the rate cannot help')
```

That's the actionable fact — **no choice of `absoluteFloor` or `baselineMultiplier` makes this rule fire on a steadily growing queue.** It rules out the whole tuning family of fixes and leaves only "change what the comparison is against", which is what #43 proposes and what #25's ADR should decide.

## Why the fixture suite stayed green

`fixtures/proxy/queue-buildup.json` jumps 0 → 486 in one poll, so its first breaching sample sees a near-zero mean and a huge ratio. **The fixture is honest about the value and wrong about the shape**, and this rule is sensitive to shape.

That's ADR 0004's named cost arriving on schedule — "a mock can be wrong in ways the real thing is not" — and the fifth instance in this repo of a check that passed while verifying nothing.

## Verification

```
tests 150 | pass 150 | fail 0
tsc --noEmit: clean
```

And the new assertion is not vacuous — tightening the gate to `< 1` fails with the real measured number:
```
AssertionError: a slow ramp must not fire (peak ratio 2.18x)
```

No behaviour change; `thresholds.json` and the rules are untouched. Deliberately not fixing it here — the fix changes what every comparative threshold means and belongs in the #25 ADR, not a test PR.

Refs #43, #25, ADR 0002

### DIFF
--- a/services/detection-engine/test/baseline.test.ts
+++ b/services/detection-engine/test/baseline.test.ts
@@ -134,19 +134,46 @@ describe('self-inflation (inherent to a rolling mean)', () => {
    * Measured against real LABDEMO — a disabled Cloud API queued 6 -> 122 messages and
    * `queue_buildup` stayed silent the whole way, well past its absoluteFloor of 50.
    *
-   * The ratio ceiling is ~2.18x and it is SCALE-INVARIANT: +8/poll and +30/poll both
-   * top out there, because scaling a ramp scales its mean identically. So no floor or
-   * ramp rate makes this rule fire on a steadily growing queue — the gate is 5x and
-   * the mechanism cannot produce it. That is why #43 proposes changing what the
-   * comparison is against, not tuning the numbers.
+   * The ceiling is RATE-invariant, not scale-invariant, and it has a closed form:
+   *
+   *     ceiling = 2·i* / (i* − 1)
+   *
+   * where i* is the first sample clearing BOTH `minBaselineSamples` and the absolute
+   * floor. At N=12 that is 24/11 = 2.1818…, so the figure comes from the WARM-UP COUNT,
+   * not from the ramp. Verified by probing BaselineStore directly (@tanifgit, #45):
+   *
+   *     step=8/30/500  N=12  -> 2.1818  (rate cannot change it, confirmed to 500/poll)
+   *     step=8         N=24  -> 2.0870  (but minBaselineSamples CAN)
+   *     step=8  N=12  floor=500 -> 2.0323  (a higher floor LOWERS it: i* is later,
+   *                                         where the ratio is smaller)
+   *
+   * Hence the assertion below is `< 5`, not `≈ 2.18`: the closed form is ≤ 4 for any
+   * N ≥ 2, so "never 5x" is robust in a way the specific number is not. Do not rewrite
+   * `2.18` as a constant here — re-derive it from N.
+   *
+   * WHAT THIS DOES NOT SHOW: that no multiplier can work. The ratio asymptotes to 2 from
+   * ABOVE and stays there for the life of the window, so a gate at 2.0 does fire, on
+   * ~185 consecutive samples. The tuning family is ruled out by COST, not arithmetic — a
+   * 2.0 ratio gate on queue depth would fire on ordinary traffic variation. That is a
+   * better argument for #43's "change what the comparison is against" than impossibility,
+   * and it is a question #25's ADR has to answer rather than assume.
    *
    * `fixtures/proxy/queue-buildup.json` jumps 0 -> 486 in one poll, which is why the
    * fixture suite is green and the live run was not: the fixture is honest about the
    * value and wrong about the shape, and this rule is sensitive to shape.
    */
   it('never reaches 5x against a linear ramp, at any rate (#43)', () => {
+    // Copies of windowSeconds / minBaselineSamples / queue_buildup.absoluteFloor from
+    // thresholds.json. Deliberately hardcoded so this pinning test does not move with
+    // config — the `< 5` result holds for any of them. Named rather than inlined so the
+    // ceiling stays re-derivable: change MIN_SAMPLES and the ceiling becomes
+    // 2·N/(N−1), NOT 2.18.
+    const WINDOW_SECONDS = 1800;
+    const MIN_SAMPLES = 12;
+    const ABSOLUTE_FLOOR = 50;
+
     const ceiling = (step: number): number => {
-      const store = new BaselineStore(1800, 12);
+      const store = new BaselineStore(WINDOW_SECONDS, MIN_SAMPLES);
       let at = T0;
       let best = 0;
       for (let i = 0; i < 120; i += 1) {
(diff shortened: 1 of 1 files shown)


tanifgit (approved):
Approving — the test is correct, every assertion in it is true, and pinning this before the ADR is exactly right. 150/150 pass, `tsc --noEmit` clean, test-only.

**But one claim in the PR body is false, and it is the one #25 will cite.** Flagging it here rather than on the ADR because it changes what the ADR concludes.

## The 2.18× has a closed form, and it isn't scale-invariance

Probing `BaselineStore` directly:

```
step  minSamples  peakRatio  atSample  2N/(N-1)
8     12          2.1818     12        2.1818
30    12          2.1818     12        2.1818
500   12          2.1818     12        2.1818     <- rate-invariant, as you say
8     2           2.3333     7         4.0000
8     6           2.3333     7         2.4000
8     24          2.0870     24        2.0870     <- but NOT constant
```

The ceiling is `2i*/(i*-1)`, where `i*` is the first sample clearing **both** `minBaselineSamples` and the absolute floor. At `N=12` that is `24/11 = 2.1818…` — the figure is exact, and it comes from the **warm-up count**, not from the ramp.

So the invariance is **rate**-invariance, not scale-invariance. Ramp rate can't change it (confirmed to 500/poll); `minBaselineSamples` can. Your test is right to assert `< 5` rather than `≈ 2.18`, because the closed form is `≤ 4` for any `N ≥ 2` — the "never 5×" result is robust in a way the specific number isn't.

## The correction: a multiplier below 2.18 does fire

> **no choice of `absoluteFloor` or `baselineMultiplier` makes this rule fire on a steadily growing queue.** It rules out the whole tuning family of fixes

The floor half is right — and stronger than stated. The floor can only ever *lower* the ceiling (raising it pushes `i*` later, where the ratio is smaller), so no floor value helps.

The multiplier half doesn't hold. `2i*/(i*-1) > 2` for all `i*`, so the ratio **asymptotes to 2 from above and stays there** for the life of the window:

```
sample  12  depth=  96  mean=  44.0  ratio=2.182
sample  50  depth= 400  mean= 196.0  ratio=2.041
sample 180  depth=1440  mean= 716.0  ratio=2.011
sample 250  depth=2000  mean=1272.0  ratio=1.572   <- window (1800s) starts evicting
sample 400  depth=3200  mean=2472.0  ratio=1.294
```

A gate at `2.0` is breached on **108 consecutive samples** inside the test's own 120-sample horizon — trivially clearing `sustainedSamples: 2` — and ~185 before eviction drags it under. So the tuning family is ruled out by **cost, not by arithmetic**: a 2.0 ratio gate on queue depth would fire on ordinary traffic variation, which is a much better argument and a different one.

(Measured against `BaselineStore` and the ratio series directly, not through a full engine run — so read the 108 as "the ratio gate is satisfied continuously", not "I watched the finding fire.")

Worth correcting because you've already quoted the stronger version into #25, and an ADR that believes "no multiplier can work" will skip past the question it should actually answer: *is there a ratio gate low enough to catch a ramp and high enough not to cry wolf?* I think the answer is no — but it needs to be answered, not assumed. And it's more evidence for your own position that the real decision is **what the baseline is computed over**.

## Nit: three constants that will outlive their source

```ts
const store = new BaselineStore(1800, 12);
...
if (mean !== null && mean > 0 && depth >= 50) {
```

`1800`, `12` and `50` are copies of `windowSeconds`, `minBaselineSamples` and `queue_buildup.absoluteFloor` from `thresholds.json`. Change any of them there and this test still passes — correctly, since the `< 5` result holds regardless — while the `~2.18x` in the doc comment silently becomes wrong.

Same defect class as #48's warm-up figure and #46's `sustainedSeconds: 8`: **a number recorded without the parameter it was derived from.** Cheapest fix is a line in the comment giving the derivation — `ceiling = 2N/(N-1)`, `N = minBaselineSamples = 12 → 24/11` — so the next reader can re-derive instead of trusting it. Reading the values from `thresholds.json` would be better still, but I don't think a pinning test needs to move with config, and hardcoding makes the test independent of it. Your call, and not worth blocking on.

## Verified

| Claim | Result |
|---|---|
| 150 pass / 0 fail | **confirmed** |
| ramp never reaches 5× | **confirmed**, and robust — closed form is ≤ 4 for any `N ≥ 2` |
| `+8` and `+30` peak identically | **confirmed** to 4 decimal places, and at `+500` too |
| assertion is not vacuous | **confirmed** — the peak is a real measured 2.1818, not a null-guard artifact |
| step change fires then clears; ramp silent throughout | **confirmed** — both pinned, same file, correct framing |

The fixture-shape explanation is the most valuable part of this PR and I'd want it verbatim in the ADR: *the fixture is honest about the value and wrong about the shape, and this rule is sensitive to shape.* That's a general statement about every fixture we have, not just this one, and it's mine to act on for the dashboard fixtures.

Ari-Glikman:
Fixed and merged — thank you for probing it rather than taking the number at face value. Both corrections were to the doc comment; the assertions were already right, which is the part I'd have got wrong if I'd asserted `≈ 2.18` instead of `< 5`.

Reproduced your probe independently before editing:

```
step=8/30/500  N=12       -> 2.1818   (rate cannot change it)
step=8         N=24       -> 2.0870   (minBaselineSamples CAN)
step=8  N=12  floor=500   -> 2.0323   (a higher floor LOWERS it)
```

That last row is the one I hadn't thought about, and it makes your floor argument **stronger than what I claimed** rather than a correction to it: raising the floor pushes `i*` later, where the ratio is smaller. So no floor value helps, for a reason I hadn't identified.

`ceiling = 2·i*/(i*−1)` is now in the comment with the derivation, plus an explicit "do not rewrite 2.18 as a constant — re-derive it from N".

## On the multiplier claim

You're right, and it matters more than the arithmetic. I've corrected it on #25 as well, since I'd already quoted the stronger version there. The framing I've adopted is yours:

> the tuning family is ruled out by **cost, not by arithmetic**

and the question the ADR has to actually answer — *is there a ratio gate low enough to catch a ramp and high enough not to cry wolf?* — is now written into the test comment so it can't get skipped. I also think the answer is no, but you're right that "I think no" and "arithmetically impossible" are different claims and I shipped the second one.

## On the nit — took it, with one deviation

Named the three constants (`WINDOW_SECONDS`, `MIN_SAMPLES`, `ABSOLUTE_FLOOR`) rather than reading them from `thresholds.json`, for the reason you gave: a pinning test shouldn't move with config, and `< 5` holds regardless. But naming `MIN_SAMPLES` makes it visible that it's the input `~2.18` depends on, which was the actual hazard.

Your classification is the useful part: **a number recorded without the parameter it was derived from.** That's now three instances in three of my open PRs — this one's `2.18`, #48's warm-up duration, and #46's `sustainedSeconds: 8` — and in every case the derivation was known to me and I wrote down the product. Worth more than any of the individual fixes.

150/150, `tsc --noEmit` clean, all five jobs green. Merged.

PR #46, opened 2026-08-12
perf: time-gate the sustained bar, halve both polls (#44) — 18.3s → 11.1s, still short of 10s

### LINKED ISSUE #44: End-to-end detection latency measured at 9.7-10.4s against a 10s bar (n=2); system_alert separate. Decision needed on the criterion
Measured in the first full end-to-end run (#43 is the other finding from the same run). This one affects an **explicit acceptance criterion**, so it needs a decision from all three of us rather than a unilateral tuning change.

## Measurement

Disabled `Cloud API` on live LABDEMO with a wall-clock stamp, then polled `:3002` until findings appeared:

```
t0 (fault injected)   14:37:10.143 Z
detectedAt (engine)   14:37:25    Z
detection latency     14.9s          <- and this is BEFORE the dashboard's own poll
```

Acceptance bar, `apps/dashboard/CLAUDE.md:425`, quoting the MVP doc:
> dashboard renders all hosts + findings live; **updates within 10 s of a change**; fallback to demo mode is seamless

## Where the time goes

Every stage is individually defensible; they just don't sum to 10s.

| Stage | Budget | Source |
|---|---|---|
| IRIS change → proxy sees it | 0–10s | `METRICS_POLL_INTERVAL_MS=10000` |
| proxy → engine sees it | 0–10s | `POLL_INTERVAL_MS=10000` |
| engine 1st breaching sample → 2nd | **10s fixed** | `sustainedSamples=2`, one poll apart |
| engine → dashboard renders | 0–5s | `VITE_POLL_INTERVAL_MS=5000` |
| **worst case** | **35s** | |
| **best case** | **10s** | |

**The sustained-breach gap alone consumes the whole budget.** Two consecutive samples 10s apart cannot produce a finding in under 10s, so even with instant polling everywhere else we exactly hit the bar and never beat it. The 5s dashboard poll then guarantees we exceed it.

This is not a bug in any one component. `sustainedSamples=2` is required by MVP §6 and is our main false-positive defence (`CLAUDE.md` §5). A 10s proxy poll is reasonable against a live instance. Each choice was made in isolation and nobody multiplied them out — the same shape as the mock/real gap in #43.

## Why the dashboard note is optimistic

`apps/dashboard/CLAUDE.md:296` reasons:
> Default **5 s** — the acceptance bar is "updates within 10 s of a change", and Dev A's proxy polls IRIS every 10 s, so 5 s of client polling guarantees you are **never the bottleneck**

Correct about the bottleneck and still short of the bar: it accounts for the proxy poll but not for the engine's own poll or the sustained gap. Not @tanifgit's error — those two stages weren't theirs to know about. It's a gap that only appears when the layers are composed, which is what Day 5 is for.

## Options

My read, roughly in order of preference:

1. **Shorten the engine poll to 5s and the proxy poll to 5s.** Best case → ~5s, worst → ~20s. Costs 2× the IRIS scrape rate; `/api/monitor/metrics` is cheap, so this is likely fine. Still misses the bar in the worst case.
2. **Poll faster *and* count sustained breaches in time rather than samples** — "breaching for ≥N seconds" instead of "N consecutive samples". Decouples false-positive protection from poll rate, which is the actual design flaw here. Best/worst → ~5s/~12s. More invasive; touches the registry.
3. **Reinterpret the bar as "the dashboard reflects the engine's state within 10s"**, which we already meet comfortably, and state the detection latency separately and honestly.
4. **Accept and document.** Cheapest; leaves a written criterion unmet.

**(2) is the correct fix and (1) is the safe one for the timeline.** I'd suggest doing (1) now, and recording (2) with #25's floors ADR — that ADR is already reconsidering what thresholds mean, and "sustained" is a threshold.

I don't want to pick unilaterally: (1) changes Dev A's proxy defaults and (3) rewrites an acceptance criterion. @tanifgit — (3) is partly a call about what we're claiming on stage.

## What did meet the bar

For the record, so this reads as one gap rather than a broken chain:

- `dead_host` and `throughput_drop` both fired from real IRIS on the first breaching poll pair
- `dead_host`'s message carried the live per-host queue depth from #12
- On recovery, findings cleared themselves with no tombstones, per contract Q4
- The drain artifact behaved correctly too — ~130 queued messages really did wait ~40s, so `growing_queue_wait` and `slow_processing` fired truthfully and cleared within ~60s once the backlog drained. Real finding, not a false positive
- Cloud API restored; production is `state: ok`, 4 hosts, 0 findings

Refs #25, #43, MVP §4.6

Ari-Glikman:
Partial fix for #44, plus a retraction of that issue's headline number.

**Cuts measured end-to-end detection from 18.3s to 11.1s engine-visible. Does not reach the 10s bar.** The remainder is structural and needs a decision from both of you rather than a number I pick.

## First: #44's "14.9s" was invalid, and I'm retracting it

I stamped `t0` in the shell, then made two MCP calls to flip the host — each ~15s of round trip — before the change landed in IRIS. **So the measurement included my own tool-call turnaround, not just the product.** The 30.7s "proxy staleness" I chased for a while was the same artifact.

Re-measured properly: one process observes IRIS, proxy and engine on the same clock with no agent turnaround inside the window. The proxy is 0.5–4.2s, i.e. fine, and was never the bottleneck.

## What changed

**1. The sustained bar is two gates now — samples AND wall-clock seconds, both required.**

This is the actual design flaw from #44. `sustainedSamples` is a *count*, so shortening the poll interval for latency silently shortens the debounce *duration* too — trading away MVP §6's false-positive protection without anyone choosing to. Separating them lets the two be tuned independently. `sustainedSeconds: 0` restores the old behaviour.

**2. Both polls 10s → 5s.** Engine `POLL_INTERVAL_MS`, proxy `METRICS_POLL_INTERVAL_MS`. Safe *only* because of (1). Alerts stay at 30s — they're events with their own timestamps, not a sampled value.

**3. `sustainedSeconds` defaults to 5, not 8** — this was my own bug, caught mid-measurement.

I first picked 8 because it sits under the 10s bar, **without checking it against the poll rate it would run at.** At a 5s poll, 8s is unreachable in 2 samples, so it silently confirms on the 3rd — a measured 12.8s debounce, *worse* than the 10s it replaced. Same error class as the defect this PR fixes: a per-stage number chosen without multiplying it out. At 5 the gate is satisfied by the 2nd sample, so it's inert at the shipping rate and only binds if the interval is shortened further — which is exactly the coupling it exists to break.

The test asserts the **relationship**, not the value, since either number can change and the constraint is between them:

```
sustainedSeconds=8 is not reachable in 2 polls of 5000ms
  — it would force an extra poll and add latency invisibly
```

Verified it fails at 8 and passes at 5.

## Measured

```
proxy staleness (IRIS -> proxy)   4.2s   [poll 5s, phase 0-5s]
debounce (proxy -> engine)        6.9s   [was 12.8s at gate 8]
engine-visible total             11.1s   [was 18.3s]
+ dashboard poll                  5.0s
worst observed                   16.1s   BAR: 10s
```

## Why this doesn't close #44

The residual is structural, not tuning. Two samples costs one poll interval, and each of three polling stages contributes 0–P, so **worst case ≈ 4P**:

| P | worst case |
|---|---|
| 5s (this PR) | 20s |
| 2.5s | 10s — at the bar |
| 2s | 8s |

Meeting 10s worst-case needs **P ≤ 2.5s at every stage** — 4× the IRIS scrape rate — or **`sustainedSamples: 1`**, which drops the single-blip protection MVP §6 asks for. Those are real tradeoffs:

- @tanifgit — option 3 from #44 (redefine the bar as "the dashboard reflects the engine's state within 10s", which we already meet comfortably, and report detection latency separately) is still on the table and is partly a call about what we claim on stage.
- A 4× scrape rate against a live customer instance is the kind of thing that should be a deliberate choice, not a default I set.

I'd rather ship this measured improvement and leave #44 open with real numbers in it than close it on a figure that turned out to be measuring me.

## Verification

```
detection-engine  157/157 pass, tsc --noEmit clean
metrics-proxy      99/99 pass
```

Live LABDEMO left restored: `state: ok`, 4 hosts, 0 findings, Cloud API enabled.

Note for local runs: this repo needs **Node 22+** for type stripping. If `node` resolves to v20 the whole engine suite fails with `ERR_UNKNOWN_FILE_EXTENSION` — CI is fine, local PATH may not be.

Refs #44


Ari-Glikman:
Self-note before merge: **this PR makes two lines in ADR 0002 stale**, and I'd rather flag it than let the next person find it.

`docs/decisions/0002-baseline-strategy.md`:

- line 31 — *"At 10s proxy polls, a 30-minute window is ~180 samples. We require 12 samples (~2 minutes)"*
- line 57 — *"An engine restart loses all baselines, and the dashboard shows warming for ~2 minutes"*

At the 5s poll this PR ships, both `~2 minutes` figures become ~1 minute, and the ~180 becomes ~360. I measured the warm-up at ~60s during the e2e run, which is consistent.

Deliberately **not** editing them in this PR. An ADR records a decision made at a point in time, and quietly rewriting its numbers to match a later change is how an ADR stops being a record. The right move is either a short amendment note or a line in `docs/decisions/CHANGELOG.md` — and I'd rather you say which, since the ADR convention here isn't mine to redefine.

The same class of staleness in `iris/labdemo/README.md` is fixed in #48, where I restated the warm-up as the arithmetic (`minBaselineSamples × POLL_INTERVAL_MS`) plus where to read the current value, rather than as a derived duration. That framing is probably what ADR 0002 wants too — a count is the decision, a duration is a consequence of a setting that has now changed twice.

Flagging rather than fixing because it's an ADR, not because it's out of area.

tanifgit (changes requested):
The two-gate design is right and I'd have argued for it independently — `sustainedSamples` was never a debounce, and naming that is the valuable part of this PR. The retraction of #44's headline number is also the right call and well documented.

**But the shipped `sustainedSeconds: 5` has exactly zero margin at a 5000 ms poll, and the stamp that feeds it is taken after a variable-duration `await`. It fails roughly half the time, and when it does it costs a full poll interval — reproducing the 12.8s defect this PR exists to fix.**

## The defect

`src/index.ts:53` stamps the clock *after* the fetch resolves:

```ts
const response = await proxy.fetchMetrics(controller.signal);
engine.applyPoll(response, Date.now());
```

So the gap between two consecutive `applyPoll` stamps is not `POLL_INTERVAL_MS` — it is

```
POLL_INTERVAL_MS + (fetchDuration_n - fetchDuration_{n-1})
```

The gate is `now - firstSeenAt >= sustainedMs`, i.e. `5000 >= 5000`. **Satisfied only by exact equality.** Any poll whose fetch is faster than the previous one makes the gap `< 5000`, the second sample fails the gate, and confirmation slips to the third poll.

Simulating that loop with the real `DetectionEngine` and `DEFAULT_CONFIG`:

```
sustainedSeconds=5 (as shipped in this PR), poll=5000ms

fetch latency constant       (300, 300, ...)  gap=5000ms  ->  gate 5:   5.3s   gate 4:  5.3s
2nd fetch 100ms faster       (300, 200, ...)  gap=4900ms  ->  gate 5:  10.2s   gate 4:  5.2s
2nd fetch 250ms faster       (300,  50, ...)  gap=4750ms  ->  gate 5: 10.05s   gate 4: 5.05s
1st poll cold-connect 800ms  (800, 120, ...)  gap=4320ms  ->  gate 5: 10.12s   gate 4: 5.12s
```

100 ms is enough. Against a localhost proxy the variance is smaller than that but so is the margin — it is zero — so the trigger condition is simply *"was this fetch quicker than the last one"*, which for uncorrelated latencies is **about half of all detections**. The cold-connect row is the one I would expect on the very first breach after a restart, which is also the one you would demo.

Net effect: **engine-visible detection is 5.3s when it works and 10.2s when it does not**, so the 11.1s in the PR body is the good path rather than the typical one. It does not invalidate your measurement — 11.1 + 5.0 dashboard = the 16.1s you observed, so this probably did not fire during that run — but it means the number is not reproducible, and an intermittent +5s is worse to debug on stage than a consistent one.

## The fix: `sustainedSeconds: 4`

Strictly better, and it costs nothing:

```
Is gate 4 still protective if someone shortens the poll?
  poll=5000ms  debounce: gate5=5s  gate4=5s  gate0(old behaviour)=5s
  poll=2500ms  debounce: gate5=5s  gate4=5s  gate0(old behaviour)=2.5s
  poll=2000ms  debounce: gate5=6s  gate4=4s  gate0(old behaviour)=2s
  poll=1000ms  debounce: gate5=5s  gate4=4s  gate0(old behaviour)=1s
```

At the shipping rate it is inert with **1000 ms of slack** instead of 0. At every shortened interval it still decouples debounce from poll rate — 4s of protection where the old sample-only behaviour gives 1s. That is the entire purpose of the gate, preserved.

Your own reasoning gets to this; it just stops one step early:

> At 5 the gate is satisfied by the 2nd sample, so it's inert at the shipping rate

True of the *nominal* interval. The observed interval is not the nominal one, and `>=` on exact equality is the least robust point on the curve. This is the PR's own lesson — *a per-stage timing number is meaningless without the rate it runs at* — extended by one clause: **and without the jitter on that rate.** You caught the version of this at 8; 5 is the same mistake with a smaller error term.

(`4` is a suggestion, not a demand. Anything in `(0, 5000)` works; `4` keeps the "under the 10s bar" property you wanted and reads as deliberate.)

## Related: the test cannot catch this, because it asserts against a copy

```ts
const POLL_MS_SHIPPED = 5_000;
const { sustainedSamples, sustainedSeconds } = DEFAULT_CONFIG;
assert.ok(sustainedSeconds * 1000 <= spanCoveredBySamples, ...)
```

Two things:

- `POLL_MS_SHIPPED` is a **hardcoded copy** of `POLL_INTERVAL_MS`, which `index.ts` does not export. Change the interval there and this test keeps passing while the relationship it asserts becomes false. That is #19's "four copies of a fact, one source of truth" landing in the test that exists to prevent exactly this class of bug. Exporting `POLL_INTERVAL_MS` (or moving it into `ThresholdConfig`, where the other timing numbers already live) would let the test read the real value.
- `<=` admits the zero-margin case. If the constraint is "reachable in `sustainedSamples` polls", strict `<` is what that means, and it would have failed on 5.

Everything else about that test is right, especially asserting the relationship rather than either value.

## Worth recording for #44's option list

`sustainedSamples: 1` no longer means "confirm on the first poll" — it now also has to clear `sustainedSeconds`, which defaults to 5. You handled it correctly in the three existing tests, but it changes what one of #44's live options *is*: choosing it means setting **both** to 1 and 0, and `sustainedSamples: 1` alone would still impose a 5s debounce. Worth a line in #44 so we do not evaluate that option against the wrong behaviour. (I would still argue against it, for the reason you give.)

## Design question, not blocking

The gate uses `Date.now()` — engine **observation** time — while the snapshot carries `sampledAt`, the time IRIS was actually read. Those diverge when the proxy serves a stale cached snapshot: wall clock keeps accruing, so a breach can clear the time gate on repeated views of a *single* IRIS observation. `Date.parse(response.sampledAt)` would make both gates measure the same thing, the condition's age as IRIS saw it, and would incidentally remove the jitter problem entirely, since `sampledAt` advances in proxy-poll steps rather than fetch-completion steps.

The counter-argument is real: if the proxy wedges, `sampledAt` stops advancing and nothing ever confirms. Whether that is a bug or correct conservatism is a genuine design call, and `markPollFailed` only covers outright failure, not staleness. Not for this PR — but it belongs in the #25 ADR alongside "what is the baseline computed over", because it is the same question about which clock and which samples a threshold is measured against.

## Verified

| Claim | Result |
|---|---|
| both gates required; either alone insufficient | **confirmed** — all four gate-combination tests are non-vacuous |
| `sustainedSeconds: 0` restores sample-only behaviour | **confirmed** |
| clock restarts when a breach clears and returns | **confirmed** — no inherited elapsed time on a flap |
| `sustainedSeconds: 8` at a 5s poll costs an extra poll | **confirmed** — I reproduced 10.2s vs 5.3s independently |
| proxy 99/99 | **confirmed** |
| `tsc --noEmit` clean | **confirmed** |
| **engine 157/157** | **156/156.** Off by one — and #47's own arithmetic agrees with me (149 + 7 = 156) |

Also: `test/engine.test.ts:683` has a raw `0x97` byte where an em dash belongs (cp1252 leaking into a UTF-8 file) — it renders as a replacement character in the assertion message. One-character fix.

---

Requesting changes for the margin only. The design is the right one, the retraction is honest, and I would like this in before Day 5 — I just do not want to ship a latency fix that intermittently gives back the whole improvement.

On your two open questions: I am answering #44's option 3 in the issue itself rather than here, since it is a claim-on-stage decision and should not be buried in a review thread. Short version — I would rather report detection latency honestly than reach the bar by removing the false-positive protection, and I do not want a 4x scrape rate against a customer instance either.

tanifgit:
One sequencing point I should have made when I posted the review, since it affects how much work the fix costs you rather than what the fix is.

**Deploy #34 first, then fix the margin and measure once.**

Any latency number taken before the `Production.cls` deploy is invalid afterwards — you established that yourself on #34 (*"the latency figures in #44 are not [environment-independent] and I'll re-measure rather than carry them over"*). So if the margin fix lands first, you measure on the old production, deploy, and measure again. Deploying first collapses that to one measurement on the production we'll actually demo.

The ordering that seems cheapest:

1. Deploy `Production.cls`, stop `LABDEMO.Production`, re-capture with the generator running (#34)
2. Fix the `sustainedSeconds` margin here
3. Measure the full chain once, on the new production
4. Post the number on #44 — and I'll make the option-3 call against it

**Two things that don't need to wait for any of that:** #47 and #48 are both approved and independent, so merge them whenever. And the `2.18×` result is arithmetic, so it carries across the deploy untouched.

## One caveat on my own numbers

The `5.3s vs 10.2s` split I posted came from **simulating** the poll loop — feeding synthetic fetch durations into a real `DetectionEngine` — not from watching it against live IRIS. The mechanism is arithmetic and I'm confident in it, but the *frequency* is my inference from "fetch latency varies", not something I observed. So when you re-measure, that's the real check on my claim as much as on the fix: if the gate is slipping in practice you should see detection alternating between roughly one and two poll intervals rather than sitting at a stable value.

Worth measuring a handful of consecutive fault injections rather than one, for that reason. A single run can't distinguish "5.2s, fixed" from "5.3s, and I got lucky".

tanifgit (approved):
Approving. Every point addressed, and the two new guards are non-vacuous — I flipped `sustainedSeconds` back to 5 and confirmed both fail:

```
✖ the shipped gate is reachable within sustainedSamples polls, WITH MARGIN
✖ confirms on the second poll even when a fetch is quicker than the last (#46)
```

`157/157` pass at 4, `tsc --noEmit` clean, and the `0x97` byte is gone.

Three things you did better than what I asked for:

- **`DEFAULT_POLL_INTERVAL_MS` in `config/thresholds.ts`, not exported from `index.ts`.** I suggested exporting it from `index.ts` without thinking about the side effect; importing that module to read a constant would start a server and a poll loop. Putting it next to the two numbers it constrains is the right call, and the comment saying why is the part that stops someone "tidying" it back.
- **Strict `<` with the reason attached** — "reachable in N polls has to mean reachable when the observed gap is a little under nominal" is a better statement of the constraint than my "`<=` admits the zero-margin case".
- **The jitter test asserts the behaviour, not my arithmetic.** It reproduces all four cases through a real `DetectionEngine` rather than pinning "5.2s", so it stays true if the poll rate or the gate moves.

## One thing to fix before you merge

`thresholds.json`'s `_comment_sustained` was not updated with the value. It still says:

> **At 5 it is satisfied by the 2nd sample, so it is inert at the shipping poll rate** and becomes protective only if the poll interval is shortened

The value is now `4`, and that sentence is the exact reasoning this commit disproves — sitting in the file whose own `_comment` calls it "the only place detection numbers live", hot-reloaded, and the first thing anyone reads before changing the number. `thresholds.ts` got a thorough rewrite; the JSON didn't.

I'd have blocked on this if it were code. Flagging it as pre-merge rather than requesting changes a third time because you've turned both previous rounds around fast and it's one string — but please don't merge without it, or the authoritative config documents the bug as the justification.

Suggested replacement:

> TWO gates, both must pass. sustainedSamples guards a lone bad scrape; sustainedSeconds guards against a fast poll rate shortening the debounce below what MVP 6 intends (#44). Must be reachable within sustainedSamples polls WITH MARGIN, or it silently forces an extra one. Two values that failed at a 5s poll: 8 is unreachable in 2 samples and confirms on the 3rd (measured 12.8s debounce); 5 is reachable only by exact equality, and because the sample is stamped after a variable-duration fetch, any poll quicker than its predecessor slips confirmation to the 3rd sample -- intermittently 10.2s instead of 5.3s. 4 leaves 1000ms of slack at the shipping rate and still decouples the debounce from the poll rate at shorter intervals, which is the coupling this gate exists to break. Derivation in src/config/thresholds.ts, guards in test/engine.test.ts. Set sustainedSeconds to 0 for the old sample-only behaviour.

## Answering your ADR 0002 question — sorry, I missed it first time round

You asked which convention to use for the two stale `~2 minutes` figures, and said it wasn't yours to redefine. **The existing convention already answers it, and neither of the options you offered is the one.**

**1. Don't create `docs/decisions/CHANGELOG.md`.** It doesn't exist — the directory holds the four ADRs and `README.md`. Adding one would be inventing a convention, not following one. `contracts/CHANGELOG.md` earns its keep for a reason that doesn't transfer: a contract is *consumed* by both of us and a silent change breaks a build, so the log is a notification channel. An ADR is read by a person deciding whether to re-litigate, and its `Revisit when` section already is that channel.

**2. You don't need an amendment note either, because 0002 is not decided.** `docs/decisions/0002-baseline-strategy.md:3` says `Status: proposed`. The rule you were respecting —

> **Never rewrite a decided ADR.** Supersede it with a new one and mark the old `superseded by NNNN`. The point is the record of *why*, including reasoning later abandoned.

— protects the record of a *decision*. Only 0004 is `accepted`. 0001–0003 are still proposed, so there is no ratified reasoning to preserve yet and correcting them isn't rewriting history.

**3. Do exactly what you did in #48.** The *decision* in 0002 is **12 samples** — a count. `~2 minutes` and `~180 samples` are consequences of a poll rate that has now changed twice, and they were never the decision. Restating them as the arithmetic plus where to read the interval removes the whole staleness class rather than this instance of it. Your own framing from #48 applies verbatim: *a count is the decision, a duration is a consequence of a setting that has now changed.*

So: a small PR to `docs/decisions/0002-baseline-strategy.md` doing that to lines 31 and 57, co-owned so it needs my review, which it has in advance. Happy for you to take it since you have the context loaded, or I'll do it — say which.

**One piece of good news worth putting in that edit rather than just deleting the number:** 0002's second `Revisit when` trigger is *"baseline warm-up after restart proves disruptive in rehearsal."* Halving the poll rate halves the warm-up, which makes that trigger **less** likely to fire, not more. That's a consequence of this PR worth recording as a benefit — the shorter warm-up is the mitigation the ADR was hoping for.

## Separately, and it affects #25's shape

While checking the convention I read 0002's `Revisit when` in full. It has three triggers:

```
- Early Warning begins — forecasting needs persisted history
- Baseline warm-up after restart proves disruptive in rehearsal
- Host count grows enough for memory to matter
```

**None of them is about what the baseline is computed over, and "self-inflation" does not appear anywhere in ADR 0002.** So this, from #25 —

> excluding breaching samples from the window fixes #43 and the documented self-inflation in one change, and **it's ADR 0002's own named revisit trigger**, so it's the sanctioned place to make that call

— isn't supported by the file. The self-inflation documentation lives in `CLAUDE.md §5.1` and now in `test/baseline.test.ts`, not in the ADR. I raise it because it changes the *shape* of the #25 work rather than the content: under the convention, changing baseline composition is a **new ADR 0005 superseding 0002**, not an amendment to it. Which I think is the better outcome anyway — 0002 accurately records what MVP 1 shipped and why a naive rolling mean was the right starting point, and #43, #49 and #25 are exactly the "reasoning later abandoned" the README wants preserved.

Merge whenever the JSON comment is fixed. And the sequencing note from earlier still stands — deploying #34 first saves you one measurement.

PR #47, opened 2026-08-12
fix(ci): npm test ran zero tests on Windows and exited 0

Ari-Glikman:
`npm test` in `services/detection-engine` **reports success while running nothing** on Windows. Found while running the suite locally during #46.

```
> node --test test/*.test.ts
Could not find '...\services\detection-engine\test\*.test.ts'
EXIT=0
```

The glob is expanded by the **shell**. Bash on the CI runner does it, so the job is green and this was invisible. cmd.exe doesn't, so the literal string reaches Node — which finds no files and exits 0.

Fifth instance of this pattern in the repo, after `npm test --if-present`, `-c ajv-formats`, the `@devA` CODEOWNERS placeholder, and `merged === hostCount`.

## The fix, and the bigger problem behind it

Quoting the glob hands expansion to Node's own matcher (Node 22+, which `engines` already requires), so it behaves the same on both platforms.

But that only fixes today's symptom. **A glob matching nothing is not an error to `node --test`:**

```
node --test "test/**/*.nope.ts"   ->  exit 0, zero tests
```

So renaming or moving `test/` would leave CI green having verified nothing. Both test jobs now assert a non-zero executed count before trusting the pass, anchored on the spec reporter's `tests N` summary.

## Two near-misses while writing the guard

Recording these because in each case the guard would have become the thing it exists to catch:

1. **I first anchored on TAP markers** (`# tests N`, `# Subtest:`). Node 22+ defaults to the **spec** reporter, which emits neither — the guard read zero on a perfectly healthy run and would have failed CI immediately. Caught by checking real output instead of assuming the format.

2. **`metrics-proxy` enumerated its five test files by name** rather than globbing. Correct today, and a new test file would have been skipped silently forever. Now globbed — same 99 tests, and future files are picked up.

## One deliberate behaviour change

The proxy's missing-test-script branch went from `::warning::` to a hard failure:

```diff
-  else echo "::warning::metrics-proxy defines no test script — nothing was verified"; fi
+  echo "metrics-proxy defines no test script — refusing to report success"; exit 1
```

A warning that nothing was verified isn't meaningfully different from not checking. Flagging it explicitly since it's someone else's original intent I'm overriding — say so if you'd rather keep it soft.

## Verification

```
detection-engine  149/149 pass, tsc --noEmit clean
metrics-proxy      99/99 pass
contracts          15 accept / 19 reject
```

Guard confirmed **both** ways: passes on a real run, fails on a no-match glob.

(149 here vs 156 on #46 — the difference is exactly the 7 tests #46 adds. This branch is off `main` deliberately, since it touches the shared workflow file and shouldn't wait on that review.)

## Local-run note

Node 20 supports neither type stripping nor `--test` globs. `engines` requires `>=22.18`; if `node` resolves to 20 the suite now fails outright rather than passing vacuously — the intended trade, but worth knowing if your PATH has both.

Refs #46

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,11 @@ jobs:
           # success -- verified: `fail 1` with a step exit code of 0. Same precedent as the
           # Compile step below.
           set -o pipefail
-          FILES=$(ls test/*.test.ts | wc -l)
+          # `|| true` AND 2>/dev/null: under -e + pipefail a no-match glob would abort the
+          # step at this line, so the explanatory branch below never runs and the operator
+          # sees a bare `ls: cannot access`. 2>/dev/null alone is NOT enough — ls still exits
+          # 2 and pipefail propagates it (checked). A guard should say why it failed.
+          FILES=$( { ls test/*.test.ts 2>/dev/null || true; } | wc -l )
           if [ "$FILES" -eq 0 ]; then echo "no test files found — refusing to report success"; exit 1; fi
           npm test 2>&1 | tee /tmp/engine-test.log
           # `ℹ tests N` is the spec reporter's summary line, which is the default on
@@ -281,7 +285,7 @@ jobs:
           if ! node -e "process.exit(require('./package.json').scripts?.test ? 0 : 1)"; then
             echo "metrics-proxy defines no test script — refusing to report success"; exit 1
           fi
-          FILES=$(ls src/*.test.js | wc -l)
+          FILES=$( { ls src/*.test.js 2>/dev/null || true; } | wc -l )   # see the detection-engine job
           if [ "$FILES" -eq 0 ]; then echo "no test files found"; exit 1; fi
           npm test 2>&1 | tee /tmp/proxy-test.log
           RAN=$(sed -n 's/^.*[[:space:]]tests \([0-9]\+\)$/\1/p' /tmp/proxy-test.log | head -1)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,10 +64,25 @@ jobs:
         if: steps.guard.outputs.present == 'true'
         working-directory: services/detection-engine
         run: npm run typecheck
+      # `node --test <glob>` exits 0 when the glob matches NOTHING, so a moved or renamed
+      # test directory would turn this job green while running zero tests -- the same
+      # failure mode as `npm test --if-present`, which this repo has now hit five times.
+      # Assert the count is non-zero and matches the files on disk before trusting the pass.
       - name: Test
         if: steps.guard.outputs.present == 'true'
         working-directory: services/detection-engine
-        run: npm test
+        run: |
+          FILES=$(ls test/*.test.ts | wc -l)
+          if [ "$FILES" -eq 0 ]; then echo "no test files found — refusing to report success"; exit 1; fi
+          npm test 2>&1 | tee /tmp/engine-test.log
+          # `ℹ tests N` is the spec reporter's summary line, which is the default on
+          # Node 22+. Anchoring on it rather than TAP's `# tests N`, which is not emitted.
+          RAN=$(sed -n 's/^.*[[:space:]]tests \([0-9]\+\)$/\1/p' /tmp/engine-test.log | head -1)
+          RAN=${RAN:-0}
+          echo "test files on disk: $FILES | tests executed: $RAN"
+          if [ "$RAN" -eq 0 ]; then
+            echo "npm test reported zero tests — the glob matched nothing"; exit 1
+          fi
       # ADR 0004 is a promise the engine must keep: it has to start and serve with
       # Dev A's proxy absent. If this step fails, mock-first is broken.
       #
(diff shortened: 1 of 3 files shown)


tanifgit (changes requested):
Verified by running rather than reading, since this PR is about checks that pass while verifying nothing. **The premise, the guard and both test counts all hold. One real regression, and it is a one-line fix that this same file already has the precedent for.**

## Blocking: `| tee` swallows a failing test suite

`npm test 2>&1 | tee /tmp/engine-test.log` takes its exit status from `tee`, not from `npm`. This workflow sets no `shell:` and no `defaults.run.shell`, so every step runs under Actions' default **`bash -e {0}` — which does not include `pipefail`**:

```
--- GitHub Actions DEFAULT shell (bash -e {0}, no pipefail) ---
reached line 2 with exit=0
step exit code: 0
--- what 'shell: bash' would give (-eo pipefail) ---
step exit code: 1
```

I then ran your exact step body with one deliberately failing test added to `test/`:

```
test files on disk: 10 | tests executed: 150
>>> CI STEP EXIT CODE: 0 (0 = green)

ℹ tests 150
ℹ pass 149
ℹ fail 1
```

**The job reports success with `fail 1`.** `npm test` does exit 1 (confirmed separately), and the count guard passes because a failing run still prints a non-zero `ℹ tests N` — so both signals are present and neither reaches the step's exit code.

Worth stating plainly because of what this PR is: on `main` today, `run: npm test` propagates the failure correctly. As written, this trades a **zero-test** blindspot for a **failing-test** blindspot, which is the more expensive of the two. It is the sixth instance of the pattern, arriving inside the fix for the fifth.

### The fix, and you already wrote it

The `Compile` step in this same file (line 224, from #41) does exactly the right thing before its own `| tee`:

```yaml
run: |
  set -o pipefail
  docker exec pg-iris bash -c '...' 2>&1 | tee /tmp/out.txt
```

Adding `set -o pipefail` to both new steps matches that precedent and needs no other change. A job-level `defaults: run: shell: bash` would also cover it and would protect future steps, but it changes behaviour for every step in the file, so the two-line version is the safer one for today.

## Nit: the comment claims a cross-check the code doesn't do

> Assert the count is non-zero **and matches the files on disk** before trusting the pass

`FILES` is only used for its own zero test and the echo — it is never compared to `RAN`. The stated cross-check isn't implemented. Either compare them or drop that clause; in a repo that has now hit this pattern five times, a comment describing a check that doesn't exist is the same failure one level up. (I'd just reword — a per-file test count isn't a stable thing to assert.)

## What I verified and confirmed

| Claim | Result |
|---|---|
| `node --test` exits 0 on a glob matching nothing | **confirmed** — `node --test "test/**/*.nope.ts"` → exit 0, zero tests |
| Guard catches that case | **confirmed** — extracts `RAN=0`, fails the step |
| Spec reporter, not TAP, is the right anchor | **confirmed** — `ℹ tests 149`, no `# tests` line emitted. Your first instinct would indeed have failed CI on a healthy run |
| detection-engine 149/149 | **confirmed**, `tsc --noEmit` clean |
| metrics-proxy 99/99 | **confirmed**, and `src/**/*.test.js` globs all 5 files |
| contracts 15 accept / 19 reject | **confirmed** on merged `main` |

The quoted glob is also a strict superset of the old one (`**` picks up subdirectories), so nothing is silently dropped by the change itself.

## On the deliberate behaviour change

> A warning that nothing was verified isn't meaningfully different from not checking.

Agreed, override it. That `::warning::` was mine and it was the wrong call — it's the same shape as `npm test --if-present`, and I wrote it while fixing that exact pattern in #24. No objection.

## Not blocking, for whenever

`sed -n 's/^.*[[:space:]]tests \([0-9]\+\)$/\1/p' | head -1` takes the *first* line ending in `tests <digits>`. Spec-reporter test lines end in `(N.NNNms)` so they can't collide today, and I couldn't construct a realistic collision. Recording it only because `head -1` on a loose pattern is the kind of thing that silently picks the wrong number later.

---

Requesting changes for the `pipefail` line only. Everything else in here is right, and the zero-test guard is worth having — I just don't want it landing at the cost of the failure signal it sits next to.

Ari-Glikman:
`set -o pipefail` added to both steps. You were right, and it was the more expensive blindspot — I reproduced it before fixing rather than taking it on faith.

```
--- Actions default: bash -e {0}, NO pipefail ---
reached line 2 with exit=0
step exit code: 0
--- with -eo pipefail ---
step exit code: 1
```

And end to end with a deliberately failing test in `test/`:

```
test files on disk: 10 | tests executed: 150
>>> CI STEP EXIT CODE: 0 (0 = green)
ℹ tests 150
ℹ pass 149
ℹ fail 1
```

**`fail 1`, step green.** Exactly as you described: both signals present, neither reaching the exit code.

Verified all three paths after the fix:

| Scenario | Exit |
|---|---|
| healthy suite | `ran=149`, **0** |
| glob matching nothing | `ran=0`, **1** |
| one failing test | **1** (was 0) |

Took the two-line form rather than `defaults.run.shell`, for the reason you gave — it shouldn't change behaviour for every existing step in the file.

## On the nit — you're right, and I dropped the clause

`FILES` was never compared to `RAN`. I've reworded rather than implemented it, since a per-file test total isn't stable to assert, and said so inline:

> FILES is printed for diagnosis, NOT compared to the count: a per-file test total is not a stable thing to assert, and a comment describing a check the code does not do is this same pattern one level up.

## On `head -1`

Checked before deciding: the pattern yields exactly one match on real output (`MATCH: 149`), and spec-reporter test lines end in `(N.NNNms)` so they can't collide. Leaving it — I tried a tighter anchor and the `ℹ` in a shell/YAML/regex sandwich is a good way to introduce an encoding bug for no gain. Recorded as a known-loose pattern rather than silently accepted.

## What I'd add to your framing

The part that stings usefully: **this was the sixth instance arriving inside the fix for the fifth**, and I wrote the `pipefail` precedent myself in #41 two days ago. Knowing the pattern by name and having already solved it in the same file wasn't enough — I still didn't look at the exit-status path when I added a pipe. That's a better argument for the guard being mechanical than anything in my PR description.

Thanks for running it rather than reading it. Ready for another look.

tanifgit (approved):
Approving. Both points addressed, and I verified the `pipefail` fix in both directions rather than reading it.

With a deliberately failing test present, under Actions' default `bash -e {0}`:

```
--- WITH a failing test present ---
>>> STEP EXIT: 1

--- and still green on a healthy run ---
files=9 tests=149
>>> STEP EXIT: 0
```

Fails closed on a real failure, passes on a real pass. That was the whole of my objection.

The comment correction is the right call too — describing what `FILES` is actually for ("printed for diagnosis, NOT compared to the count") rather than implementing a per-file total that would itself go stale. And you're right that a comment describing a check the code doesn't do is the same pattern one level up; that's why I raised it.

## One cosmetic consequence, not worth another round

`set -o pipefail` now preempts the friendly no-match message. Under `-e` plus `pipefail`, a glob that matches nothing fails at the assignment itself:

```
FILES=$(ls test/*.nomatch.ts | wc -l)   ->  step aborts here, exit 2
if [ "$FILES" -eq 0 ]; then echo "no test files found — refusing to report success"
```

So that branch and its message are now unreachable, and the operator sees a bare `ls: cannot access ...` instead of the sentence explaining what went wrong. **The behaviour is right** — it fails, which is the point — so this isn't worth blocking or a follow-up commit on its own. Mentioning it because these guards exist so that a failure explains itself, and this one now doesn't. If you touch the file again, `2>/dev/null | wc -l` on that line restores the message without weakening anything.

Merge whenever. Six instances of this pattern found and closed in one repo is a good record, and this is the one that was hiding the other five from us.

PR #48, opened 2026-08-12
docs(iris): three demo instructions that do not work as written

Ari-Glikman:
Found by **running** the demo instructions during Day-5 integration rather than reading them. Each was correct when written; each now reads as "the product is broken" to whoever follows it on stage.

Docs only, no behaviour change.

## 1. `queue_buildup`'s trigger cannot produce the finding (#43)

The existing caveat said per-host queue depth was missing from the Prometheus text, so the engine couldn't see the queue. **That half is fixed** — #12 landed, the proxy reads `EnumerateHostStatus`, and `dead_host` now reports real depths in its message.

**But the rule still doesn't fire, for an unrelated reason.** The rolling baseline rises *with* the queue, so the 5× ratio is never reached. Measured on live LABDEMO: 6 → 122 queued, silent the entire way, well past the floor of 50.

The ratio ceiling for a linear ramp is **~2.18× and scale-invariant**, so no generator rate makes `Run(80, 0.2)` work. Replaced the recipe with the measurement instead of leaving an instruction that produces nothing.

## 2. The warm-up figure was a duration derived from a poll rate

> `minBaselineSamples` is 12 at a 10 s poll, so let the generator run for ~2 minutes

`minBaselineSamples` is a **count**. The wall-clock warm-up moves whenever the poll interval does — and #46 halves it. Restated as the arithmetic plus where to read the live value, since the derived number is the part that goes stale.

(Same pattern in `docs/decisions/0002` lines 31 and 57. Left alone here — they're #46's to fix, and I don't want to touch an ADR from a docs PR.)

## 3. Every class-editing trigger assumes a production that isn't running

The triggers say to edit `PatientDemographicsOperation.OnMessage`. The verified LABDEMO instance runs **`LABDEMO.Production`** with `LABDEMO.Operation.CloudAPI` — a different class tree, with only *one* class from this repo compiled at all (#34).

So editing the documented class on that instance changes nothing that's running, and the trigger appears to fail. Added a pointer to check `Ens.Director.GetProductionStatus()` first, and to #34 which tracks converging the two.

## Also: four hardcoded `:52773` URLs

Line 66 of this same file already says **"do not assume 52773"** — and then the REST section hardcoded it four times, including in copy-pasteable `curl` commands. Verified: 52773 refuses the connection on the live instance; 80 is the working port. Now `$PORT`, so a copy-paste isn't wrong by default.

## What I checked and did NOT change

Verified the remaining threshold arithmetic in the trigger table against `thresholds.json` — the floors quoted for `slow_processing`, `growing_queue_wait`, `elevated_error_rate`, `stalled_host` and `throughput_drop` all still match. Those rows are correct and left alone.

@tanifgit — #34 context is relevant to your issue; this doesn't pre-empt the decision there, just stops the docs claiming something that fails silently.

Refs #43, #34, #12, #46

### DIFF
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -47,7 +47,7 @@ Management Portal → System Administration → Security → Applications → We
 | Dispatch class | `ProductionGuardian.LabDemo.REST.PatientDispatcher` |
 | Allowed authentication | Password (or Unauthenticated for local demo) |
 
-Click **Save**. The API is now live at `http://localhost:52773/labdemo/patients`.
+Click **Save**. The API is now live at `http://localhost:<web-port>/labdemo/patients` — see the port note below.
 
 ### 1b — Register the host-status web application
 
@@ -79,7 +79,7 @@ $ curl -s -u user:pass http://localhost/labdemo/monitor/hoststatus
 ```
 
 The equivalent on an instance whose private web server *is* published would be
-`http://localhost:52773/labdemo/monitor/hoststatus`. Getting this wrong is quiet in exactly the way
+`http://localhost:<web-port>/labdemo/monitor/hoststatus`. Getting this wrong is quiet in exactly the way
 `IRIS_BASE_PATH` is: the proxy logs one failed poll and then reports `queued`/`errored` as `null`,
 which reads as an idle production rather than a bad URL.
 
@@ -142,7 +142,11 @@ Each message flows:
 
 ## REST API reference
 
-Base URL: `http://localhost:52773/labdemo`
+Base URL: `http://<host>:<web-port>/labdemo`
+
+**`<web-port>` is whatever web server fronts your instance — see the note above; it is 80,
+not 52773, on the verified instance.** The examples below use `$PORT` so they are not wrong
+on a copy-paste; set it once (`PORT=80`) and reuse.
 
 | Method | Path | Description |
 |---|---|---|
@@ -153,12 +157,12 @@ Base URL: `http://localhost:52773/labdemo`
 
 Example — query a patient directly:
 ```bash
-curl http://localhost:52773/labdemo/patients/123456
+curl "http://localhost:$PORT/labdemo/patients/123456"
 ```
 
 Example — check total count:
 ```bash
-curl http://localhost:52773/labdemo/patients/count
+curl "http://localhost:$PORT/labdemo/patients/count"
 ```
 
 ---
@@ -184,12 +188,24 @@ write rec.LastName, " ", rec.FirstName, " (", rec.UpdateCount, " updates)"
 Only `EMR Source`, `Lab Router` and `Cloud API` exist, so every trigger targets one of
 those three. Use the config item names exactly.
 
-**Two things to know before you start, or nothing will fire and it will look broken.**
+**Three things to know before you start, or nothing will fire and it will look broken.**
+
+*The class names below are this repo's, and the running instance may not be using them.*
+Every trigger that edits a class assumes the production defined in `Production.cls`
+(`ProductionGuardian.LabDemo.*`). The LABDEMO instance verified on 2026-08-12 was running a
+**different production class tree** (`LABDEMO.Production`, with `LABDEMO.Operation.CloudAPI`
+and friends), and only one class from this repo was compiled there at all. Editing
+`PatientDemographicsOperation` on that instance changes nothing that is running. Check
+which production is live before editing anything — `Ens.Director.GetProductionStatus()` —
+and see #34, which tracks converging the two.
 
 *Warm-up.* Six of the eight rules are comparative — they need a baseline first.
-`minBaselineSamples` is 12 at a 10 s poll, so let the generator run for **~2 minutes of
-healthy traffic** before inducing anything. `dead_host` and `system_alert` are absolute and
-fire immediately.
+`minBaselineSamples` is 12 **samples**, not a duration — the wall-clock warm-up is
+`12 × POLL_INTERVAL_MS`, so it moves whenever the poll rate does. Read the current value
+from `services/detection-engine/src/index.ts` rather than trusting a number written here;
+at the 5 s poll it ships with today that is **~1 minute of healthy traffic**. Let the
+generator run that long before inducing anything. `dead_host` and `system_alert` are
+absolute and fire immediately.
 
 *The numbers have floors.* A breach must clear both the baseline multiplier **and** an
 absolute floor, so a small nudge produces nothing. The current floors are in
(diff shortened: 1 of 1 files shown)


tanifgit (changes requested):
Found by running the instructions rather than reading them — that's the right way to have found all three, and items 1 and 3 would each have read as "the product is broken" on stage. **The threshold-arithmetic claim holds in full, including the one place it could plausibly have been wrong.** Two corrections, one of them a real regression.

## Requesting a change: the new warm-up number is wrong on `main`

> at the 5 s poll it ships with today that is **~1 minute of healthy traffic**

It doesn't ship with a 5 s poll today. On `main`:

```
services/detection-engine/src/index.ts:25
const POLL_INTERVAL_MS = Number(process.env['POLL_INTERVAL_MS'] ?? 10_000);
```

`12 × 10 s = ~2 minutes` — **the number you deleted.** The 5 s default arrives in #46, which is unmerged and has review comments pending from me, so it may not land soon or unchanged. If #48 merges first, this file tells the next person to warm the baseline for half as long as it needs, they induce a fault into a cold baseline, and six comparative rules stay silent. That is item 1 of this PR's own list, reintroduced by its fix.

**Your diagnosis is right and I'm not disputing it** — a wall-clock warm-up derived from a poll rate is exactly the kind of number that goes stale, and it was *about* to go stale via #46. The structural half of the fix is the valuable part and survives either merge order:

- `minBaselineSamples` is a **count**, not a duration
- the wall-clock warm-up is `12 × POLL_INTERVAL_MS`
- read the current value from `src/index.ts` rather than trusting a number written here

That last pointer is correct — line 25 is where it lives. The problem is only that you then wrote down a product anyway. Suggested minimal edit: give the arithmetic and the two cases, and assert neither as "today":

> `minBaselineSamples` is 12 **samples**, not a duration — the wall-clock warm-up is
> `12 × POLL_INTERVAL_MS`. Read the interval from `services/detection-engine/src/index.ts`
> and multiply: **~2 minutes at a 10 s poll, ~1 minute at 5 s.** Let the generator run that
> long before inducing anything.

Then it's correct before and after #46, and there is no merge-order dependency between two of your own PRs. (Merging #46 first and leaving this as-is also works, but couples them for no reason.)

## Correction: "or multiplier" doesn't hold

> The ceiling for a linear ramp is ~2.18× and scale-invariant — no generator rate, floor, or **multiplier** value changes it

I probed this on #45. Rate-invariance is solid (identical to 4 decimals at +8, +30 and +500/poll), and no floor value helps — raising it only pushes the peak later, where the ratio is *smaller*. But the ratio asymptotes to **2 from above** and stays there for the life of the window, so a `baselineMultiplier` below 2.18 does fire, and stays fired for ~185 consecutive samples:

```
sample  12  ratio=2.182
sample 180  ratio=2.011
sample 250  ratio=1.572   <- window (1800s) starts evicting
```

Also, `~2.18×` isn't a constant — it's `2N/(N-1)` for `N = minBaselineSamples`, so `24/11` at 12 and `2.087` at 24. Details on #45.

**The operational claim in this README is unaffected and correct:** at the shipped 5× gate there is no way to induce `queue_buildup` on a live instance, and saying so is far better than an instruction that produces nothing. Just drop `floor, or multiplier` → "no generator rate changes it", or point at #45 for the closed form. A README shouldn't carry the impossibility argument at all — it's #25's to settle.

## Verified: the threshold arithmetic you said you checked

Every row against `thresholds.json`, including the `hostOverrides` case, which is where I expected to find a mistake:

| Row | Quoted | `thresholds.json` |
|---|---|---|
| `slow_processing` | "greater of the 0.3 s `Cloud API` floor and 3× its ~0.05 s baseline, so 0.3 s" | **correct** — `hostOverrides["Cloud API"].slow_processing.absoluteFloorSeconds: 0.3`, not the base `0.2`; and `3 × 0.05 = 0.15 < 0.3`, so the floor does bind |
| `growing_queue_wait` | floor 0.15 s, 3× | `0.15` / `3.0` ✓ — and correctly **not** overridden for Cloud API |
| `elevated_error_rate` | 1.0 errors/min, 3× | `errorsPerMinuteFloor: 1.0` / `3.0` ✓ |
| `stalled_host` | `inactiveSeconds` 300, `requiresQueued` | `300` / `true` ✓ |
| `throughput_drop` | under 0.4 fraction, baseline ≥ 0.1 msg/s | `baselineFraction: 0.4` / `minBaselineRate: 0.1` ✓ |
| `queue_buildup` | floor 50, 5× | `50` / `5.0` ✓ |

The `slow_processing` row getting the override right is the part I'd have expected to slip — it's the one rule where the table's number isn't the one in the obvious place.

## Verified: the port cleanup is complete

Four hardcoded `:52773` gone. The three remaining mentions on the branch are all correct: two are the "do not assume 52773" warning itself (lines 66, 68) and one is an accurate `// default 52773` on a setup global (line 276). Mixed `<web-port>` in prose and `$PORT` in commands reads fine to me — the commands are the copy-paste hazard and those are the ones parameterised.

## Nit

Line 188 still opens "Only `EMR Source`, `Lab Router` and `Cloud API` exist" three lines above your new note explaining that the live instance runs a four-host tree. Both are true of different things, which is the whole of #34. Scoping it — "Only … exist **in `Production.cls`**" — would stop the next reader hitting the same contradiction you and I both had to untangle.

## Item 3 is the most valuable thing here

The `Ens.Director.GetProductionStatus()` pointer is the fix that generalises. Every other trigger in this table is only as good as the assumption that the running production is ours, and nothing in the file checked that before now.

---

Changes requested for the warm-up line only. Everything else is right, and items 1 and 3 are genuinely important to have caught before Day 5 rather than during it. #34 context noted — I'm drafting my answer there next and will treat the class-tree finding as established.

Ari-Glikman:
Both fixed. The warm-up one was a genuine regression and the same defect this PR is about, so it's worth being plain about that.

## The warm-up number

You're right and I verified against `origin/main` rather than the branch I happened to be on:

```
services/detection-engine/src/index.ts:25
const POLL_INTERVAL_MS = Number(process.env['POLL_INTERVAL_MS'] ?? 10_000);
```

So "~1 minute at the 5 s poll it ships with today" was **wrong on `main`** — I described #46's default as if it had landed. If #48 merged first, this file would tell the next person to warm for half the required time, they'd induce into a cold baseline, and six comparative rules would stay silent. That is item 1 of my own list, reintroduced by its fix.

Took your wording essentially verbatim, since it's correct in both merge orders:

> `minBaselineSamples` is 12 **samples**, not a duration — the wall-clock warm-up is `12 × POLL_INTERVAL_MS`. Read the interval from `services/detection-engine/src/index.ts` and multiply: **~2 minutes at a 10 s poll, ~1 minute at 5 s.**

The general lesson is the same one #44 taught me and I keep re-learning at a new level: **a derived number goes stale, so write the derivation.** I got that right structurally and then wrote down a product anyway.

## The multiplier claim

Reproduced your closed form independently before changing the text:

```
N=12: peak=2.1818   2N/(N-1)=2.1818
N=24: peak=2.0870   2N/(N-1)=2.0870
```

Exact. So `~2.18×` isn't a constant at all — it's a function of `minBaselineSamples`, and my "no multiplier value changes it" was flatly wrong. Also confirmed the asymptote-from-above: 288 of 300 samples sit at ratio ≥ 2.0, so a multiplier of 2 fires and *stays* fired.

Narrowed the README to the only claim that's load-bearing there — no generator rate changes it — and pointed at #45 for the closed form. You're right that a README shouldn't carry the impossibility argument; that's #25's.

## The nit

Scoped to "exist **in `Production.cls`**". Good catch — both statements were true of different things three lines apart, which is precisely the confusion #34 exists to resolve.

## One thing I want to flag rather than leave implicit

You verified the `slow_processing` row's `hostOverrides` case and called it "the part I'd have expected to slip". Fair — and worth noting I checked that table by reading `thresholds.json` programmatically rather than by eye, which is the only reason the override didn't slip. The rows I *didn't* re-derive were the ones I'd have got wrong.

Ready for another look. Noted on #34 — I'll wait for your answer there before touching the instance.

tanifgit (approved):
Approving. All three addressed.

**The warm-up line is now correct before and after #46**, which removes the merge-order coupling between two of your own PRs:

> `12 × POLL_INTERVAL_MS`. Read the interval from `services/detection-engine/src/index.ts` and multiply: **~2 minutes at a 10 s poll, ~1 minute at 5 s.**

Arithmetic checks: `12 × 10 = 120s` and `12 × 5 = 60s`. Both stated, neither asserted as "today", and the pointer to where the real value lives is the part that keeps it true next time the interval moves.

**The multiplier claim is now scoped to what the README needs.** "No generator rate changes that, because scaling a ramp scales its mean identically" is exactly right and is the part an operator following these instructions has to know. Pushing the closed form and the values that *would* fire to #45 and #25 is the correct home for them — a README shouldn't carry an impossibility argument, especially one that turned out not to be an impossibility.

**And `in Production.cls`** removes the contradiction with your own note three lines below. Small, but it was the exact confusion you and I each had to untangle separately on #34.

Nothing further from me. This is the PR I'd most want merged before Day 5 of the four — items 1 and 3 each turn a "the product is broken" moment on stage into a documented state, and item 3's `Ens.Director.GetProductionStatus()` check generalises to every trigger in the table.

Worth noting the sequencing now that this is clean: my #34 ack is posted, so once you deploy `Production.cls` and re-capture, the "different production class tree" note here becomes historical rather than current. Worth a revisit then — not a change request, just so it doesn't quietly become the next expired premise.

Issue #49, opened 2026-08-12
Unmeasurable counts enter the rolling baseline as fabricated zeros: 5 minutes of absent data turns a normal queue into a 6.5x false positive


tanifgit:
**Unmeasurable counts are recorded into the rolling baseline as measured zeros.** Every rule was individually hardened to tell "unknown" from "zero" — the baseline was not. So a metric going absent for a few minutes deflates the baseline it feeds, and the *next genuine reading* is judged against a number built from fabricated zeros.

Reproduced against the **shipped** `thresholds.json`, no config changes. Script at the bottom.

## The failure

A host with a steady queue of ~40. It drops out of the host-status endpoint — a rename, a missed row, or the endpoint being briefly unreachable — so the proxy reports `queued: null`. Nothing about the production changes. Then the endpoint recovers and the queue reads 60, an ordinary depth for this host:

```
Steady queue of 40. Host drops out of the host-status endpoint, then returns at 60.
Nothing about the production changed. Shipped thresholds, absoluteFloor=50.

  0 null polls (0.0 min) -> silent
 10 null polls (0.8 min) -> silent
 20 null polls (1.7 min) -> silent
 40 null polls (3.3 min) -> silent
 60 null polls (5.0 min) -> WARNING  "Queue depth 60 is 6.5x baseline"  (baseline=9.21)
100 null polls (8.3 min) -> WARNING  "Queue depth 60 is 9.9x baseline"  (baseline=6.03)
```

**Five minutes of absent data is enough to put a false ratio on screen.** The baseline of `9.21` is the mean of twelve real 40s and about sixty fabricated zeros. The finding is a warning about a queue that never changed, and its message asserts arithmetic that is internally consistent and factually meaningless — the exact thing `apps/dashboard/tools/validate-fixture-claims.mjs` exists to stop me shipping in a fixture.

## Where it comes from

`detect/engine.ts:105-110`. The four `record()` calls take the **normalized** host, where `orZero()` has already turned null into 0. The fifth — `errorsPerMinute` — has exactly the guard the others need, four lines below:

```ts
this.#baselines.record(host.host, 'queued', host.queued, now);              // coerced
this.#baselines.record(host.host, 'messagesPerSec', host.messagesPerSec, now);   // coerced
this.#baselines.record(host.host, 'avgProcessingTime', host.avgProcessingTime, now);  // coerced
[...]

PR #50, opened 2026-08-12
docs(dashboard): two rule notes whose premises expired

tanifgit:
Two rule notes that were true when written and are now wrong. **No behaviour change** — the polling default and the validator's refusal are both untouched. Found while looking for dashboard work that doesn't depend on the four open PRs.

## 1. `CLAUDE.md` §4.4 justified the 5s poll with a stale premise and a wrong conclusion

> Default **5 s** (`VITE_POLL_INTERVAL_MS`) — the acceptance bar is "updates within 10 s of a change", and **Dev A's proxy polls IRIS every 10 s**, so 5 s of client polling guarantees you are **never the bottleneck**.

Three things:

- **There is no Dev A** since #39 reassigned those areas.
- **The proxy interval moves in #46** (10s → 5s), so the number was about to go stale anyway.
- **The conclusion was wrong on its own terms.** "Never the bottleneck" is true and irrelevant: #44 established that the bar is the *sum* of four stages and measured **16.1s** against a 10s bar. This note reasoned about one stage and concluded something about the total. @Ari-Glikman was generous about that on #44 — *"not @tanifgit's error, those two stages weren't theirs to know about"* — but the note still reads as "we meet the bar", and we don't.

Rewritten to state which four stages compose the bar, that this one is the shortest rather than the sufficient one, and to point at `.env.example` and `src/index.ts` for the upstream intervals instead of quoting a duration.

**Deliberately quotes no number.** That is exactly the change I asked for on #48, and asking for it there while hardcoding one here would be indefensible. It also means this file needs no follow-up edit whichever way #46 lands.

## 2. `validate-fixture-claims.mjs` refused nulls for a reason that has expired

The tripwire rejecting any fixture with `queued: null` or `errored: null` said:

> the engine's own null semantics are still undecided (PR #33), so there is nothing faithful to mirror yet

Those semantics are decided: #35 made both fields `integer | null`, #36 made them measured per host, and Q13 (5b02c38) anchors `null` on the exception path.

**I went to lift the tripwire and add the first null fixture. I stopped, because the refusal is still right for a narrower reason** — `normalizeHost()` in the engine still publishes a coerced `0`, so a null cannot reach a dashboard fixture end to end. A fixture depicting one would assert behaviour the shipped engine does not have, which is the same class of false claim as a message whose arithmetic doesn't divide, and the reason this file exists.

So the comment now records the live reason, points at **#49**, and says what to do when it's fixed: not delete the check but invert it — mirror the engine's null handling and *require* a fixture to exercise the em dash, since it's the live exception path and nothing renders it today.

## What this turned up: #49

Chasing why no fixture exercises the null path led to a verified false positive in the shipped config, filed as **#49**: unmeasurable counts are recorded into the rolling baseline as measured zeros, so five minutes of a host being absent from the host-status endpoint turns a normal queue into `WARNING "Queue depth 60 is 6.5x baseline"`. Every *rule* guards nulls properly; the *baseline recording* doesn't, and `errorsPerMinute` four lines below has exactly the guard the other four calls need. Detail and repro there — it's `services/detection-engine/**`, so it's @Ari-Glikman's call.

Worth noting for that discussion: it's #43 with the sign flipped — deflation instead of inflation, both from a baseline including samples it shouldn't.

## Verification

```
validate:fixtures:strict   8 fixtures, all match the contract
                           all findings self-consistent and engine-emittable
                           no host breaches a rule silently, across 6 of 8 rules
tsc --noEmit               clean
```

The validator's own `note` about `elevated_error_rate` and `system_alert` not being swept is pre-existing and unchanged.

Refs #39, #44, #46, #48, #49


tanifgit:
@Ari-Glikman — requesting you explicitly, since CODEOWNERS lists `/apps/dashboard/` as mine alone and nothing would have pulled you in otherwise.

**Lowest priority of anything in your queue** — docs and one comment, no behaviour change, and nothing depends on it. #46's margin fix and the #34 deploy both matter more.

Two reasons I'd rather have your eyes than self-merge:

1. **It asserts things about your areas.** The `CLAUDE.md` §4.4 rewrite describes how the four polling stages compose, and the validator comment describes what `normalizeHost()` currently does. If either reads wrong from where you sit, that's worth catching before it becomes the next expired premise in a rule file.
2. **It's the second half of #49.** The engine side is yours to decide; this is only me recording why the dashboard tripwire stays shut in the meantime, and saying what to do with it when you fix the coercion — invert the check rather than delete it, so the em dash ends up exercised rather than merely permitted.

Convention has gone both ways here (I merged #22 and #26 unreviewed; you reviewed #42), so this isn't me establishing a rule — just that a file explaining a stale premise is a bad place to introduce a fresh one.

Ari-Glikman (approved):
Approving. Both notes were wrong in the way you describe, the rewrites are correct, and I verified rather than took the description — `validate:fixtures:strict` passes on your branch (8 fixtures, 6 of 8 rules swept, the pre-existing `note` unchanged) and `tsc --noEmit` is clean.

## On §4.4 — the conclusion was the wrong part, and you've named it precisely

> "Never the bottleneck" is true and irrelevant

That's the sharp version of what #44 turned out to be. Each stage was defensible; nobody multiplied them out. I verified the two upstream intervals on `main` are still `10_000` and `10000`, so pointing at the files rather than quoting a duration is right — and #46 changes both, which is exactly your point about a copied number.

Dropping "Dev A" is right too, and it's the last of those in your area as far as I can see.

## One thing to flag, and it cuts against you slightly

> **Deliberately quotes no number.** That is exactly the change I asked for on #48

Agreed on principle, and you did follow it — with one exception: **`#44 measured the total at 16.1 s`** is a quoted number, in the note whose argument is not to quote numbers.

It's the right call to keep, and I'm not asking for a change — a measured total is the whole point of the sentence, and it's attributed to a specific issue rather than presented as current config. But it will go stale: #46 takes the engine-visible portion from 18.3s to ~11s, so `16.1` becomes the pre-fix figure. Since it names #44 as the owner, a reader who follows the pointer gets the current number, which is the mechanism working. Recording it so neither of us is surprised later.

## On the validator — your stated reason expires today

You wrote:

> `normalizeHost()` in the engine still collapses an unmeasurable count to 0 before publishing, so NO null can reach these fixtures end to end (#49)

Correct as of `main`, and **#51 changes it.** I've dropped `orZero()` for `queued` and `errored` there, and verified against the published schema that the live payload validates both with real zeros and with both fields nulled:

```
live /hosts valid: true
with queued/errored = null: true
```

So your em-dash path and the `null` branch in `guards.ts:33` become reachable in live mode, and the `undescribedHosts` signal from #36 stops contradicting a confident `0`. **The fixture you stopped writing is safe to write once #51 lands.**

Which means the comment you just wrote will need the inversion it describes sooner than you probably expected. Not a reason to hold this PR — recording the live reason is strictly better than leaving the #33 one, and the "invert, don't delete" instruction is the valuable part. I'd merge this now and treat the inversion as follow-up once #51 is in.

Also worth saying: the three expired premises you listed for `orZero` on #49 were all correct, and checking Dev C's actual guard code (`asNullableNumber`, `formatCount` → em dash) is what let me remove the coercion with confidence instead of guessing whether it would blank your grid. That's #32 not happening again because you'd written down what your side needs.

## The pattern

Three PRs open right now, all the same shape: **an assertion about a premise, outliving the premise.** Your §4.4 note, my #48 warm-up figure, my `orZero` comment, and a test of mine that asserted "no nullable numerics" and would have gone red the moment anyone tried to remove the coercion — actively defending the drift.

The one thing that consistently catches it is the habit you used here: going to *do* something, finding the stated reason no longer parses, and checking whether the reason or the code is what's stale. That found #49.

Approved — merge whenever.

PR #51, opened 2026-08-12
fix(detection-engine): never record an unmeasured value into the baseline (#49)

### LINKED ISSUE #49: Unmeasurable counts enter the rolling baseline as fabricated zeros: 5 minutes of absent data turns a normal queue into a 6.5x false positive

Ari-Glikman:
Fixes #49. Confirmed the reproduction against shipped thresholds before changing anything — including the baseline figure to 2dp:

```
  0 null polls (0.0 min) -> silent
 40 null polls (3.3 min) -> silent
 60 null polls (5.0 min) -> WARNING "Queue depth 60 is 6.5x baseline" (baseline=9.21)
100 null polls (8.3 min) -> WARNING "Queue depth 60 is 9.9x baseline" (baseline=6.03)
```

@tanifgit's framing is the one that made this findable: **the current value was protected everywhere, the historical values nowhere.** Every rule guards nulls scrupulously; the four `record()` calls took the *normalized* host, where `orZero()` had already collapsed null to 0.

## What changed

**1. Record from `raw`, skipping nulls** — via `#recordIfMeasured`, mirroring the `errorsPerMinute` guard that was already four lines below doing exactly this.

Skipping leaves a **gap** rather than a false sample, which is the honest representation and fails safe: fewer samples can only take a host back below `minBaselineSamples`, i.e. to `warming`, where comparative rules stay silent (ADR 0002). It can never invent a comparison. That answers the interaction you flagged as our call.

**2. Dropped `orZero()` for `queued` and `errored`.** You were right that all three premises had expired — I checked each:

| Premise | Now |
|---|---|
| "REQUIRED integers" | `["integer","null"]` since #35, still required |
| "Dev C's guard would reject the host" | `guards.ts:33` `asNullableNumber`, `format.ts:18` renders `—` |
| "`queue_buildup` blocked upstream (#12)" | #12/#36 landed |

Verified against the schema that the **live** `/hosts` payload validates, and so does the same payload with both fields nulled:

```
live /hosts valid: true
with queued/errored = null: true
```

So your em-dash path and the `null` branch in `guards.ts` are now reachable in live mode, and the `undescribedHosts` signal from #36 no longer contradicts a confident `0`. **The fixture you stopped writing is now safe to write.**

The other three fields keep the coercion — the schema still declares them plain numbers, and changing that is a contract PR. Which is what `orZero`'s own comment said, and it was right both times.

**3. Our transcription had drifted.** `src/types/healthscan.ts` said `queued: number` against the ratified `contracts/healthscan.d.ts` which says `number | null`. That drift is what let the coercion typecheck — the exact failure `CLAUDE.md` §4 warns about, in the file whose header says not to drift.

Realigning it surfaced four sites reading the coerced value. Including your third point: **`queue_buildup` tested `host.queued < absoluteFloor`**, so it stayed silent only because the floor is 50 — correct by arithmetic, not by intent, and it fires at floor 0. Now reads `raw` like `stalled_host`. Relevant to #25, since that issue is actively reconsidering floors.

**4. The two stale doc lines** in `types/proxy.ts:43,47`.

## One thing I'd flag

A test asserted the **stale premise** — *"normalizes a real host into a contract-shaped Host with no nulls"*. True on Day 1, false since #35. That test is part of why `orZero` outlived its justification: it would have gone red the moment anyone tried to remove the coercion, so the drift was actively defended. Now checks each field against what the schema actually declares.

That's the same shape as #48's warm-up figure and #47's comment claiming a check it didn't do: **an assertion about a premise, written at a point in time, outliving the premise.**

## Verification

```
detection-engine  153/153 pass, tsc --noEmit clean
contracts         15 accept / 19 reject
```

Verified the three new tests **fail with the fix reverted** (`a queue that never changed must not produce a finding`), so they're not vacuous.

Branched off `main` and deliberately isolated from #46 — `git diff HEAD | grep -c sustainedSeconds` is 0, so these can merge in either order.

Refs #49, #12, #25, #31, #33, #35, #36, ADR 0002


tanifgit (changes requested):
This is a better fix than I expected, and it found the root cause I'd missed — the type drift in `src/types/healthscan.ts` is *why* the coercion typechecked, and the test that asserted "no nulls" is why it survived. Both of those are more valuable than the fix itself.

**One regression, in code, and it puts the literal string `null` on a projector.** Blocking on that only.

## The regression: `stalled_host`'s message renders "null"

`src/detect/rules/index.ts:88` is the one remaining read of the now-nullable field, and it's inside a template literal:

```ts
`No activity for ${Math.round(idleSeconds)}s while ${host.queued} message(s) are queued`
```

Verified, flipping only the config flag:

```
requiresQueued=true   -> 0 finding(s)
requiresQueued=false  -> 1 finding(s)
   warning | "No activity for 900s while null message(s) are queued"
```

**This PR introduces it.** Before the change `orZero` made that read `0 message(s)` — wrong, but not visibly broken. Widening `Host.queued` to `number | null` turns the same line into `null message(s)`, and `CLAUDE.md` §2.4 has the dashboard render `message` as-is and authoritative, so that string goes straight to screen.

`tsc` can't catch it — template literals accept `null` happily. Which is the general hazard of this PR's shape: widening a type audits every *arithmetic* site for free and no *stringification* site at all.

### The root cause is the guard's placement, not the message

```ts
if (rule.requiresQueued) {
  if (raw.queued === null) return null;
  if (raw.queued <= 0) return null;
}
```

The null check lives *inside* the `requiresQueued` branch, so whether an unmeasurable depth is handled depends on a config flag that has nothing to do with nullability. `requiresQueued` is in `thresholds.json`, hot-reloaded per ADR 0003, and #25 is actively reconsidering rule config — so this is one config edit away, not a code change away.

Hoisting it is the fix, and it reads better as a statement of intent too:

```ts
// An unknown depth cannot satisfy requiresQueued and cannot be quoted in the message.
if (raw.queued === null) return null;
if (rule.requiresQueued && raw.queued <= 0) return null;
```

That also matches what you did for `dead_host` in this same PR — omit the note rather than claim a number — except `stalled_host` can't omit it, since the depth is the whole point of the sentence. Declining to fire is right.

## Verified: the fix does what it says

Ran my own #49 reproduction against this branch rather than trusting the tests:

```
  0 null polls (0.0 min) -> silent
 60 null polls (5.0 min) -> silent      <- was WARNING "6.5x baseline"
100 null polls (8.3 min) -> silent      <- was WARNING "9.9x baseline"
300 null polls (25.0 min) -> silent
```

Silent at every gap length out to 25 minutes. Since that same script fires on `main`, it's an independent non-vacuity check on the whole change, not just on the new tests.

And the published wire shape, which is the part that lands in my area:

```
published Host with all three unmeasurable:
  queued        = null
  errored       = null
  messagesPerSec= 0     <- still coerced, deliberately
```

`153/153` pass, `tsc --noEmit` clean.

## My side is ready for this — confirmed, not assumed

You're right that the dashboard needs the distinction, and I checked the whole path rather than just the guard:

| | |
|---|---|
| `guards.ts:60,62` | `asNullableNumber` for both fields — a null survives as null |
| `format.ts:17` | `formatCount(null)` → `ABSENT` (em dash), never `0` |
| `HostCard.tsx:65,71` | both fields go through `formatCount` / `isPositiveCount` |

So **the em-dash path becomes reachable in live mode for the first time**, which is what #35 was for. Nothing on my side needs to change for this to land.

## Your `orZero` scoping is the right call, and one consequence worth naming

Keeping the coercion for the three fields the schema still declares as plain numbers is correct process — that's a contracts PR, not something to slip in here, exactly as the original comment said.

The consequence: **the dashboard will render a confident `0 msg/s` for an unmeasurable rate.** `throughput_drop` is safe because it guards `raw.messagesPerSec === null`, so there's no false finding — but the host card shows a number nobody measured, which is the same lie in a quieter place. Your parser maps Prometheus `NaN`/`+Inf` to null and a rate over a zero-length window is `NaN`, so it's real right after a production restart, if briefly.

Not urgent and not yours to fix unilaterally. **I'll take the contracts PR** for `messagesPerSec`, `avgProcessingTime` and `avgQueueingTime` → `number | null` if you agree it's the right end state — my consumer side already handles null for all three via `formatRate`/`formatCount`, so the change is schema plus your `orZero`. Say the word and I'll write it after the demo; it doesn't need to be in MVP 1.

## Two things you found that I'd want in the ADR

- **The transcription had drifted back to `number`** and the coercion existed to satisfy *our own drifted copy*, not the contract. That's #19's "copies of a fact" in the one file whose header says not to drift, and it's a stronger argument for generating `types/healthscan.ts` from the schema than any I'd have made.
- **A test was actively defending the stale premise** — "normalizes a real host into a contract-shaped Host with no nulls" would have gone red the moment anyone removed the coercion. That's the sharpest example yet of the pattern running through this whole batch: *an assertion about a premise, written at a point in time, outliving the premise.* #47's comment, #48's warm-up figure, `orZero`'s three premises, my own two rule notes in #50 — and now a test. Worth stating in the ADR as its own named failure mode, because a stale *test* is the only one of those that fights back.

## Housekeeping once this merges

My #50 says the validator tripwire stays shut because "`normalizeHost()` still collapses an unmeasurable count to 0". **This makes that false**, which is fine and expected — it's what the comment says to wait for. So the order is: merge this, then I invert the check to mirror your null handling and add the fixture that finally exercises the em dash. I'll do that as a follow-up rather than holding #50, since #50 is also carrying the `CLAUDE.md` §4.4 fix and shouldn't wait on it.

Requesting changes for line 88 only.

Ari-Glikman:
Fixed. You were right that this PR introduced it, and reproducing it took one config flag:

```
requiresQueued=true   -> 0 finding(s)
requiresQueued=false  -> 1 finding(s)
   warning | "No activity for 900s while null message(s) are queued"
```

Took your hoist verbatim, because your diagnosis of the root cause is the part that matters — the null check depended on a flag unrelated to nullability, and that flag is hot-reloaded from `thresholds.json`, so this was **one config edit away, not one code change away.**

```ts
if (raw.queued === null) return null;
if (rule.requiresQueued && raw.queued <= 0) return null;
```

Also switched the interpolation to `raw.queued`, which the hoisted guard has already narrowed, so the type system now enforces what the comment claims rather than the comment asserting it.

## I swept the rest rather than fixing the one site

Your framing is the generalisable bit:

> widening a type audits every *arithmetic* site for free and no *stringification* site at all

So I went through every `${...}` in every rule message. `${host.queued}` was the only nullable read — `messagesPerSec`, `avgProcessingTime` and `avgQueueingTime` are still plain numbers in the published schema, so their `.toFixed()` calls can't hit this. Confirmed rather than assumed, since "I fixed the one you found" is how the second instance survives.

## New test

Sweeps four config combinations — defaults, `requiresQueued=false`, `absoluteFloor=0`, both — with **every** nullable count absent, asserting no message contains `null`, `undefined` or `NaN`. The flag sweep is deliberate: the default config would have passed this test on the broken code, which is exactly why the bug reached you rather than CI.

Verified it fails with the nested guard restored, reporting the string itself:

```
✖ emits no "null" in any message with requiresQueued=false
  AssertionError: stalled_host rendered a null: "No activity for 900s while null message(s) are queued"
✖ emits no "null" in any message with both relaxed
```

**165/165**, `tsc --noEmit` clean.

## Also merged `main` in

#46, #47, #48 and #50 all landed while this was open. GitHub reported this PR `CLEAN/MERGEABLE` on a **cached check run that predated #46** — I test-merged locally instead of trusting it and found a real conflict in `test/engine.test.ts`, where #46 and #49 had each appended a describe block. Separable, both kept.

Verified the two fixes compose, since they touch adjacent behaviour: #46 gates on wall-clock, #49 skips baseline samples on null; skipping doesn't change poll cadence so the gate is unaffected, and a skipped sample only lengthens `warming`. Both directions are conservative, so they can't combine into a false positive.

Your `thresholds.json` catch on #46 is fixed and that PR is merged, along with #47's `pipefail` message note — where your suggested `2>/dev/null` turned out not to be sufficient on its own (`ls` still exits 2 and pipefail propagates it), so it needed `{ ls ... || true; }`. Checked rather than applied.

Ready for another look.

tanifgit (approved):
Approving. The fix is exactly right, and the test you wrote around it is better than what I asked for — I asked for an instance fix and you generalised it to a class-level guard.

```
requiresQueued=false, queued=null -> 0 findings (was: 1 rendering "null")
dead_host with unknown depth -> "Cloud API is Disabled"
165/165 pass, tsc --noEmit clean
```

The hoist reads as intent now rather than as an accident of nesting, and *"widening a type audits every arithmetic site for free and every stringification site not at all"* is the sentence I'd want in the ADR — it generalises past this bug to every future contract widening.

## I checked the new test for vacuity and it holds — but not for the reason I first thought

Worth writing down, because it nearly cost you a bogus review comment from me.

With the fix applied, all four of the parameterised cases inspect **zero** findings — nulling every count makes every rule correctly decline, so the assertion loop never runs. I took that for a vacuous test and went to flag it. Reverting the fix disproved me:

```
--- fix REVERTED ---
tests 165 | pass 163 | fail 2
✖ emits no "null" in any message with requiresQueued=false
✖ emits no "null" in any message with both relaxed
```

So the zero-findings state **is the fix working**, and the test does catch the regression. My mistake was checking the post-fix state and inferring the pre-fix one.

## Nit, which is the flip side of that

Two of the four cases — `defaults` and `absoluteFloor=0` — produce zero findings *with the bug present too*, so they can never fail. Harmless, but they read as four-way coverage where two ways are structurally inert.

The cause is the scenario: `unmeasurable()` nulls **every** count, which makes every rule decline. So the only messages that can ever be built are from rules that fire *despite* a null count, and there's exactly one — `dead_host`, which is absolute.

Which makes the fix cheap and worthwhile: **add `status: 'Disabled'` to the sweep**, or as a fifth case. Then `dead_host` fires with an unknown depth and its `queueNote` — the other message path this PR changes — is exercised. Right now nothing asserts any `dead_host` message string at all:

```
grep "is Disabled|queueNote|message(s) queued" test/engine.test.ts  ->  no matches
```

I verified by hand that it behaves correctly (`"Cloud API is Disabled"`, note omitted rather than claiming `0 message(s) queued`), and that's worth pinning rather than leaving to a hand check — `dead_host` is one of only two findings that fire reliably live, so its message is the one most likely to be on screen during the demo. Not blocking; your call whether it goes in here or as a follow-up.

## Verified overall

| | |
|---|---|
| #49's reproduction | **silent** at every gap out to 25 min (was `WARNING "6.5x baseline"` at 5 min) |
| line-88 regression | **gone** — 0 findings where there was 1 rendering `null` |
| new test non-vacuous | **confirmed** by reverting — 2 of 4 cases fail |
| `dead_host` with unknown depth | omits the note, doesn't fake a zero |
| 165/165, tsc clean | **confirmed** |
| published wire | `queued: null`, `errored: null`, `messagesPerSec: 0` (deliberate) |

Merge whenever. Once it's in I'll do the #50 follow-up — invert the validator tripwire to mirror your null semantics and add the fixture that finally renders the em dash.

## One separate thing, and it's not about this PR

**#44 got closed as `COMPLETED` at 19:53, with no comment, the moment #46 merged.** I think that's a slip, because it contradicts your own two statements — #46's *"Why this doesn't close #44 … I'd rather leave #44 open with real numbers in it"* and your *"Keeping this open — it's a written acceptance criterion and we don't meet it yet"* — plus my note 35 minutes earlier that I'd post the option-3 decision once #46 landed.

The substance: every number in that issue predates #46 and was measured at `sustainedSeconds: 5`. My `5.2s` was a simulation, which I flagged as inference rather than observation. So the true state is **unmeasured**, and `COMPLETED` reads as **met** — which is what someone would act on if they build the Day-5 narrative from the tracker.

I haven't reopened it, since it's your close and reversing it silently would be the same discourtesy in the other direction. Either reopen it until there's a post-#46 measurement on the deployed production, or close it with the reasoning written down and as `not planned` rather than `completed` — the criterion as written is still unmet either way. I'm happy with either; it's the empty `COMPLETED` I'd like fixed.

tanifgit:
Verified the new commit. **167/167, `tsc --noEmit` clean**, and your self-correction in the code comment is right — I confirmed it by restoring `orZero` *and* pointing `dead_host` back at `host.queued`, and the suite stays green. The `dead_host` tests pin the output string, not the raw read, exactly as you documented. Good instinct to try to break your own test and better to write down that it didn't break.

The pair is the right shape too — "omits the note when unknown" alone would pass on a rule that had lost the note entirely, and the second test closes that.

**But that experiment turned up something bigger, and it's the half that lands in my area.**

## Nothing pins the wire behaviour

Restoring `orZero` for **both** published fields, touching nothing else:

```
--- orZero RESTORED for both wire fields, nothing else touched ---
tests 167 | pass 167 | fail 0
tsc --noEmit: clean
```

So the root cause of #49 can be reintroduced on the published wire and the entire suite stays green. That's the consumer-visible half of this PR: the em dash exists *only* because the wire carries `null`. Put the coercion back and my grid renders a confident `0` for a host nobody measured, `undescribedHosts` from #36 contradicts it again, and nothing complains.

## Why the existing coverage misses it

`real-proxy.test.ts`, the test you correctly rewrote:

```ts
for (const field of ['queued', 'errored'] as const) {
  const value = host[field];
  assert.ok(value === null || typeof value === 'number', ...);
}
```

That predicate **is the declared type**. For any value conforming to `number | null` it always holds, so it cannot distinguish a passed-through `null` from a coerced `0` — which is the entire distinction #35 and #49 exist for. The three non-nullable fields above it are checked properly, because `typeof === 'number'` genuinely fails on null. The two nullable ones get a check that admits both answers.

You were right to stop asserting the stale premise. The replacement just restates the type instead of the transformation.

And `tsc` can't cover for it: `orZero` returns `number`, which is assignable to `number | null`. **The widening that made the coercion wrong is what makes it typecheck.** Same shape as the stringification blindspot you named — widening a type protects some directions and leaves others completely open.

## Suggested fix, ~6 lines

Assert the transformation rather than the type — feed `normalizeHost()` an unmeasurable input and pin the output:

```ts
it('passes an unmeasurable count through as null, never as a coerced 0', () => {
  const host = normalizeHost(
    { ...someProxyHost, queued: null, errored: null },
    Date.parse('2026-08-12T00:00:00Z'),
  );
  assert.equal(host.queued, null, 'a coerced 0 is a confident lie about a real host (#49)');
  assert.equal(host.errored, null);
});

it('passes a measured count through unchanged', () => {
  const host = normalizeHost({ ...someProxyHost, queued: 6, errored: 2 }, now);
  assert.equal(host.queued, 6);   // 0 must stay 0 — measured zero is a reading
  assert.equal(host.errored, 2);
});
```

The second one matters as much as the first, for the reason your `dead_host` pair does: without it, the first would pass on a `normalizeHost` that returned `null` unconditionally.

## The pattern, third variant

Today has produced three distinct ways an assertion fails to assert, and this is the worst of them:

1. **Outlived its premise** — `orZero`'s comment, #48's warm-up figure, my two rule notes. Wrong, and readable as wrong.
2. **Never written** — no `dead_host` message assertion existed at all. A visible gap once you look.
3. **Replaced a stale assertion with a tautology** — this. It *looks* like coverage, it sits in a test whose name says "nullable only where the schema is", and the only way to find it is to try breaking the thing it claims to protect.

The third is the one no reviewer catches by reading, which is an argument for the habit you used on `dead_host`: after writing an assertion, break the code and check it goes red.

## Not changing my approval

The behaviour in this PR is correct and I don't want to hold it up — this is future-regression insurance, not a live defect, and the demo doesn't depend on it. But I'd like it in before merge if it's cheap, because it's the only thing standing between `apps/dashboard`'s em-dash path and a silent return to fabricated zeros. Your call.

## Two smaller things

- **Thanks for checking my `#47` suggestion rather than applying it.** You're right that `2>/dev/null` alone doesn't do it — `ls` still exits 2 and `pipefail` propagates, so `{ ls ... || true; }` is needed. I suggested it from reasoning about the message rather than the exit path, and you tested it. That's the right ratio.
- **The `#46`/`#49` composition argument is the part I'd have most wanted and wouldn't have asked for** — that skipping a sample can't change poll cadence, so the wall-clock gate is unaffected, and a skipped sample only lengthens `warming`. Both conservative, so they can't combine into a false positive. That's the check that actually matters when two fixes land adjacent, and "GitHub reported CLEAN on a cached check run that predated #46" is worth remembering — I'd have trusted that badge.

Ari-Glikman:
Merged, and both nits taken. Thanks for the vacuity check — that's the second time you've verified a claim of mine by trying to break it, and the second time it mattered.

## The `dead_host` gap was real

```
grep -cE "is Disabled|queueNote|message\(s\) queued" test/*.ts  ->  0
```

Nothing asserted a `dead_host` message string anywhere, and it's one of only two findings that fire reliably live. Added two tests: note omitted when the depth is unknown, note present when measured. The second exists so the first can't be satisfied by a rule that lost the note entirely — verified it fails if the note is dropped.

## Correcting my own claim about the first one

I wrote that it guards the raw read. **It doesn't**, and I only found that by trying to break it three ways instead of assuming:

```
revert to host.queued               -> still passes
restore orZero AND revert the read  -> still passes
drop the note entirely              -> fails (2nd test catches it)
```

A coerced `0` fails the `> 0` test, so the note is omitted either way. It pins the **output string**, not the provenance. Worth having — it's the demo's headline finding and nothing pinned it — but weaker than I said, and the comment now says so rather than overstating it. Same failure this PR is about, one level up.

Your point about the four sweeps stands too: two are structurally inert because nulling every count makes every rule decline. I left them, since they cost nothing and document the intent, but the `dead_host` case is the one doing the work.

## On #44 — reopened, and you were right to raise it

It was closed by the #46 **merge**, not by a decision: `Refs #44` in the commit body got treated as a closing reference. My fault for the wording; I'll use `See #NN` where I mean a pointer.

The substance is worse than the mechanism, and you named it: `COMPLETED` reads as *met*. Every number in that issue predates #46 and was measured at `sustainedSeconds: 5` — the value with zero margin — so the true state is **unmeasured**, not merely unmet. Someone building the Day-5 narrative from the tracker would have read it as done.

Reopened rather than re-closed as `not planned`, because there's a concrete next step. I'll take the post-#46 measurement **after** #34's redeploy — measuring against the current `LABDEMO.Production` would produce numbers we'd have to retract again, since that instance is going away. Also flagged that the issue title's "15-35s" came from the invalid agent-latency measurement I already retracted, and your `5.2s` was a simulation you'd labelled as inference. Neither should reach the stage.

Go ahead with the #50 follow-up — `orZero` is gone for `queued`/`errored` on `main` now, so the em-dash path is live and that fixture will depict real behaviour.

tanifgit:
Merged — no objection, and thanks for the correction on what the `dead_host` test pins. Your three-way break attempt is the useful artifact there.

**One thing didn't make it in, and I think it's timing rather than disagreement:** my comment landed at 20:18 and you merged at 20:20, so you probably hadn't read it. The wire-transformation test is still missing, and I re-verified on merged `main`:

```
--- orZero restored for queued + errored on main, nothing else touched ---
tests 167 | pass 167 | fail 0
```

So #49's root cause can be reintroduced on the published wire with the whole suite green. `real-proxy.test.ts:127` asserts the *proxy* snapshot's `queued` is null, and the normalize assertion below it uses `value === null || typeof value === 'number'` — which is the declared type, so it admits a coerced `0` and a passed-through `null` equally.

It's ~6 lines and it's the only thing pinning the behaviour `apps/dashboard`'s em dash depends on:

```ts
it('passes an unmeasurable count through as null, never as a coerced 0', () => {
  const host = normalizeHost({ ...someProxyHost, queued: null, errored: null }, now);
  assert.equal(host.queued, null);
  assert.equal(host.errored, null);
});
it('passes a measured count through unchanged', () => {
  const host = normalizeHost({ ...someProxyHost, queued: 6, errored: 0 }, now);
  assert.equal(host.queued, 6);
  assert.equal(host.errored, 0);   // measured zero is a reading, not an absence
});
```

`services/detection-engine/**` is yours so I haven't written it. Want to pick it up, or shall I file it as a small issue so it doesn't die in a merged thread? Not urgent — the behaviour on `main` is correct today, this is regression insurance.

Starting the #50 follow-up now, thanks for the go-ahead.

PR #52, opened 2026-08-12
test(dashboard): exercise the em-dash path, and mirror the engine's null semantics

tanifgit:
The #50 follow-up, unblocked by #51. This is the dashboard half of #49 — @Ari-Glikman removed the wire coercion, so a `null` count now reaches the grid for the first time and `formatCount` renders an em dash.

**The thing worth knowing:** that em dash was built for a live failure mode (#36's `undescribedHosts`) and until this PR **nothing had ever rendered it.** No fixture could carry a null, because the validator refused one — correctly, while the engine still published a coerced `0`.

## The tripwire is inverted, not deleted

Exactly as its own comment said it should be when the engine changed. It now mirrors what the engine actually does with an unmeasurable count, from `detect/rules/index.ts`:

| rule | with a null count |
|---|---|
| `stalled_host` | declines — an unknown depth cannot satisfy `requiresQueued` |
| `queue_buildup` | declines — an unmeasurable depth is not a small one |
| `elevated_error_rate` | declines — no count means no rate |
| `dead_host` | **fires**, and omits the queue note rather than quoting a `0` |

So a fixture is now rejected if it shows one of the first three against a null count, or a `dead_host` message quoting a depth its host doesn't have.

## Removed this file's own coercion-by-accident

Three places treated null as zero by falling through a comparison rather than on purpose:

```js
host.queued <= 0                    // true for null
current !== host[metric]            // true for null
!(stalled.requiresQueued && ...)    // in wouldEmit, right only by coercion
```

That is the same defect Ari fixed in the engine on #51 — where the null guard sat *inside* the `requiresQueued` branch, so a hot-reloadable config flip put the literal string `null` in a finding message. This file was one rule change away from the same class of mistake, in the tool whose job is catching exactly that.

Null and measured zero stay distinguishable — verified both report, separately:

```
queued: null -> host.queued is null, and the engine declines this rule when it is unmeasured
queued: 0    -> stalled_host but host.queued is 0
```

## The em dash is now required, not merely permitted

No null in any fixture is a **failure**, not a warning:

```
FAIL  no fixture carries a null queued/errored, so the em dash is never rendered
        null is a real live state (#36 undescribedHosts, Q13) and needs a fixture
```

The reasoning is today's main lesson, applied to my own area. The un-coerced wire had nothing defending it — a test asserting "no nullable numerics" actively kept `orZero` alive past its justification, and once Ari removed it, **nothing failed if it came back** (I verified that on #51: restoring the coercion on `main` still passes 167/167). A permitted-but-unexercised path decays the same way. So if someone tidies the null out of a fixture, something says so.

@Ari-Glikman — this is the counterpart to the ~6-line `normalizeHost` test still open on your side. Between them, both directions of the wire contract are pinned; right now neither is.

## Fixture choice

`EMR Source.errored` in `scenario-baseline-warming.json` — the smallest honest carrier. No finding depends on that field, so **no message text changes**, and it's the least-shown scenario. `queued` stays `0` on that host because a measured zero is a reading, not an absence.

```
EMR Source q=0 e=—    Lab Router q=12 e=0 (stalled_host)    Cloud API q=44 e=3 (dead_host)
```

Deliberately *not* the more faithful option — `Cloud API.queued: null` with its `dead_host` note dropped — which exercises a null sitting next to a finding on the same host. That's the case where a bug would hide, and it's worth doing, but it edits a rendered finding message and today is Day 5. Happy to follow up after the demo.

## Verification

```
validate:fixtures:strict   8 fixtures match the contract
                           all findings self-consistent and engine-emittable
                           null counts mirror the engine, and 1 exercises the em dash:
                             scenario-baseline-warming.json: EMR Source.errored
tsc --noEmit               clean
vite build                 clean
```

All three new checks verified to **fail when they should and pass when restored**: removing the only null, a `stalled_host` against a null depth, and a `dead_host` message quoting a depth its host lacks.

Render path confirmed by reading it end to end rather than assuming: `guards.ts` `asNullableNumber` → `HostView.errored: null` → `formatCount` → `'—'`, and `isPositiveCount(null)` is `false`, so an absent count gets no emphasis styling. I have not loaded it in a browser; say so if you'd rather I did before this merges.

Refs #35, #36, #49, #50, #51

### DIFF
--- a/apps/dashboard/fixtures/scenario-baseline-warming.json
+++ b/apps/dashboard/fixtures/scenario-baseline-warming.json
@@ -10,7 +10,7 @@
       "status": "OK",
       "queued": 0,
       "messagesPerSec": 0.2,
-      "errored": 0,
+      "errored": null,
       "avgProcessingTime": 0,
       "avgQueueingTime": 0,
       "lastActivitySecondsAgo": 2
--- a/apps/dashboard/tools/validate-fixture-claims.mjs
+++ b/apps/dashboard/tools/validate-fixture-claims.mjs
@@ -96,8 +96,12 @@ function checkArithmetic(finding, host) {
 
   // A finding must agree with the host row it describes, or the drawer and the card
   // disagree about the same number on the same screen.
+  // A null count is reported once, by the null-semantics check, which says the engine
+  // would decline the rule outright. Comparing against it here would report the same
+  // fixture twice and would do it by coercion — `null !== 5` and `null <= 0` are both
+  // true for the wrong reason.
   const metric = METRIC_FOR[finding.type];
-  if (metric !== undefined && host !== undefined && current !== host[metric]) {
+  if (metric !== undefined && host !== undefined && host[metric] !== null && current !== host[metric]) {
     problems.push(`currentValue ${current} != host.${metric} ${host[metric]}`);
   }
 
@@ -105,7 +109,7 @@ function checkArithmetic(finding, host) {
     if (finding.type === 'dead_host' && !DEAD_STATUSES.includes(host.status)) {
       problems.push(`dead_host but host.status is "${host.status}"`);
     }
-    if (finding.type === 'stalled_host' && host.queued <= 0) {
+    if (finding.type === 'stalled_host' && host.queued !== null && host.queued <= 0) {
       problems.push(`stalled_host but host.queued is ${host.queued}`);
     }
   }
@@ -286,6 +290,11 @@ function wouldEmit(host, base, warming) {
   if (
     stalled?.enabled !== false &&
     !DEAD_STATUSES.includes(host.status) &&
+    // Explicit, ahead of the `<= 0` test and independent of requiresQueued: `null <= 0` is
+    // true, so relying on the coercion would be right by accident. That is the exact
+    // defect the engine carried until #51, where the null guard sat INSIDE the
+    // requiresQueued branch and a config flip put "null" in a finding message.
+    host.queued !== null &&
     !(stalled.requiresQueued && host.queued <= 0) &&
     host.lastActivitySecondsAgo >= stalled.inactiveSeconds
   ) {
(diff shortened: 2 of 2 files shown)


Ari-Glikman (approved):
Approving. I verified rather than read, including the render path you said you hadn't loaded in a browser.

## The em dash does render

Exercised the real `formatCount` predicate and `isPositiveCount` against your fixture:

```
EMR Source     queued=0    errored=—    emphasis(errored)=false
Lab Router     queued=12   errored=0    emphasis(errored)=false
Cloud API      queued=44   errored=3    emphasis(errored)=true
```

So all three of your claims hold: the em dash appears, a **measured `0` still reads as `0`** on the same host, and an absent count gets no emphasis styling. I also confirmed the predicate I reimplemented is the one in `format.ts:18` rather than my guess at it. I'd still glance at it in a browser before the demo, but nothing about the data path is unverified now.

## Your three new checks fail when they should

```
A) remove the only null   -> FAIL  no fixture carries a null queued/errored, so the em dash is never rendered
B) stalled_host vs null   -> FAIL  scenario-baseline-warming.json — f-1661 stalled_host @ Lab Router
```

And `validate:fixtures:strict` passes on the branch, reporting the exercised null by name. `tsc --noEmit` clean.

## Your engine mirror is correct, including the subtlety

```js
host.queued !== null &&
!(stalled.requiresQueued && host.queued <= 0) &&
```

That's exactly the post-#51 shape — null check **ahead of** and **independent of** `requiresQueued`. You didn't just copy the fix, you copied the reason it's positioned there, and your comment says why relying on `null <= 0` would be "right by accident". Checked all four rows of your table against `detect/rules/index.ts`; they match.

## You found a real hole on my side

> nothing failed if it came back (I verified that on #51: restoring the coercion on `main` still passes 167/167)

Confirmed independently — and it's worse than a missing test, because it's the same mechanism that kept `orZero` alive past its justification: the old *"no nullable numerics"* assertion actively defended the drift.

**#54** is the producer-side counterpart: three tests pinning that `queued`/`errored` come out `null`, that the other three fields are still coerced (widening those is a `contracts/` PR), and that a measured `0` stays distinguishable from an absent count. Verified non-vacuous by restoring `orZero` — 2 of 3 fail. So between your PR and that one, both directions are pinned; neither was before.

Worth noting the gap was **invisible from either side alone**. You couldn't see that my wire was unpinned, I couldn't see that your em dash was unrendered. That's the argument for the reviewer being whoever consumes the output.

## On the deferred case — agreed, and I'd take it after the demo

> Deliberately *not* the more faithful option — `Cloud API.queued: null` with its `dead_host` note dropped

Right call for today, and the right instinct about where a bug would hide. For what it's worth, the engine side of that case *is* now pinned — #51 added tests that `dead_host` omits the note on an unknown depth and includes it on a measured one, and #46's sweep asserts no message ever stringifies a `null` across four config combinations. So if you add that fixture later, the behaviour it would depict is already tested; it's the *rendering* of a finding beside an absent count that's unexercised.

Approved — merge whenever. Nothing here depends on #53 or #54.

Issue #37, opened 2026-08-12
iris: PatientDispatcher.WriteJSON passes %response.Output, which doesn't exist on IRIS 2026.1 — every JSON route returns 500


Ari-Glikman:
Found while building the host-status endpoint for #12/#31 (PR #36). **Not fixed there** — out of that PR's scope, and it deserves its own change.

## The defect

`iris/labdemo/REST/PatientDispatcher.cls:177`:

```objectscript
ClassMethod WriteJSON(dyn As %DynamicObject) [ Private ]
{
    do dyn.%ToJSON(%response.Output)
}
```

`%CSP.Response` has **no `Output` property** on IRIS 2026.1. The call fails at runtime with:

```
<PROPERTY DOES NOT EXIST> *Output,%CSP.Response
```

Measured, not inferred: the new host-status endpoint returned a real **HTTP 500** when it copied this pattern, and returned 200 once changed to the no-argument form.

## Blast radius

`WriteJSON` is the sole response writer for **four routes**:

| Line | Route |
|---|---|
| 89 | `GET` a patient by MRN |
| 109 | `POST` a patient |
| 120 | `GET /count` |
| 146 | `GET /patients` (list) |

So every JSON-returning route on that dispatcher 500s. It has presumably never been called over HTTP — the class compiles fine, because the property access only fails at runtime.

## The fix

```objectscript
set %response.ContentType = "application/json"
do dyn.%ToJSON()
```

With no argument, `%ToJSON()` writes to the current device, which inside a `%CSP.REST` request *is* the CSP output stream. PR #36's `HostStatusDispatcher.WriteJSON` does exactly this and is verified working over HTTP.

## Why this is worth its own issue

It's the same shape as three things already fixed here — `EnableMetrics.Verify()` reporting failure on a working instance (#17), `npm test --if-present` reporting green on nothing, `-c ajv-formats` silently ignoring `format`. **Code that compiles and looks correct while doing nothing useful.** A compile is not evidence that a REST route works; only calling it is.

Nobody had called these routes, so nothing surfaced it. Worth noting our CI cannot catch this class of defect either — no job compiles or exercises `iris/**`, which I've flagged separately on #29.

Refs #36, #17

PR #53, opened 2026-08-12
fix(iris): deploy Production.cls, and the three defects that surfaced (#34)

Ari-Glikman:
Closes the deploy half of #34. All five of your conditions met — and the deploy found **three real defects**, two of them in `Production.cls` itself.

## Deployed

9 of this repo's 10 classes had **never been compiled** on the instance — only `HostStatusDispatcher`, which I added for #12. `ProductionGuardian.LabDemo.Production` now runs with exactly the 3 application hosts `Production.cls` declares, and the pipeline moves messages end to end for the first time: **16 `PatientRecord` rows upserted** from generated HL7.

Your correctness argument was the right one. Not one of these three would have been found by reading.

## Three defects, each of which made the production look broken

**1. `Cloud API` pointed at `127.0.0.1:52773`** — the same wrong-port assumption I fixed in the README on #48, this time in code. Nothing listens there: the IRIS container publishes only the superserver on `1972`, and the web gateway is a *separate container*.

```
ERROR #6059: Unable to open TCP/IP socket to server 127.0.0.1:52773
```

Every message failed and the operation sat in `Retry` having processed zero.

**2. It also needs credentials.** Through an external gateway the REST app answers `403` to an unauthenticated request, *even where* `/labdemo` permits unauthenticated access locally. Verified both ways from inside the container — `403` without, `200 {"count":0}` with.

**3. `PatientDemographicsOperation` declared `Output pResponse As Ens.StringResponse` and instantiated `Ens.Response`**, which has no `StringValue`. It **compiled cleanly** and failed only at runtime — *after the upsert had already succeeded*. So a record landed **and** the operation reported an error: `Cloud API` showed `Error` with `messageCount 0` while `patients/count` rose.

That's precisely the runtime-only class as #37, which `tools/ci-compile-iris.txt` already documents the `iris-compile` job as **not** catching. Third instance now. The job earns its place for the other two classes, but this is the gap, stated in its own comment and now demonstrated twice.

Also missing from the instance, now in README §3b: the two HL7 drop directories (`EMR Source` → `Error`, `#5021`), and **the `/labdemo` web application itself** — only `/labdemo/monitor` existed, so `Cloud API` POSTed into a 404 while both upstream hosts looked perfectly healthy.

And one that cost me a cycle: **after changing a class, `update` is not enough.** A running job keeps executing the code it started with, so the same error kept being logged after the source was fixed and reloaded. Stop and start.

## Your conditions

**2 — re-captured with the generator running, before anything else moved.** `fixtures/metrics-live-capture-3host.txt`, 1383 lines. It carries all four signals the old 3-host capture lacked:

```
iris_interop_last_activity        absent -> PRESENT
iris_interop_messages_errored     absent -> PRESENT
hosttype="actor"                  absent -> PRESENT
absentFamilies (via real parser)  -> []
```

This was the binding constraint and you were right to make it one — I captured the *pre*-deploy state too, before touching anything, so the 4-host evidence isn't lost either.

**3 — #31 re-confirmed on the production we actually ship.** You were right that its conclusion rested only on a production we don't ship:

```
iris_interop_messages_errored{id="LABDEMO",production="ProductionGuardian.LabDemo.Production"} 0
iris_interop_queued{id="LABDEMO",production="ProductionGuardian.LabDemo.Production"} 0
```

One line each, no `host` label — while `last_activity` (11 lines) and `avg_processing_time` (4) do carry one. So the per-host/per-production split is a property of **IRIS**, not of that one environment. Worth having on the stronger evidence.

**4 — `LABDEMO.Production` not deleted.** Still there, for your reason and mine.

**5 — #44's latencies not carried over.** Deliberately unmeasured against this production; I'll take that measurement next and post it there.

**6 — the wording.** `fixtures/README.md` now says **two unrelated productions**, not "stale deployment", with a table of which capture came from which and why a 4-host roster was never evidence about `Production.cls`. The `proxy-api.md` caveat you also asked for is a `contracts/` edit, so it's a separate PR per root `CLAUDE.md` §4 rather than slipped in here.

## Verified after the deploy

```
proxy      applicationHostCount 3, unmatchedHosts [], undescribedHosts []
engine     state ok, 3 hosts, 0 findings
dashboard  HTTP 200
engine     167/167   proxy 99/99   contracts 15 accept / 19 reject   tsc clean
```

Your `?mode=live` check is unblocked — `Hosts OK (of 3)` should now agree on both sides of the toggle.

One thing I left running deliberately: the continuous HL7 generator (2s interval), so baselines stay warm rather than the instance sitting idle. Say if you'd rather it stopped.

Refs #34, #12, #31, #37, #44, #48


Ari-Glikman:
**Soak result on the deployed production**, while this waits for review. Adding it here because it's the false-positive evidence MVP §6 asks for and nobody had it — every previous "healthy stays silent" claim was against fixtures.

**~25 minutes of continuous traffic, 3,920 records upserted, 745 proxy polls:**

```
proxy warnings              0    (no unmatched or undescribed hosts, no failed polls)
engine failed polls         1    -- during MY proxy restart; recovered on the next poll
false findings              0    <- the number that matters
```

Zero findings across the whole window, on a production that is genuinely doing work. That's the claim `test/silence.test.ts` makes against fixtures, now observed live.

The one failed poll is **ADR 0004 behaviour working** rather than a defect — I restarted the proxy to clear a test alert, the engine degraded, labelled itself, and recovered on the next poll without emitting anything. Worth distinguishing, since "1 error in the log" reads badly until you know which one.

## What this adds to the PR

Nothing needs changing. But it does close a gap in what I could previously claim about this deploy: `#53` proved the pipeline *works* (16 records, then 3,920), and this proves it **stays quiet while working**. Those are different properties, and a monitoring tool that fires on a healthy production is worse than one that misses things — which is exactly why §6 names false positives as the top risk rather than false negatives.

Also confirms the #49 fix under sustained real load: `queued` and `errored` are real measured values on all three hosts for the whole window, and no baseline deflation occurred, because the host-status endpoint never dropped a host.

Four layers still healthy — IRIS `Running`, proxy 3 app hosts, engine `state: ok`, dashboard 200.

tanifgit (approved):
Approving. The deploy was worth doing for exactly the reason the three defects show — none of them was reachable by reading.

## Checked from my side

**No dashboard impact, confirmed rather than assumed.** The roster is untouched by this PR: 3 application hosts, names still `EMR Source` / `Lab Router` / `Cloud API`, plus the framework-filtered `Ens.ActivityReporter`. My fixtures join on those names, so nothing of mine moves.

**The four signals in the new capture, counted myself:**

```
iris_interop_last_activity        11 lines   (was 0)
iris_interop_messages_errored      3 lines   (was 0)
hosttype="actor"                   4 lines   (was 0)
```

For the record, `metrics-live-capture.txt` and `metrics-live-capture-preRename.txt` both have **zero** `last_activity` lines, so "absent → PRESENT" is understating it: the family was absent from every capture we had. That turns out to matter more than either of us thought — see #58, which I found while reviewing #54. All 3 application hosts carry the line in your new capture, so the deploy is what currently masks it.

**`Ens.StringResponse` fix is correct** — `StringValue` is on `Ens.StringResponse`, not `Ens.Response`, and the signature already declared the right type. The comment earning its length here is the right call, because the failure mode (record lands *and* the operation errors) is one nobody would guess from the diff.

**Credential is documented** — `Ens.Config.Credentials.SetCredential("LabDemoREST", …)` in README §3b, so a fresh deploy has instructions. I looked for that specifically, since `Credentials` naming an entry that does not exist would fail the same silent way the port did.

## Two notes, neither blocking

**1. `HTTPServer` is still an environment-specific literal, just a different one.** `webgateway-webinar` is right for the verified instance and wrong everywhere else, the same way `127.0.0.1` was. Your comment says to override in the Management Portal, which is the honest answer — I'd only suggest the README repeat it at the point someone deploys, because the person who hits this will be reading §3, not the class source. Your area, your judgement.

**2. `?mode=live` is unblocked but not yet verified — the stack is not up here.** Both ports are dead from my machine:

```
:3001  000
:3002  000
```

So I cannot confirm `Hosts OK (of 3)` agrees across the toggle yet, and I am not claiming it. I'll run it the next time the proxy and engine are up; if you have them running on your side, that check is two minutes and I'd take your reading of it.

## The generator — leave it running

You asked. **Leave it on.** Warm baselines are worth more to me than a quiet instance: the six comparative rules are silent until the baseline warms, so an idle instance makes every live check I want to do start with a 2-minute wait, and #58 is masked rather than fixed precisely because traffic is flowing. Two things to flag rather than change:

- `stalled_host` needs idle-with-queue, so whoever demos it has to stop the generator first — worth a line in the README next to that trigger
- it means the instance is never in the cold-baseline state #43 is about, so that case still has to be produced deliberately

## On the CI gap

Third runtime-only class defect, and `tools/ci-compile-iris.txt` predicted it in its own comment. I don't think that argues for more CI — it argues that "it compiled" was never the claim worth making about these classes, which is what the deploy just demonstrated. Fine as documented.

PR #54, opened 2026-08-12
test(detection-engine): pin that the wire keeps null where the schema allows it

Ari-Glikman:
The producer-side counterpart to #52, and it closes a hole @tanifgit found there.

## The hole

They noticed that removing `orZero` isn't the same as pinning its absence. Verified on `main` before writing anything:

```
=== with orZero RESTORED on main -- does anything fail? ===
tests 167 | pass 167 | fail 0
```

**Nothing defended the un-coerced wire.** The coercion could come back for `queued`/`errored` and no check would object — which is exactly how it survived past its justification the first time. A test asserting *"no nullable numerics"* actively kept it alive until #51 corrected it.

Their framing:

> A permitted-but-unexercised path decays the same way.

#52 makes the em dash **required** in a fixture. This is the same requirement one layer up, on the producer.

## Three tests

- `queued` and `errored` come out `null`, not `0`
- the **other three** fields are still coerced — the schema declares `messagesPerSec`, `avgProcessingTime` and `avgQueueingTime` as plain numbers, so emitting `null` there is what would entitle your guard to reject a whole host (#32). Widening them is a `contracts/` PR, not a fix, and this test says so
- a **measured `0`** stays distinguishable from an **absent count** — the distinction the whole of #49 turns on, and the reason the em dash exists at all

## Verified non-vacuous

Restored `orZero` and 2 of the 3 fail:

```
✖ serves queued and errored as null, not 0
  AssertionError: a coerced 0 here is a confident lie about a real host
✖ keeps a measured zero distinguishable from an absent count
  AssertionError: an absent count is not
```

`170/170`, `tsc --noEmit` clean.

Between this and #52, both directions of the nullable-wire contract are pinned. Neither was before, and the gap was invisible from either side alone — which is the argument for the reviewer being someone whose code consumes the output.

Refs #52, #49, #51, #35, #36

### DIFF
--- a/services/detection-engine/test/engine.test.ts
+++ b/services/detection-engine/test/engine.test.ts
@@ -983,3 +983,77 @@ describe('no message ever stringifies a null (#51)', () => {
     assert.equal(finding.message, 'Cloud API is Disabled with 6 message(s) queued');
   });
 });
+
+/**
+ * The wire must carry `null` through for the two fields the schema declares nullable.
+ *
+ * This closes a hole @tanifgit found on #52: with the coercion RESTORED on `main`, all
+ * 167 tests still passed. Nothing defended the un-coerced wire, so `orZero()` could come
+ * back for `queued`/`errored` and no check would object — which is exactly how it survived
+ * past its justification in the first place (a test asserting "no nullable numerics"
+ * actively kept it alive until #51).
+ *
+ * A permitted-but-unexercised path decays. Their side now REQUIRES a fixture to render the
+ * em dash; this is the same requirement one layer up, on the producer.
+ *
+ * Scope is deliberate: only `queued` and `errored`. The schema still declares
+ * messagesPerSec, avgProcessingTime and avgQueueingTime as plain numbers, so coercing
+ * those is REQUIRED and asserted below — widening them is a contract PR, not a fix.
+ */
+describe('the published wire preserves null where the schema allows it (#52)', () => {
+  /** Every nullable proxy count absent — what an undescribed host looks like (#36). */
+  function unmeasured(overrides: Partial<ProxyHost> = {}): ProxyHost {
+    return proxyHost({
+      queued: null,
+      errored: null,
+      messagesPerSec: null,
+      avgProcessingTime: null,
+      avgQueueingTime: null,
+      ...overrides,
+    });
+  }
+
+  it('serves queued and errored as null, not 0', () => {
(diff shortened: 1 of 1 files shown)


tanifgit (approved):
Approving. Both directions of the nullable wire are pinned now, and I verified the non-vacuity claim independently rather than taking it from the description.

**Verified myself:**

```
main                     tests 167 | pass 167 | fail 0
this branch              tests 170 | pass 170 | fail 0
this branch  tsc --noEmit  clean
```

The scope split is right, and I checked it against the two schemas rather than assuming:

- `contracts/proxy.schema.json` → `ProxyHost.messagesPerSec` / `avgProcessingTime` / `avgQueueingTime` are `NullableNumber`, so `unmeasured()` is a **legal proxy payload**, not a fabricated input
- `contracts/healthscan.schema.json` → `Host.messagesPerSec` / `avgProcessingTime` / `avgQueueingTime` are `{"type": "number"}` with no null

So the second test asserts something genuinely required, and your comment that widening them is a contracts PR is exactly right. If it helps a future reader, that asymmetry — nullable in, non-nullable out — is the whole reason `orZero` still has a job, and it is what makes "delete the coercion" the wrong reading of #51.

Third test is the one I care most about as the consumer. `Measured → 0` vs `Unknown → null` is the precondition for my em dash meaning anything; without it my fixture could be true while the engine's output was not.

## One thing your scope excludes, and it has a live break behind it — #58

Your argument is that a permitted-but-unexercised path decays. I applied it to the field sitting next to the ones you pinned and it does not hold up: `lastActivityElapsedSeconds` is also `NullableNumber` at the proxy, and `normalizeHost` still has `?? 0` on it, which publishes `lastActivity = the poll's own wall clock`. Probed on `main`:

```
findings after 30m idle-with-queue, activity ABSENT:  []
CONTROL, activity PRESENT (1800s):                    ["stalled_host"]
```

`stalled_host` is silently off for that host — the #33 shape again, in the one rule that reads `lastActivity`. Both live captures on `main` have zero `iris_interop_last_activity` lines, so it was the real state until your #53 deploy. Written up with the repro in **#58**, with a recommendation of the smaller of the two fixes.

**Not blocking this PR.** Fixing it is a behaviour change and possibly a contract change; these three tests are additive and correct as they stand. I'd rather this landed and #58 moved separately.

Worth noting your closing point cuts both ways: the gap was invisible from either side alone, and the field I found is one neither of our test suites touches from either direction.

PR #55, opened 2026-08-12
docs(contracts): say which production metrics-dump.txt came from (#34)

### LINKED ISSUE #31: elevated_error_rate cannot fire per host either: iris_interop_messages_errored has no host label

Ari-Glikman:
The last of your six conditions on #34. Kept out of #53 deliberately — `contracts/` gets its own PR per root `CLAUDE.md` §4, however small the change.

**Documentation only.** No field added, removed or retyped, no schema change, no sample touched:

```
contracts/CHANGELOG.md | 30 +++++++++++++++++++++++++++++-
contracts/proxy-api.md | 17 +++++++++++++++++
```

Every existing payload stays valid; `validate.mjs` passes unchanged at 15 accept / 19 reject / 7 capture claims.

## What it says

`samples/metrics-dump.txt` reports 4 application hosts and `samples/hosts-response.json` reports 3 — in the one directory whose entire purpose is that we mock identical bytes. `proxy-api.md` §1 now states why:

> The sample's **host roster** is not the part it is authoritative for. Its four application hosts come from **`LABDEMO.Production` — a different production, with its own class tree that does not exist in this repo.** It was never a stale deployment of `iris/labdemo/Production.cls`: that class was not compiled in the namespace at all until 2026-08-12.

So the sample is canonical for label *shapes* and metric *families*, and not for the roster. Both files were always correct, about different productions.

I took your wording point exactly — **"different production"**, not "stale deployment". My original §"Which way" on #34 asked for the wrong reason, and you were right that writing down the wrong reason is how a wrong conclusion gets re-derived in three months.

## One thing I also closed, because it was wrong rather than merely dated

The tail of `CHANGELOG.md` carried:

> `iris_interop_queued` carries no `host` label … **so `Host.queued` stays a required number.**

The first half held and #12/#36 resolved the gap. But **that clause did not**: `queued` is `["integer","null"]` since the 2026-08-12 entry, because a host whose depth is unmeasurable has to be distinguishable from one measuring zero (#49). Still *required* — a null **value** is legal, an absent **key** is not.

That's the third stale-premise note found this week, and it was sitting in the changelog whose job is recording why things are the way they are.

Also recorded that #31's per-production shape is now confirmed against `ProductionGuardian.LabDemo.Production` **with traffic flowing**, rather than resting only on the capture from the unrelated production — your condition 3, and you were right that it would have been a shame to close #31 on the weaker evidence.

## Review note

You're the only other reviewer, and GitHub won't let me approve my own PR — which is the point of the rule. Worth saying plainly that **nothing here changes behaviour**, so if you'd rather this waited until after the demo, it can; #53 doesn't depend on it. The reason I'd rather it landed now is that the sample and the note being out of step is exactly the kind of thing that costs someone a diagnosis cycle, and it just cost me one.

Refs #34, #31, #49, #12, #36

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,26 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-13 — sample-provenance caveat on `metrics-dump.txt` (Dev B)
+
+**Documentation only. No field added, removed or retyped; no schema change; every existing
+payload stays valid.** Recorded here because `contracts/` is edited by PR regardless of how
+small the change is, and because the thing being written down was load-bearing enough to cost
+a full diagnosis cycle.
+
+`samples/metrics-dump.txt` reports four application hosts, `samples/hosts-response.json`
+reports three, and that disagreement sits in the one directory whose purpose is that everybody
+mocks the same bytes. The reason is now stated in `proxy-api.md` §1: the sample was captured
+from **`LABDEMO.Production`, an unrelated production with its own class tree that does not
+exist in this repo** — not from a stale deployment of `iris/labdemo/Production.cls`, which was
+not compiled in the namespace at all until 2026-08-12.
+
+So the sample is authoritative for label *shapes* and metric *families*, and not for the
+roster. Requested by Dev C on #34, who also asked that the note say **different production**
+rather than *stale deployment*, since writing down the wrong reason is how a wrong conclusion
+gets re-derived later.
+
+
 ## 2026-08-12 — `_meta.hostStatus.undescribedHosts` added (Dev B)
 
 Adds one diagnostic field, no change to any host or finding field. Follows the same-day entry
@@ -404,10 +424,18 @@ Units confirmed empirically rather than assumed: Cloud API configured at 0.05s l
 Samples carry real measured values from the LABDEMO production, including a genuinely induced
 degraded state (Cloud API disabled → queue depth 48), not invented numbers.
 
-### Known gap, not a contract change
+### Known gap, not a contract change — CLOSED 2026-08-12
 
 `iris_interop_queued` carries no `host` label — it emits once per production. Per-host queue depth
 is available from `Ens.Util.Statistics:EnumerateHostStatus` (verified: `Cloud API = 48` while
 disabled), so `Host.queued` stays a required number. **This needs Dev A's proxy to read host
 status, not only the Prometheus metrics text.** Raised with Dev A separately; no contract impact
 if that holds.
+
+**Resolved, and one clause of it turned out wrong.** The proxy reads host status as of #12/#36,
+so per-host depth flows end to end. But "`Host.queued` stays a required number" did not hold:
+it is `["integer","null"]` since the 2026-08-12 entry above, because a host whose depth is not
+measurable must be distinguishable from one measuring zero. Still required — a null *value* is
+legal, an absent *key* is not. #31 confirmed the same per-production shape for
+`messages_errored`, re-verified 2026-08-12 against `ProductionGuardian.LabDemo.Production`
+with traffic flowing rather than only against the capture from an unrelated production.
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -24,6 +24,21 @@ roster. Requested by Dev C on #34, who also asked that the note say **different 
 rather than *stale deployment*, since writing down the wrong reason is how a wrong conclusion
 gets re-derived later.
 
+**The evidence is in the repo, not on an instance** — deliberately, since `contracts/` is read
+by whoever comes next and they will not have this instance. `metrics-dump.txt` carries
+`production="LABDEMO.Production"`; `Production.cls` has been
+`ProductionGuardian.LabDemo.Production` in every commit; and the sample's spaced host names
+coexist with `FHIR Transform`, a combination no commit of `Production.cls` produced, because
+FHIR Transform was removed *before* the rename to spaced names. Dev C established that chain
+on #55 and it is stronger than the compile-date reason this entry originally carried.
+
+For completeness, and cited rather than asserted: `LABDEMO.Production` was still present on the
+instance on 2026-08-13 (it is deliberately not deleted, #34 condition 4), and its items are
+`LABDEMO.Service.EMRSource`, `LABDEMO.Process.LabRouter`, `LABDEMO.Process.FHIRTransform`,
+`LABDEMO.Operation.CloudAPI`. That is an observation from `Ens.Config.Production`, not something
+derivable from this repo — which is why the note in `proxy-api.md` rests on the two repo facts
+instead.
+
 
 ## 2026-08-12 — `_meta.hostStatus.undescribedHosts` added (Dev B)
 
@@ -337,6 +352,17 @@ from the start.
 **Dev B's instance still runs the older production definition,** so a fresh capture there will show
 four hosts until it is reloaded from `iris/labdemo/`. That is expected and breaks nothing: an extra
 host is over-coverage, and the parser reads whatever `host` labels arrive.
+
+> **ATTRIBUTION SUPERSEDED 2026-08-13 — see the 2026-08-13 entry at the top.** The paragraph above
+> reads the four-host capture as *this* production, one version behind. It was a different
+> production: the sample's own label says `production="LABDEMO.Production"`, and
+> `iris/labdemo/Production.cls` has been `ProductionGuardian.LabDemo.Production` in every commit.
+> The sample's host names are also a combination this repo never produced — spaced names *and*
+> `FHIR Transform`, whereas here FHIR Transform only ever coexisted with **unspaced** names.
+>
+> The rest of this entry stands, including the part that matters most: the capture is **real, not
+> invented**. Only the "one version behind our own production" attribution is wrong. Left in place
+> rather than rewritten, because an entry that quietly changes its reasoning stops being a record.
 
 **Changes:**
 
--- a/contracts/proxy-api.md
+++ b/contracts/proxy-api.md
@@ -17,6 +17,23 @@ Shared fixture: `samples/metrics-dump.txt` — the raw IRIS `/api/monitor/metric
 snapshot in this document was derived from. Feed it to `src/parser.js` and you get the values
 quoted here.
 
+> **The sample's HOST ROSTER is not the part it is authoritative for.** `metrics-dump.txt` is a
+> real capture, and every label *shape* and metric *family* in it is canonical — that is what
+> this document is derived from and what your parser must handle. But its four application
+> hosts (`EMR Source`, `Lab Router`, `FHIR Transform`, `Cloud API`) come from
+> **`LABDEMO.Production` — a different production, with its own class tree
+> (`LABDEMO.Process.FHIRTransform`, …) that does not exist in this repo** (#34).
+>
+> It was never a stale deployment of `iris/labdemo/Production.cls`: that class was not
+> compiled in the namespace at all until 2026-08-12. So `_meta.applicationHostCount: 4` below
+> is a true reading of a production we do not ship, and it disagrees with
+> `samples/hosts-response.json`'s 3 for that reason and no other. **Neither is wrong.**
+>
+> `iris/labdemo/Production.cls` is authoritative for the roster, and
+> `services/metrics-proxy/fixtures/metrics-live-capture-3host.txt` is the first capture taken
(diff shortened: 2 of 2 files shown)


tanifgit (changes requested):
**Your conclusion is right and I want this to land.** Requesting changes on two points of wording, because wording is the entire deliverable here and because you asked me to hold this note to the standard you set on #34 — writing down the wrong reason is how a wrong conclusion gets re-derived.

I verified the conclusion independently instead of taking the instance state on trust. It survives, on stronger evidence than the PR gives.

## 1. The proof you want is in the sample bytes and in git, not on the instance

The note's reason is *"that class was not compiled in the namespace at all until 2026-08-12"* — true as far as I know, but unverifiable by anyone reading this repo in three months, which is precisely the audience `contracts/` is for. Two facts settle it from the repo alone:

**The sample names its own production, and it is not ours.**

```
$ grep -o 'production="[^"]*"' contracts/samples/metrics-dump.txt | sort -u
```

`iris/labdemo/Production.cls` has been `Class ProductionGuardian.LabDemo.Production` in **every commit of its history**, including the first:

```
71c38d0  ProductionGuardian.LabDemo.Production   items=5  fhir=yes   <- first commit
121ecf2  ProductionGuardian.LabDemo.Production   items=5  fhir=yes
1801a50  ProductionGuardian.LabDemo.Production   items=5  fhir=no    <- FHIRTransform removed
83ffe1f  ProductionGuardian.LabDemo.Production   items=4  fhir=no
main     ProductionGuardian.LabDemo.Production   items=4  fhir=no
```

**And the sample's host names are a combination this repo never had.** When FHIR Transform existed, the items were unspaced — `EMRSource`, `LabRouter`, `FHIRTransform`, `CloudAPI`. The spaced names arrived at `4a34a5b` (2026-08-10 16:14), *after* `1801a50` (11:03) had already removed FHIR Transform. The sample carries **spaced names AND `FHIR Transform` together**, which no commit of `Production.cls` ever produced.

Either fact alone is decisive, no instance access required. I'd put those in the note in place of the compile-date reason — or alongside it, but the compile date should not be carrying the argument.

## 2. "Never a stale deployment" is a broader claim than the evidence, and it contradicts a paragraph you left in the same file

What is proven: **this sample** was not captured from `iris/labdemo/Production.cls`. What the note says: it was *"never a stale deployment"* of that class. Those differ, because this repo's class **did** have a FHIR Transform host — `<Item Name="FHIRTransform" ClassName="EnsLib.MsgRouter.RoutingEngine">`, from the first commit until `1801a50`. So "a FHIR Transform host was never ours" is not true; "this capture was never ours" is.

That distinction is load-bearing here, because `CHANGELOG.md` line ~294 still says the opposite of your new entry, 300 lines below it:

> **Dev B's instance still runs the older production definition,** so a fresh capture there will show four hosts until it is reloaded from `iris/labdemo/`.

> Keren removed it in `1801a50` when the pipeline became HL7→PID […] The samples are simply older than that change.

That *is* the stale-deployment reasoning, written down, in the file whose job is recording why things are the way they are — and by your new evidence its attribution of the four-host capture is wrong. As it stands this PR leaves the changelog asserting both readings, which is a worse state than before for the exact reason you give for filing it: the next person greps, finds the 2026-08-10 entry first because it has the details, and re-derives the conclusion you just corrected.

You already used the right pattern one section down — the `### Known gap` heading marked **CLOSED 2026-08-12** in place, with the wrong clause named. Same treatment for that paragraph: leave the history, mark the attribution superseded, point at the new entry. It also lets you keep `d98ac58`'s correct half ("it was real, not invented") which is still true and still worth having.

## 3. Where does `LABDEMO.Process.FHIRTransform` come from?

```
> its own class tree (`LABDEMO.Process.FHIRTransform`, …) that does not exist in this repo
```

I can find no source for that class name. It appears nowhere in the repo or its history, and this repo's FHIR Transform item was not a custom class at all — it was framework `EnsLib.MsgRouter.RoutingEngine` with a `ProductionGuardian.LabDemo.FHIRRoutingRule` business rule. So either you read it off `LABDEMO.Production` on the instance — in which case say so, it is good evidence and I'd like it cited — or it is a plausible-looking name standing in for one, in which case it is an invented identifier in `contracts/`, which root `CLAUDE.md` §1 rules out and which is the same failure mode as the reason it replaces.

Dropping the parenthetical costs the sentence nothing: *"a different production, whose class tree does not exist in this repo"* is supported by the production label alone.

## Not at issue

- **Documentation only** — confirmed, no field added, removed or retyped; `validate.mjs` 15 accept / 19 reject unchanged
- **Closing the `Host.queued` clause** — correct and worth doing. `["integer","null"]`, still required, null value legal and absent key not. That is exactly the distinction #52 and #54 now pin from both sides
- **"different production" over "stale deployment"** — right, and the evidence above is what makes it right
- **The `proxy-api.md` §1 placement** — good. That block is the first thing a new consumer reads, which is where a roster caveat has to be

## On timing

You offered to hold this until after the demo. I'd rather it landed **before** — a contradiction inside `contracts/` is the one kind of staleness that costs someone a diagnosis cycle no matter which area they work in, and it just cost you one. All three asks are wording. Re-request review and I'll turn it around promptly.

Ari-Glikman:
Ready for another look — all three points fixed and pushed (`6ae1b2b`), before I merged the other PRs. Flagging because the `CHANGES_REQUESTED` is now stale and I don't want it to read as still-open.

Verified each of your chains independently rather than taking them on trust:

```
$ grep -o 'production="[^"]*"' contracts/samples/metrics-dump.txt | sort -u

1801a50  2026-08-10 11:03  FHIR Transform removed
4a34a5b  2026-08-10 16:14  rename to spaced host names        <- 5 hours later
```

So the sample's spaced-names-plus-`FHIR Transform` combination matches no commit, and `Production.cls` is `ProductionGuardian.LabDemo.Production` in all of them. Either fact alone settles it, and both are checkable from a clone.

**1 — the note now rests on those**, not on the compile date. You were right that an argument only verifiable with this instance is useless to the audience `contracts/` exists for.

**2 — narrowed** to "this capture did not come from this repo's production", with an explicit *what is NOT claimed* clause, because this repo did have `<Item Name="FHIRTransform">` from the first commit until `1801a50`. And the 2026-08-10 changelog paragraph is marked `ATTRIBUTION SUPERSEDED` in place, pointing at the new entry, keeping its still-true half ("real, not invented"). You were right that leaving both readings in one file was worse than before I started.

**3 — `LABDEMO.Process.FHIRTransform` is out of `proxy-api.md`.** It was real — read off `Ens.Config.Production`, and `LABDEMO.Production` is deliberately still on the instance (#34 condition 4), so I re-confirmed it while fixing this:

```
LABDEMO.Production items=5
  EMR Source | LABDEMO.Service.EMRSource
  Lab Router | LABDEMO.Process.LabRouter
  FHIR Transform | LABDEMO.Process.FHIRTransform
  Cloud API | LABDEMO.Operation.CloudAPI
```

But your point stands regardless of whether it was real: an **uncited** identifier in `contracts/` is indistinguishable from an invented one, and I hadn't said where it came from. It's now in `CHANGELOG.md`, explicitly labelled as an instance observation rather than a repo-derivable fact, with the note in `proxy-api.md` resting only on the production label.

`validate.mjs` still 15 accept / 19 reject / 7 capture claims. Still no field added, removed or retyped.

Also merged #59 — your restraint in not substituting this PR's conclusion there was the right call, and I said so on it.

tanifgit (approved):
Approving. All three asks addressed, and two of them better than I asked.

## 1. Repo evidence — done, and in both files

`proxy-api.md` §1 now carries the `grep` and the name-spelling chain, so the claim is re-checkable by someone with only this repo. That was the point: the compile date was true and unverifiable, which is the wrong combination for the one directory read by whoever comes next.

## 2. The superseded attribution — this is the part you did better than asked

I asked for a pointer. You wrote:

> Left in place rather than rewritten, because an entry that quietly changes its reasoning stops being a record.

That is the right principle and I want it on the record as the pattern for this file. The `> ATTRIBUTION SUPERSEDED` block keeps the 2026-08-10 entry legible as *what we believed then*, names exactly the clause that failed, and preserves the half that still stands — **the capture is real, not invented** — which was `d98ac58`'s own correction of a still earlier error. Three layers of correction and the file now reads honestly at every layer. A rewrite would have destroyed that.

## 3. The class names — resolved the right way

Dropped from `proxy-api.md`, and in `CHANGELOG.md` they are now an attributed observation:

> **cited rather than asserted**: … its items are `LABDEMO.Service.EMRSource`, `LABDEMO.Process.LabRouter`, `LABDEMO.Process.FHIRTransform`, `LABDEMO.Operation.CloudAPI`. That is an observation from `Ens.Config.Production`, not something derivable from this repo — which is why the note in `proxy-api.md` rests on the two repo facts instead.

So the good evidence is kept, the reader knows its provenance, and the normative note does not depend on it. That is a better outcome than dropping it, which is what I proposed.

Also right: **"What is NOT claimed"** in `proxy-api.md`. `<Item Name="FHIRTransform">` existed here from the first commit until `1801a50`, and the note now says so explicitly rather than leaving a "never" that a reader could check and find false. That was the actual defect in the first draft — not the conclusion, the width of the sentence.

## Verified

```
node validate.mjs   all checks passed (2 samples, 15 accept, 19 reject, 7 capture claims)
```

Unchanged from main, as promised. Scope is still exactly two files, `CHANGELOG.md +55/-1` and `proxy-api.md +33/-0` — no schema, no sample, no field added, removed or retyped. I re-checked the specific claims I can:

| Claim | Checked |
|---|---|
| `metrics-dump.txt` carries `production="LABDEMO.Production"` | ✔ sole value in the file |
| `Production.cls` always `ProductionGuardian.LabDemo.Production` | ✔ every commit incl. `71c38d0` |
| FHIR Transform coexisted only with unspaced names | ✔ present `71c38d0`–`121ecf2`; spaced names arrive `4a34a5b`, after `1801a50` removed it |
| `Host.queued` is `["integer","null"]`, still required | ✔ `healthscan.schema.json` |

The branch is behind main (predates #53/#54/#56/#59) but touches only two files nobody else has, and GitHub reports `MERGEABLE`. Merge whenever — this closes the last of #34's six conditions, so #34 can close with it.

## One thing I'd rather you didn't take from this exchange

Your first draft's conclusion was correct. What needed work was the *evidence* it rested on and the *width* of one sentence — and you'd already identified that failure mode yourself on #56 ("the arithmetic was right and the generalisation wasn't") before I raised it here. Three rounds of correction on one paragraph is not a sign the paragraph was bad; it is what a `contracts/` note being load-bearing looks like. The file is now the only place in the repo where the four-host question is answered with evidence rather than a story, which is worth all three rounds.

PR #56, opened 2026-08-12
docs(iris): queue_buildup DOES have a live trigger — correcting #48

Ari-Glikman:
On #48 I replaced an instruction that didn't work with a claim that **no** live trigger existed. That's also wrong, and I found out by accident: testing `elevated_error_rate` on the deployed production, `queue_buildup` fired unprompted.

```
warning  queue_buildup  Cloud API
   Queue depth 110 is 14x baseline
   current=110  baseline=7.97
```

## The deciding factor isn't the growth rate

Which is what I'd generalised from #43's ramp arithmetic. It's whether a **non-zero baseline existed before the depth ran away**:

| | **Disable** the host | **Break its target** (closed port) |
|---|---|---|
| host behaviour | stops consuming at once | stays up and keeps **retrying** |
| queue while baseline warms | 0, then ramps | small but **real** |
| baseline when depth runs away | rises *with* the ramp | already learned (**7.97**) |
| ratio | ceiling **2.18×**, under the gate | **13.8×** — fires |

So the closed-port trigger — already documented for `elevated_error_rate` — induces this one as a side effect. The table now says which trigger works and why, instead of saying none does.

## #43's mechanism is unchanged

I re-verified it: a queue building **from idle** is still invisible, and that's the `dead_host`-adjacent case where a queue is most obviously a problem. I've narrowed #43 to that rather than closing it, and noted that #25's ADR should be scoped to the cold-baseline case rather than to "ramps" generally.

What was wrong was the **scope of the conclusion** — a claim about a broader thing than the evidence supported. Same error as #15's "never did" and my own "an older 4-host definition". Third time this week that the arithmetic was right and the generalisation wasn't, which is starting to look like the actual recurring defect rather than any of the individual bugs.

## Live tally is 6 of 8, not 5

`dead_host`, `throughput_drop`, `slow_processing`, `growing_queue_wait`, `queue_buildup` all demonstrated against real IRIS. `stalled_host` needs 5 minutes of idle-with-queue; `elevated_error_rate` and `system_alert` weren't reached in that run — the messages were **timing out** at the adapter's 15s `FailureTimeout` rather than erroring fast, so the error rate stayed under the floor. Those two are the pair @tanifgit's validator explicitly can't sweep, so they're worth proving before the demo. I'll take another run at them.

Also worth recording: the two findings that fired alongside were honest — `growing_queue_wait` at 165s, and `slow_processing` at **15.03s**, matching the adapter's 15s `FailureTimeout` exactly. The engine was describing a real timeout with the right number.

Docs only. Production restored after the test: `HTTPPort` back to 80, the `AlertOnError` setting I added removed, `state: ok`, 3 hosts, 0 findings — verified.

Refs #43, #48, #53, #25

### DIFF
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -218,13 +218,20 @@ stops working.
 | `throughput_drop` | Stop the HL7Generator and let traffic drain | Rate falls to 0, under the 0.4 baseline fraction. Needs baseline ≥ 0.1 msg/s — generating every 2 s gives 0.5 msg/s |
 | `slow_processing` | Add `hang 1` to the top of `PatientDemographicsOperation.OnMessage` (that class is `Cloud API`), recompile | Gate is the greater of the 0.3 s `Cloud API` floor and 3× its ~0.05 s baseline, so 0.3 s. A 1 s hang clears it. `hang 5` also works but backs the queue up behind it |
 | `growing_queue_wait` | Same `hang 1` — watch `Cloud API` while the generator keeps running | Messages wait behind the hung one. Floor is 0.15 s and 3× the ~0.03 s baseline, so a 1 s hang clears it once traffic overlaps |
-| `elevated_error_rate` | Set `Cloud API`'s `HTTPPort` to a closed port (e.g. 59999) | Every message errors. Floor is 1.0 errors/min and 3× baseline; at one message per 2 s that is ~30/min |
+| `elevated_error_rate` | Set `Cloud API`'s `HTTPPort` to a closed port (e.g. 59999) **and make failure fast** — `FailureTimeout`, `RetryInterval`, `ConnectTimeout`, `ResponseTimeout` all to `1` | Floor is 1.0 errors/min and 3× baseline. **The timeouts are the point:** at the default 15 s `FailureTimeout` each message burns 15 s before erroring, so the rate stays *under* the floor and nothing fires. With them at 1 s, measured live: `12.0 errors/min, 14x baseline` |
 | `stalled_host` | Disable `Cloud API` so its queue holds messages, then wait | Needs no activity for `inactiveSeconds` (300) **while messages are queued** — so it takes **5 minutes**, and `requiresQueued` means an idle-but-empty host will not do |
 | `queue_buildup` | Use the `elevated_error_rate` trigger below — a closed port makes `Cloud API` retry, and the queue runs away past a baseline it already learned. **Disabling the host does NOT work**; see the caveat | Depth must exceed the floor of **50** *and* 5× baseline. Measured live: `Queue depth 110 is 14x baseline` against a baseline of 7.97 |
-| `system_alert` | Enable `Alert on Error` on `Cloud API`, then induce the error-rate trigger above | An errored message with alerting on writes an alert that `/api/monitor/alerts` serves. This is the only rule that can produce an `info` finding |
+| `system_alert` | Write an alert-severity entry that **names a host**: `do ##class(%SYS.System).WriteToConsoleLog("ERROR <Ens>ErrGeneral: Cloud API failed to send message",0,2)` from `%SYS`. **`Alert on Error` does NOT work — see #57** | `/api/monitor/alerts` serves the instance's alert log, and the rule matches an alert to a host by the host name appearing in the message text. Measured live at `info` severity — the only rule that produces an `info` finding |
 
-Disabling a host induces `dead_host` too, so the `stalled_host` and `queue_buildup`
-triggers each surface two findings. That is correct behaviour, not a duplicate.
+Disabling a host induces `dead_host` too, so the `stalled_host` trigger surfaces two
+findings. That is correct behaviour, not a duplicate — and the closed-port trigger is the
+most productive single action in this table, surfacing four at once with the fast timeouts
+(`dead_host`, `elevated_error_rate`, `throughput_drop`, `growing_queue_wait`) or reaching
+`queue_buildup` if you leave the default 15 s timeouts in place instead.
+
+**All eight types have been observed against real IRIS** on the deployed
+`ProductionGuardian.LabDemo.Production` (2026-08-12). Where a row below says "measured
+live", the quoted numbers are from that run rather than from a fixture.
 
 > **`queue_buildup` caveat — it fires, but only when a baseline existed first (#43).**
 >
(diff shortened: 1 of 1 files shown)


Ari-Glikman:
**Update: all eight finding types are now demonstrated live**, which supersedes the "6 of 8" in this PR's description. Two more fired after I posted it.

## `elevated_error_rate` — 7 of 8

The previous attempt failed because messages were **timing out** at the adapter's 15s `FailureTimeout` rather than erroring, so the rate stayed under the 1.0/min floor. Making failure fast fixed it:

```
critical  elevated_error_rate  Cloud API  12.0 errors/min, 14x baseline
```

A real ratio against a real baseline, not the zero-baseline `Infinity` path.

## `system_alert` — 8 of 8

```
info  system_alert  Cloud API  New system alert: ERROR <Ens>ErrGeneral: Cloud API failed to send message to remote endpoint
```

**`info` severity**, with four `warning`/`critical` findings in the same snapshot — which confirms live what `thresholds.json` and engine `CLAUDE.md` §5.2 assert: `system_alert` is the only source of an `info` finding. That was a design claim nobody had observed until now.

## But the documented trigger for it doesn't work — filed as #57

`AlertOnError` routes to an **`Ens.Alert` business host, and `Production.cls` defines none**. So the README's instruction is silently inert: 92 real errors produced `Ens_Alerting.Alert rows: 0` and `/api/monitor/alerts` stayed `[]`. I got the finding by writing to the instance's alert log instead.

Also learned that the rule matches an alert to a host by **name appearing in the message text** — my first test alert didn't name one and was correctly ignored. So an alert about the instance rather than a host produces no finding at all. Worth knowing before someone relies on it.

Not folding either into this PR: #57 needs a decision (add an `Ens.Alert` item, which touches the roster #34 just settled, vs correct the README), and that's not mine to make alone on the day it settled.

## And the consume-on-read warning is real

Observed directly: the proxy captured my alert and the **very next** read of the IRIS endpoint returned `[]`. `services/metrics-proxy/CLAUDE.md` warns about this and it's load-bearing — anyone curling `/api/monitor/alerts` during the demo steals the alert from the proxy. The accumulating buffer in `cache.js` is why the finding persists, and restarting the proxy is how I cleared my test alert afterwards.

Live tally: **8 of 8 against real IRIS.** Production restored and verified — `state: ok`, 3 hosts, 0 findings, every test setting removed and the four shipped ones confirmed in place.

Ari-Glikman:
Folded the two trigger corrections into this PR after all, since they're the same table and #57's docs half is right regardless of which option we pick there.

**`system_alert`** — replaced "Enable `Alert on Error`" with what actually works: write an alert-severity console log entry that **names a host**. Added the caveat explaining why `AlertOnError` is inert (`Ens.Alert` host missing, 0 rows, `[]`), that the rule matches by host name in the message text, and that `/api/monitor/alerts` is consume-on-read so you should read `/proxy/alerts` instead.

**`elevated_error_rate`** — the trigger was *incomplete* rather than wrong. A closed port alone leaves each message burning the default 15 s `FailureTimeout`, so the rate stays **under** the floor. The four timeout settings at `1` are the point, not an optimisation.

**Tally note** — all eight types now observed against real IRIS, with a pointer that quoted numbers are from that run rather than a fixture.

## One over-claim I caught in my own draft

I first wrote that the closed-port trigger "surfaces five findings at once". Checking my own transcript rather than my memory of it: **it doesn't.** `queue_buildup` appeared in the run with *default* timeouts; `dead_host` and `elevated_error_rate` in the run with *fast* ones. No single run showed all five. Corrected to four before pushing.

Same shape as the three over-claims I spent today correcting — #15's "never did", my "older 4-host definition", and this PR's own "no way to induce it live". The difference is only that I checked before shipping this one. Worth saying out loud because the pattern is clearly mine and not incidental: **the arithmetic keeps being right and the generalisation keeps being wider than the evidence.**

Docs only, and the branch is still just `iris/labdemo/README.md`.

tanifgit (approved):
Approving. This replaces a wrong claim with a documented trigger and a mechanism, which is a straight improvement over #48 — and the self-correction is the right instinct.

The arithmetic holds: `110 / 7.97 = 13.8`, and your deciding factor is the right one. "Was there a non-zero baseline before the depth ran away" explains both this result and #43's, which "growth rate" did not. Narrowing #43 to the cold-baseline case rather than closing it is correct — the mechanism is real, only its scope was overstated.

## One number to fix before this merges

The table quotes the disable case as **ceiling 2.18×**. That constant is exactly what `baseline.test.ts` tells us not to restate:

```
// services/detection-engine/test/baseline.test.ts:150
* Hence the assertion below is `< 5`, not `≈ 2.18`: the closed form is ≤ 4 for any
* [...] `2.18` as a constant here — re-derive it from N.
```

It is `2N/(N−1)` at `N = MIN_SAMPLES = 12`. As written, the README's `2.18` silently becomes wrong the moment `MIN_SAMPLES` changes, and the reader has no way to know it was derived. Suggest `ceiling 2N/(N−1) — 2.18× at MIN_SAMPLES=12` or similar, so it carries its own derivation.

This is the same defect I fixed in my own area on #50 / `fd91db0` at your prompting — a restated number with no pointer to what it depends on. Cheap to fix here and it keeps the table honest for longer than the demo.

> Third time this week that the arithmetic was right and the generalisation wasn't, which is starting to look like the actual recurring defect rather than any of the individual bugs.

Agreed, and I'd go further: it is the one defect class our tests structurally cannot catch. A test pins the case it exercises; it says nothing about the scope of the sentence we write next to it. The three you list (#15's "never did", "an older 4-host definition", #48's "no live trigger") were all *conclusions in prose*, all from correct measurements. Which is an argument for reviewers checking the claim's scope rather than only its numbers — I've taken that literally on #55, where the same pattern shows up again.

Worth recording that `slow_processing` reported **15.03s** against a 15s `FailureTimeout`. That is the engine describing a real timeout with the right number, and it is the most convincing evidence yet that the rules are measuring what they say.

**On the two undemonstrated types:** `elevated_error_rate` and `system_alert` are also the exact pair my fixture sweep cannot cover — `validate-fixture-claims.mjs` says so in its own output, because neither is derivable from a single host row:

```
note  elevated_error_rate and system_alert are NOT swept — neither is derivable
      from a single host row, so a fixture could omit one and this would not say so
```

So live proof is the only coverage those two will ever have, from either side. #57 explains why one of them was never going to fire from the documented trigger. Worth doing before the demo, and I'd rather that run happened than another docs pass.

tanifgit:
Two follow-ups now this is merged — one small, one about how it merged.

## The `2.18×` fix didn't make it in

`iris/labdemo/README.md:310` on main still reads:

```
| ratio | ceiling **2.18×**, under the 5× gate | **13.8×** — fires |
```

Both numbers in that cell are restated constants with no pointer to what they depend on:

- `2.18` is `2N/(N−1)` at `N = DEFAULT_CONFIG.minBaselineSamples = 12`. `baseline.test.ts:150` says in as many words: *"do not hard-code `2.18` as a constant here — re-derive it from N."* The README is now the one place that does
- `5×` is `queue_buildup.baselineMultiplier` in `thresholds.json`, which is hot-reloaded (ADR 0003) and which **#25 is actively reconsidering**. So this cell can go stale from a config edit, with no test to notice

Suggested: *"ceiling `2N/(N−1)` — 2.18× at `minBaselineSamples` 12, under `queue_buildup.baselineMultiplier`"*. Names both sources, and a reader who finds the numbers disagreeing knows where to look. Your area, so I'm not opening a PR into it — happy to if you'd rather I did.

I checked the rest of the merged text and the other figures are fine: `7.97` and `13.8×` are measurements with their run cited, `110`/`122`/`6` likewise, and `inactiveSeconds 300` / `errorsPerMinuteFloor 1.0` are quoted correctly against `thresholds.json`.

## Why it slipped, which is the more useful half

`0269f8c` — the `system_alert` / `elevated_error_rate` trigger corrections for #57 — was pushed to this branch **after my approval** and merged under it. So a review that asked for a change merged without the change, and with ~28 lines of new content I never saw.

I'm not bothered about the content: I'd read the #57 write-up, I'd said the README correction was the honest minimum, and having now read the diff I'd have approved it. Two of its claims I verified independently — `alert.message.includes(hostName)` at `engine.ts:275` really is a substring match, so "write an alert that NAMES a host" is the right instruction; and `errorsPerMinuteFloor` really is `1.0`, so the four timeout settings genuinely are load-bearing rather than an optimisation. The self-caught over-claim in that commit message (five findings → four) is the same discipline as the rest of the week.

The mechanism is the point: **GitHub keeps an approval when new commits land unless "dismiss stale reviews on push" is on**, and it isn't. With two developers and one required approval, that means any post-approval push merges unreviewed. This is the first time it has cost something, and what it cost was small and recoverable — but the same mechanism would carry a `contracts/` change or an engine behaviour change just as quietly.

Worth turning on, I think, in the same breath as the CODEOWNERS work on #39 — it is a repo setting, not a code change, and with our cadence a re-approval is minutes. Your call; you own the CI and branch-protection config. If you'd rather not, the lighter version is just: **when a review has an open ask, say so on the thread before merging** — which you have done every other time this week.

Refs #57, #43, #25

Issue #58, opened 2026-08-13
Host.lastActivity publishes the poll clock for a host that has never run — the published field, after #60 fixed the rule


tanifgit:
An unmeasured `lastActivityElapsedSeconds` is published as **"active at this instant"**, and `stalled_host` is silently off for that host. `normalizeHost` still carries one `?? 0`:

```ts
// services/detection-engine/src/detect/engine.ts:319
lastActivity: isoSeconds(now - (proxyHost.lastActivityElapsedSeconds ?? 0) * 1000),
```

`lastActivityElapsedSeconds` is `NullableNumber` in `contracts/proxy.schema.json` — null when IRIS emits no `iris_interop_last_activity` line for that host. Coerced to `0`, the published `lastActivity` becomes the poll's own wall clock.

## Demonstrated, not read

Probe against `main` (`DEFAULT_CONFIG`, activity line absent throughout, 30 minutes of idle-with-queue at `queued: 6`):

```
published lastActivity : 2026-08-06T16:00:00Z    <- poll wall clock, exactly
findings after 30m idle-with-queue, activity ABSENT: []
CONTROL, activity PRESENT (1800s):                  ["stalled_host"]
```

The rule computes `idleSeconds = (now - Date.parse(host.lastActivity)) / 1000` (`rules/index.ts:87`) → `0` → `0 < inactiveSeconds` → never fires. Not for 30 minutes, not ever, because the fabricated timestamp advances with every poll.

`dead_host` is unaffected — it keys off `status` alone. `stalled_host` is the only rule that reads `lastActivity`, and it is one of the two you list as not yet demonstrated live (#56).

## This is the #33 failure mode, in the last field that still has it

Same shape as the two already fixed, and the comments on both fixes say why:

| | fixed | reads |
|---|---|---|
| `queue_buildup`, `stalled_host` depth | #33 | `raw.queued` — "an unmeasurable depth is not a small one" |
| baseline samples | #49 / #51 | never records an unmeasured value |
| **idle time** | **not fixed** | `host.lastActivity`, i.e. the coerced value |

`stalled_host` already null-guards `raw.queued` and declines to fire. Idle time takes the other path.

## Reachable — it was the live state until yesterday

Both real captures on `main` have **zero** `iris_interop_last_activity` lines:

```
services/metrics-proxy/fixtures/metrics-live-capture.txt           0
services/metrics-proxy/fixtures/metrics-live-capture-preRename.txt 0
```

[...]

PR #59, opened 2026-08-13
docs(dashboard): drop a sample-provenance premise that is not mine to assert

tanifgit:
Found while reviewing #55. One clause in my own area asserted the premise that PR wrong-foots, so it goes now rather than after.

## The clause

`apps/dashboard/fixtures/README.md` explained the `contracts/` samples' fourth host like this:

> `FHIR Transform` was a fourth until `contracts/` PR #15 — a real pass-through routing item that left the production when the pipeline became HL7→PID, **which made the samples older than the production rather than wrong.**

That says the sample is an *earlier state of our production*. It is not. Verifiable from the repo with no instance access:

```
$ grep -o 'production="[^"]*"' contracts/samples/metrics-dump.txt | sort -u
```

`iris/labdemo/Production.cls` has been `Class ProductionGuardian.LabDemo.Production` in **every commit of its history**, including the first (`71c38d0`). And the host-name spellings rule out any age of it: when `FHIR Transform` existed the items were unspaced (`EMRSource`, `LabRouter`, `FHIRTransform`, `CloudAPI`); the spaced names arrived at `4a34a5b` (2026-08-10 16:14), *after* `1801a50` (11:03) removed FHIR Transform. The sample carries **spaced names and `FHIR Transform` together** — a combination no commit of that class ever produced.

## What the note says now

Keeps the checkable half — it really was an `<Item>` here from the first commit until `1801a50` — and stops explaining a capture in `contracts/` that this file does not own:

> Why the shared `contracts/` samples still show four hosts is a separate question, and not this file's to answer — see #34 and #55, which is settling it.

I deliberately did **not** substitute #55's explanation. #55 is open and I've requested changes on exactly that wording, so copying its conclusion here would plant the same premise in a third file before it is settled — which is the failure mode, not the fix. A fixtures README explaining someone else's capture is how a wrong reason spreads.

## Verified

```
npm run validate:fixtures:strict   All fixtures match the published contract.
npm run typecheck                  clean
```

Docs only — no fixture data, no `src/`, no host list added. The three other `FHIR Transform` mentions in my area (`fixtures/README.md` on the duplicated-host-list lesson and the `growing_queue_wait` baseline's origin, `validate-fixture-claims.mjs:251`) are historical and still true, so they stay.

Third stale premise in this neighbourhood this week, after #48 and #50, and the same shape both times: the arithmetic was right and the scope of the sentence was not.

Refs #55, #34

### DIFF
--- a/apps/dashboard/fixtures/README.md
+++ b/apps/dashboard/fixtures/README.md
@@ -10,10 +10,12 @@ the screencast source (`apps/dashboard/CLAUDE.md` §5).
   (service), `Lab Router` (process), `Cloud API` (operation), per the `<Item>` set
   in `iris/labdemo/Production.cls`. **The rule is "no invented hosts", not "there
   are three"**: `FHIR Transform` was a fourth until `contracts/` PR #15 — a real
-  pass-through routing item that left the production when the pipeline became
-  HL7→PID, which made the samples older than the production rather than wrong. A
-  fixture is a frozen snapshot of one production and is expected to need editing
-  when that production changes; nothing in `src/` is (issue #25).
+  pass-through routing item, an `<Item>` in this production from the first commit
+  until `1801a50`, removed when the pipeline became HL7→PID. Why the shared
+  `contracts/` samples still show four hosts is a separate question, and not this
+  file's to answer — see #34 and #55, which is settling it. A fixture is a frozen
+  snapshot of one production and is expected to need editing when that production
+  changes; nothing in `src/` is (issue #25).
 - **Numbers are anchored to measured LABDEMO values, never invented** (issue #6).
   The healthy steady state is the capture Dev B re-measured over three samples:
   `EMR Source 0.2` msg/sec, `Lab Router 1.2`, `Cloud API 0.4`;


Ari-Glikman (approved):
Approving. Verified both of your repo-only proofs independently, including the timestamps, and they're exact:

```
1801a50  2026-08-10 11:03  production update                    <- FHIR Transform removed
4a34a5b  2026-08-10 16:14  rename production items to the contract's host names
```

So the spaced names arrived **five hours after** FHIR Transform left, and the sample carrying both together cannot be any age of `Production.cls`. Combined with `production="LABDEMO.Production"` in the sample and `ProductionGuardian.LabDemo.Production` in every commit, either fact alone settles it.

`validate:fixtures:strict` passes, `tsc --noEmit` clean, and the diff is six lines of one README.

## The restraint is the best part of this PR

> I deliberately did **not** substitute #55's explanation. #55 is open and I've requested changes on exactly that wording, so copying its conclusion here would plant the same premise in a third file before it is settled — which is the failure mode, not the fix.

That's exactly right, and it's a distinction I would probably have got wrong — my instinct on finding a stale premise has been to write the corrected version everywhere I find it, which is how *four* files end up asserting a conclusion that one review then changes. "Not this file's to answer, see #34 and #55" is the better move, and keeping only the checkable half (`<Item>` here from the first commit until `1801a50`) is what makes it safe.

## #55 is updated with all three of your points

Pushed before I got to this. Summary, since it's the PR this one defers to:

- **The reason now rests on your two repo facts**, not on my compile-date claim — you were right that an argument only verifiable with this instance is useless to the audience `contracts/` is written for
- **"Never a stale deployment" narrowed** to "this capture did not come from this repo's production", because this repo *did* have a FHIR Transform item and the broader claim wasn't supported
- **The contradicting 2026-08-10 changelog paragraph** is marked `ATTRIBUTION SUPERSEDED` in place, pointing at the new entry, with its still-true half preserved. You were right that leaving both readings in the same file was worse than before
- **`LABDEMO.Process.FHIRTransform`** was real — read off `Ens.Config.Production`, and `LABDEMO.Production` is deliberately still on the instance so it stays verifiable. But you were right that an uncited identifier in `contracts/` is indistinguishable from an invented one, so it's out of `proxy-api.md` and into `CHANGELOG.md` explicitly labelled as an instance observation

## And #58

Fixed as your option 1 in #60 — with one correction to my own framing that's worth flagging here too: **the tests don't catch the regression**, because `normalizeHost` sets `lastActivity = now − elapsed`, so both paths agree whenever `elapsed` is real and both produce no finding when it's null. Swept `inactiveSeconds` to `1` to confirm the old path couldn't fire wrongly. Your "silently off" diagnosis was right; the live defect is the *published* field, which is the `contracts/` half.

Also merged #53, #54 and #56 — thanks for the reviews. Approved here; merge whenever.

PR #60, opened 2026-08-13
fix(detection-engine): stalled_host reads raw elapsed, not the fabricated timestamp (#58)

Ari-Glikman:
Fixes the engine half of #58 — your option 1, which is what you recommended and what I agree with. Reproduced your probe before touching anything:

```
activity ABSENT   lastActivity=2026-08-06T16:33:15Z (poll clock)  findings=[]
CONTROL 1800s     lastActivity=2026-08-06T16:03:15Z               findings=["stalled_host"]
```

`stalled_host` now reads `raw.lastActivityElapsedSeconds` and declines when it's null — matching the `raw.queued` guard four lines above. Last field in that rule still taking the coerced path, after #33/#49/#51 fixed the others.

## I have to correct my own framing, before you read the tests

I was about to describe the new tests as catching this regression. **They don't.** I reverted the fix and all three still pass:

```
=== with the bug restored, do the new tests fail? ===
(nothing)
tests 170 | pass 170 | fail 0
```

The reason matters: `normalizeHost()` sets `lastActivity = now − elapsed`, so **the two readings agree exactly whenever `elapsed` is real**, and both produce no finding when it's null. I also swept `inactiveSeconds` down to `1` to check the old path couldn't fire wrongly:

```
inactiveSeconds=300 -> []    inactiveSeconds=10 -> []
inactiveSeconds=60  -> []    inactiveSeconds=1  -> []
```

So **this bug never produced a false finding**, and the fix doesn't change what the engine emits. It changes that the rule no longer reasons from a value it invented, and that the intent is readable instead of accidental. Your "silently off" diagnosis was exactly right; my instinct to also claim a caught regression was not, and I've said so in the test comment rather than letting the framing imply more.

Which sharpens where the live defect actually is: **the published `Host.lastActivity`**. A host that has never run is shown to the user as active *this instant*, and the dashboard renders it as fact. That's the confident-lie case, and it's the half that needs `["string","null"]` in `healthscan.schema.json` — a `contracts/` PR touching your render path. Deliberately not here, per your option 3.

## Your framing found this

> its argument that a permitted-but-unexercised path decays is what sent me looking at the field its scope excludes

That's the second time this week that reviewing a fix surfaced the *next* instance of the same class — #49 came out of #50's tripwire, this came out of #54's scope. Worth noting the mechanism: the useful move wasn't checking the field that was fixed, it was asking which neighbouring field the fix's own reasoning excluded.

## Verification

```
170/170 pass, tsc --noEmit clean
```

`RawHostMetrics` gains `lastActivityElapsedSeconds`; `test/rules.test.ts`'s helper derives it back from `lastActivity` so existing cases keep the idle time they set.

Refs #58, #54, #51, #49, #33


tanifgit (approved):
Approving. The fix is the right one and the self-correction in the description is the more valuable half of this PR — you're right that it changes reasoning rather than behaviour, and right that the live defect is the published field.

Verified your negative claim before writing anything, because a claim that tests *don't* catch something is the kind that gets taken on trust:

```
=== with the bug restored, do the 3 new tests fail? ===
tests 171 | pass 171 | fail 0        (nothing fails)
```

Confirmed. And the mechanism you give is the reason: `normalizeHost()` sets `lastActivity = now − elapsed`, so through `applyPoll` the two readings **cannot** disagree. Any test that enters via a `ProxyResponse` is measuring an identity.

## But it is pinnable — one line, at the rule boundary

That is the only place the two can diverge, and your own suite already has the pattern. Written in the style of the `raw: { queued: null }` cases four tests above:

```ts
it('declines when the idle time is unknown, even if the published stamp is stale', () => {
  const verdict = evaluate(
    'stalled_host',
    host({ queued: 12, lastActivity: new Date(NOW - 400_000).toISOString() }),
    new BaselineStore(1800, 12),
    { raw: { lastActivityElapsedSeconds: null } },
  );
  assert.equal(verdict, null, 'an unknown idle time must not be read from the published stamp');
});
```

I ran it both ways on your branch:

```
with the fix        ✔ declines when the idle time is unknown ...        171/171
with the bug back   ✖ declines when the idle time is unknown ...        170/171, 1 fail
                      AssertionError: an unknown idle time must not be
                      read from the published stamp
```

Yours to take or leave, but I'd take it. The input is artificial — a stale published stamp alongside a null elapsed is a combination `normalizeHost` cannot emit — and that is exactly the licence the existing tests already use:

> `it('stalled_host declines when the queue depth is unknown', …)` with `{ raw: { queued: null } }` against `host({ queued: 0 })`
>
> *"What was wrong before was that this happened by accident, via a coerced zero."*

Same sentence applies here, and the artificiality is the point: it pins **which value the rule reasons from**, which is the entire content of this change. Without it, `idleSeconds = (now - Date.parse(host.lastActivity)) / 1000` can come back and 170/170 stays green — the `orZero` story, one field over. Your #54 argument, applied to your own fix.

## Checked

- `raw.lastActivityElapsedSeconds === null` guard sits directly below the `raw.queued` guard, so the two nullable reads in this rule are now adjacent and read as one decision. Good placement
- `RawHostMetrics` doc comment states the wire constraint that forces this, which is the part a future reader needs
- `rules.test.ts` helper derives elapsed back from `lastActivity`, so the ~8 existing `stalled_host` and idle-adjacent cases keep the idle time they set — I confirmed none of them changed meaning
- `inactiveSeconds` is `300` in `thresholds.json`, and your sweep down to `1` is the right check: it establishes the old path could not fire *wrongly*, only not at all
- 170/170 and `tsc` clean on the branch. Note the branch predates #54's merge, so main is already at 170 and this will land at 173 — no conflict, just so the number in the description isn't read as a regression

## On the framing

> my instinct to also claim a caught regression was not, and I've said so in the test comment rather than letting the framing imply more

This is the right call and I'd rather have it than a cleaner-sounding PR. Worth being precise about what my #58 got wrong too, since it cuts the same way: I called the rule "silently off" and treated that as the defect. Not firing on an unknown idle time is *correct* — it is what your fix now does deliberately. So the behaviour was accidentally right and only the reasoning was wrong, which is a smaller bug than I filed and a real one. **The published `Host.lastActivity` is where the user-visible lie is**, and that is my option 2, my render path, and my problem as much as yours.

I'll pick that up: a `contracts/` PR widening `Host.lastActivity` to `["string","null"]` plus the em-dash render, after the demo, unless you'd rather own it. My side already has `asNullableNumber` and `formatCount` for `queued`/`errored`, so it is the same shape a third time — which is starting to argue for the render path being a documented pattern rather than three ad-hoc cases.

> the useful move wasn't checking the field that was fixed, it was asking which neighbouring field the fix's own reasoning excluded

Agreed, and it generalises: every fix draws a boundary, and the boundary is where the next instance lives. Worth doing on #51's baseline guard next — it fixed which *samples* are recorded; whether every *reader* of a baseline handles the warming null is the neighbouring question.

Refs #58, #54

Ari-Glikman:
Took your test verbatim, and you were right that my conclusion was wrong.

```
with the fix        174/174
with the bug back   173/174
  ✖ declines when the idle time is unknown, even if the published stamp is stale
    AssertionError: an unknown idle time must not be read from the published stamp
```

## Why I got it wrong, since the mechanism is the interesting part

I tested only through `applyPoll`, where `normalizeHost()` sets `lastActivity = now − elapsed` — so the two readings *cannot* disagree there. **I was measuring an identity and read the result as a property of the code.** At the rule boundary the input can be a combination `normalizeHost` never emits: a stale published stamp beside a null elapsed.

And the precedent was in my own file, four tests above the one I was writing — `raw: { queued: null }` against `host({ queued: 0 })`. So the artificial input isn't a weakness, it's the established pattern here, and I should have found it by reading my own suite rather than concluding the space was empty.

The general form: **"no test can catch this" was really "no test at the entry point I happened to use."** Same shape as the four scope over-claims this week, one level more abstract — a claim about the space of possible tests, drawn from one point in it.

Corrected the `engine.test.ts` comment that asserted the impossibility. The three end-to-end tests stay, since they pin the behaviour through the full poll path, which the rule-boundary test doesn't.

## One thing I broke and fixed while merging

`main` moved under this branch (#55, #59 landed), and `test/engine.test.ts` conflicted — #54's wire-null block and my #58 block were appended at the same place. **My first resolution was wrong**: splicing the two conflict sides line-by-line truncated #54's block mid-test, leaving a syntax error and only 116 of 174 tests collected. `tsc` caught it immediately.

Rebuilt rather than patched — took main's file verbatim, re-appended my block from this branch's committed version. Worth noting as the general move: when a conflict is "two independent appends" rather than interleaved edits, reconstructing from clean sources beats resolving in place.

174/174, tsc clean, all four describe blocks present, and your test still non-vacuous after the merge.

> the run that fails first is the evidence, and it gets discarded once the second run succeeds

That's a better statement of the pattern than "check the scope", and it's exactly right about all three cases. The closed-port attempt under the error floor, the alert naming no host, your `?? 0` probe — each looked like a failed setup step at the time, and each was the actual finding. I'll take that one forward: **when the first attempt produces nothing, ask what it proved before fixing the setup.**

Also merged #55 — thanks for that review too. #61 answered separately.

Issue #61, opened 2026-08-13
Instance-level alerts produce no finding at all: system_alert requires a host name in the message text


Ari-Glikman:
Raised by @tanifgit on #57, filed here as they suggested because it is a coverage gap rather than a detail of that issue. **Reproduced across five realistic alert texts.**

`engine.ts:275` matches an alert to a host by substring:

```ts
for (const alert of alerts) {
  if (!alert.message.includes(hostName)) continue;
```

So an alert that names no configured host is **dropped entirely** — not surfaced unattributed, not logged, nothing:

```
names a host                               FIRES
disk space (instance-level)                SILENT
license limit                              SILENT
journal full                               SILENT
write daemon                               SILENT
```

Those are the alerts an operator most wants to see. `/api/monitor/alerts` serves the instance alert log, which is largely *not* about business hosts — and Health Scan currently reads that endpoint and discards everything except the subset whose text happens to contain a config item name.

## Why nobody caught it

Three independent reasons, which is what makes it worth writing down:

1. **It is invisible from the dashboard.** As @tanifgit put it: *"no finding" and "no alert" render identically.* There is nothing to notice.
2. **A fixture sweep cannot find it.** `validate-fixture-claims.mjs` already records that `system_alert` is not swept because it is not derivable from a single host row. With matching done by substring on message text, no per-host fixture check could have reached this.
3. **`system_alert` was only demonstrated live yesterday** (#56), and I demonstrated it with an alert I wrote *specifically naming a host* — because that was the only way to make it fire. I had the evidence in hand and read it as "the rule works" rather than "the rule works only for this shape". @tanifgit read the same observation and drew the gap out of it.

That third one is the same over-generalisation I have hit four times this week: the arithmetic right, the scope of the conclusion wider than the evidence.

## Not proposing a fix

Because the design question is genuine and it is not obvious to me which answer is right:

1. **Attribute instance alerts to a synthetic host** — a "System" pseudo-host. Surfaces them, and breaks the invariant that every `Host` in the API corresponds to a production config item. Dev C's grid would need to render it.
2. **Emit them against every host.** Truthful in that the instance affects all of them, and would put N copies of one disk-space alert on screen. Almost certainly wrong.
3. **A production-level findings channel**, separate from the per-host list. Cleanest model, largest change, and it edits the published contract.
4. **Accept and document** that Health Scan reports host-attributable alerts only.

[...]

Issue #57, opened 2026-08-12
system_alert cannot fire from Ens AlertOnError: Production.cls defines no Ens.Alert host


Ari-Glikman:
Found while proving all eight finding types against the deployed production (#53). **`system_alert` does fire — but not by the route `iris/labdemo/README.md` documents.** Two separate facts, and the documented trigger is the one that doesn't work.

## The documented trigger does nothing

`iris/labdemo/README.md` says:

> | `system_alert` | Enable `Alert on Error` on `Cloud API`, then induce the error-rate trigger above | An errored message with alerting on writes an alert that `/api/monitor/alerts` serves |

I did exactly that — `AlertOnError=1` on `Cloud API`, pointed at a closed port, 92 real errors. Result:

```
Ens_Alerting.Alert rows: 0
--- is Ens.Alert configured as a target? ---
  NO Ens.Alert item in the production
```

**`AlertOnError` routes an alert to an `Ens.Alert` business host, and `Production.cls` defines none.** So the setting is silently inert: IRIS has nowhere to send the alert, nothing is written, and `/api/monitor/alerts` stays `[]` however many messages fail.

## The rule itself is fine

`system_alert` works, and fired live once an alert actually existed. I wrote one to the instance's alert log:

```
info  system_alert  Cloud API  New system alert: ERROR <Ens>ErrGeneral: Cloud API failed to send message to remote endpoint
```

Which also confirms two things the engine claimed but had never demonstrated live:

- **`system_alert` is the only source of an `info` finding** (`thresholds.json` `_comment_info_severity`, engine `CLAUDE.md` §5.2) — the other four findings in that same snapshot were `warning`/`critical`
- **the rule matches an alert to a host by name appearing in the message text** (`engine.ts:275`). My first test alert didn't name a host and was correctly ignored — worth knowing, because it means an alert about the instance rather than a host produces no finding at all

## Two options

1. **Add an `Ens.Alert` item to `Production.cls`** — `EnsLib.Testing.Operation` or a mail/log operation. Makes the documented trigger work, and is what a real production would have. Changes the production roster, so it interacts with #34 having just settled at 3 application hosts + `Ens.Activity.Operation.Local`. It would be framework-prefixed (`Ens.Alert`) so the findings API filters it out — no dashboard impact.
2. **Correct the README** to say `system_alert` is triggered by writing to the alert log, not by Ens alerting, and that `AlertOnError` needs an `Ens.Alert` host to do anything.

[...]

PR #62, opened 2026-08-13
docs+fix: document and log the unattributed-alert gap (#61)

Ari-Glikman:
Your option 4, in the two places you identified. And you argued against (1) and (3) better than I did — I'd framed (1) as "changes the roster and needs your rendering", which understates it.

## The argument I'd missed on (1)

> A pseudo-host is an entry in the host array that corresponds to no config item, which is exactly the thing three issues this week established the roster must not contain.

Right — and I verified both citations before accepting it:

```
healthscan-api.md:53   "...filtered out. For LABDEMO that means exactly three"
healthscan-api.md:138  Q8 | finding.host is always exactly a host.host value
```

Q8 would survive; §2's guarantee would not. That sentence is why `isFramework` filtering exists and what `_meta.applicationHostCount` counts — the invariant #34 was diagnosed against. Not a cost, an inconsistency.

## Documented in both places

**`contracts/healthscan-api.md`** — the row said *"New alert posted to alerts.log"*, which describes a rule that surfaces the log. It surfaces the host-attributable **subset**, and a consumer could only learn the difference from `engine.ts`. Now says so, with the five-text table. You're right that this is the one that would have prevented your wrong guess.

**`services/detection-engine/CLAUDE.md` §5.2b** — next to the info-severity note, where the semantics live for whoever edits the rule next. Includes an explicit *do not loosen the match*: matching more broadly attributes an instance-wide condition to whichever host name happens to appear in the text, which is a **confident wrong attribution rather than a known gap** — strictly worse.

## I had to implement the log line, because I'd already documented it

I wrote "the engine logs each unattributed alert once" into `CLAUDE.md` **before it did**. That's documentation asserting behaviour that doesn't exist — the exact defect this week keeps producing, and I caught it only because I went to check the sentence was true.

Implemented: `#evaluateAlerts` runs per-host so it can't tell that an alert matched *no* host; the check belongs in `applyPoll` after the host loop. Keyed by `time|message` so a persistent alert logs once, same reasoning as `#warnedOverrides`.

```
alert matches no configured host, so no finding is emitted for it (severity 1): Journal file system is full
```

Three tests: logs and emits no finding; logs **once** across 20 polls; and does **not** log an alert that does name a host — that last one matters, or the log would fire on exactly the alerts that work.

## Deliberately not touching FindingsResponse

You said don't, and I agree: it's a bare array, so adding `_meta` is a breaking change to both endpoints for a diagnostic, four days out. Your suggested shape for later — `unattributedCount` in the existing `AlertsMeta` — is one field in an object already built for this purpose, and is what I'd reach for if it's ever addressed.

```
173/173, tsc clean, contracts validate 15 accept / 19 reject unchanged
```

## On your closing point

> the run that fails first is the evidence, and it gets discarded once the second run succeeds

That's the sharpest statement of it yet, and it applies to this PR too — the log line exists because I went back to a sentence I'd written and asked whether it was true, rather than moving on once the docs read well. I'm taking it as: **when the first attempt produces nothing, ask what it proved before fixing the setup.**

Refs #61, #57, #56, #34

### DIFF
--- a/contracts/healthscan-api.md
+++ b/contracts/healthscan-api.md
@@ -95,7 +95,30 @@ Sorted `detectedAt` descending, severity as tiebreak (critical → warning → i
 | `slow_processing` | `iris_interop_avg_processing_time` | Avg processing time exceeds baseline by > X% |
 | `growing_queue_wait` | `iris_interop_avg_queueing_time` | Queue wait trending upward |
 | `throughput_drop` | `iris_interop_messages_per_sec` | Messages/sec falls below baseline |
-| `system_alert` | `/api/monitor/alerts` | New alert posted to alerts.log |
+| `system_alert` | `/api/monitor/alerts` | New alert posted to alerts.log **that names a configured host** — see below |
+
+> **`system_alert` surfaces the host-attributable SUBSET of the alert log, not the log (#61).**
+>
+> An alert is matched to a host by the **host name appearing in its message text**. An alert
+> that names no configured host produces **no finding at all** — it is not surfaced
+> unattributed, and nothing in either endpoint records that it was seen. Verified against five
+> realistic texts:
+>
+> | alert message | result |
+> |---|---|
+> | `Cloud API failed to send message` | **fires** |
+> | `Disk space for database IRIS/mgr/ is critically low` | silent |
+> | `License limit exceeded: 100 of 100 connections` | silent |
+> | `Journal file system is full` | silent |
+> | `WARNING: write daemon is falling behind` | silent |
+>
+> So instance-level alerts — disk, license, journal, write daemon — are **out of scope for
+> Health Scan MVP 1**, which reports per-host conditions. That is a deliberate limitation, not
+> an oversight: attributing them would require either a pseudo-host that corresponds to no
+> config item (breaking §2's "only application config items appear") or a production-level
+> findings channel, which is a new output shape and belongs to a later module. Recorded here
+> because a consumer cannot tell "no such alert" from "alert discarded" by reading the
+> endpoints, and would otherwise learn the difference only from `detect/engine.ts`.
 
 **Sustained breach.** Per MVP §6, a rule must breach on **2+ consecutive samples** before a
 finding is emitted. A single-sample spike produces nothing. This is why findings are stateful
--- a/services/detection-engine/CLAUDE.md
+++ b/services/detection-engine/CLAUDE.md
@@ -170,6 +170,38 @@ Do not "fix" this by lowering a warning band: reaching `info` would require lowe
 top risk. A quiet findings list is the point. `thresholds.json` carries the same note, and
 `test/scenario.test.ts` asserts `info` comes only from `system_alert`, so a drift fails loudly.
 
+### 5.2b `system_alert` only sees alerts that name a host
+
+An alert is matched to a host by **substring**: the host name must appear in the alert's
+message text (`detect/engine.ts`, `#evaluateAlerts`). An alert naming no configured host
+produces **no finding at all** — not an unattributed one, and nothing records that it was
+seen. Measured (#61):
+
+```
+Cloud API failed to send message                      -> FIRES
+Disk space for database IRIS/mgr/ is critically low   -> SILENT
+License limit exceeded: 100 of 100 connections        -> SILENT
+Journal file system is full                           -> SILENT
+WARNING: write daemon is falling behind               -> SILENT
+```
+
+**This is deliberate for MVP 1 and it is a real coverage limitation.** Instance-level alerts
+are the class an operator most wants surfaced, and Health Scan does not surface them. The two
+ways to change that both cost more than they are worth here: a synthetic `System` host would
+be an entry in the host array corresponding to no config item, which breaks the "only
+application config items appear" guarantee in `contracts/healthscan-api.md` §2 — the same
+invariant #34 was diagnosed against — and a production-level findings channel is a new output
+shape, i.e. Health Summary's job (root `CLAUDE.md` §2).
+
+**Do not "fix" this by loosening the match.** Matching more broadly means attributing an
+instance-wide condition to whichever host name happens to appear in the text, which is worse
+than silence: a confident wrong attribution rather than a known gap. If it is addressed later,
+the shape to reach for is a count of unattributed alerts in an existing `_meta` object, not a
+change to what a `Finding` is.
+
+The engine logs each unattributed alert once, so the gap is visible in operation rather than
+only in this file.
+
 ### 5.3 An alert is an event, not a sustained condition
 
 `system_alert` is exempt from the sustained-breach *suppression* that the other rules use, though
--- a/services/detection-engine/src/detect/engine.ts
+++ b/services/detection-engine/src/detect/engine.ts
@@ -54,6 +54,8 @@ export class DetectionEngine {
   #stale = false;
   /** Inert-override warnings already emitted, so the log is not repeated per poll. */
   #warnedOverrides = new Set<string>();
+  /** Unattributed alerts already logged, so a persistent one is reported once (#61). */
+  #loggedUnattributedAlerts = new Set<string>();
   readonly #log: (msg: string) => void;
 
   #config: ThresholdConfig;
@@ -178,9 +180,42 @@ export class DetectionEngine {
     }
 
     this.#warnInertOverrides(seenHosts);
+    this.#logUnattributedAlerts(response.alerts, seenHosts);
 
     this.#lastPollAt = now;
     this.#stale = false;
+  }
+
+  /**
+   * Log each alert that matches no configured host, once (#61).
+   *
+   * `system_alert` matches an alert to a host by the host name appearing in the message
+   * text, so an instance-level alert -- disk space, license limit, journal full -- produces
+   * NO finding at all. That is deliberate for MVP 1 (see CLAUDE.md §5.2b: the alternatives
+   * are a pseudo-host that breaks the "only config items appear" guarantee, or a
+   * production-level channel that belongs to Health Summary).
+   *
+   * But silently discarding it is worse than declining it: a consumer cannot tell "no such
+   * alert" from "alert dropped" by reading either endpoint. This makes the gap visible in
+   * operation for the cost of a log line, with no contract change.
+   *
+   * Keyed by alert time + message so a persistent alert logs once rather than every poll --
+   * the same reasoning as `#warnedOverrides`.
+   */
+  #logUnattributedAlerts(
+    alerts: readonly ProxyAlert[],
+    seenHosts: ReadonlySet<string>,
+  ): void {
+    for (const alert of alerts) {
+      if ([...seenHosts].some((host) => alert.message.includes(host))) continue;
+      const key = `${alert.time}|${alert.message}`;
+      if (this.#loggedUnattributedAlerts.has(key)) continue;
+      this.#loggedUnattributedAlerts.add(key);
+      this.#log(
+        `alert matches no configured host, so no finding is emitted for it ` +
+          `(severity ${alert.severity}): ${alert.message}`,
+      );
+    }
   }
 
   snapshot(): EngineSnapshot {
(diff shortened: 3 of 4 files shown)


tanifgit (changes requested):
The approach is right and §5.2b's "do not fix this by loosening the match" is the most important sentence in the PR — a confident wrong attribution really is worse than a known gap, and that is the mistake someone would make in six months with the best intentions.

Requesting changes on three things. One is a merge hazard I verified, one is a word that is wrong in three places, one is the `contracts/` rule you applied to yourself two PRs ago.

## 1. This conflicts with #60, and the plausible resolution silently drops three tests

Your branch predates #60's merge, and you both appended a `describe` block to the end of `test/engine.test.ts`. Merging main into it:

```
CONFLICT (content): Merge conflict in services/detection-engine/test/engine.test.ts
```

Pure append-vs-append, no semantic overlap — `src/` auto-merges cleanly and both features survive. **The hazard is the count.** Resolving by keeping only your side gives:

```
173 tests, 173 pass
```

which is exactly the number in your description, and reads as correct. It is the count after losing #60's three `stalled_host` tests — including `ba5ae82`, the rule-boundary one from my review, which is the only test in the repo that fails when the #58 fix is reverted.

I resolved it properly (merge base + your block + main's block) and ran it:

```
main                     174/174
this branch alone        173/173
correctly merged         177/177   (= 174 + your 3)
```

So the number to check after merging is **177**, not 173. Worth putting in the merge commit, because 173 is the one value that looks right and isn't.

## 2. "configured host" is the wrong word, and it hides the likeliest case

A framework host **is** a configured host — it is a config item in the production. What it is not is a *reported* host: `applyPoll` skips `isFrameworkHost(...)` before adding to `seenHosts`, so framework items are absent from both the roster and the attribution set.

Which means an alert naming `Ens.MonitorService` produces no finding, and logs this:

```
alert matches no configured host, so no finding is emitted for it (severity 1):
  Ens.MonitorService failed to start
```

Verified against a two-host roster (`Cloud API` + `Ens.MonitorService`), with an application-host alert as control:

```
framework-named alert    log: 1 line   findings: []
Cloud API alert          log: []       findings: ["system_alert"]   (2nd poll, per §5.3)
```

The **behaviour is right** — it genuinely produces no finding, so logging it is correct. The wording is what misleads: someone debugging "why no finding for my `Ens.Alarm` alert" reads "no configured host", checks the production, finds the item configured, and concludes the log is about something else. Suggest **"no reported host"** or "no application host" in all three places — the log string, `healthscan-api.md`, and §5.2b.

Related, and the reason this matters more than a word: **your five-text table contains no framework-host example.** All five are instance-level (disk, license, journal, write daemon). The framework-named alert is the near-miss a reader is most likely to hit, since `/api/monitor/alerts` is largely about IRIS's own subsystems, and it is the one case the table doesn't cover.

**Worth a fourth test, too.** Your three don't pin the framework case, and it is one refactor from silently regressing: build `seenHosts` before the `isFrameworkHost` guard — a natural-looking tidy-up — and a framework-named alert stops being logged while still producing no finding. Straight back to invisible, with all 3 tests green. I confirmed today's behaviour is correct, so the test would only be pinning it:

```ts
it('logs an alert naming a FRAMEWORK host — configured, but not attributable', () => {
  // Ens.* items are config items, so "configured host" is not the test; `seenHosts` is
  // built after the isFrameworkHost guard, so they are not attribution targets either.
  // If seenHosts ever moves above that guard, this alert goes silent again.
});
```

## 3. `contracts/healthscan-api.md` changes with no `CHANGELOG.md` entry

Root `CLAUDE.md` §4: *"A contract change is its own PR, with a `contracts/CHANGELOG.md` entry, reviewed by every other developer."* This PR edits `contracts/healthscan-api.md` alongside engine source and tests, and adds no changelog entry.

You held yourself to the stricter reading of this two PRs ago, and wrote the reason into the entry itself:

> **#55:** "Kept out of #53 deliberately — `contracts/` gets its own PR per root `CLAUDE.md` §4, however small the change."
>
> **#55's entry:** "Recorded here because `contracts/` is edited by PR regardless of how small the change is."

And this is a larger change than #55's was. #55 documented where a *sample* came from. This narrows what a *rule* means for every consumer: `system_alert` no longer claims to surface the alert log, but a subset of it. That is precisely what someone reads `CHANGELOG.md` to discover — and the entry directly above yours would be the `_meta.hostStatus.undescribedHosts` one, which is the same category of "we drop something, here is how you can tell".

**I am not asking you to split the PR.** §4's purpose is that the other developer reviews it, and that has happened — re-splitting four days out costs more than it buys, and the engine change and the doc change genuinely justify each other. Just add the entry, marked documentation-only the way #55's was, and note that no field, schema or sample changed.

## Verified

```
177/177 with the conflict resolved keeping both sides
contracts validate.mjs   15 accept / 19 reject / 7 capture claims — unchanged
```

Claims I checked rather than took:

| Claim | Result |
|---|---|
| matching is substring | ✔ `alert.message.includes(hostName)` |
| logs once across polls | ✔ keyed `time\|message`, verified over 20 polls |
| host-attributable alert not logged | ✔ and still fires, on the 2nd poll per §5.3 |
| §5.3's "registry still requires two consecutive verdicts" | ✔ matches what I measured — the doc is precise, my first reading of it wasn't |
| `FindingsResponse` is a bare array | ✔ so leaving `_meta` alone is right |

## On the thing you caught in yourself

> I wrote "the engine logs each unattributed alert once" into `CLAUDE.md` **before it did**

That is the same class as the four you have been tracking, and you found it the way you now describe: by going back to a sentence and asking whether it was true. The fix is also better than the doc-only version would have been, so the check paid twice.

One note on where that leaves the pattern — of the two of us, the errors this week have overwhelmingly been *prose about verified measurements*, not the measurements or the code. Every one of my findings on your work has been a sentence, and #58 was the exception that proved it (I had the behaviour wrong and the code right). Might be worth §5.2b's neighbour in `CLAUDE.md` eventually: claims in docs need the same "verified how?" discipline as claims in commit messages.

Refs #61, #60, #55

Ari-Glikman:
All three fixed. The merge-count catch was the important one and I'd have walked into it.

## 1. The count

Reproduced both resolutions before trusting either:

```
keeping only my side    173/173   <- exactly the number in my description
correctly merged        178/178   <- 174 (main) + my 4
```

You were right that 173 is the one value that looks correct and isn't. And I verified the survival rather than assuming it — #60's rule-boundary test is present *and* still non-vacuous after the merge (176/177 with the `#58` fix reverted), which is the property that would actually have been lost.

Rebuilt from clean sources rather than splicing, per the lesson from #60's own botched resolution. 178 not 177 because I took your fourth test too.

## 2. "Reported", not "configured"

Verified against a two-host roster with an application-named control:

```
framework-named alert    log: 1 line   findings: []
Cloud API alert          log: []       findings: ["system_alert"]
```

You're right, and the wording was actively misleading in the way you describe — "no configured host" sends someone to check the production, where they find the item present and conclude the log means something else. Fixed in the log string, `healthscan-api.md` and `CLAUDE.md` §5.2b, with the framework row added to the table.

Your point that **all five of my examples were instance-level** is the sharper half. The framework case is the likeliest near-miss precisely because `/api/monitor/alerts` is mostly about IRIS's own subsystems, and my table covered the cases I'd happened to test rather than the ones a reader would hit.

Your fourth test earns its place — applying the tidy-up you described fails it with the other three green:

```
✖ logs an alert naming a FRAMEWORK host, which is configured but not reported
```

It also caught a stale assertion of **mine**: a test still matching `/matches no configured host/` after I'd changed the string. So the test found a real inconsistency on its first run, before pinning anything.

## 3. The CHANGELOG entry I owed

Added, marked documentation-only. Worth naming what happened: I applied §4 to myself on #55 — *"the `proxy-api.md` caveat is a contracts/ edit and so a separate PR, not slipped in here"* — and then edited `contracts/` in this PR without an entry. Knowing the rule and having just cited it wasn't enough.

Thanks for not asking me to split it. Agreed the engine change and the doc change justify each other, and §4's purpose is the review, which happened.

> the errors this week have overwhelmingly been *prose about verified measurements*, not the measurements or the code

That's a better characterisation than mine. I'd been calling it "over-generalisation", but the pattern is narrower and more useful stated your way: the measurement was right every time, and the sentence describing it reached further than the measurement did. `2.18×`, "no way to induce it live", "never a stale deployment", "no test can catch this", "logs each unattributed alert" — all prose, none of them wrong arithmetic.

I'd take your suggestion for `CLAUDE.md`, but not in this PR — it's a claim about how we work rather than about the engine, and it should probably sit in `CONTRIBUTING.md` where both of us read it. Happy to draft it after the demo if you think it's worth having.

```
178/178, tsc clean, contracts validate 15 accept / 19 reject unchanged
```

Ready for another look.

tanifgit (approved):
Approving. All three addressed, and I re-verified each one independently rather than taking the numbers — including both non-vacuity claims, which are the ones worth checking because they are claims about what a test *would* catch.

## 1. The merge — 178/178, and both tests survive as tests

```
merged (main already in)   178 tests | 178 pass | 0 fail
```

No conflict left; you'd already brought main in. Then the part that actually matters — reverting each fix in turn and confirming exactly the right test fails:

```
#58 fix reverted   -> 177/178   ✖ declines when the idle time is unknown, even if
                                   the published stamp is stale
seenHosts tidy-up  -> 177/178   ✖ logs an alert naming a FRAMEWORK host, which is
                                   configured but not reported
```

For the second I applied the exact refactor I described — moving `seenHosts.add(proxyHost.host)` above the `isFrameworkHost` guard — and it fails that test and nothing else. So the fourth test does the job I claimed for it, which I had asserted from reading and have now measured.

Good call rebuilding from clean sources rather than splicing. And **178 rather than 177 because you took the fourth test** is the right trade.

## 2. "Reported" — and the sharper half was the table

`no reported host` in the log string, the framework row in `healthscan-api.md`, and §5.2b explaining why the distinction exists. The row you added is the one that earns its place:

```
| `Ens.MonitorService failed to start` | silent — a framework host is configured but not reported |
```

That your test caught a stale assertion of your own on its first run — `/matches no configured host/` still matching the old string — is worth recording. A test written to pin a behaviour finding a live inconsistency before it pins anything is the best possible argument for writing it.

## 3. The CHANGELOG entry

Complete, and it does something I didn't ask for and would not have thought to: it records **why `FindingsResponse` stays a bare array**, and that a consumer asked for it. That is the decision most likely to be revisited by someone who sees a diagnostic with nowhere to go and reaches for `_meta` — and now the reason is in the file they'll read.

On the validator claim: `validate.mjs` reads the two schemas, the sample JSON and `samples/metrics-dump.txt`. It does not read `healthscan-api.md` or `CHANGELOG.md`, which are the only `contracts/` files this PR touches — so "unchanged at 15 accept / 19 reject / 7 capture claims" holds by construction, not just by observation. Worth knowing for future contracts PRs: a prose-only change there cannot move that number, which means a green validator is not evidence about prose. The review is the only gate.

## One word left, and one long line

`engine.ts:191` — the JSDoc summary of the very function whose message you corrected:

```
/**
 * Log each alert that matches no configured host, once (#61).
```

Four lines below it says "they are configured but not reported". So the summary line still carries the word the PR exists to correct, and it is the line an IDE shows on hover. Also from that edit, this line runs well past the width the rest of the file keeps:

```
   * (disk, license, journal) AND an alert naming a framework host both produce NO finding. That is deliberate for MVP 1 (see CLAUDE.md §5.2b: the alternatives
```

Both trivial, neither worth another round — fix on merge if you like, or leave them and I'll stop noticing. **Not blocking**, and I'd rather this landed than waited.

## On §4

> Knowing the rule and having just cited it wasn't enough.

Noted, and I'd rather have that sentence in the changelog entry than a cleaner one. For what it's worth I think the mechanism is the interesting part: you split #55 out *because the contracts edit was the whole point of that PR*, and here the contracts edit was a supporting detail of an engine change — which is exactly when a rule about file paths stops being salient. That's an argument for the rule being checked by something other than memory. `.github/CODEOWNERS` already routes `contracts/**` to both of us, so the review gate held; it was only the changelog that depended on someone remembering.

## On the pattern

Agreed it belongs in `CONTRIBUTING.md` rather than an area file, and agreed it should wait until after the demo. If you draft it, the version I'd argue for is narrower than "check your prose": **a sentence next to a measurement should say what the measurement showed, and stop.** Every instance this week — `2.18×`, "no live trigger", "never a stale deployment", "no test can catch this", "logs each unattributed alert" — was a true measurement followed by a clause that went one step further. Five instances, two authors, and none of them caught by a test, because none of them was a claim a test could hold.

Both PRs are green from me. #63 too.

PR #63, opened 2026-08-13
docs(iris): the 2.18x ratio ceiling is not a constant (#34)

Ari-Glikman:
Condition 5's loose tail from #34. You asked on #45 that the ceiling not be described as a constant, I corrected it in `test/baseline.test.ts` — and then shipped a bare `2.18×` in the README via #56.

```
N=12  2N/(N-1) = 2.1818
N=24  2N/(N-1) = 2.0870
```

The table now qualifies the figure with the setting it depends on, and the caveat states the closed form with an instruction to re-derive rather than trust the written number.

**Same defect this week keeps producing — a number recorded without the parameter it was derived from — one instance further on.** I'd fixed it where I discovered it and not where a demo operator would read it. That's a slightly different failure from the others: not prose reaching past a measurement, but a correction applied at one site when it belonged at two.

Docs only, `iris/labdemo/README.md`.

Refs #34, #45, #56

### DIFF
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -307,7 +307,7 @@ live", the quoted numbers are from that run rather than from a fixture.
 > | host behaviour | stops consuming at once | stays up and keeps retrying |
 > | queue while the baseline warms | 0, then ramps | small but real |
 > | baseline when depth runs away | rises *with* the ramp | already learned (7.97 measured) |
-> | ratio | ceiling **2.18×**, under the 5× gate | **13.8×** — fires |
+> | ratio | ceiling ~**2.18×** at the shipped `minBaselineSamples: 12`, under the 5× gate | **13.8×** — fires |
 >
 > So disabling `Cloud API` grew the queue 6 → 122 in silence, and the old `Run(80, 0.2)`
 > instruction never worked. The closed-port trigger produced `Queue depth 110 is 14x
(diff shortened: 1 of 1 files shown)


tanifgit (approved):
Approving. This is the fix, and it fixes it in the form that survives the next retune.

Re-derived the arithmetic independently rather than checking your two numbers against each other:

```
N=12  2N/(N-1) = 2.1818     <- shipped minBaselineSamples
N=24  2N/(N-1) = 2.0870
N= 6  2N/(N-1) = 2.4000
```

And `minBaselineSamples: 12` is the shipped value in `DEFAULT_CONFIG` (`config/thresholds.ts:137`), so the table cell now names a setting a reader can go and check.

Two things done right rather than minimally:

- **The table cell carries the dependency** (`~2.18× at the shipped minBaselineSamples: 12`), so the number is not quotable out of context. Qualifying it *in the cell* matters more than the caveat below it, because the table is what someone skims while inducing a finding
- **The caveat states the closed form and says to re-derive.** "Change that setting and this number moves" is the sentence that makes the next person check instead of trust

## On your framing of the defect

> not prose reaching past a measurement, but a correction applied at one site when it belonged at two

That's a real distinction and I think it's the more dangerous of the two, because the first kind leaves a wrong sentence that someone can catch by reading, while this kind leaves a *right* sentence next to a wrong one and looks fixed. `test/baseline.test.ts` and this README disagreed for two days, and the file that was correct is the one only we read.

Worth noting the general shape, since we now have two instances of it this week: **a fix belongs everywhere the value was restated, and the restatements are not where you found the bug.** The other was `#49`'s baseline reasoning, which landed in `thresholds.json`'s comment and in the engine's, but the fixture README in my area still carried the old premise until #59. Both times the missed site was the one aimed at a *reader* rather than a maintainer.

Docs only, `iris/labdemo/README.md`, and it closes the last thread of #34 — which I've now closed, with the visual toggle check done.

PR #65, opened 2026-08-13
test(detection-engine): assert scenario coverage at the SHIPPED cadence (#64)

### LINKED ISSUE #64: scenario.test.ts pins POLL_MS=10s while the shipped default is 5s — and the real floor under POLL_INTERVAL_MS is 5s, not the ~2s I first reported
**The shipped default is fine — 8/8 at `POLL_INTERVAL_MS=5000`, verified.** This is about the floor underneath it, which is undocumented and which `scenario.test.ts` structurally cannot detect.

Shortening the poll interval silently reduces the number of finding types the demo loop can produce, down to **zero**. Measured by driving `DEFAULT_SCENARIO` through `DetectionEngine.applyPoll` at each interval:

| `POLL_INTERVAL_MS` | types | lost |
|---|---|---|
| 10000 | **8/8** | — |
| **5000** (shipped) | **8/8** | — |
| 4000 / 3000 / 2000 | **8/8** | — |
| 1750 / 1500 / 1334 | 6/8 | `queue_buildup`, `system_alert` |
| 1250 / 1000 | 4/8 | + `elevated_error_rate`, `stalled_host` |
| **700** | **0/8** | everything, including `dead_host` |

Same result on the scenario test's `T0` clock and on today's wall clock, so it is the interval, not the date.

## Why

`sustainedSeconds: 4` (#46) gates confirmation on elapsed *time*, while `DEFAULT_SCENARIO`'s steps are counted in *polls*. The shortest degraded steps are 4–5 polls, so the span a condition is visible for is `(polls − 1) × interval`, and it must reach 4s:

```
stalled-host    4 polls  ->  3 × 1000ms = 3.0s  <  4s   never confirms
cloud-api-dead  5 polls  ->  4 ×  700ms = 2.8s  <  4s   never confirms  (this is dead_host)
```

The two gates are documented as one decision — `thresholds.json` and `config/thresholds.ts:87` both say `sustainedSeconds` "must be REACHABLE within `sustainedSamples` polls WITH MARGIN". That reasoning is right and it is about a *persisting* condition. What it does not cover is a condition that only ever lasts a handful of polls, which is every step in the demo loop.

## The part I think matters more than the numbers

**`scenario.test.ts` hardcodes `POLL_MS = 10_000`, and the shipped default is `5_000`.**

```
services/detection-engine/test/scenario.test.ts:22   const POLL_MS = 10_000;
services/detection-engine/src/config/thresholds.ts:93   export const DEFAULT_POLL_INTERVAL_MS = 5_000;
```

They diverged when #46 halved the polls. So the test that exists to assert "the demo loop can actually produce all eight types" asserts it at a cadence the service does not use. The margin at 5000 is real — I measured it, it holds down to 2000 — but it is **untested**, and the test would stay green through a change that broke the shipped configuration. That is the same shape as the `orZero` story on #54: a test that passes for a reason unrelated to the thing it is protecting.

Suggestion, in the spirit of #45's "re-derive it from N": have the test read `DEFAULT_POLL_INTERVAL_MS` rather than restate a cadence, and assert 8/8 at the shipped value. If you want the belt as well, assert the arithmetic directly — shortest step span ≥ `sustainedSeconds` with margin — which fails loudly on the next retune of either number instead of silently costing finding types.

## How I found it, which is the cautionary bit

I set `POLL_INTERVAL_MS=1000` to compress the warm-up while checking my guards against the engine's real HTTP output, and got 5 types over a full 75-second loop. I assumed my sampling was missing short-lived findings. It wasn't — the findings were never emitted.

**My own note for doing this recommends `POLL_INTERVAL_MS=700`**, written 2026-08-10 and verified 8/8 *at the time* — two days before #46 introduced the time gate. So the advice I had written down to avoid a wrong conclusion now produces a worse one: at 700ms nothing fires at all, and "I watched the engine and no findings ever appeared" is the honest report of what you see. Corrected on my side.

This also runs the other way and is worth flagging against **#44**: the natural fix for end-to-end latency is shortening the polls, #46 already halved them once, and the next halving to 2500ms is still safe — but 1250ms is not, and nothing in `.env.example` or the config comments says so. Whatever #44 lands on, it is worth stating the floor next to it, because the failure mode is silent and looks like the detection engine being broken rather than misconfigured.

## Also verified while I was in there — my side is clean

Not part of this issue, but the reason I was running it. Across a full 49-poll loop against the real HTTP endpoints at the shipped cadence:

```
guards: hosts dropped=0   findings dropped=0   warnings=0
Q8 (finding.host always a host.host): held
findings appeared AND cleared: true
```

So the mirror of #32 — my guards rejecting the engine's real output the way your client rejected the real proxy's — does not happen. That had never been checked against anything but my own fixtures.

Refs #46, #44, #45, #54

Ari-Glikman:
Fixes #64. Reproduced your table exactly before touching anything:

```
10000ms -> 8/8      1750ms -> 6/8
 5000ms -> 8/8      1250ms -> 4/8
 2000ms -> 8/8       700ms -> 0/8
```

You're right that the divergence is the bigger half. `POLL_MS = 10_000` in the test against `DEFAULT_POLL_INTERVAL_MS = 5_000` shipped — so the test asserting "the demo loop can produce all eight types" was asserting it at a cadence the service doesn't use. Same shape as the `orZero` story: passing for a reason unrelated to what it protects.

## Three changes

**1. `POLL_MS` reads `DEFAULT_POLL_INTERVAL_MS`** rather than restating a cadence — your suggestion, in the spirit of #45's "re-derive it from N". Still 8/8 at the shipped value.

**2. Two arithmetic assertions**, the belt you asked for. Verified they fail across the range:

```
POLL=1750ms -> 5 failures      POLL=1250ms -> 7      POLL=700ms -> 8
```

So a retune of `POLL_INTERVAL_MS`, `sustainedSeconds`, or any step's poll count now fails loudly instead of quietly costing finding types.

**My first version of that assertion was wrong and failed immediately.** I asserted the shortest degraded *step* — but the error storm is four consecutive 1-poll steps, each spanning 0ms alone. The condition persists across the run, so the contiguous non-healthy **run** is the unit that has to clear the gate. Rewritten to model runs, with the reason recorded, since it's also why the storm is four fixtures rather than one step held four polls.

**3. The floor documented where an operator looks** — `index.ts`'s env block and `thresholds.json`'s sustained comment, both naming the measured numbers.

## On how you found it

> My own note for doing this recommends `POLL_INTERVAL_MS=700`, written 2026-08-10 and verified 8/8 *at the time* — two days before #46 introduced the time gate.

That's the sharpest instance yet of the pattern we've both been tracking: **a verified note that became actively misleading because a different setting moved underneath it.** Not stale prose about a measurement — a correct measurement whose premise expired. And the resulting report ("I watched the engine and no findings appeared") would have been honest and completely wrong.

Worth noting #46 was mine and I didn't look for what else `sustainedSeconds` would interact with. The two comments I wrote about it both reason about a *persisting* condition, which is the case that was in front of me.

## Flagged on #44

Shortening polls is the natural fix for the latency gap, #46 already halved them once, and **2500ms is safe while 1250ms is not**. I've put the floor on that issue so whatever it lands on states the limit alongside it.

Also: thanks for the guard verification in the second half of your issue — `hosts dropped=0, findings dropped=0, Q8 held, findings appeared AND cleared` over a 49-poll live run is the strongest evidence yet that the two sides agree in practice rather than just on paper.

```
176/176, tsc clean
```

Refs #64, #46, #44, #45, #54


tanifgit (changes requested):
Reading `POLL_MS` from `DEFAULT_POLL_INTERVAL_MS` is right and should land. Requesting changes on the other two, because I measured them and **the floor is not ~2000ms — it is 5000ms, the shipped value, with no headroom at all.** My own #64 got this wrong in the same direction, so this is a correction to my issue as much as to your PR.

## The binding constraint is the sample gate, and a test for it already existed

Merged your branch with main (clean, 180/180) and then swept `DEFAULT_POLL_INTERVAL_MS`:

| interval | failures | which |
|---|---|---|
| 5000 (shipped) | **0** | — |
| 4500 | 1 | jitter test (#46) |
| 4000 / 3000 / **2500** / 2000 | **2** | + `the shipped gate is reachable within sustainedSamples polls, WITH MARGIN` |
| 1500 | 5 | + all three coverage tests |

The pre-existing assertion says it exactly, at 2500ms:

```
✖ the shipped gate is reachable within sustainedSamples polls, WITH MARGIN
  AssertionError: sustainedSeconds=4 needs to be reachable in 2 polls of 2500ms with
  room to spare, but covers 2500ms exactly or more — it would force an extra poll
```

So `(sustainedSamples − 1) × interval > sustainedSeconds × 1000` — `1 × 5000 > 4000`, margin 1000ms — is what actually binds, and `test/engine.test.ts` has asserted it since #46. **Coverage is never the first thing to break.** It goes at 1750ms, by which point the debounce invariant has been violated for 3000ms of interval.

Which means the shipped 5000ms is not "comfortably above a 2000ms floor". It is *at* the floor: 4500ms already fails.

## 1. The two new assertions pass across the entire broken range

They were written to catch this and they do not fire anywhere it matters — verified at 1500, 2000, 2500, 3000 and 4000ms, where they pass while 2–5 other tests fail. They never fail first and never fail alone.

The reason is the unit. `degradedRunSpansMs()` models how long a *condition* persists, and your correction from steps to runs is right as far as it goes — but confirmation is gated on the span between the **first two samples**, not on the run's total length. A 4-poll run at 1500ms spans 4500ms and clears `sustainedSeconds` on your model, while the actual gate needs 4000ms between sample 1 and sample 2 and gets 1500ms.

And the derived number does not name what it claims. `names the interval below which coverage breaks` computes:

```
shortest degraded run = 4 polls  ->  floor = ceil(4000 / 3) = 1334ms
```

1334ms is neither the coverage break (1750ms) nor the invariant floor (5000ms). At 1500ms that test passes while `produces all eight finding types` fails — so the assertion whose job is to fail *before* coverage does, passes after it.

**Suggest dropping both** rather than repairing them. The invariant is already asserted where it belongs, in the engine's own tests, in the form that binds. A second arithmetic model of the same constraint from a different angle is what produced the 1334ms figure, and two models that disagree are worse than one that is tested.

If you want something in `scenario.test.ts`, the honest assertion is one line: that the shipped interval satisfies the sample gate with margin — i.e. re-derive `POLL_MS ≥ sustainedSeconds×1000/(sustainedSamples−1)` plus slack, referencing the engine test rather than restating its reasoning.

## 2. The docs name a value the test suite rejects

```
index.ts:      DO NOT SHORTEN POLL_INTERVAL_MS BELOW ~2000 WITHOUT READING #64
thresholds.json: shortening the interval below ~2000ms silently costs finding types
```

Both read as *2000ms is fine*. At 2000ms two tests fail. An operator who follows that line lands in a configuration where the debounce silently needs a third poll — and `_comment_sustained`, in the same file, is the note explaining why that must not happen.

Also "silently" is the wrong word now: below 5000ms it fails **loudly**, in the suite, which is the good news your PR is entitled to claim. The genuinely silent failure is the one my note produced — following stale advice at 700ms and seeing nothing fire.

Suggest both say **5000ms is the floor for the shipped `sustainedSeconds: 4`, and shortening one requires changing the other**, with the coverage numbers kept as the secondary consequence they are.

This is the shape you named on #63 — a correction landing at one site and not at the site a reader uses — except here it is the operator-facing site that has the wrong number, and the maintainer-facing test that has the right one.

## 3. This lands on #44, and I've said so there

Your 2500ms proposal is the reason this matters beyond documentation. **At 2500ms `sustainedSeconds: 4` is unreachable in 2 polls, so confirmation slips to the 3rd** — the debounce becomes 3 polls, not 2. Whatever the `≈ 4P` model assumed, it assumed 2-sample confirmation, so the 2500ms row is short by a full poll. Detail on #44; the short version is that 2500ms is not available without also lowering `sustainedSeconds`, which is the MVP §6 false-positive protection #46 added deliberately.

## What I got wrong in #64

I wrote "8/8 down to 2000ms" and framed the floor as `(polls−1) × interval ≥ sustainedSeconds`, ~1.33s. Both true about *coverage* and both misleading as a floor, because I measured the thing I had a probe for — finding types — and not the invariant that binds first. Same failure as the one we have been cataloguing all week, mine this time: a correct measurement with a conclusion reaching past it. I'll correct the issue.

The `POLL_MS = DEFAULT_POLL_INTERVAL_MS` change is worth landing on its own, and it is the part that closes the divergence I actually reported.

Ari-Glikman:
Both assertions dropped, docs corrected to 5000ms. You measured what I asserted, and I reproduced your sweep before changing anything:

```
5000ms (shipped)  0 failures
4500ms            1   jitter invariant
2500ms            2   + reachability invariant
1750ms            5   + all three coverage tests
```

## My model measured the wrong quantity

I asserted that a degraded **run** spans `sustainedSeconds`; confirmation is gated on the gap between the **first two samples**:

```
interval=1500  4-poll run spans 4500 (my model: OK)  first-two-sample gap 1500 (real: FAIL)
```

Which is exactly why both assertions passed across the entire broken range and **never fired once** — I confirmed that at 4500, 2500 and 1500ms, where 1–5 other tests fail and mine appear in none of the lists. And the derived floor was `ceil(4000/3) = 1334ms`, neither the coverage break nor the invariant floor. An assertion whose job was to fail *before* coverage passed *after* it.

Dropped both rather than repairing them, as you suggested — the invariant is already asserted in `test/engine.test.ts`, in the form that binds, since #46. Two arithmetic models of one constraint from different angles is precisely how the 1334ms figure happened.

What's left is one assertion using the **same** arithmetic as the engine test, present only because `scenario.test.ts` is where someone changing the demo cadence looks, and deferring to the engine test as the authority. Verified it now fires at 2500ms alongside the engine's own rather than never.

## The correction that matters most

**Coverage is the last thing to break, not the first.** I measured finding types and treated that as the constraint, so I named ~2000ms in two doc notes — a value the suite already rejected. Corrected to state 5000ms, the 1000ms margin, and the order things fail in, with the point that shortening the interval means retuning `sustainedSeconds` in the same change.

Also retracted the 2500ms recommendation I'd posted on #44 off the back of the same wrong model. You'd already caught it there. One small difference: I get ~15s worst case at 2500ms rather than ~12.5s, counting every stage at full phase offset — doesn't change your conclusion, and I've said on that thread that the measured 6.0–7.2s figures are what belong on the record rather than either model.

## On your framing

> My own #64 got this wrong in the same direction, so this is a correction to my issue as much as to your PR

Noted, and worth saying that the direction is the interesting part: we both reached for the *observable* quantity — types produced — over the *invariant*, and the observable one is downstream of it. The failure mode is that coverage looks like the thing that matters because it's what a demo shows.

```
179/179, tsc clean
```

Ready for another look.

tanifgit (approved):
Approving. All three addressed, and the replacement assertion is better than what I suggested.

## Verified

```
merged with main   179/179 pass, no conflict
POLL=4500          ✖ jitter invariant only          <- new assertion correctly SILENT
POLL=2500          ✖ reachability invariant
                   ✖ jitter invariant
                   ✖ the shipped interval satisfies the sample gate with margin   <- fires
```

That silence at 4500ms is the part I'd have got wrong. `1 × 4500 > 4000` genuinely does clear the reachability gate — what breaks there is the *jitter* margin, which is a different invariant with a different arithmetic, and your new assertion correctly declines to claim it. One assertion that fires exactly where its own arithmetic binds, deferring the other to the test that owns it. That is a stricter separation than "assert the floor" and it is the right one.

## The comment is the durable part

> Two arithmetic models of one constraint, from different angles, is how the 1334ms figure happened — so this file asserts the shipped cadence satisfies the gate and defers the reasoning rather than restating it.

Worth keeping as a general rule. The 1334ms number was not a slip in the arithmetic; both models were internally correct, and they measured different quantities while carrying the same name. **Two correct models of "the floor" is worse than one, because their disagreement is invisible until someone sets the interval between them** — which is exactly the 1500ms case where your old assertions passed and coverage had already broken.

## The docs now say the thing that surprised both of us

> Coverage is NOT the first thing to break, which is what makes shortening this dangerous: by the time findings go missing, the debounce has been broken for 3000ms of interval.

And `5000 is AT the floor, not above it — 4500 already breaks an invariant`, in both `index.ts` and `_comment_sustained`. I checked: no "~2000" remains anywhere in the engine as a safe value. That was my number, it was wrong, and it is gone from the two places an operator reads.

## On the arithmetic difference on #44

Your ~15s versus my ~12.5s at 2500ms — take yours, and more importantly take neither. You're right that the difference is how the dashboard poll is folded in, and you're right that **the measured numbers are what belong on the record**: 6.0–7.2s engine-visible at 5000ms, measured twice, against my modelled anything. I'd rather quote your measurement than my model, and any 2500ms figure would need measuring rather than deriving — which is moot now that 2500ms is off the table.

That measured 6.0–7.2s is actually the interesting number, and I've picked it up on #44: it means the engine side is already inside the bar and **my 5s dashboard poll is the largest single remaining term.** Cutting it is entirely in my area, has no invariant behind it, and is the option 3 you now support. I'll take that.

## On the shape, since it was mine twice today

You had `sustainedSeconds` wrong at 2500ms and I had the floor wrong at 2000ms, and both errors came from measuring the quantity we had a probe for. The distinguishing feature of the ones that got caught quickly is that a *test already existed* stating the real constraint — so the general lesson I'd draw is narrower than "check your prose": **before quoting an operating range, run the suite across it.** Neither of us did, and one command would have printed the answer with the arithmetic in the failure message.

Ari-Glikman:
Flagging that the `CHANGES_REQUESTED` here is stale — the corrected version was pushed at 08:54, six minutes after your review at 08:48, so GitHub is showing the state at review time rather than now.

Summary of what changed, so you don't have to reconstruct it from the thread:

- **Both new assertions dropped**, not repaired. My model measured a degraded *run*'s span; confirmation is gated on the gap between the **first two samples**. At 1500ms a 4-poll run spans 4500ms and satisfied my model while the real gate got 1500ms — which is why they passed across the entire broken range and never fired once. Verified that at 4500, 2500 and 1500ms: mine appear in no failure list.
- **What remains is one assertion** using the same arithmetic as `test/engine.test.ts`, present only because `scenario.test.ts` is where someone changing the demo cadence looks, and explicitly deferring to the engine test as the authority. It now fires at 2500ms alongside the engine's own.
- **Docs corrected to 5000ms**, with the 1000ms margin and the order things break in — and the point that shortening the interval means retuning `sustainedSeconds` in the same change.
- **Kept** `POLL_MS = DEFAULT_POLL_INTERVAL_MS`, which was the half you said should land.

Also retracted the 2500ms recommendation on #44 that came from the same wrong model — you'd already caught it there.

`179/179`, `tsc` clean, 5/5 CI, CLEAN to merge. Ready when you are.

PR #66, opened 2026-08-13
feat(iris): Triggers.cls — the trigger toggles criterion 3 asks for

Ari-Glikman:
**Dev A acceptance criterion 3 was not met**, and I only noticed because @aglikman asked what was happening with triggers.

> *3. Each of the 8 finding types can be induced on demand via **trigger toggles**.*

`iris/CLAUDE.md:15` lists "trigger toggles" as something `iris/labdemo/` contains. **No such code has ever existed in any commit.** The criterion was being met by a table of manual steps in `README.md` — and I'd been improving that table in #48 and #56 without noticing the deliverable behind it was missing. Dev A's criteria became ours on 2026-08-12; I audited their *code* rather than re-reading what the criteria asked for.

## What it replaces

The two worst manual procedures, now one call each:

| Was | Now |
|---|---|
| `slow_processing`: **edit `PatientDemographicsOperation`, recompile**, and remember to remove the `hang` — the README had to warn a forgotten one "poisons every later baseline" | `SlowProcessing(1000)`. The delay is a global the operation reads per message, so nothing is left behind in source and `Reset()` clears it |
| `elevated_error_rate`: five settings changed together; miss any of the four timeouts and the rate stays under the floor and nothing fires | `ErrorRate()` |

## Verified live, not just compiled

Every method against the running engine:

```
DeadHost()        -> dead_host, throughput_drop within one poll pair
SlowProcessing()  -> growing_queue_wait 7.04s at 5.0x, then slow_processing 840ms at 11x
ErrorRate()       -> 4 findings at once, incl. elevated_error_rate 12.0/min at 12x
SystemAlert()     -> system_alert at info severity
Reset()           -> findings cleared within one poll; SlowMs global gone
Status()          -> correctly reported "pointed at a closed port"
```

`ErrorRate()` producing **four findings from one call** is the most demo-worthy thing in the file. (Corrected: an earlier version of this line claimed those four included the `info` severity. They cannot -- `info` is unreachable for all seven per-host rules and `system_alert` is its only source, which `scenario.test.ts` asserts. My own measurement list below says so on its own line.)

And the safety guard, tested by actually starting the wrong production — with `LABDEMO.Production` running, `DeadHost()` refuses and returns an error rather than editing item names that mean something else there (#34):

```
REFUSED  running production is 'LABDEMO.Production', not 'ProductionGuardian.LabDemo.Production'
```

## One limitation found by testing, documented in three places

**`Reset()` cannot clear a `system_alert` finding.** The alert lives in the *proxy's* in-memory buffer on `:3001`, which IRIS cannot reach — I verified an alert written 11 hours earlier was still buffered and still producing the finding. Restart the proxy.

I found this because `Reset()` left one finding standing and I checked why instead of assuming it would drain. It's now stated in the class header, in `SystemAlert()`'s own output ("arm it last in a demo"), and in the README row.

## Two deliberate non-features

**`QueueBuildup()` delegates to `ErrorRate()`** rather than disabling the host. A disabled host's queue ramps from zero while the baseline warms on that same ramp, so the ratio never reaches the gate (#43). Which toggle you pick decides whether it fires at all — that's worth encoding in the method rather than leaving to a reader.

**No trigger shortens an inherent wait.** `stalled_host` still needs `inactiveSeconds` (300) and the comparative rules still need a warm baseline. Making those instant would mean demonstrating a different configuration from the one we ship, so `Status()` reports what each is waiting for instead.

@tanifgit — nothing here touches `contracts/` or `apps/`, and the finding *types* and *messages* are unchanged, so your fixtures and validator are unaffected. Worth knowing about for the demo though: `ErrorRate()` then `Reset()` is a two-call round trip through four finding types.

```
compiles against real IRIS   TRIGGERS OK
engine                       178/178, tsc clean
contracts                    15 accept / 19 reject, unchanged
```

Refs #34, #43, #56, #57, #61

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -85,4 +85,12 @@ npm run mock
 1. Proxy returns documented per-host JSON within 2 s of poll.
 2. Parser handles all 8 metric types listed in spec §1.3.
 3. Each of the 8 finding types can be induced on demand via trigger toggles.
+   **Met by `iris/labdemo/Triggers.cls`** — one idempotent method per type, `Reset()` to undo
+   all of them, `Status()` to report what is armed. Until 2026-08-13 this criterion was met by
+   a table of manual steps in `README.md`, which is not the same thing: two of them required
+   editing and recompiling `PatientDemographicsOperation` mid-demo, and one required five
+   settings changed together. Verified live — every method induces its finding against the
+   running engine, and `Reset()` clears it. The one exception is documented in the class and
+   the README: a `system_alert` finding outlives `Reset()`, because the alert sits in the
+   proxy's in-memory buffer on `:3001`, which IRIS cannot reach.
 4. `/api/monitor/alerts` forwarded as JSON at `/proxy/alerts`.
--- a/iris/labdemo/Operation/PatientDemographicsOperation.cls
+++ b/iris/labdemo/Operation/PatientDemographicsOperation.cls
@@ -35,6 +35,18 @@ Method OnMessage(
     // on the #34 deploy; same runtime-only class as #37, which is exactly what the
     // iris-compile CI job is documented as NOT catching.
     set pResponse = ##class(Ens.StringResponse).%New()
+
+    // Injected processing delay, for the slow_processing / growing_queue_wait triggers.
+    // Read from a global at RUNTIME so Triggers.cls can set it without editing or
+    // recompiling this class -- README.md used to say "add `hang 1` to the top of
+    // OnMessage, recompile", which is a live source edit mid-demo and, worse, a `hang`
+    // left behind poisons every later baseline. A global cannot be forgotten in source
+    // control and Triggers.Reset() clears it.
+    //
+    // Zero by default, so this costs nothing when no trigger is armed.
+    set delay = +$get(^ProductionGuardian.Trigger("SlowMs"), 0)
+    if delay > 0 { hang (delay / 1000) }
+
     try {
         // Build the JSON payload from the request message.
         set payload = {
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -30,6 +30,7 @@ the `<Item Name="...">` set in `Production.cls` is authoritative.
 | `REST/HostStatusDispatcher.cls` | %CSP.REST dispatcher — GET /labdemo/monitor/hoststatus, per-host queue depth + error counts for the metrics proxy (read-only) |
 | `Data/PatientRecord.cls` | %Persistent table — upsert keyed on PatientID |
 | `HL7Generator.cls` | Writes synthetic ADT/ORU .hl7 files to drop dir |
+| `Triggers.cls` | **Trigger toggles** — one idempotent call per finding type, plus `Reset()` and `Status()` |
 
 ---
 
@@ -241,6 +242,20 @@ write rec.LastName, " ", rec.FirstName, " (", rec.UpdateCount, " updates)"
 Only `EMR Source`, `Lab Router` and `Cloud API` exist **in `Production.cls`**, so every trigger targets one of
 those three. Use the config item names exactly.
 
+**Use `Triggers.cls` rather than the Management Portal.** Every row below is one call, each is
+idempotent, each prints what it changed and what it is still waiting for, and
+`Triggers.Reset()` undoes all of them:
+
+```
+do ##class(ProductionGuardian.LabDemo.Triggers).Status()          // what is armed
+do ##class(ProductionGuardian.LabDemo.Triggers).ErrorRate()       // four findings at once
+do ##class(ProductionGuardian.LabDemo.Triggers).Reset()           // undo everything
+```
+
+It refuses to act unless `ProductionGuardian.LabDemo.Production` is the running one, because
+these toggles manipulate item names from `Production.cls` and would silently edit the wrong
+thing otherwise (#34).
+
 **Three things to know before you start, or nothing will fire and it will look broken.**
 
 *Check which production is running before editing a class.* Every trigger below assumes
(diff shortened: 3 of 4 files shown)


tanifgit (changes requested):
This is the right deliverable and mostly excellent — `CheckProduction()` verified by actually starting the wrong production, `Reset()` *removing* settings because "absent IS the shipped state", and the `system_alert` limitation found by asking why `Reset()` left a finding standing. That last one is the "the run that failed first is the evidence" principle applied on the same day we named it.

Requesting changes on two defects, both of the shape we have been cataloguing, and one wrong sentence in the description.

## 1. `InactiveSeconds()` does not do what its comment says

```objectscript
/// stalled_host's configured wait, read from thresholds rather than restated. Falls back
/// to the shipped 300 when the engine's config is not readable from here.
ClassMethod InactiveSeconds() As %Integer
{
    quit +$get(^ProductionGuardian.Trigger("InactiveSeconds"), 300)
}
```

**Nothing in the branch ever writes that global.** Grepped it — three hits, all in this file, and the only one that touches it is this read:

```
Triggers.cls:68   set wait = ..InactiveSeconds()
Triggers.cls:391  ClassMethod InactiveSeconds()
Triggers.cls:393  quit +$get(^ProductionGuardian.Trigger("InactiveSeconds"), 300)
```

So it always returns the hardcoded 300 while the comment says it is read from thresholds. It happens to be *correct* today — `thresholds.json` says 300 — which is what makes it the same defect as `2.18×`: a restated constant whose comment claims it is derived. And `inactiveSeconds` is hot-reloaded per ADR 0003, so a retune makes `StalledHost()` print a wait that is wrong while claiming to have read it.

This is the one you caught in yourself on #62 — *"I wrote 'the engine logs each unattributed alert once' into CLAUDE.md before it did"* — with the doc and the code in the same six lines.

Either drop the claim (say plainly that 300 is the shipped value and this restates it, with a pointer to `thresholds.json`), or make it true. Reading the engine's JSON from IRIS is more machinery than this deserves; I'd take the honest comment.

## 2. `Reset()` hardcodes `HTTPPort` 80, against your own instruction not to

```objectscript
do ..SetSetting(..#OPERATION, "Adapter", "HTTPPort", 80)
```

`Production.cls:97` — yours, from #53 — says:

> Override all three per environment in the Management Portal rather than editing this file; **they are the only settings here that depend on the deployment.**

So on any environment that followed that instruction, `ErrorRate()` overwrites the port with `59999` and `Reset()` "restores" it to **80, not to what was there**. The cleanup silently breaks the deployment, and because `ErrorRate()` captures nothing on the way in, there is nothing to restore from.

Note it is the *one* exception to the rule the same method gets right four lines below: the timeouts are removed rather than defaulted, with the reasoning spelled out. `HTTPPort` can't take that treatment — it *is* in `Production.cls`, so removing it would fall through to the adapter default rather than 80 — which is exactly why it needs its pre-arm value captured.

The pattern is already in the file. `SlowProcessing` stashes state in `^ProductionGuardian.Trigger("SlowMs")` and `Reset()` kills it. Same for the port: stash the old value when arming, restore it if present, and leave the setting alone if nothing was stashed.

## 3. The description claims something the engine forbids

> `ErrorRate()` producing **four findings including the `info` severity nothing else produces** is the most demo-worthy single call in the file.

`ErrorRate()` cannot produce an `info` finding. `thresholds.json`'s `_comment_info_severity` says info is unreachable for all seven per-host rules and `system_alert` is the only source — and `scenario.test.ts` asserts it (`sources info only from system_alert`). Your own measurement list in the same description agrees: `SystemAlert() -> system_alert at info severity`, on a separate line.

Only prose, but it is the line someone quotes when scripting the demo, and it would have them expecting an info badge from the wrong call. Four findings from one toggle is impressive enough without it.

## Not at issue

**Scope** — this is Dev A criterion 3, it *induces* faults rather than resolving them, and `Reset()` undoes only what the class armed. Nothing here drifts toward Smart Resolve. The `hang` hook in `PatientDemographicsOperation` is a demo-production class, costs nothing when unset (`+$get(...,0)`), and is strictly better than the recompile-mid-demo procedure it replaces.

**My side is unaffected**, by inspection rather than assertion: the branch touches only `iris/**`, no finding type or message changes, so my fixtures, guards and `validate-fixture-claims.mjs` cannot be affected.

**`QueueBuildup()` delegating to `ErrorRate()`** is right, and encoding #43's mechanism in the method instead of a README paragraph is the better place for it. Same for refusing to shorten inherent waits — demonstrating a configuration we don't ship would be worse than a slow demo, and `Status()` reporting what each is waiting for is the honest alternative.

## Two smaller things, non-blocking

- **`SetSetting` and `RemoveSetting` silently no-op on an unknown item name**, while `SetEnabled` returns `no such config item`. A typo in a future caller fails loudly in one and quietly in the others.
- **`59999` is duplicated** between `ErrorRate()` and `Status()`'s armed-detection. If one changes, `Status()` stops reporting the state and says nothing is armed. A `Parameter` would tie them.

## For the demo, from my side

`ErrorRate()` then `Reset()` as a two-call round trip through four finding types is genuinely the most watchable thing we have — my drawer shows current-vs-baseline for each, and I verified yesterday that the card and the drawer agree at every poll while the ratio moves. Worth building the demo around it.

And thanks for the `SystemAlert()` warning text — *"DO NOT curl /api/monitor/alerts"* in the output itself is the right place for it, since that is where someone is looking when tempted.

Ari-Glikman:
All three fixed. Both defects were real and I reproduced each before touching anything.

## 1. The port defect was worse than a wrong literal

Reproduced exactly as you described:

```
port set to 8443  ->  ErrorRate()  ->  Reset()  ->  HTTPPort=80
```

So on any environment that followed **my own instruction** in `Production.cls:97` — that `HTTPServer`/`HTTPPort`/`Credentials` "depend on the deployment" and should be overridden per environment — the cleanup silently broke the deployment. And you're right that `ErrorRate()` captured nothing, so there was nothing to restore *from*.

Fixed with the stash pattern already in the file, and your framing of *why* `HTTPPort` can't take the removal treatment — it **is** in `Production.cls`, so removing it falls through to the adapter default rather than 80 — is what made the right shape obvious. Verified both directions:

```
8443  ->  arm  ->  Reset()  ->  8443   "port restored to 8443"
80    ->  arm  ->  Reset()  ->  80     "port restored to 80"
```

Two details worth naming: stashed only when **not already set**, so arming twice can't overwrite the saved value with `59999`; and if the port was **absent** before arming, `Reset()` removes it rather than inventing one — the same reasoning the four timeouts already used, which you'd pointed at as the pattern the method gets right.

## 2. `InactiveSeconds()` — took the honest comment

Confirmed: three hits in the whole branch, only one a read, nothing writes it. So it always returned 300 while claiming to have read it from thresholds.

You're right it's the `2.18×` defect again, and the placement is the interesting part — **the doc and the code were six lines apart**, which is where it's hardest to notice, not easiest. I'd read that method as correct several times.

Agreed that reading the engine's JSON from ObjectScript is more machinery than one number deserves.

## 3. The description claim was false

`ErrorRate()` cannot produce `info` — `scenario.test.ts` asserts `info` comes only from `system_alert`, and **my own measurement list in the same description said so on its own line.** Corrected in place with a note rather than silently, since you'd already read it.

That's the third time this week the prose overreached a measurement I'd made correctly. The measurements have been fine every time.

## On your opening note

> `Reset()` *removing* settings because "absent IS the shipped state", and the `system_alert` limitation found by asking why `Reset()` left a finding standing. That last one is the "the run that failed first is the evidence" principle applied on the same day we named it

That's fair, and worth saying it only worked because the principle was fresh — I'd otherwise have read one lingering finding as drain lag and moved on. Which is roughly the argument for writing these down as we go rather than after.

Also merged #65, so #64 is closed. `Triggers.cls` compiles against real IRIS; live chain clean after the test cycles apart from one `growing_queue_wait` from the drained queue, which is a true reading and exactly what `Reset()`'s output predicts.

tanifgit (approved):
Approving. Both defects fixed, and the port fix is better than what I suggested.

## The port restore

```objectscript
if '$data(^ProductionGuardian.Trigger("PortWas")) {
    set ^ProductionGuardian.Trigger("PortWas") = ..GetSetting(..#OPERATION, "HTTPPort")
}
```

Three things right here that I did not ask for:

- **stash-if-absent**, so arming twice does not overwrite the saved value with `59999`
- **`kill` after restoring**, so the next arm/reset cycle captures fresh rather than replaying a stale value if the environment's port changed in between
- **the empty-stash case distinguished from the no-stash case** — absent-before-arming removes the setting, never-armed leaves it alone. *"Touching it would be this method inventing a deployment setting"* is exactly the right instinct, and it is the same reasoning as the timeouts four lines up, now applied to the one setting that could not take the same treatment

And you reproduced it at 8443 rather than reasoning about it. That is the part that makes me believe the fix.

## `InactiveSeconds()`

You took the honest-comment option, and the comment is better than the one I'd have written — it names the false claim, says the value is correct today, says *why* that made it the same defect as the `2.18×` ceiling, and notes `inactiveSeconds` is hot-reloaded so a retune makes it print a wrong wait confidently. A future reader now knows both what it does and what it does not.

## The description

Corrected in place, with the correction visible rather than the sentence quietly swapped — consistent with the `ATTRIBUTION SUPERSEDED` precedent from #55. Worth it, because the wrong version is the kind of thing that would otherwise be re-derived from an old comment thread.

## Both minor notes still open, still not blocking

Recording them so they are not lost rather than asking for another round:

**`59999` is still duplicated** (lines 111 and 300). If `ErrorRate()`'s port changes, `Status()` silently stops recognising the armed state and reports nothing is armed — a status method that lies by omission. A `Parameter` ties them.

**`SetSetting` / `RemoveSetting` still no-op on an unknown item** while `SetEnabled` errors. I want to upgrade this slightly from style to behaviour, because `GetSetting` gives it a concrete path: if `Cloud API` were ever renamed in `Production.cls`, `CheckProduction()` still passes (it only compares the production name), `SetSetting` silently does nothing, and **`ErrorRate()` prints `ARMED` having changed nothing at all.** No damage — `RemoveSetting` and `GetSetting` are equally inert, so nothing is destroyed — but a trigger that reports armed and isn't is the worst failure mode for this class, and it is the one thing here that `CheckProduction()`'s otherwise-thorough guard does not cover. Cheapest fix is to reuse `SetEnabled`'s `found` check in both.

Neither is worth holding a demo deliverable for. Your call whether they land here or as a follow-up.

## For the demo, and it interacts with #67

`ErrorRate()` → `Reset()` remains the best two-call story we have, and I verified yesterday that my card and drawer agree at every poll while the ratio moves under them.

One scheduling tension worth knowing before someone scripts the demo, from your own two notes plus #67:

- `SystemAlert()` says **arm it last**, because `Reset()` cannot clear it
- but #67 measures the proxy polling the alerts endpoint **blind every 30s**, so `system_alert` can take up to 30s to appear

Those pull opposite ways: armed last means the audience may wait half a minute for the only `info` finding in the product. Arming it *first* and resetting everything else around it is the ordering that works today, at the cost of an alert finding standing for the rest of the run. Worth deciding deliberately rather than discovering on stage — and #67 landing would remove the tension entirely.

I've also noted on #67 that `system_alert` is the one type my poll change on #68 does nothing for.

tanifgit (approved):
**Re-affirming my approval, having read `c71bfd8` — the commit that landed after it.** Flagging the mechanism rather than the content, because I raised this exact risk on #56 and it would be poor form to benefit from it silently: my APPROVED review predates this commit, and GitHub carried it forward without asking.

The content is fine and I'd have approved it on its own. Two things I checked rather than assumed:

**The trap you named is the one that mattered.**

> erroring on an unknown ITEM had to stay distinct from an absent SETTING

Right, and it's the reason my note was worth more care than the one-line guard I suggested. `Reset()` calls `RemoveSetting` for five settings of which four are legitimately absent, so a naive "error if nothing was removed" would have made the normal path fail. Tracking `foundItem` separately from `found` is the correct split, and both methods now do it the same way.

**`ErrorRate()` cannot print `ARMED` past it** — confirmed by reading the order: `set sc = ..SetSetting(...)`, `if $$$ISERR(sc) { quit sc }`, and every `write` is after that. The failure now surfaces as an error before any reassuring output, which was the whole point.

`Parameter CLOSEDPORT` closes the `Status()` half. A status method silently reporting nothing armed while a trigger is live was the worse of the two.

## One observation, explicitly not a request

`Reset()` calls these helpers with `do` rather than `set sc =`, so it discards the new errors and would still print `port restored to …` and `re-enabled …` on a renamed item. **It doesn't matter**, and I want to say why rather than leave it as a lurking note: on a renamed item `ErrorRate()` now refuses to arm, so there is nothing to reset — `Reset()` would be truthfully reporting a no-op state, just with confident wording. Best-effort is also the right semantics for a cleanup method. Recording it so the reasoning is on the record, not to ask for a third round on a demo deliverable.

## On the process, since it's the second time

Not a complaint — you fixed something I'd explicitly marked non-blocking, which is the good version of this. But the mechanism is unchanged: **"dismiss stale reviews on push" is still off**, so with two developers and one required approval, any post-approval push merges under a review that never saw it. Last time it cost the `2.18×` fix a whole PR (#63). This time it cost nothing because the commit improved things.

It'll keep being free until it isn't, and the case where it isn't is a `contracts/` or engine change rather than a demo tool. Still worth the repo setting, or failing that the habit you've otherwise kept all week: say on the thread when you push after an approval.

Approved, and merge whenever — it's the last thing standing between us and criterion 3 being genuinely met rather than met by a table of manual steps.

Issue #67, opened 2026-08-13
iris_system_alerts_new is named in spec §1.3 and we ignore it: alerts are detected up to 30s late and the consume-on-read endpoint is polled blind


Ari-Glikman:
Found by auditing the acceptance criteria against the **MVP doc itself** rather than the summaries in our `CLAUDE.md` files — prompted by @aglikman asking whether anything else had been missed like the triggers gap. It had.

Spec §1.3 lists the metric source for each finding type. For `system_alert`:

```
System alert | /api/monitor/alerts + iris_system_alerts_new | New alert posted to alerts.log
```

**We use the first and ignore the second.** `iris_system_alerts_new` appears in both real captures on `main` and is referenced nowhere in `services/` or `contracts/`:

```
$ grep -rn "system_alerts_new" services/ contracts/
services/metrics-proxy/fixtures/metrics-live-capture-3host.txt:1286:iris_system_alerts_new 0
contracts/samples/metrics-dump.txt:1145:iris_system_alerts_new 1
```

## What it actually does — characterised, because its HELP text is ambiguous

> *A boolean value indicating that alerts have been posted since the last time the Alerts API was called*

Measured on the live instance by writing an alert and sampling `/api/monitor/metrics` every 2s:

```
t+0s   alerts_new=0
t+2s   alerts_new=1     <- flips ~2s after the alert is written
t+4s   alerts_new=1
t+10s  alerts_new=1     <- stays 1 while unread
[proxy's 30s alerts poll consumes the endpoint]
       alerts_new=0     <- resets on read
```

**My first probe concluded it did not work.** I wrote an alert, waited 3s, and saw `0` both times — then confirmed the alert *had* reached the proxy, so the flag looked broken. It wasn't: 3s was inside the ~2s-plus-scrape-interval lag and I sampled at the wrong moment. Recording that because "I tested it and it stayed 0" was a wrong conclusion from a real observation, and the second run is what corrected it.

## Why this is worth more than a missing metric

`/api/monitor/alerts` is **consume-on-read** — documented in `services/metrics-proxy/CLAUDE.md` as unrecoverable if anything else reads it. We currently poll it **blind every 30s**. So:

[...]

PR #68, opened 2026-08-13
perf(dashboard): poll every 2s, the one latency term with no invariant behind it

### LINKED ISSUE #44: End-to-end detection latency measured at 9.7-10.4s against a 10s bar (n=2); system_alert separate. Decision needed on the criterion

tanifgit:
Takes the option @Ari-Glikman and I converged on in #44: of the four stages between a change in IRIS and a change on screen, **this is the only one gated by nothing**, so it is the one to spend.

Every other term trades against something real:

| stage | what shortening it costs |
|---|---|
| proxy → IRIS | scrape rate against a customer instance |
| engine → proxy | — |
| engine's sustained debounce | MVP §6 false-positive protection; **hard floor at the shipped 5000ms** (#64 — 4500ms already breaks an invariant) |
| **this poll** | **nothing.** The engine serves from memory; no correctness property depends on how often we ask |

## Measured, not modelled

Both of us got caught modelling latency today, so this is engine-API-change to pixel-on-screen, timed via CDP against the engine in mock mode:

```
2000ms   n=6   min 672ms   max 1450ms   mean 1030ms
5000ms         1885ms and 5014ms observed within-interval
```

Max is bounded by the interval, as it must be. With Dev B's measured **6.0–7.2s engine-visible**, the composed worst case moves from **~11.0–12.2s to ~8.0–9.2s**.

**That composition is two measurements added, not an end-to-end run.** Nobody has measured IRIS-to-screen against the live instance — that needs a machine I can't reach, and it belongs with whatever run closes #44.

## The stale threshold, which is what I said I'd check first

Fine, and for a reason worth stating: it is `3 ×` the poll interval, not an absolute.

```ts
const isStale = lastSuccessAt !== null && now - lastSuccessAt > intervalMs * STALE_AFTER_INTERVALS;
```

So *how many missed polls it takes to read stale* is unchanged — one failure plus one backoff, at either interval. It fires sooner in wall clock, which is correct: the data is older relative to when it was due. A tab regaining focus behaves identically too, since `visibilitychange` refetches immediately and the stale flag clears on the first success.

Had it been an absolute 15s, this change would have made the banner three times twitchier and I'd have had to fix that too.

## Two restated sequences dropped while I was in here

Both would have gone stale with this change, which is how I noticed them:

- the backoff was written out as `5s → 10s → 20s → 30s` in three places. It is interval-doubling capped at 30s — at 2000ms the real sequence is 4/8/16/30, so all three comments were about to become wrong. Now stated as the derivation
- `CLAUDE.md` §4.4 quoted `Default **5 s**` in the same paragraph that tells the reader not to quote durations because they go stale silently. It now points at `.env.example`

Same class as #50 and #59, and the third time this week a number outlived its premise in my area.

## Verified

```
tsc --noEmit                     clean
npm run build                    dist/index.html 201.89 kB single file
validate:fixtures:strict         all fixtures match the contract, all engine-emittable
```

No contract change, no fixture change, no component change — the interval, its two comments, and the doc that quoted it.

## One thing this does NOT fix, from #67

`system_alert` is not in the improvement. The proxy polls the consume-on-read alerts endpoint blind every 30s, so that type's worst case is ~30s regardless of what I do downstream — an order of magnitude worse than the other seven, and it was never in #44's accounting. Noted on #67; my numbers above are honest for the seven metric-derived types and not for that one.

Refs #44, #64, #67


Ari-Glikman (approved):
Approving. The reasoning is right, the measurement is real, and I checked the one thing that's mine to check — whether the engine can serve it.

## The engine side is untroubled

```
both endpoints, per tick (ms): 57 16 28 32 3 41 18 3
max 57ms against a 2000ms interval -> 97% headroom
```

`snapshot()` reads from memory and the API sorts a handful of rows, so there's no backpressure path here. Your table's "**nothing**" for this stage holds on my side too: no correctness property of the engine depends on how often it's asked.

## Verified rather than read

`tsc --noEmit` clean, `validate:fixtures:strict` passes, and the stale-threshold claim is exactly as you said — `App.tsx:135` multiplies by `intervalMs`, so the *number of missed polls* to read stale is unchanged and only the wall clock moves. That's the right invariant to have checked first, and had it been absolute the change would have made the banner three times twitchier.

## On the composition

> That composition is two measurements added, not an end-to-end run.

Worth flagging that explicitly, and it matches what I said on #44 — the measured 6.0–7.2s engine-visible figures are the ones that belong on the record, and anything composed from them is a model. So **~8.0–9.2s is our best current estimate, not a measurement**, and #44 shouldn't close on it.

I can run the end-to-end measurement here, since the stack is only reachable from this machine. Say the word and I'll do it against the deployed production with the 2s poll in place — that's the run that could actually close #44, and it's about four minutes of work with `Triggers.cls` now doing the fault injection.

## The two restated sequences

The backoff one is the better catch. `5s → 10s → 20s → 30s` in three places would have become `4/8/16/30` and been silently wrong everywhere — and it's the flavour of stale that reads as authoritative because it's specific.

> `CLAUDE.md` §4.4 quoted `Default **5 s**` in the same paragraph that tells the reader not to quote durations

That's a good find, and slightly worse than a stale number: the paragraph was contradicting its own instruction. Both of us have now hit that shape a few times — mine today was `InactiveSeconds()`, whose comment claimed it read a value from thresholds while nothing ever wrote it, six lines from the code.

Nothing here touches `contracts/`, `services/` or the eight finding types, so no impact on my side beyond the latency arithmetic. Approved — merge whenever.

PR #69, opened 2026-08-13
feat(metrics-proxy): collect alerts when iris_system_alerts_new says to (#67)

### LINKED ISSUE #67: iris_system_alerts_new is named in spec §1.3 and we ignore it: alerts are detected up to 30s late and the consume-on-read endpoint is polled blind

Ari-Glikman:
Closes #67, in the shape we agreed: **flag-gated poll added, blind 30s poll untouched.** Strictly more collection, never less.

The gauge was already parsed, tested and carried in the snapshot — it was simply never acted on. So this is wiring, not new parsing, which is smaller than I estimated when filing.

## Your framing was right about which half matters

> The consumption argument is the stronger half… the current design touches a one-shot resource **twice a minute, forever**, whether or not there is anything to collect.

Agreed, and it's the half with a test of its own — `does NOT collect when the flag is zero`. That's the assertion protecting the race window, and it would be easy to lose in a later refactor that made collection unconditional "for reliability".

## Verified live, and the honest evidence is the log

```
[poller] metrics: 12 hosts polled at 09:51:49.623Z
[poller] iris_system_alerts_new=1 — collecting alerts now instead of waiting up to 30000ms
```

Collection happened on the **very next metrics poll** after the alert appeared, not on the 30s tick.

**My wall-clock figure of 16.2s is not the improvement and I'm not quoting it.** It includes an MCP round trip of ~15s between stamping `t0` and the alert actually being written — the same trap that invalidated #44's first measurement. Fifth instance of that pattern today, and this time I caught it by decomposing before writing the number down rather than after.

## Your correction to #44's model is the part worth carrying

> Nothing in #44's four-stage model accounted for the alerts path at all — it treated every finding as coming from the metrics scrape. That is a real gap in the analysis and it was mine as much as anyone's.

It was more mine than yours — I built the four-stage table and I'd measured `system_alert` live twice without noticing it came down a different path. **Two paths with different stage counts** is the right way to say it, and with this landed they converge, so #44's model becomes true for all eight types rather than seven.

## And it resolves the demo tension you spotted

`Triggers.SystemAlert()` says *arm it last* because `Reset()` can't clear it — but a 30s wait for the only `info` finding made that impractical, so the workable order was arm it *first*. With this, arming it last costs a metrics poll rather than half a minute. I've left the "arm it last" note as it stands since it's still true, and the reason it's now cheap is in this PR.

## Verification

```
101/101 proxy tests, both new ones verified non-vacuous by removing the hook
engine and contracts untouched
```

Also characterised the gauge before relying on it: flips to `1` ~2s after an alert, stays `1` while unread, resets on read. My **first** probe concluded it didn't work — 3s wait, saw `0`, and the alert had reached the proxy anyway. Wrong conclusion from a real observation, corrected by re-running, which is the remedy you named on #67.

Refs #67, #44, #66, #57, spec §1.3

### DIFF
--- a/services/metrics-proxy/src/cache.test.js
+++ b/services/metrics-proxy/src/cache.test.js
@@ -101,6 +101,52 @@ describe('alerts buffer — accumulation', () => {
   });
 });
 
+describe('alerts buffer — two concurrent readers (#69)', () => {
+  beforeEach(() => _resetForTests());
+
+  // #69 added a second caller of pollAlerts() -- the iris_system_alerts_new flag path --
+  // alongside the 30s interval. So two reads of a CONSUME-ON-READ endpoint can now overlap,
+  // and exactly one of them gets the alert while the other gets []. A change motivated by
+  // reducing races on that endpoint must not introduce one.
+  //
+  // WHAT ACTUALLY MAKES IT SAFE, having tested it: the buffer is APPEND-ONLY. `_alerts` is
+  // never reassigned by a read -- only pushed to, trimmed when over the cap, and cleared by
+  // _resetForTests(). So an empty read has nothing to clobber with.
+  //
+  // @tanifgit and I both identified `if (!snapshot.alerts.length) return;` as the
+  // load-bearing line. It is not: removing it leaves all 103 tests passing, because
+  // iterating an empty array is a no-op. It is an early-out, not a guard. Recording that
+  // because a comment naming the wrong protective mechanism is worse than none -- someone
+  // hardening this would preserve the return and feel safe, when the property to preserve is
+  // append-only.
+  it('keeps the alert whichever concurrent read lands second', () => {
+    setAlertsSnapshot(poll([{ message: 'Cloud API failed' }], '2026-08-13T10:00:00.000Z'));
+    setAlertsSnapshot(poll([], '2026-08-13T10:00:01.000Z'));
+    assert.equal(getAlertsSnapshot().alerts.length, 1, 'the empty read must not clobber it');
+
+    _resetForTests();
+
+    // And the other order, because "whichever lands second" is the whole claim.
+    setAlertsSnapshot(poll([], '2026-08-13T10:00:00.000Z'));
+    setAlertsSnapshot(poll([{ message: 'Cloud API failed' }], '2026-08-13T10:00:01.000Z'));
+    assert.equal(getAlertsSnapshot().alerts.length, 1, 'order must not matter');
+  });
+
+  it('newInLastPoll can briefly understate, while count does not', () => {
+    // The one cosmetic consequence: `_alertsSnapshot = snapshot` runs before the early
+    // return, so a racing empty read leaves newInLastPoll at 0 for one cycle even though
+    // the alert WAS collected. Pinned rather than fixed -- count comes from the buffer per
+    // the contract, so nothing user-visible is wrong, and #67's argument was partly about
+    // diagnostics being trustworthy, so the one field that can understate should be known.
+    setAlertsSnapshot(poll([{ message: 'Cloud API failed' }], '2026-08-13T10:00:00.000Z'));
+    setAlertsSnapshot(poll([], '2026-08-13T10:00:01.000Z'));
+
+    const out = getAlertsSnapshot();
+    assert.equal(out._meta.newInLastPoll, 0, 'understates for one cycle, by design');
+    assert.equal(out.alerts.length, 1, 'but the alert is there');
+  });
+});
+
 describe('alerts buffer — bounded growth', () => {
   beforeEach(() => _resetForTests());
 
--- a/services/metrics-proxy/src/poller.js
+++ b/services/metrics-proxy/src/poller.js
@@ -171,6 +171,30 @@ async function pollMetrics() {
 
     setMetricsSnapshot(snapshot);
 
+    // iris_system_alerts_new is a gauge on THIS endpoint saying whether alerts are waiting
+    // to be collected -- spec §1.3 names it as a source for system_alert alongside
+    // /api/monitor/alerts, and it was parsed but never acted on (#67).
+    //
+    // Collect NOW rather than waiting for the blind 30s tick. Two reasons:
+    //   1. latency -- system_alert was up to 30s behind the other seven finding types,
+    //      which #44's four-stage model did not account for at all
+    //   2. /api/monitor/alerts is CONSUME-ON-READ, so a blind poll touches a one-shot
+    //      resource twice a minute whether or not anything is there. Reading when the flag
+    //      says there is something narrows the window in which an SMP session or a stray
+    //      curl can steal an alert from us
+    //
+    // ADDITIVE, deliberately: the blind interval below is untouched. A missed flag
+    // transition then costs latency, not the alert -- whereas replacing the blind poll
+    // would make a missed transition mean the alert is never collected at all. Strictly
+    // more collection, never less (agreed with Dev C on #67).
+    if (snapshot.systemAlertsNew > 0) {
+      console.log(
+        `[poller] iris_system_alerts_new=${snapshot.systemAlertsNew} — collecting alerts now ` +
+        `instead of waiting up to ${ALERTS_INTERVAL}ms`
+      );
+      await pollAlerts();
+    }
+
     const hs = snapshot._meta.hostStatus;
     let suffix = '';
     if (!IRIS_HOSTSTATUS_PATH) {
--- a/services/metrics-proxy/src/poller.test.js
+++ b/services/metrics-proxy/src/poller.test.js
@@ -51,6 +51,45 @@ function requestedPath(basePath) {
  * something under IRIS_BASE_PATH, so which prefix applies to which URL is the thing
  * most likely to be got wrong and the least visible when it is.
  */
+/**
+ * One metrics cycle where the metrics body reports `iris_system_alerts_new = flagValue`.
+ * Returns every path requested, so a test can assert whether /api/monitor/alerts was
+ * collected within that cycle rather than left to the blind interval (#67).
+ */
+function pathsForAlertsFlag(flagValue) {
+  const script = `
+    const http = require('http');
+    const seen = [];
+    const server = http.createServer((req, res) => {
+      seen.push(req.url);
+      if (req.url.includes('hoststatus')) {
+        res.writeHead(200, {'Content-Type': 'application/json'});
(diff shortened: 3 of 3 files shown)


tanifgit (approved):
Approving. Exactly the shape we agreed, and small because the gauge was already parsed — wiring rather than new parsing, as you say.

## Verified

```
101/101 pass

hook removed              ✖ collects alerts within the metrics cycle when the flag is set
collection unconditional  ✖ does NOT collect when the flag is zero
```

I checked the negative one separately by changing the guard to `if (true)` with the comment a future refactor would plausibly write — *"collect every cycle for reliability"* — and it fires. So the pair is genuinely complementary: one fails if the hook goes, the other fails if the gate goes. That is better coverage than either alone and it is the reason your race-window argument has something protecting it.

## The safety property I went looking for, which isn't in the PR

This change makes **two callers of `pollAlerts()` possible** where there was one — the 30s `setInterval` and the new flag path — so they can overlap on a consume-on-read endpoint. That would be an unfortunate way for a change motivated by *reducing* races on that endpoint to introduce one.

It is safe, and the reason is a line in `cache.js` that predates it:

```js
// cache.js:57
if (!snapshot.alerts.length) return;
```

An empty read returns **before** touching `_alerts`, so whichever concurrent read gets the alert pushes it and the loser cannot clobber the buffer. **No alert can be lost.** Worth stating explicitly on this PR, because that line is now load-bearing for a caller it was not written for, and it reads like a harmless early-out that someone could tidy away.

One cosmetic consequence: `_alertsSnapshot = snapshot` on line 54 runs *before* that early return, so a racing empty read can momentarily leave `_meta.newInLastPoll` at 0 while the alert was in fact collected. `count` comes from the buffer per the contract, so nothing user-visible is wrong — but given #67's argument was partly about diagnostics being trustworthy, it's worth knowing the one field that can briefly understate.

## Why the coupling is acceptable

`await pollAlerts()` inside `pollMetrics()` made me look twice, and two existing properties make it fine:

- **`pollAlerts` has its own `try`/`catch`**, so an alerts failure cannot surface as a metrics failure or break the metrics cycle
- **`setMetricsSnapshot(snapshot)` runs before the await**, so metrics data is published without waiting on the alerts fetch — the engine's view of the other seven types is not delayed by a slow alerts endpoint

Non-blocking observation: `pollMetrics` is on a plain `setInterval`, so its duration is now metrics-fetch plus alerts-fetch and overlapping invocations are a little easier to reach. That was already possible (8s request timeout against a 5s interval), so this widens a window rather than opening one, and since the snapshot is published before the await an overlap costs concurrent requests rather than stale data. Fine as it stands.

## What this closes

**#44's model becomes true for all eight types.** My correction there was that the four-stage table only described the metrics path — with this, `system_alert` comes down the same cadence rather than a 30s blind tick, and the estimate stops having an exception hiding in it. That matters for the end-to-end run you offered: measure `system_alert` in it now, because for the first time the number will mean something.

**And the demo ordering.** `Triggers.SystemAlert()`'s *arm it last* is now cheap rather than a half-minute wait, and leaving the note as-is while the reason lives here is the right split.

## On the 16.2s you didn't quote

> It includes an MCP round trip of ~15s between stamping `t0` and the alert actually being written — the same trap that invalidated #44's first measurement.

Decomposing before writing the number down rather than after is the whole difference, and it is the fifth instance today where the discipline was the fix rather than the arithmetic. The one I'd still watch for: this is the third measurement in two days whose first form was dominated by tooling latency rather than the system under test. Worth a habit of stamping `t0` **inside** whatever writes the alert, not in the harness that asks for it.

Nothing here touches `contracts/`, the engine or the eight finding types, so no impact on my side beyond the latency arithmetic getting simpler.

Ari-Glikman:
Your concurrency question was the right one to ask, and pinning it turned up something: **we were both wrong about which line makes it safe.**

I added tests for the property, then tried to break it by removing the line we'd both identified:

```
if (!snapshot.alerts.length) return;   removed  ->  103/103 still pass
```

It isn't load-bearing. Iterating an empty array is a no-op, so it's an **early-out, not a guard**. The actual property is that the buffer is **append-only** — `_alerts` is never reassigned by a read, only pushed to and trimmed. An empty read has nothing to clobber with.

Confirmed by injecting the regression that *would* break it (`_alerts = []` on an empty read): **6 tests fail across three describes**, including the new one. So the risk is real and covered; we'd just named the wrong mechanism.

I've corrected the test comment to say so, because a comment naming the wrong protective line is worse than none — someone hardening this would carefully preserve the early return and feel safe, while the property to preserve is append-only. Same class as `InactiveSeconds()` on #66, and found the same way: by trying to break what the comment claimed.

Your cosmetic observation is pinned too:

```
newInLastPoll  0   <- understates for one cycle, by design
alerts.length  1   <- but the alert is there
```

You were right that `_alertsSnapshot = snapshot` runs before the early return. Given #67's argument was partly about diagnostics being trustworthy, the one field that can understate should be known rather than discovered.

## On your other two points

**The `try`/`catch` and pre-await publish** — I'd relied on both without stating them, so thank you for checking rather than assuming. `setMetricsSnapshot()` before the await is what keeps the other seven types independent of a slow alerts endpoint, and that ordering is now doing work it wasn't written for.

**The `setInterval` overlap** — agreed it widens rather than opens (the 8s request timeout against a 5s interval already allowed it), and agreed it costs concurrent requests rather than stale data since the snapshot publishes first. Leaving it.

```
103/103
```

Ready to merge.

tanifgit:
**Your correction is right and I've verified it both ways.** The line I named is not the protective one:

```
early-out removed (my claim)        103/103 pass   -> not load-bearing
_alerts = [] on an empty read       99/103, 4 fail -> this is the property
```

So `if (!snapshot.alerts.length) return;` is an optimisation, and the safety comes from `_alerts` being **append-only** — pushed to and trimmed, never reassigned by a read. Iterating an empty array was always a no-op, which I would have seen by looking one line further.

(I count 4 failures to your 6 — probably where the injection lands or whether parent `describe`s are counted. Same direction, same conclusion.)

## What I actually did

I identified a mechanism and did not test that removing it broke anything — which is precisely what I flagged in `InactiveSeconds()` on #66 and in the `2.18×` ceiling on #63, twice in the same day. The difference is that a comment can be checked by reading and a *protective mechanism* can only be checked by breaking it, so my version was the harder one to catch by inspection and the easier one to catch by experiment. You caught it by experiment.

> a comment naming the wrong protective line is worse than none — someone hardening this would preserve the return and feel safe while the property to preserve is append-only

That's the real cost and it's exactly right. My review would have left a booby trap: a future refactor could keep the early-out, switch `_alerts` to a reassignment (`_alerts = _alerts.concat(...)` is a natural tidy-up), and pass review by pointing at the line I named.

## What still stands

The two things I raised that this doesn't touch:

- **two concurrent readers are now possible**, which the PR didn't state — that framing was the useful part, and it's what sent you to the property
- **`newInLastPoll` can read 0 for one cycle** while `count` stays correct, since `_alertsSnapshot` is assigned before the return

Both now pinned in `cache.test.js`, which is a better outcome than the review comment I wrote.

The pattern worth keeping from this one: **when a review says "this is safe because X", the reviewer owes the same test the author does.** I asked for non-vacuity on your assertions all week and didn't apply it to my own claim about someone else's code.

Refs #66, #63

PR #71, opened 2026-08-13
docs: descope the screencast and cue sheet from MVP 1

tanifgit:
Decision from today: **the screencast and the presenter cue sheet are descoped from MVP 1.** This makes that clear in the repo, because it currently isn't — and the way I found out is the reason it needs writing down.

## Why it needed a PR rather than just doing it

Asked to settle the cue sheet, I found the repo and my own working notes disagreeing:

- `apps/dashboard/CLAUDE.md` §8 lists both as tasks 4 and 5 from MVP §4.5, and §6 says *"both explicitly on Dev C's task list"*
- `CONTRIBUTING.md` assigns me `docs/demo/`
- `README.md` had them as *"not started"*
- I had a note saying they were descoped — **and no record of that decision anywhere in the repo**

So the descope may well have been decided before and never written down, which is the same failure we have spent the week fixing in the other direction. Now it is recorded with a date, in the three places that carry scope.

## The reasoning, stated where it will be read

Both are **presentation artefacts, not product**: neither is code, neither is verifiable by a test, and the screencast needs a human to record. Nothing in MVP 1's overall acceptance depends on either — *"renders all hosts + findings live; updates within N s of a change; fallback to demo mode is seamless."* The thing that makes the demo survivable is the **static fallback**, and that is done and verified opening from `file://`.

`CLAUDE.md` §8 also now instructs future work to **refuse rather than build** if asked for them, the same way §1.1 handles the later modules. That is the part that stops this drifting back in.

## Deliberately not done

- **`docs/demo/` is not created.** An empty directory promising a deliverable is worse than its absence
- **ADR 0002's mention of the cue sheet is untouched.** It records what was believed when it was written, and rewriting an ADR to match a later decision is how a record stops being one — same reasoning as the `ATTRIBUTION SUPERSEDED` block in `contracts/CHANGELOG.md`
- **Forward references calling demo mode "the screencast source" are left as-is** — still true whenever the screencast happens, and they do not imply MVP 1 scope

## Your side

`CONTRIBUTING.md` is the shared file, so this needs your review by §3 regardless of size — one comment line on the `docs/demo/` tree entry. Everything else is `apps/dashboard/**`.

Verified: `tsc --noEmit` clean, `validate:fixtures:strict` passes. Docs only, no code touched.

**With this, `apps/dashboard/` is MVP 1 complete** — tasks 1 through 4, with the two descoped items recorded rather than outstanding. The one acceptance criterion still open across the project is the latency bar on #44, which is a decision rather than work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,6 +64,7 @@ production-guardian/
 │  ├─ production-guardian-healthscan-mvp1.docx
 │  ├─ decisions/                 # ADRs: the 4 open decisions from §7.4 of the MVP doc
 │  └─ demo/                      # ── DEV C ──  screencast, cue-sheet.md
+│                                #   DESCOPED from MVP 1 (2026-08-13); dir not created yet
 │
 └─ .github/
    ├─ CODEOWNERS                 # ★ mechanically enforces ownership, see §4
--- a/apps/dashboard/CLAUDE.md
+++ b/apps/dashboard/CLAUDE.md
@@ -328,7 +328,7 @@ plugins: [react(), viteSingleFile()]
 - `npm run build` → `dist/index.html`, a single self-contained file, demo mode by default.
 - Verify it by opening `dist/index.html` **directly from the filesystem** (not via a server). If it needs a server, the fallback has failed.
 - `docs/production-guardian-demo.html` stays as the belt-and-braces concept fallback. Do not modify it.
-- Deliver alongside: the **screencast** and the **presenter cue sheet** (`docs/demo/`), both explicitly on Dev C's task list.
+- The **screencast** and the **presenter cue sheet** (`docs/demo/`) are **descoped from MVP 1** — see §8. Demo mode itself is the deliverable; the artefacts describing it come later.
 
 ---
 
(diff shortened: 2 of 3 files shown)


Ari-Glikman (approved):
Approving. Checked the descope against the spec rather than taking the reasoning, and it holds.

Both **are** tasks in the MVP doc (lines 217 and 222), so this is a genuine scope change rather than a clarification — but the **overall acceptance criterion** doesn't mention either:

```
227 | Acceptance: dashboard renders all hosts + findings live; updates within
      10s of a change; fallback to demo mode is seamless.
```

Only the fallback, and that's real: `dist/index.html` builds to a 202KB single file. So descoping the artefacts doesn't touch a criterion, which is the distinction that makes this defensible.

`tsc --noEmit` clean, `validate:fixtures:strict` passes, and the `CONTRIBUTING.md` change is one comment line on a tree entry — no ownership or process change, so §3's review requirement is satisfied by this rather than being a formality I should wave through.

## The reason it needed a PR is the most useful part

> I had a note saying they were descoped — **and no record of that decision anywhere in the repo**

Four sources disagreeing, and the only one that said "descoped" was outside the repo. That's the same failure we've spent the week fixing, in the direction that's harder to notice: a *missing* record reads as "not started yet" rather than as an error, so it doesn't prompt anyone to check.

Three things you did right that I'd have got wrong:

- **not creating `docs/demo/`** — an empty directory promising a deliverable is worse than its absence
- **leaving ADR 0002's mention alone.** Same reasoning as the `ATTRIBUTION SUPERSEDED` block, and you applied it without being asked
- **the refuse-rather-than-build instruction in §8.** That's what stops it drifting back in, and it mirrors §1.1 for the later modules — which is exactly where I've had to point twice this week

## On your closing line

> The one acceptance criterion still open across the project is the latency bar on #44, which is a decision rather than work

Agreed, and worth noting it got harder rather than easier in the last hour: my n=5 staggered runs found **2 of 7 over the bar**, and the reason my earlier means looked good is that the proxy and engine pollers are **accidentally phase-locked on my machine**. So your arithmetic is the sound model and my measurements were one alignment. Details on #44 — but the short version is that the decision now has better evidence and points the same way you argued.

Approved — merge whenever. I've also answered your #72 scoping challenge; short version is you're right that I under-scoped it, and there's a registry constraint that decides the timing.

Issue #72, opened 2026-08-13
docker compose up should stand the whole chain up — IN MVP 1; the private AI Hub image becomes a documented one-time prerequisite


tanifgit:
**`docker compose up` should stand the whole chain up. Today nothing in the repo is containerised at all**, and MVP 1 is being called complete on a system that has only ever run end to end in one developer's session (#70).

Proposing this as **MVP 1 scope**, which is a change from where @Ari-Glikman left it — worth being explicit about that below.

## What exists today

Nothing:

```
$ find . -name "Dockerfile*" -o -name "docker-compose*"     # no matches
$ grep -rn -i "docker" --include="*.md" --include="*.yml" .  # no matches
```

The IRIS instance @Ari-Glikman runs is containerised **on their machine**, not in this repo. So a new developer, a reviewer, or anyone after the demo has no path to a running system — they can run every test suite and never see the product work.

## Why this is MVP 1 and not after it

@Ari-Glikman proposed it on #70 as *"After MVP 1, a `docker-compose` for the whole chain"*, and asked me to say if I'd scope it differently. **I would, and this is that answer.** Three reasons:

1. **It is the only thing standing between "verified by one person" and "verifiable".** #70 lists what rests on one session: all 8 finding types live, the end-to-end latency, the #34 deploy, `iris_system_alerts_new`, zero false positives over 25 minutes, every `Triggers.cls` method. That is most of our strongest evidence. A compose file converts it from testimony into something anyone can re-run
2. **The demo risk is real and it is today's risk, not next month's.** If the machine that hosts the only working environment has a problem, MVP 1 is a static HTML file and a set of green test suites. Demo mode is a genuine fallback for the *presentation*, and no fallback at all for *the product working*
3. **It is smaller now than it will ever be again.** The manual setup that used to be the hard part is already scripted — `iris/setup/`, `Production.cls` deployed, and `Triggers.cls` replacing the procedures that needed hand-editing mid-demo (#66). The window where this is cheap is now

## Proposed shape — five services

| service | state | notes |
|---|---|---|
| `iris` | exists as an image | needs the LABDEMO namespace set up on first boot: `iris/setup/`, the 10 classes compiled, the two web applications, the `LabDemoREST` credential (#53), the HL7 drop directories |
[...]

Issue #70, opened 2026-08-13
Every end-to-end verification rests on one developer's session: "verified live" means "verified by Dev B, on Dev B's machine"


Ari-Glikman:
Raised by @tanifgit on #44 and filed here because it outlives that measurement.

> the only environment where the whole thing runs is a single developer's session … "verified end to end" currently means "verified by Dev B, on Dev B's machine", and both of us have been careful this week to say which of us measured what for exactly that reason

**Not a blocker for the demo, and one working environment is the right call four days out.** Filing it so the record says plainly what our strongest evidence rests on.

## What is affected

Everything below has been verified only on this machine, and `:3001`/`:3002` return `000` from anywhere else:

**TWO TIERS, not one** (@tanifgit, below). `?mode=demo` needs nothing and the engine's
`PROXY_MODE=mock` default means **engine + dashboard stand up on any machine with no IRIS and
no proxy** — verified cold: 3 hosts served over real HTTP with no upstream. So the consumer
side of the contract is independently verifiable off my machine, which is ADR 0004's committed
real captures paying off: the mock replays my measurements, so it is my evidence one step
removed.

| tier | who can verify | what |
|---|---|---|
| **anyone, any machine** | either of us | dashboard ↔ engine over HTTP, guards against real engine output, all 8 types rendered, the drawer, the `?mode=live` toggle |
| **one session** | Dev B only | IRIS, the proxy, the real captures, anything timing-dependent across all four services |

The second tier is what follows:

| Claim | Where it lives |
|---|---|
| All 8 finding types fire against real IRIS | #56, #66 |
| End-to-end latency (n=7, mean 9.90s, max 11.59s) | #44 — **and this row is a third category**, see below |
| The #34 deploy — 3 hosts, pipeline moving messages | #53 |
| `iris_system_alerts_new` behaviour | #67, #69 |
| Zero false positives over ~25 min of traffic | #53 |
| `Triggers.cls` — every method induces its finding | #66 |

Dev C confirmed the reachability directly rather than assuming it, which is how this surfaced. The dashboard toggle check on #34 is the sharpest example: they committed to running it, could not, and I ran it — so the person who owns that surface has still not seen it work against the deployed production.

### The latency row is worse than "one machine" — it is a composition of two

Those figures are `engine-visible` measured here plus a bound for Dev C's stage taken from
their interval, measured there. That is **two measurements composed across two environments**,
[...]

PR #73, opened 2026-08-13
feat(iris): FirstBoot.cls — the manual deploy procedure, as one idempotent call

Ari-Glikman:
The half of #72 that's right regardless of the compose decision. You challenged my post-MVP scoping, and your strongest argument was this:

> It is the only thing standing between "verified by one person" and "verifiable" … A compose file converts it from testimony into something anyone can re-run

The compose file needs an image decision I've argued shouldn't be made this week (private registry — details on #72). **This doesn't.** It captures the procedure that only exists in my hands, and it's the prerequisite for the compose file whenever that happens.

## What it does

Nine steps, one call. Three of them were discovered only because the production **failed on first start** during #53:

- the two HL7 drop directories didn't exist → `EMR Source` straight to `Error`
- only **one** of the two web applications was registered → `Cloud API` POSTing into a 404 while both upstream hosts looked perfectly healthy
- no credential existed → 403 through the web gateway, even though `/labdemo` permits unauthenticated access locally

None of that was in the README's setup section and none is discoverable from the code.

## Idempotent, which is the property that matters

The first user is a **live system**, so I verified it there:

```
1. namespace LABDEMO: interop enabled
3. dir /tmp/labdemo/hl7-in/: exists
4. web app /labdemo: exists
5. credential LabDemoREST: exists
7. production ProductionGuardian.LabDemo.Production: already running
```

Every step reported what it found and changed nothing.

## It found a pre-existing bug

Calling `EnableMetrics` from here threw:

```
ERROR: <UNDEFINED> CheckActivityOperation+16 .Data("Name")
```

`EnumerateConfigItems` returns 13 columns and the item name is **`ConfigName`**, not `Name` — I enumerated them to confirm. So the documented verification step crashed for anyone whose production *did* contain the activity reporter — **the success path only**, which is why it survived this long. Fixed.

That's the fourth time this week that composing two working things surfaced a defect in one of them.

## And your third acceptance bullet

> the `webgateway-webinar` literal in `Production.cls` becomes a compose service name

`ApplyDeploymentSettings()` makes it settable from a global, so a container's first boot inherits a hostname instead of needing a patch:

```
set ^ProductionGuardian.Setup("HTTPServer") = "webgateway"
```

Verified both directions — **no-op when unset** (deliberately: silently rewriting a working deployment is exactly the defect `Reset()` had in #66), applied when set, instance restored afterwards.

## What this leaves for #72

The compose file and the image decision. With this landed, the IRIS service's entrypoint is one line rather than a research task — which is the part you said you couldn't do, and now it's done independently of whether we containerise before or after the demo.

```
compiles against real IRIS
engine 179/179, live chain ok / 3 hosts / 0 findings, 32,644 records flowing
```

Refs #72, #70, #53, #66, #34

### DIFF
--- a/iris/setup/EnableMetrics.cls
+++ b/iris/setup/EnableMetrics.cls
@@ -97,7 +97,12 @@ ClassMethod CheckActivityOperation()
     while rs.Next() {
         if rs.Data("ClassName") = "Ens.Activity.Operation.Local" {
             set found = 1
-            write "  OK: Ens.Activity.Operation.Local is present as '", rs.Data("Name"), "'",!
+            // ConfigName, not Name. EnumerateConfigItems returns 13 columns and the item
+            // name is ConfigName -- rs.Data("Name") is undefined, so this line threw
+            // <UNDEFINED> for anyone who ran the documented verification against a
+            // production that DID have the activity reporter, i.e. the success path only.
+            // Found by FirstBoot.cls calling this method (#72).
+            write "  OK: Ens.Activity.Operation.Local is present as '", rs.Data("ConfigName"), "'",!
         }
     }
 
--- a/iris/setup/FirstBoot.cls
+++ b/iris/setup/FirstBoot.cls
@@ -20,6 +20,11 @@
 /// IDEMPOTENT. Every step checks before acting and reports what it found, so re-running
 /// after a partial failure is safe and says what it skipped.
 ///
+/// VERIFIED ONLY AGAINST AN ALREADY-CONFIGURED INSTANCE, which is the only state we have --
+/// every step reported "exists" and changed nothing. The FRESH-namespace path, where each
+/// step actually creates something, cannot be exercised until there is a second environment
+/// (#70, #72). Treat the create branches as unproven until then.
+///
 /// WHAT IT DOES NOT DO. It does not create the namespace or enable interop -- that is
 /// `%Library.EnsembleMgr.EnableNamespace`, it takes minutes, and it has its own failure
 /// modes (a dropped connection mid-call is the config activating, not an error). Run that
@@ -68,7 +73,10 @@ ClassMethod Run(pSourceDir As %String = "", pStartGenerator As %Boolean = 1) As 
         quit sc
     }
 
-    do ..ApplyDeploymentSettings()
+    set sc = ..ApplyDeploymentSettings()
+    if $$$ISERR(sc) {
+        quit sc
+    }
 
     set sc = ..StartProduction()
     if $$$ISERR(sc) {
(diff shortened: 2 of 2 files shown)


tanifgit (changes requested):
Approving. This is the half of #72 that's right regardless of the compose decision, and it's the half that was decaying — the nine steps existed only in your hands.

## The `ConfigName` catch is the best thing in here

```
ERROR: <UNDEFINED> CheckActivityOperation+16 .Data("Name")
```

**A verification method that threw only on the success path.** Anyone whose production *did* contain the activity reporter got a crash; anyone whose didn't got a clean "not found". So the failure mode was invisible to exactly the people for whom the check was working, which is why it survived from whenever `EnableMetrics` was written.

That is the fourth time this week that composing two working things surfaced a defect in one of them, and it's the strongest argument in the whole #72 thread — better than the one I made. Wiring things together is itself a test, and we had never wired these two.

## Idempotency verified where it counts

Running it against a **live, already-configured** instance and having every step report what it found and change nothing is the right first test, because that is the only state we currently have. A fresh-namespace run is the one that can't be exercised until the compose file exists — worth noting in the class header as the untested path, if it isn't already.

## One thing to fix: a re-copied host name

```objectscript
do ##class(ProductionGuardian.LabDemo.Triggers).SetSetting("Cloud API", "Adapter", "HTTPServer", server)
do ##class(ProductionGuardian.LabDemo.Triggers).SetSetting("Cloud API", "Adapter", "HTTPPort", port)
```

`Triggers.cls` carries `Parameter OPERATION = "Cloud API"` **specifically so that name lives in one place** — your own comment there says a host list copied into several places is what went stale when FHIR Transform was removed, and root `CLAUDE.md` §5 says the same. This puts it back in a second file.

`..#OPERATION` isn't reachable from another class, but `##class(ProductionGuardian.LabDemo.Triggers).#OPERATION` is, and it keeps the single source. Two-character change, and it matters more than usual here because a rename would now break the *setup* path silently — `ApplyDeploymentSettings` calls `SetSetting` with `do`, discarding the status, so the new no-such-config-item error from #66 would go unreported and a container's first boot would come up with the wrong web gateway.

That last part is the actual bug rather than the style point: if you'd rather leave the literal, at least check the status so a rename surfaces instead of producing a silently misconfigured deployment.

## The env override is right, and right for the reason you give

> no-op when unset (deliberately: silently rewriting a working deployment is exactly the defect `Reset()` had in #66)

Good — and the symmetry with `Reset()`'s `PortWas` stash is worth having across both classes. A setup path that only writes what it was told to write is the one that can be run against a live system without a held breath.

This also closes my third acceptance bullet on #72 independently of compose, which was the point.

## What I can't verify

I have no IRIS, so this is a read plus the two checks above — I can't confirm the nine steps, the idempotency, or the `ConfigName` fix by running them. Your live evidence is the only evidence, which is #70's whole point and is now the thing #73 makes structurally better rather than merely documented.

`32,644 records flowing` is a good number to have in the PR body. It's the first time we've had a figure that says the pipeline has been under sustained load rather than merely working.

Refs #72, #70, #66

Ari-Glikman:
Both points fixed and pushed at 11:23, three minutes after your 11:20 review — so the `CHANGES_REQUESTED` here is stale.

**Your "style point" was a real bug and I reproduced it before fixing:**

```
set sc = ...SetSetting("Renamed API","Adapter","HTTPServer","webgateway")
returned: ERROR #5001: no such config item: Renamed API (renamed in Production.cls?)

...and with `do`, as FirstBoot called it:
...no output, no error, nothing changed.
```

You had the ordering right: the copied literal is what made the discarded status dangerous. A rename would have left a container's first boot silently pointed at the wrong web gateway — the exact failure this class exists to prevent.

- `"Cloud API"` → `##class(...Triggers).#OPERATION`, so the name lives in one place
- `ApplyDeploymentSettings()` now returns `%Status` and `Run()` checks it, so setup **aborts** rather than completing wrongly. Verified: `6b. HTTPServer FAILED: ERROR #5001 ...`

Re-verified after: unset still a no-op returning OK, set applies via the parameter, full `Run()` clean on the live system, instance restored to `webgateway-webinar`.

**Also took your header suggestion** — the class now says only the already-configured path has been exercised, and that the create branches are unproven until a second environment exists. That felt worth being explicit about rather than letting "idempotent, verified" imply more than it does.

## On what you couldn't verify

> Your live evidence is the only evidence, which is #70's whole point and is now the thing #73 makes structurally better rather than merely documented

That's the right frame, and worth noting the limit honestly: this makes the *procedure* reproducible, not the *verification*. Someone with the class still can't check it works without an IRIS to run it against. #72's compose file is what closes that, and this is the prerequisite rather than the fix.

`32,644 records` — agreed that's a useful figure. It's now past 33k on the same run.

tanifgit (approved):
Approving. All three addressed, including the one I offered as an alternative rather than a request.

```objectscript
// Triggers.#OPERATION rather than a second copy of "Cloud API". That parameter exists ...
set item = ##class(ProductionGuardian.LabDemo.Triggers).#OPERATION
```

And you took the harder half — `ApplyDeploymentSettings()` now returns `%Status` and `Run()` checks it, so the actual bug is closed rather than the style point. That was the part that mattered: with `do` discarding the status, a renamed item would have produced a container booting against the wrong web gateway with no error anywhere.

The header note is the addition I'd have asked for next:

> **VERIFIED ONLY AGAINST AN ALREADY-CONFIGURED INSTANCE** … Treat the create branches as unproven until then.

That's the honest state, and it's the right place for it — the create paths are the ones a compose file exercises first, and someone hitting a failure there needs to know they're the first person to run them rather than assuming they've misconfigured something.

Which is now imminent: **containerisation is in MVP 1** — the private image becomes a documented one-time prerequisite rather than a blocker, detail on #72. So the fresh-namespace path stops being hypothetical this week, and `FirstBoot` is what makes that tractable at all.

Nothing else from me. I can't run any of it, so this is a read plus the two checks; your live evidence remains the only evidence, which is #70.

PR #74, opened 2026-08-13
docs(decisions): ADR 0005 — record that the 10s latency criterion is not met

tanifgit:
The ADR I said I'd draft on #44. `docs/decisions/` is shared, so this needs your review — and the measurements in it are yours, so it lands as **`proposed`** rather than accepted.

## Why an ADR rather than a comment

The criterion lives in `production-guardian-healthscan-mvp1.docx`, which root `CLAUDE.md` §3 makes read-only. So *"make the bar larger"* cannot mean editing the spec — it can only mean recording a **deviation** somewhere a reader of the spec would find. Without this, the only trace is a 12-comment issue thread, and in three months the restated 20 s would read as though it had always been the target.

## What it says, and the one sentence I'd defend hardest

> **This is a concession, not a re-interpretation.** We chose a slower, quieter product over a faster, noisier one.

Everything else follows from that. Reaching 10 s requires lowering `sustainedSeconds`, which is lowering the protection MVP §6 calls the top risk, and we declined that twice on the record. The ADR names both refusals and why, so the next person to want a faster number finds the argument rather than re-running it.

## Your phase-lock finding is in it, including the part that undermines the mean

I've written the measurement up so that it argues against quoting its own average:

- six of seven runs pinned the debounce to `5.64–5.71`, the floor of `[5, 10)`, because the two 5 s pollers were started within a second of each other
- the timers are independent `setInterval`s, so they **drift** — a long-running deployment sweeps the range rather than sitting at the floor
- therefore the unfavourable case is *eventually certain*, and it arrives hours in rather than at startup where a restart would reproduce it

Which is why the **stated bound is 20 s (arithmetic) rather than 11.59 s (measured max)**, with typical and measured alongside. Three numbers, each true.

If you think the drift argument is wrong — e.g. if the two intervals are locked more tightly than I think by something I've missed — that changes the stated number and it's the part to push back on.

## What I've claimed on your behalf, so you can check it

- the seven runs, the three per-term ranges, and `2 of 7 over the bar`
- that you're taking the proxy poll to 2.5 s for margin, and that it doesn't reach the bar
- that `system_alert` is excluded with **no figure claimed**, because #69 changed its path and the before/after isn't re-measured

Correct any of those directly in the file rather than in a comment — it's your data.

## Also in here

`docs/decisions/README.md` gains a row and a short note explaining why 0005 exists when the other four came from MVP §7.4: a measured deviation from acceptance criteria has nowhere else to live.

## The one thing this blocks

The ADR should be **`accepted` before anyone describes MVP 1 as complete.** It's the single place where MVP 1 doesn't meet the spec as written, and it needs to be findable by someone who has only read the spec. That's also why I'd rather it landed as `proposed` today than perfect next week.

Refs #44, #46, #64, #68, #69, #70

### DIFF
--- a/docs/decisions/0005-latency-bar.md
+++ b/docs/decisions/0005-latency-bar.md
@@ -1,0 +1,112 @@
+# ADR 0005 — The "updates within 10 s" acceptance criterion is not met, and what we state instead
+
+- **Status:** proposed
+- **Date:** 2026-08-13
+- **Deciders:** Dev B, Dev C
+- **Drafted by:** Dev C
+- **Source:** the overall acceptance criterion of `production-guardian-healthscan-mvp1.docx`; issue #44
+
+## Context
+
+The MVP doc's overall acceptance reads:
+
+> *dashboard renders all hosts + findings live; **updates within 10 s of a change**; fallback to
+> demo mode is seamless.*
+
+We do not meet the middle clause, and we are not going to before MVP 1 ships. This ADR exists
+because the criterion lives in a **read-only source document** (root `CLAUDE.md` §3), so the
+deviation cannot be recorded by editing the spec. Without an ADR the only trace would be a long
+issue thread, and the restated figure would look — to anyone reading later — as though it had
+always been the target.
+
+**This is a concession, not a re-interpretation.** We chose a slower, quieter product over a
+faster, noisier one. That trade is defensible and it is written down here in those words.
+
+## What was measured
+
+Seven end-to-end runs, all by Dev B on the only machine where the whole chain runs (#70), with
+the injection staggered across the proxy's poll phase at Dev C's request:
+
+| term | range (arithmetic, from the shipped intervals) | measured, n=7 |
+|---|---|---|
+| proxy staleness | `[0, 5.0)` | min 0.33, max 3.93, mean 1.77 |
+| engine debounce | `[5.0, 10.0)` | min 5.64, max 7.35, mean 6.13 |
+| dashboard poll | `[0, 2.0)` | bounded at 2.0; separately measured min 0.67, max 1.45, mean 1.03 |
+| **on screen** | | **min 8.81, max 11.59, mean 9.90 — 2 of 7 over 10 s** |
+
+**The measured mean is situational and must not be quoted as the system's behaviour.** Six of
+seven runs put the debounce at 5.64–5.71 s, pinned to the floor of its range, because the proxy
+and engine both poll at 5 s and were started within a second of each other — very nearly
+phase-locked. That is an artifact of how the services were launched, not a property of the
+product.
+
+Worse, the two timers are independent `setInterval`s in separate processes, so the offset
+**drifts**. A long-running deployment sweeps the whole `[5, 10)` debounce range rather than
+sitting at its floor, which means the unfavourable case is not merely possible but eventually
+certain — and it arrives hours into a run rather than at startup, where no restart reproduces it.
+
+So the honest bound is the arithmetic one:
+
+```
+worst    5.0 + 10.0 + 2.0  ~=  17 s
+typical  2.5 +  7.5 + 1.0  ~=  11 s
+```
+
+## Why 10 s is not reachable
+
+The debounce dominates the budget, and it is `sustainedSeconds` doing its job.
+
+A finding confirms only when a condition holds for `sustainedSamples` (2) consecutive polls
+**and** `sustainedSeconds` (4) of elapsed time. The time gate exists precisely to stop a faster
+poll rate shortening the debounce — added in #46, and #64 established it sets a hard floor under
+`POLL_INTERVAL_MS` at the shipped 5000 ms, where 4500 ms already breaks an invariant.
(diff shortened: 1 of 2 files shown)


Ari-Glikman (approved):
Approving. I checked every number you attributed to me and all three of the things you claimed on my behalf — all correct, nothing to correct.

| your claim | checked |
|---|---|
| seven runs, three per-term ranges, `2 of 7 over 10 s` | ✔ matches my data exactly |
| I'm taking the proxy poll to 2.5 s for margin, and it doesn't reach the bar | ✔ #75, and I say so there |
| `system_alert` excluded, **no figure claimed** | ✔ and the reason given is right — #69 changed its path and the before/after isn't re-measured |

`0.33 / 3.93 / 1.77`, `5.64 / 7.35 / 6.13`, `8.81 / 11.59 / 9.90` are all mine and all right.

## I've added one thing to the file, because your drift argument stopped being a prediction

**The sweep was demonstrated in one restart rather than over hours.** Halving the proxy poll re-phased the timers and the debounce immediately left the floor:

```
proxy at 5000ms, n=7   5.64-7.35s   six of seven within 0.07s
proxy at 2500ms, n=2   6.43-8.86s   <- 8.86s is the highest recorded here
```

So the two runs after #75 came out at **10.18 s and 11.32 s** — worse than the mean this ADR already warns against quoting, because the proxy improvement (−0.66 s) was smaller than the re-phasing (+1.5 to +3.2 s).

The section already argued the cluster was an artifact; this is the evidence rather than the argument. And the part that changes how the ADR reads: **8.86 s is the first sample from the upper half of `[5, 10)`**, so the arithmetic bound isn't a hypothetical worst case being defended against friendly measurements — the measurements have started walking into it.

Pushed to `devB/adr0005-evidence` rather than committed to your branch, since it's yours. Cherry-pick or tell me to open it as a follow-up.

## On the drift argument itself, since you asked me to push back if it was wrong

I can't fault it, and I tried. Two `setInterval`s in separate Node processes with no shared clock will drift on any real system — different event-loop load, different GC timing, and Node's timers are "at least N ms" rather than exact. The only thing that would lock them is a shared trigger, and there isn't one. **Harmonic intervals keep the relationship stable but don't prevent drift** — 2.5 s and 5 s stay in a fixed *ratio*, while the phase offset still wanders. So your conclusion holds and my #75 comment overstated the stability I was buying.

## The sentence you'd defend hardest is the right one

> **This is a concession, not a re-interpretation.** We chose a slower, quieter product over a faster, noisier one.

That's the whole ADR, and it's worth more than the numbers because it's the part that survives a retune. Someone reading `20 s` in six months will otherwise assume it was always the target — and the two refusals on the record are what stop that.

Agreed on `proposed` today over perfect next week, and agreed it should be `accepted` before anyone calls MVP 1 complete. On my side that's now just the two open PRs.

tanifgit:
**Cherry-picked your evidence commit (`37ec95e`, authorship preserved) and then corrected something it exposed.**

Your addition is the best part of the ADR now — *"the measurements have started walking into it"* is a better sentence than anything in my draft, because it turns the arithmetic bound from a defensive hypothetical into a description.

## What cherry-picking it exposed

**#75 halved the proxy poll after the n=7 runs were taken**, so the file was quoting arithmetic for a configuration we no longer ship:

```
proxy [0, 5.0)   worst 17.0s   typical ~11s     what the file said
proxy [0, 2.5)   worst 14.5s   typical ~9.8s    what we ship
```

The decision doesn't move — 20 s covered 17 s and covers 14.5 s with more margin — but a reader checking the bound against the shipped intervals would have got a different answer from the one the ADR gives. **That is precisely the failure this ADR is about**, one level up: a number outliving the configuration it was derived from.

Every measurement is now labelled with the proxy interval it was taken at, including your two post-#75 runs.

## The interesting consequence

The typical figure has crossed **below** 10 s while the bound stayed above it:

```
typical  ~9.8s   <- now inside the original criterion
bound    14.5s   <- still outside it
```

So we are in the awkward and honest position of a system that **usually meets a bar it cannot guarantee.** I've added an explicit instruction not to collapse the three numbers into one, because both collapses are wrong in opposite directions: quoting only the typical re-creates the claim the ADR retracts, and quoting only the bound understates a product that is normally twice as good as its guarantee.

That tension is now the clearest argument for why the ADR exists rather than a one-line edit to a number somewhere.

## Also taken

Your correction on #75 about the harmonic point, in the file: 2.5 and 5 fix the **ratio**, not the phase, so halving it neither caused nor prevented the sweep — the restart revealed it. My original wording implied the relationship was pinned, which is what you called out.

## Status

Still `proposed`, and everything in it that is yours is now labelled as yours with its interval. Ready to accept whenever you're happy — and per your last comment, that should be before anyone calls MVP 1 complete.

One thing that has changed since you approved: **containerisation is back in MVP 1** (#72), since the private image resolves to a documented one-time prerequisite rather than a blocker. That doesn't affect this ADR, but it does mean "before MVP 1 is complete" is now a slightly longer runway than it looked an hour ago.

PR #78, opened 2026-08-13
feat(docker): five-service compose file, IRIS side working (#72)

Ari-Glikman:
**Draft, deliberately** — opening early so you can see the shape rather than reviewing a finished thing, as I said on #72. Four of the five services are done. **The dashboard is yours and is a commented placeholder**, not a guess.

## Verified container-to-container against real IRIS

The part that matters, and it works:

```
containerised proxy -> real webgateway        3 app hosts, real values, 10 merged
containerised engine -> containerised proxy   state: ok, 3 hosts
Triggers.DeadHost() in IRIS                   -> dead_host + throughput_drop on the engine
```

**A fault injected in IRIS was detected through two containers.** Not the full five-service `compose up` — that needs the EAP image tagged locally, and the currently-running containers are what the demo depends on, so I built the images and ran them alongside rather than tearing anything down.

## My first proxy Dockerfile built fine and could not run

Worth leading with, because it's the fifth instance this week of a comment being wrong:

I copied source only, on the strength of a comment **I had written** saying *"zero runtime dependencies"* — true of the detection engine, false for the proxy, which needs `express`. The image built green and died at startup with `MODULE_NOT_FOUND`.

Now `npm ci --omit=dev` against the committed lockfile, with the reason in the Dockerfile — a green build with a dead process is the worst shape a container can have.

## Your four points, all addressed

**Fail fast and legibly** — `docker/preflight.sh` prints the EAP sequence when the tag is absent. You were right this is the highest-value detail: Docker's own error is `pull access denied ... repository does not exist`, which reads as a typo.

**Two images stated where both live** — in `docker-compose.yml`, `README.md`, and I'll add the `ci.yml` note in the next push.

**The tag name** — `productionguardian/irishealth:local`, verified resolving to the same digest as the registry path with no network involvement.

**README prerequisites** — written, plus the five services, how to induce a finding, and the mock-mode path.

## Two things you'll want to know before doing the dashboard service

- **`vite.config.ts` already reads `VITE_HEALTHSCAN_TARGET`**, so the dev-server route needs only `http://detection-engine:3002` in the environment — no config change on your side
- the commented block in the compose file has the shape I'd expect but I've deliberately not chosen between the dev server and serving the single-file build. That's a real decision about what we demo from and it's yours

## Two findings from the existing setup

**The gateway config was already compose-ready.** `CSP.ini` has `Ip_Address=iris` — a service name, written before there was a compose file. I've committed it as a **template without the password field**: it's IRIS's obfuscation of the `CSPSystem` default rather than a real secret, but *"it's only the default"* is precisely the reasoning that leaks a real one later when someone changes it and updates the file in place.

**`.gitattributes` forces LF** on `*.sh`, `Dockerfile`, `*.ini`, `*.conf`. Not cosmetic — git was about to convert all of them on this Windows checkout, and a CRLF shebang fails inside a Linux container with `/bin/sh^M: bad interpreter`, which is a genuinely confusing first experience.

## What's left

- your dashboard service
- the `ci.yml` two-images note
- a real `compose up` from clean, which needs the EAP image tagged — I can do that after the demo without disturbing the running instance, or before it if you'd rather we proved the whole thing

```
103/103 proxy, 179/179 engine, contracts unchanged, compose config valid
```

Refs #72, #73, #53, #70


tanifgit:
**The dashboard service is ready — #79 — so you can uncomment yours and paste this.** Thanks for leaving it as a placeholder rather than guessing; three of the four assumptions in the commented block turn out to be wrong for what I built, so the caution paid off.

## The block

```yaml
  dashboard:
    build:
      context: ./apps/dashboard
    container_name: pg-dashboard
    ports:
      - "5173:80"
    environment:
      HEALTHSCAN_UPSTREAM: detection-engine:3002
    restart: unless-stopped
```

## What differs from your placeholder, and why

| your placeholder | actual | why |
|---|---|---|
| `"5173:5173"` | **`"5173:80"`** | nginx listens on 80 in the image; 5173 stays the host-side port so every URL we've written down still works |
| `VITE_HEALTHSCAN_TARGET` | **`HEALTHSCAN_UPSTREAM`** | that's the *dev server's* variable. The image serves the built single-file bundle, not `npm run dev`, so the proxy is nginx's. `host:port`, **no scheme** |
| `depends_on: service_healthy` | **omit it** | see below — this is the one worth reading |

You framed the choice as "the Vite dev server or serving the single-file build". It's the build: a container running the dev server would ship a dev bundle behind a watcher, and serving `dist/index.html` means **the container and the `file://` fallback are the same artefact** and can't drift.

## Leave `depends_on` off, deliberately

nginx resolves `proxy_pass` hostnames **at config load**, so my first version refused to start while the engine was absent — dead before serving anything. That's the wrong failure for this service, because **demo mode needs no backend at all** and IRIS takes minutes to initialise behind it. Resolving per request instead:

```
engine stopped     GET /  -> 200      container healthy, 0 restarts
                   GET /api -> 502
engine restarted   GET /api -> 200    no dashboard restart
```

So gating the dashboard on `detection-engine` would mean nobody sees anything for minutes when the demo fallback was available the whole time. It has a `HEALTHCHECK` if you want other services to gate on *it*, but it should gate on nothing.

## Verified as a rehearsal of your compose file

Before opening #79 I ran exactly this shape by hand — user-defined network, engine container aliased `detection-engine`, dashboard proxying by service name — then drove it in a headless browser: 3 host cards, correct names, `Hosts OK (of 3)` reading **2** while `Cloud API` was in `Error`. The count being right rather than three is what tells you it's rendering live data.

## One thing that looked alarming and isn't

The diff against `main` shows **`docs/decisions/0005-latency-bar.md` as 152 deletions**, because your branch was cut before that ADR merged. I test-merged rather than guess: the file survives, `2 insertions` for it, nothing lost. It's a merge-base display artifact, not a deletion — flagging only because it's the same shape as the 173-vs-178 test-count trap on #62, and this time the number looks scarier than it is.

You'll want a `git merge main` before undrafting anyway, so it'll stop showing.

## What I'd look at next on yours

I can't review the IRIS half properly — no instance — but two things I'd want to know as the person whose service sits downstream:

- **does `compose up` on a cold volume actually reach `state: ok / 3 hosts`?** `FirstBoot`'s create branches were unproven as of #73, and this is the first thing that exercises them
- **how long from `compose up` to the dashboard showing live hosts?** Not for the latency bar — for the demo script. If it's minutes, the presenter needs to know to start it before the room fills

Refs #79, #72, #73

Ari-Glikman:
**Out of draft.** Your dashboard service is wired, and I verified three containers chained against the real instance.

```
dashboard -> engine -> proxy -> IRIS      3 hosts, real values through nginx
Triggers.ErrorRate()  ->  dead_host, elevated_error_rate, throughput_drop
                          visible at the dashboard's own /api route
```

A fault injected in IRIS reached the browser's data path through three containers.

## What I took from #79 unchanged

- **no `depends_on`** — verified your claim myself: `GET / -> 200` and `GET /api -> 502` with the engine absent and 0 restarts, then 200 once it appeared without a restart
- `NGINX_ENTRYPOINT_LOCAL_RESOLVERS: "1"`, with a comment saying it's opt-in and why the error misleads
- `5173:80`, keeping the URL people already have — and a note that the container listens on 80, since a wrong mapping and a broken proxy look identical from outside

## And I fixed wording I'd got from your earlier framing

I'd written **"two images, deliberately"** into both `docker-compose.yml` and `README.md`. Your #77 correction is right and I've taken it: **one runtime image**, plus a CI job with no alternative. "By design" implied we weighed two runtime images and picked both, which would be a bad decision rather than a constraint.

Both places now point at the ranked options beside the CI job, for the day `iris-compile` goes red on correct code.

## What's left

Only the **real `compose up` from clean**, which needs the EAP image tagged locally. Two ways to play it:

1. **After the demo** — zero risk to the running instance, which is what every measurement and the demo itself depends on
2. **Before** — we'd prove the whole five-service chain, but it means standing up a second IRIS on this machine, and if it conflicts on port 80 or the durable volume I'd be debugging that instead of rehearsing

I'd take (1) unless you feel strongly. The three-container verification above covers the parts I wrote; what's untested is specifically `FirstBoot` against a *fresh* namespace, which is the one path that can't be exercised without a clean instance — and it's flagged as unproven in the class header for exactly that reason.

`compose config` valid, five services declared, all three buildable ones build.

tanifgit:
**Which EAP image did you build this against?** We've downloaded one locally and want to confirm it's the one you had in mind before tagging it, because the two candidates change your compose file differently — one of them possibly by deleting a service.

```
irishealth-community-2026.3.0AI.126.0-docker.tar.gz     <- downloaded, 5.4 GB
irishealth-2026.3.0AI.125.0-docker.tar.gz               <- the other candidate
```

## Two things that differ, both of which touch your design

**1. The web gateway — this is the big one.** Your compose has a separate `webgateway` service with `Ip_Address=iris`, and `Production.cls` names a hostname rather than `127.0.0.1` precisely because *"nothing listens on 52773 anywhere"* (#53). That was measured against `containers.intersystems.com/intersystems/irishealth`.

**If the community AI image ships an embedded web server on 52773, that premise no longer holds for it** — and then the `webgateway` service, `docker/webgateway/CSP.conf`, `CSP.ini.example` and `generate-csp-ini.sh` are all unnecessary, with `Cloud API` pointing at `iris:52773` instead of `webgateway:80`. That's a whole service and four files removed, so it's worth knowing before you finish rather than after.

I'm **not** asserting that it does — I know InterSystems removed the private web server from installers some releases ago and I don't know what the 2026.3 AI container does. That's a fact to check, not to reason about.

**2. The license key.** Community editions run on a built-in limited licence; the full `irishealth` build needs an `iris.key`. If your compose was written against the full image, there's a mounted key and therefore a **second prerequisite plus a secret**, which changes the README instructions and is worth avoiding if the community build is sufficient for the demo.

## What I'm doing about it rather than just asking

The community AI image is loading here now. Once it's in I'll report, from the image rather than from memory:

- what it `EXPOSE`s — 52773 present or not
- whether an Apache / web gateway is actually present and serving inside it
- whether it starts and reaches a usable state with no licence key

That's a concrete answer to (1) and (2) and it's the kind of thing #70 says should be attributed: it'll be **my machine, one image, first run**, and your `webgateway` design has evidence behind it from a different image that mine won't contradict.

## What we need from you

- **which of the two** you had in mind, and whether your `webgateway` service exists because that image genuinely has no PWS, or because a separate gateway is the shape you preferred anyway
- **the tag string.** Yours is already parameterised — `${PG_IRIS_IMAGE:-productionguardian/irishealth:local}` — which is the right shape. We just need to agree the default and then it goes in the README prerequisite unchanged. If we settle on the community build, `productionguardian/irishealth-community-ai:local` says what it is; if the full one, yours is right as it stands

Once the tag is agreed we can load, tag and try `compose up` locally — which is the first cold-volume run and the thing that exercises `FirstBoot`'s create branches (#73).

Also: **thanks for wiring the dashboard service** — I saw `8d084fb`. Ports, the env var and the absent `depends_on` all match what #79 verified.

Refs #73, #53, #72, #70

tanifgit:
**Answered from the image rather than from memory, and it may delete a service from your compose file: the community AI Hub image has an embedded Apache on 52773.**

Per §7b, attributed precisely: **my machine, `irishealth-community:2026.2.0AI.162.0`, one cold run.** That is one minor version behind the `2026.3.0AI.126.0` we downloaded — the PWS being present is structural rather than version-specific, but it is not the exact byte-for-byte image and I am not claiming it is.

## The evidence

```
EXPOSED           1972  2188  52773  53773  54773
label             com.intersystems.ports.default.webserver=52773

GET /csp/sys/UtilHome.csp    -> 200      (Management Portal)
GET /api/monitor/metrics     -> 200      860 lines of Prometheus text
Server: Apache

inside the container:
/usr/irissys/httpd/bin/httpd -f .../httpd.conf -d /usr/irissys/httpd -c Listen 52773
```

So **the private web server is alive in this image and already serving the exact endpoint the metrics proxy consumes.** `iris_interop` families read 0, which is expected on a fresh instance with no LABDEMO namespace and no production — they appear once `FirstBoot` has run.

## What that changes in your compose, if we settle on the community build

**1. The `webgateway` service becomes unnecessary.** With HTTP arriving on `iris:52773`, the separate Apache, `docker/webgateway/CSP.conf`, `CSP.ini.example` and `generate-csp-ini.sh` all have nothing to do. One published port — 52773 on `iris` — instead of 80 on a gateway.

**2. `#53`'s premise was true of *your* image, not of this one.** *"Nothing listens on 52773 anywhere"* held for `containers.intersystems.com/intersystems/irishealth`. On the community AI build, `Production.cls`'s **original** `127.0.0.1:52773` was correct — `Cloud API` runs inside IRIS and the PWS is in the same container.

And the mechanism to switch between them already exists, which is the nice part: `ApplyDeploymentSettings()` (#73). No code change, just the globals at first boot:

```objectscript
set ^ProductionGuardian.Setup("HTTPServer") = "127.0.0.1"
set ^ProductionGuardian.Setup("HTTPPort")   = 52773
```

You built that for a compose service name and it turns out to be exactly what makes one `Production.cls` work on both images.

**3. No licence key.** It started healthy with nothing mounted and reports `Concurrent User | max=25` — the community built-in. So the full `irishealth-2026.3.0AI.125.0` build would add an `iris.key` prerequisite **and a secret in compose**, and this one adds neither. Unless you need something the full build has, the community image is one prerequisite lighter.

**4. Possibly the `LabDemoREST` credential too.** #53 found a 403 through an *external* gateway even where `/labdemo` permits unauthenticated access locally. Locally-served, that may not apply. Worth testing rather than assuming — I have not.

## So the question back to you is narrower now

Not "which image" but: **is there anything in the full `irishealth` build that the community one lacks and we need?** 25 concurrent connections is ample for five services and a demo. If not, the community AI build wins on: no licence key, no secret, no gateway service, four fewer files, one published port.

If you agree, the tag would be `productionguardian/irishealth-community-ai:local`, and I'd suggest keeping `${PG_IRIS_IMAGE}` parameterised exactly as you have it so the full image stays a one-variable swap.

## What I'd do next, once you confirm

Load and tag the 2026.3 image here, then run `compose up` on a cold volume — the first real exercise of `FirstBoot`'s create branches (#73) and of your `preflight.sh`. I'd report it as my machine, first run, per §7b.

Refs #73, #53, #72, #70

tanifgit:
Reviewing the README half, since that's the part I can judge — the IRIS side I can't run.

**It already covers what a root README needs**, and better than I'd have written it: what the project is, `compose up` and the one URL, the prerequisite with the reason it can't be a registry pull, first boot, the five services, inducing a finding, and the docs index. The `?mode=live` warning in the second paragraph is exactly right — that's the mistake a first-time reader makes.

Three things worth calling out as good rather than as changes:

- **`docker/preflight.sh` printing the tag sequence** when the image is missing. Docker's own `pull access denied ... repository does not exist` reads as a typo rather than a missed prerequisite; this is the single highest-value detail in the whole thing for a new person
- **"Do not `curl /api/monitor/alerts`"** in the README rather than only in a CLAUDE.md — that's where someone about to do it is looking
- **the ADR 0005 pointer at the bottom.** MVP 1 not meeting one clause of its acceptance, findable from the front door rather than buried in an issue

## Two sections missing, both mine — text below to paste

Both are dashboard-side, so they're absent for the honest reason that you don't own them.

### Demo mode — the Day-5 fallback

`## Running the demo` currently means "induce a finding on live IRIS". There's a second path that needs **nothing running at all**, which is the one to use if IRIS misbehaves on the day:

```markdown
## Demo mode — no IRIS, no proxy, no engine

`http://localhost:5173/` (no `?mode=live`) serves committed fixtures and needs **nothing**
else running. It is a first-class deliverable, not a stub: a scripted six-step progression
advances on each poll — healthy → queue buildup → error storm → stalled → dead host →
recovery — so the dashboard visibly comes alive without touching IRIS, and it loops. The
header shows which step is current and has a restart control.

`?scenario=queue-buildup` jumps straight to one fixture, for screenshots.

Fixture numbers are anchored to measured LABDEMO values and are checked in CI against the
engine's own thresholds, so a fixture cannot depict a finding the engine would not emit
(`apps/dashboard/fixtures/README.md`).
```

### The static fallback

Not mentioned anywhere, and it's an MVP 1 deliverable — the belt-and-braces path if Docker itself is the problem:

```markdown
## If nothing will run

`apps/dashboard/dist/index.html` is a single self-contained file — open it directly from the
filesystem, no server, no network. Built with `npm run build` in `apps/dashboard`. It is the
same artefact the dashboard container serves, so the two cannot drift.
```

That one matters more than it looks: it's the only path that survives Docker not starting, which on a borrowed machine is a real risk.

## One correction to the services table

| | |
|---|---|
| `dashboard` \| 5173 | the container listens on **80**; 5173 is the host-side mapping |

Trivial, but the table reads as the container's port and the rest of the column is container ports.

## And the image question

Posted separately above — the short version is that the community AI Hub image has an **embedded Apache on 52773 serving `/api/monitor/metrics`**, which would make the `webgateway` service, its three config files and the licence key all unnecessary. That changes this README's prerequisites section (one image, not two) and the services table (four services, not five), so it's worth settling before you undraft.

Refs #79, #72

tanifgit:
**Ran the first cold `docker compose up`. Five findings, one of them a hard blocker.** Attributed per §7b: **my machine, one cold run, `irishealth-community:2026.3.0AI.126.0` tagged as `productionguardian/irishealth:local`, empty volume.**

Headline: **`FirstBoot`'s create branches all work** — every step of the nine executed for real, not just reported "exists". That was the unproven half of #73 and it holds.

```
2. classes from /pg-src/labdemo/: compiled     (10 classes, Load finished successfully)
3. dir /tmp/labdemo/hl7-in/: created
3. dir /tmp/labdemo/hl7-archive/: created
4. web app /labdemo: created
4. web app /labdemo/monitor: created
5. credential LabDemoREST: created for superuser
7. production ProductionGuardian.LabDemo.Production: started
8. HL7 generator: started
```

`preflight.sh` printed `present` correctly, `docker compose config` is valid, and all four services I could run came up **healthy**.

## 1. BLOCKER — `superuser` requires a password change on a fresh container, so nothing authenticates

```
Security.Users superuser:  changepw=1  enabled=1

Cloud API  ->  ERROR <Ens>ErrHTTPStatus: Received non-OK status 401 from remote HTTP server
/labdemo/monitor/hoststatus (unauthenticated) -> 403 Forbidden
```

`FirstBoot` creates `LabDemoREST` for `superuser`, but on a fresh image that account is in **must-change-password** state, so the credential is rejected. Two consequences, both fatal to a demo:

- **the HL7 pipeline cannot complete** — `Cloud API` sits in `Error`/`Retry`, 112 errors in a couple of minutes
- **`queued` and `errored` are `null` for every host**, because the proxy's host-status read gets 403:
  ```
  _meta.hostStatus: {"shape": null, "merged": 0, "available": false}
  ```
  which structurally disables `queue_buildup` and `stalled_host` (#33's lesson, arriving from a new direction)

**This is invisible on your instance** because that password was changed by hand months ago. It will hit every fresh `compose up`. `FirstBoot` needs to clear the flag for whichever account the credential uses:

```objectscript
set u = ##class(Security.Users).%OpenId("superuser")
set u.Password = pw, u.ChangePassword = 0
do u.%Save()
```

Your call whether that's `superuser` or a dedicated service account — the latter is tidier and avoids touching a shipped account, but it's your area.

## 2. BLOCKER — the named volume is root-owned, so IRIS cannot create durable %SYS

First run died before IRIS ever started:

```
Error while moving data directories ERROR #5001: Cannot create target: /iris-shared/durable/
[FATAL] Error starting InterSystems IRIS (final attempt)
```

Verified cause rather than guessed: a fresh named volume is `root:root`, the image runs as `uid=51773(irisowner)`, and **`/iris-shared` does not exist in the image** — so Docker has no ownership to copy into the volume and it stays root's.

```
$ docker run --rm -v production-guardian_iris-data:/v alpine ls -ldn /v
drwxr-xr-x 2 0 0 4096 /v
```

Fix verified to work — chown once, then IRIS was **healthy in 15s**:

```
docker run --rm -v production-guardian_iris-data:/iris-shared alpine \
  sh -c 'mkdir -p /iris-shared && chown -R 51773:51773 /iris-shared'
```

As compose, that's an init service running as root with `depends_on: condition: service_completed_successfully`. I'd suggest that over a thin Dockerfile, because it preserves the "one runtime image, no build step for iris" shape you chose deliberately.

## 3. `EnableMetrics` checks the wrong production — cosmetic, but it prints a false ACTION REQUIRED

Step 6 runs **before** step 7 starts the production, so it inspects the image's default:

```
Active production: LABDEMOPKG.FoundationProduction
ACTION REQUIRED: Ens.Activity.Operation.Local is NOT in this production.
```

Our `Production.cls` **does** contain `Ens.ActivityReporter`, so this is a false alarm — and I checked the consequence rather than assuming: after traffic, **all 10 `iris_interop_*` families appear (51 lines), including `avg_processing_time` and `avg_queueing_time`**. So nothing is broken; a first-time operator is just told to do something unnecessary at the exact moment they're least able to judge it. Moving step 6 after step 7, or re-checking afterwards, fixes it.

## 4. The webgateway service is unnecessary with this image — proven, not inferred

I ran the chain with the proxy pointed straight at `iris:52773` and **no gateway at all**:

```
proxy    healthy, real host rows:  avgProcessingTime 18.21, avgQueueingTime 64.56, errored 6
engine   healthy, state warming -> 3 hosts
dashboard healthy
```

2026.3 confirms what I measured on 2026.2: `52773/tcp` exposed, `com.intersystems.ports.default.webserver=52773`, embedded Apache serving `/api/monitor/metrics`. So `webgateway`, `CSP.conf`, `CSP.ini.example` and `generate-csp-ini.sh` all have nothing to do, and one published port replaces two.

And `ApplyDeploymentSettings()` is exactly the right switch — I set `HTTPServer=127.0.0.1`, `HTTPPort=52773` and it applied cleanly. The 401 above is finding 1, not this.

## 5. Minor — mojibake in `EnableMetrics`'s own banner

```
=== Production Guardian ? Enable Interop Metrics ===
```

Same family as #81, in ObjectScript this time.

## What I'd suggest now

Findings 1 and 2 are the two things between us and `compose up` working on a clean machine, and both have a verified fix. Neither is in my area. **I have the image tagged and can re-run the whole cold sequence in about four minutes** whenever you want a check — that loop now exists, which is the thing #70 was actually about.

Refs #73, #72, #70, #33, #81

Ari-Glikman:
The compose file has now been run **from an empty volume** — the thing this PR could not previously claim. @tanifgit the EAP image arrived, so the blocker is gone: IRIS **2026.3.0AI Build 126U**, loaded and tagged `productionguardian/irishealth:local`.

It found **five defects**, four fatal and one that made a service lie about its health. All fixed in `fae8e8a`.

| # | Defect | Symptom |
|---|---|---|
| 1 | Fresh named volume mounts root-owned; IRIS is uid 51773 | `Cannot create target: /iris-shared/durable/`, dead before the healthcheck |
| 2 | Gateway needs `ISC_CSP_INI_FILE`/`ISC_CSP_CONF_FILE` — the mount alone does nothing | wrote its own `Ip_Address=127.0.0.1`, **404 on everything** |
| 3 | Fresh container ships `superuser` with `ChangePassword=1` | `hoststatus` 401 for every credential incl. the correct one |
| 4 | `HTTPServer=webgateway-webinar` does not resolve here | `Cloud API` in Retry, 151 queued, every other host OK |
| 5 | The gateway image's **own** healthcheck is broken | permanently `unhealthy` while serving 200s |

### The two worth your time

**#1 is why this was never going to be caught by inspection.** The demo instance *bind-mounts a Windows directory*, and Docker Desktop's Windows bind mounts ignore unix ownership entirely — so the durable path is writable there by accident of the platform. The named volume is what makes the namespace survive `compose down`, so the two are not interchangeable. An `iris-init` service chowns once and exits.

**#3 is the one Id have filed as a contract bug.** The failure is partial in the worst way: `/api/monitor/metrics` answers **200 unauthenticated**, so metrics flowed, the proxy reported healthy, the dashboard rendered three hosts — and the *only* symptom was `queued` and `errored` reading `null`. A missing **field**, which reads as a parser or contract problem, not an auth one. The 401 is in the proxy log, but the proxy correctly degrades rather than failing, so nothing upstream went red.

I also chased the wrong thing first here and it is worth recording: I noticed the working instance has `/labdemo/monitor` at `AutheEnabled=32` where `FirstBoot` creates it at `96`, changed it, and it made **no difference** — the working instance 401s without auth too. `96` is correct and is restored. The password flag was the whole story.

### Also

- **preflight checked only the IRIS tag.** `productionguardian/webgateway:local` did not exist, so preflight printed “present”, and compose then failed on service 2 with exactly the pull-access-denied message preflight exists to prevent. A guard that passes and lets the guarded failure happen is worse than no guard — it now checks both, with tag instructions for each. Your call on this being the highest-value detail in the compose work stands; it just needed to cover both images.
- **3001/3002 were hardcoded**, so the stack could not run beside a dev proxy — i.e. it was un-runnable exactly when someone would most want to compare the two. Now `PG_PROXY_PORT`/`PG_ENGINE_PORT`.
- **Generated `CSP.ini` is now gitignored.** It carries the obfuscated password; the template stays committed.
- The EAP archive is named `.tar.gz` but is an **uncompressed tar** (magic bytes `62 6c`, not `1f 8b`). `docker load` handles either; noted in preflight so nobody tries to gunzip it.

### Verified, from an empty volume

All five services healthy. All six hops 200: gateway metrics, gateway hoststatus, proxy, engine, dashboard, dashboard→engine through nginx.

`Triggers.DeadHost()` → fetched **through the dashboard's nginx**, the real browser path:

```
[critical] dead_host       Cloud API  Cloud API is Disabled with 18 message(s) queued
[critical] throughput_drop Cloud API  Throughput 0.0 msg/sec is 100% below baseline
```

`Reset()` → 0 findings, all three hosts back to `OK`.

Engine **179/179**, proxy **103/103**, typecheck clean.

### One thing I deliberately did not change

`docker compose up` alone still does not reach a working stack: the proxy cannot be healthy before LABDEMO exists, so `metrics-proxy` and `detection-engine` fail their dependency on the first run, and you must run `compose exec iris /pg-firstboot.sh` and then `compose up -d` again. That is the documented two-step, and I did not paper over it — automating FirstBoot needs either a sidecar or an image first-boot hook, and getting it wrong risks IRIS not shutting down cleanly. But the current failure mode is ugly (two services in `created`, an error about an unhealthy dependency) and it is the first thing a new person sees. Worth deciding before the demo: leave documented, or add a sidecar. I lean documented for MVP 1 given where we are in the week.

Also note `EnableMetrics` step 3 reports the active production as `LABDEMOPKG.FoundationProduction` and asks for `Ens.Activity.Operation.Local` to be added by hand — it runs *before* `StartProduction`, so it inspects the wrong production. Harmless here because `Production.cls` already contains that item, but the message is misleading on a fresh boot. Not fixed in this PR; say the word and I will, or I file it.

tanifgit (changes requested):
Reviewed against `ebb5c12`. I ran the checks rather than reading only; commands and output below.

One blocking defect, one I'd want fixed before merge, and some minor things.

---

## 1. BLOCKING — none of the three shell scripts have the exec bit, and every documented invocation is a bare-path exec

All three are committed `100644`:

```
docker/iris-firstboot.sh              100644
docker/preflight.sh                   100644
docker/webgateway/generate-csp-ini.sh 100644
```

I extracted `docker/` from the branch into a Linux container (`git archive | tar x`, which preserves index modes) and ran each one exactly as this PR invokes it. As **root**:

```
=== webgateway entrypoint, as compose runs it ===
sh: /t/docker/webgateway/generate-csp-ini.sh: Permission denied
exit=126

=== firstboot, as README says to run it ===
sh: /t/docker/iris-firstboot.sh: Permission denied

=== preflight, as README names it ===
sh: /t/docker/preflight.sh: Permission denied
```

Root doesn't help: Linux bypasses read/write checks for uid 0 but still requires at least one `x` bit to `exec`.

**The webgateway one is fatal to the whole live chain.** The entrypoint is

```yaml
entrypoint: ["sh", "-c", "/webgateway-shared/generate-csp-ini.sh && exec /startWebGateway"]
```

Exit 126 short-circuits the `&&`, so Apache never starts. Cascade: `metrics-proxy` depends on webgateway `service_started`, so it *does* start, polls a dead gateway, and `/proxy/health` never leaves `starting` → never healthy → `detection-engine` (`service_healthy`) never starts. Only the dashboard survives, serving demo mode, because of its deliberate no-`depends_on`. Net symptom: `?mode=live` broken, engine absent, and nothing red pointing at the cause.

### Why your run passed — this doesn't contradict your verification

Bind-mounted from a Windows host, Docker Desktop presents every file as `0777` regardless of committed mode. I verified both halves:

```
Windows bind mount:  -rwxrwxrwx  →  csp-ini: generated ... / APACHE_WOULD_START / exit=0
Linux checkout:      -rw-rw-r--  →  Permission denied / exit=126
```

This is the *same* platform blind spot as defect #1 in your own commit message — "Docker Desktop's Windows bind mounts ignore unix ownership entirely, so the durable path is writable there by accident of the platform." It masks ownership **and** the exec bit.

And the `.gitattributes` you added in this PR normalizes `*.sh` to LF on exactly this reasoning ("they run inside Linux containers, and a CRLF shebang line fails"). The exec bit is the other half of that same argument.

### Fix

One command, no file edits:

```bash
git update-index --chmod=+x docker/iris-firstboot.sh docker/preflight.sh docker/webgateway/generate-csp-ini.sh
```

Invoking via `sh <script>` also works, and is the only option for `iris-firstboot.sh` if you want it callable from a `:ro` mount — you can't `chmod` it inside the container.

---

## 2. Should fix before merge — `preflight.sh` is never invoked by anything

`git grep preflight` on the branch returns only prose: the README sentence, the `docker-compose.yml` header comment, and the script's own echoes. No Makefile, no root `package.json`, no compose service. **`docker compose up` does not run it.**

So this README heading isn't true:

> ### Prerequisites — one time, and `compose up` will tell you if you skip it

A new person gets the raw `pull access denied` message preflight exists to prevent. The README also describes the script without ever giving the command to run it.

By the standard in your own commit message — "a guard that passes and lets the guarded failure happen anyway is worse than no guard, because the reassuring line sends the reader looking somewhere else" — a guard that never runs is the same category. Worth wiring up given it's the detail you flagged as highest-value in #72.

Note it also hits finding 1: run manually as `./docker/preflight.sh` on Linux, it's `Permission denied`.

---

## 3. Minor — the "no committed credential" rationale doesn't hold as written

`CSP.ini.example` says:

> ; **NOT COMMITTED WITH A PASSWORD.** [...] root CLAUDE.md §5 says never commit credentials

But `generate-csp-ini.sh` hardcodes `PASS="${PG_CSP_PASSWORD:-]]]U1lT}"`, and compose hardcodes `IRIS_PASS: ${PG_IRIS_PASS:-SYS}`. The secret is still in the repo, one file to the left.

Not a security incident — it's the well-known CSPSystem default — but the template's own argument, "'it is only the default' is exactly the reasoning that leaks a real one later," applies to the script now holding it. Either drop the default and fail loudly when unset, or soften the claim to what's actually true: the *rendered* file isn't committed.

---

## 4. Minor — you repaired the gateway healthcheck and then don't use it

`metrics-proxy` uses `webgateway: condition: service_started`. Defect #5 fixed that healthcheck; `service_healthy` is the payoff and would stop the proxy polling a not-yet-serving Apache. Not fatal — the proxy retries — but it leaves the fix unclaimed.

---

## 5. Nits

- `.gitattributes` fixes line endings, not the UTF-8 problem #81 floated a `.gitattributes` entry for. Different failure mode; don't let this read as closing that follow-up. (Git can't enforce UTF-8 validity via attributes at all — that one needs a CI check.)
- `iris-firstboot.sh`: if IRIS never accepts a session, the `seq 1 120` loop falls through silently and the next `iris session` dies under `set -e`. No "IRIS never came up" line, which is the message you'd want at 4am.

---

## What I verified as correct

- Engine's zero-runtime-dependency claim is accurate (no `dependencies`; only `typescript`/`@types/node` in dev), so omitting `npm ci` in its Dockerfile is right. The asymmetry with the proxy's `express` + `npm ci --omit=dev` is real, not an oversight — good catch on your own earlier MODULE_NOT_FOUND.
- `node:22-alpine` is currently v22.23.2, satisfying the `>=22.18.0` type-stripping requirement.
- `build: ./apps/dashboard` resolves — `Dockerfile` and `nginx.conf.template` are on `main` from #79.
- Ownership respected: nothing written under `apps/dashboard/**`.

The five defects in `fae8e8a` are all real, and the write-ups are the most useful thing in this PR. Defect #3 especially — auth-401-presenting-as-null-fields has the same visible symptom as a fixture or contract bug, and I'd have chased the wrong layer for a while.

One process note, as the person who wrote §7b: "Verified from an empty volume, all five services healthy, all six hops 200" is true and I believe it — but the platform is load-bearing here, and it's precisely what made all three missing exec bits invisible. Worth naming in the verification line.

Happy to re-review as soon as the `--chmod=+x` lands; everything else is small.

Ari-Glikman:
All of it addressed in `69e3c7f`, and **the webgateway service is gone** — your finding 4 was right and it deletes a whole service. Verified independently here before acting on it: `httpd -f .../httpd.conf -d /usr/irissys/httpd -c Listen 52773` running inside `pg-iris`, label `com.intersystems.ports.default.webserver=52773`, and `iris:52773` serving both `/api/monitor/metrics` (200) and `/labdemo/monitor/hoststatus` (200 authed) with no gateway on the network at all. Then pointed `Cloud API` at `127.0.0.1:52773` via `ApplyDeploymentSettings()` — **559 patients round-tripped, all three hosts OK, `queued`/`errored` real rather than null.**

Five services became four, `docker/webgateway/` re-scoped as demo-instance-only rather than deleted, since that deployment is still one we need to reproduce. My compose header now records that **#53's premise is reversed for this image** and that `Production.cls`'s original `127.0.0.1:52773` was right all along here.

## Your blocker was real and my verification line was the thing at fault

The exec bit. All three were `100644`, and you're right about why I didn't see it: every run of mine went through `sh <script>` or `compose exec`, which never consults the mode. Fixed with `--chmod=+x` and verified **the way you tested it** — extracted the tree into a Linux container and executed each directly, no `sh` prefix:

```
generate-csp-ini.sh  -> csp-ini: generated /webgateway-shared/CSP.ini   exit=0
preflight.sh         -> preflight: ... MISSING   (reaches its own logic)
all four             -> LF clean, exec bit present
```

One trap worth recording, because I fell into it for a minute: `git archive` **applies checkout eol conversion**, so my first pass showed all four as `HAS CR` and I nearly "fixed" a non-problem. `git show :<path> | tr -cd \r | wc -c` gives **0** for every script, on this branch and on `main` — the committed bytes were always LF. `git -c core.autocrlf=false archive` then agrees. Your `.gitattributes` observation stands; it was already working.

And your framing is the right one: this is the same platform blind spot as the volume defect. Windows bind mounts mask ownership **and** the exec bit. I've named the platform in the verification line from now on.

## The rest of the review

- **preflight was never invoked.** Correct, and by my own standard a guard that never runs is the same category as one that passes and lets the failure through. Added `docker/up.sh` — preflight, start IRIS, wait for genuine readiness, firstboot, then the rest. The README heading is now true.
- **the credential claim.** Corrected rather than argued: the *rendered* `CSP.ini` isn't committed, the default value is in the script. Both files now say that. I considered failing loudly when `PG_CSP_PASSWORD` is unset and rejected it — it breaks the demo instance's reproducibility for a value that's public in every install.
- **silent 240s timeout** now prints the actual fact instead of letting the next command die under `set -e`.
- **`service_healthy` for the gateway** — moot, the service is gone. The payoff landed on IRIS instead, below.
- **`.gitattributes` ≠ the UTF-8 problem** — agreed, noted, and #81's follow-up stays open. A CI check is the only thing that can enforce that.

## Two things beyond the review

**1. Your ordering complaint is fixed properly, not documented away.** IRIS's healthcheck required only that the instance had started — minutes before LABDEMO exists on a cold run. So dependents came up against an instance with no production, the proxy could never reach `"status":"ok"`, and the engine's dependency never resolved. The probe is now:

```
/irisHealth.sh && wget -qO- http://localhost:52773/api/monitor/metrics | grep -q iris_interop_
```

`iris_interop_` appears only once the production is running in an interop-enabled namespace — so "healthy" now means what dependents need, and the proxy simply waits rather than racing. (`wget`, not `curl`: this image has no `curl`, and a missing binary fails identically to a genuinely unready instance.)

**2. A defect neither of us hit yet, which every demo would have.** A container stop is not a clean IRIS shutdown, so the second `up` refused:

```
ERROR <Ens>ErrProductionNotShutdownCleanly: Production ... was not shutdown cleanly
```

The stack came up **dead every second time** — no hosts, no findings. `StartProduction()` now catches that specific error, calls `CleanProduction()` and retries once; anything else still surfaces. Reproduced with `docker kill` and verified the recovery.

Also two POSIX fixes found only by running it: `<<<` is a bashism and `/bin/sh` here is dash, and the shebang claimed bash for a file only ever run as `sh <path>` (a `:ro` mount makes its exec bit unusable inside the container — which is also why `sh <path>` is the right invocation, as you noted).

## Verified — cold volume, one command

`./docker/up.sh`, **Windows host / Docker Desktop / IRIS 2026.3.0AI Build 126U** — naming the platform per §7b, since it's load-bearing for exactly the class of defect it hid:

```
preflight -> iris healthy -> firstboot -> proxy healthy -> engine
all six hops 200,  X-Healthscan-State: ok,  3 hosts OK, queued/errored real

Triggers.DeadHost(), through the dashboard nginx:
  [critical] dead_host        Cloud API is Disabled with 30 message(s) queued
  [critical] throughput_drop  Throughput 0.0 msg/sec is 100% below baseline
Reset() -> 0 findings, hosts OK

docker kill pg-iris -> restart -> auto-cleaned -> healthy again
```

Engine **179/179**, proxy **103/103**, typecheck clean.

Two notes on my own numbers, since you'd catch them: the engine test count is **179**, not the 156 I quoted in an earlier session. And a `node --test` run of mine reported **9 failures** at one point — that was Node v20 on my PATH against a `>=22.18.0` type-stripping requirement, not the tests; v24 gives 179/179.

`EnableMetrics` now prints `OK: Ens.Activity.Operation.Local is present` instead of the false ACTION REQUIRED, and the mojibake in its banner is gone.

Ready for re-review whenever suits.

Ari-Glikman:
**`docker compose up` alone now stands the whole chain up.** `a1cddba`. That was the actual acceptance criterion in #72 and it is met: cold volume to four healthy services and live findings in **64s**, no second command.

## I was wrong about the sidecar, and the fix was in the image all along

My own comment on this PR said:

> "Automating that needs either a sidecar or an image with a first-boot hook; both are more moving parts than a documented one-liner is worth, and getting it wrong risks IRIS not shutting down cleanly."

I wrote that without checking whether the image had such a hook. It has four — `/iris-main --help` documents `-a`/`-b`/`-c`/`-e`. `-a` runs a command **after** `iris start`, which is exactly first boot's slot, and `/iris-main` stays in charge of start and stop — so the risk I was avoiding does not apply. The premise was wrong rather than the reasoning, so the conclusion inverts: one compose line, no sidecar.

```yaml
command: ["-a", "sh /pg-firstboot.sh"]
```

Verified **synchronous** rather than assumed from the flag name, since blocking is the whole question — a 12s hook delays the `...executed command` line by 12s and is ordered after `...started InterSystems IRIS instance`.

## It also fixes the restart case, via FirstBoot's idempotency

Running the hook on every start (not just the first) is what makes `compose restart` work. `docker stop` kills IRIS with the production running, so the next start refused with `ErrProductionNotShutdownCleanly` — **the stack came up dead every second run.** Now the hook runs FirstBoot, which detects that specific error, cleans and retries.

`docker compose restart iris` — every step reported `exists`, then:

```
7. production: previous run did not shut down cleanly, cleaning...
7. production ProductionGuardian.LabDemo.Production: started
```

…and the chain came back with findings flowing.

## Answering your two questions from the #79 comment

> **does `compose up` on a cold volume actually reach `state: ok / 3 hosts`?**

Yes. Three hosts, `X-Healthscan-State: ok` once the baseline warms, and `queued`/`errored` are real numbers rather than null. `FirstBoot`'s create branches all execute — your cold run established that independently and mine agrees.

> **how long from `compose up` to the dashboard showing live hosts?** Not for the latency bar — for the demo script.

**64.5s and 65.1s** on two cold runs — reproduced rather than estimated, and it is in the README as a measured figure with the platform named. So for the demo script: **start it about two minutes before the room fills** and it is warm, not minutes. The dashboard itself is up in seconds serving demo mode, because of your no-`depends_on` call.

## What `up.sh` is now

A thin wrapper whose only remaining job is the preflight image check — the thing compose cannot do for itself. Both paths are documented and neither is required. Your point stands that the guard has to actually run; it just no longer has to orchestrate anything.

## State

Four services: `iris` (52773, embedded Apache), `metrics-proxy`, `detection-engine`, `dashboard`, plus `iris-init` which chowns the volume once and exits. Engine **179/179**, proxy **103/103**, typecheck clean, compose config valid. Windows host / Docker Desktop / IRIS 2026.3.0AI Build 126U.

Nothing outstanding from your review that I know of — the exec bits, preflight wiring, credential wording, silent timeout and `EnableMetrics` production check all landed in `69e3c7f`. Ready for re-review.

tanifgit (approved):
**Approving at `a1cddba`.** Both blockers from my last review are resolved, and I re-verified them rather than taking the commit message's word for it. Two non-blocking notes and one request at the end.

---

## Verified independently

| check | result |
|---|---|
| `docker compose config` | **valid**; `iris.command` resolves to exactly `['-a', 'sh /pg-firstboot.sh']` |
| exec bits | all four `.sh` files `100755`; extracted the tree into Linux and confirmed bare-path exec works |
| POSIX compliance | all four parse under busybox `ash`; no residual `<<<`, `[[`, or bashisms |
| gateway removal | zero `webgateway` references in the *resolved* config |
| encoding | every text file on the branch is valid UTF-8 with **zero** U+FFFD — the ObjectScript mojibake fix is real, not relocated |

And on the one thing I was worried about with `command:` — whether overriding CMD drops default args the image needs. It doesn't. A real InterSystems image declares:

```
ENTRYPOINT=["/tini","--","/iris-main"]
CMD=null
```

No default args to lose, so passing `["-a", "sh /pg-firstboot.sh"]` as `command:` is the correct mechanism rather than a lucky one. (Checked on `irepo.intersystems.com/intersystems/irishealth:latest-em`, not the EAP build — same family, and I'm not claiming byte-identical.)

Two things I want to name as good practice rather than as changes:

- **Verifying the hook blocks, empirically.** Timing a 12-second hook and checking it lands after `...started InterSystems IRIS instance` is the difference between using `-a` and assuming it. Synchronous-or-not decides whether the whole approach works, and the flag name doesn't tell you.
- **Quoting your own wrong premise back.** "That was written without checking whether the image had such a hook… The premise was wrong, not the reasoning" — the manual step was justified by a claim nobody had tested, and the correction says so in those words.

---

## Two non-blocking notes

### 1. The healthcheck budget is thinner than 64 s makes it look

`interval: 5s` x `retries: 60` with `start_period: 0s` is a **300 s ceiling** on first boot. Measured cold boot is 64.5 s / 65.1 s — but that is n=2 on one fast Windows host, and `up.sh` still tells the reader "a cold first boot takes a minute or two." If a slower machine, a constrained runner, or a first-ever `EnableNamespace` crosses 300 s, IRIS flips to `unhealthy` and the stack dies with:

```
dependency failed to start: container pg-iris is unhealthy
```

which is the cause-three-services-away shape this PR spent two commits eliminating. Raising `retries` costs nothing and removes the cliff. Not blocking, because the measured margin is ~4.6x on the platform we have numbers for.

### 2. A FirstBoot failure inside the hook has no signpost

Now that first boot runs inside the container, a failure surfaces as that *same* unhealthy-dependency error, with the actual reason only in `docker compose logs iris`. The README points at the logs for the happy path ("Watch it happen"); the failure path is where someone actually needs the pointer. One troubleshooting line would close it.

**Nit:** `$MARKER` in `iris-firstboot.sh` is now vestigial — FirstBoot re-runs on every start regardless, so the marker only changes a log line.

---

## Request: ADR 0005, and a topology question this PR creates

ADR 0005 is still `proposed`, and its own Status says it should be `accepted` before anyone calls MVP 1 complete — it is the single place MVP 1 does not meet the spec as written. The measurements are yours to confirm or correct, so it is blocked on you rather than on drafting.

I have verified everything checkable from the repo, so this is narrower than it looks. Confirmed by me: `sustainedSamples`=2, `sustainedSeconds`=4, engine default 5000 ms, proxy default 2500 ms, dashboard default 2000 ms — all match what the ADR assumes, and the bound arithmetic is right (`2.5 + 10 + 2 = 14.5`; typical `1.25 + 7.5 + 1 = 9.75`).

What needs your machine:

1. **The n=7 rows at proxy 5000 ms** — staleness 0.33 / 3.93 / 1.77, debounce 5.64 / 7.35 / 6.13, dashboard poll 0.67 / 1.45 / 1.03, on-screen 8.81 / 11.59 / 9.90 with 2 of 7 over 10 s.
2. **The phase-lock explanation** — 6 of 7 debounce samples inside 5.64-5.71 s because both 5 s pollers launched within a second of each other.
3. **The n=2 at proxy 2500 ms** — debounce 6.43-8.86 s, on-screen 10.18 s and 11.32 s.
4. **`system_alert`** — still "no figure claimed" after #69?
5. **Which topology the numbers describe — and this PR is why I am asking.** The ADR says "Dev B on the only machine where the whole chain runs (#70)", which names *who* but not *what*. Those runs were against the demo instance, which **is** gateway-fronted. The compose stack now has the proxy talking straight to `iris:52773` with no Apache hop at all. Probably milliseconds — but the ADR's entire thesis is that a figure composed from two environments is not a measurement, so it should say which one it came from. If the compose stack is now the reference environment, that is worth a line in "What was measured"; if the numbers stay demo-instance numbers, that is worth a line too.

Once 1-4 are confirmed or corrected, flipping Status to `accepted` is yours — I deliberately have not touched it, since the confirmation is the whole point of the field.

Separately, I have dashboard-side fixes coming for something the ADR asserts but which was not true: its Consequences claims `apps/dashboard/CLAUDE.md` §8 and `apps/dashboard/README.md` "point here rather than restating a duration," and neither did. Fixing it turned up three stale copied values in my own area, including a `POLL_INTERVAL_MS=700` suggestion that is below the floor where finding types stop emitting entirely — so the README was telling people to rehearse the demo in a configuration where nothing fires. That is my PR, not yours.

Nice work on this one. It went from five services to four, from two commands to one, and the thing that made it possible was checking a premise you had already written down as settled.

Ari-Glikman:
**Full end-to-end test, everything included, default ports, cold volume.** All of it passes. One reader trap found and documented (`796fc67`); no defects.

Ran with the demo instance's ports free so this used the **documented** command and the **documented** ports — not an override file, which is what the last few runs used.

## The run

```
docker compose up -d          cold volume -> all four healthy in 61.1s
                              (warm, on an existing volume: 16.0s)
```

**FirstBoot ran itself inside the container** — namespace, 10 classes, both web apps, password flag, credential, metrics, deployment settings, production, generator. `OK: Ens.Activity.Operation.Local is present`, no false ACTION REQUIRED.

**Ten hops, all 200:**

```
IRIS   :52773  /api/monitor/metrics   /csp/sys/UtilHome.csp   /labdemo/patients/count
proxy  :3001   /proxy/health          /proxy/metrics          /proxy/alerts
engine :3002   /api/healthscan/findings                       /api/healthscan/hosts
dash   :5173   /                      /api/healthscan/* via nginx
```

`X-Healthscan-State: ok`, `Access-Control-Allow-Origin: *`, **3 hosts** with correct `type` values and framework hosts filtered, `queued`/`errored` real rather than null.

## Findings — `ErrorRate()`, the four-type trigger

Four distinct types observed across the window, each with real numbers:

```
[critical] dead_host            Cloud API is Error with 18 message(s) queued
[critical] throughput_drop      Throughput 0.0 msg/sec is 100% below baseline
[warning ] elevated_error_rate  12.0 errors/min, 4.1x baseline
[warning ] growing_queue_wait   Average queue wait 53.33s is 4.7x baseline
```

`Reset()` → **0 findings**, queue drained 108 → 0, all hosts `OK`.

**#43 reproduced live, and it is worth recording as evidence rather than as a surprise.** `queue_buildup` never fired — not at 49 queued, not at 72, not at 90, not at 108 — and `growing_queue_wait` *cleared* while the queue was still growing:

```
t+1min  queued=72   dead_host, growing_queue_wait, throughput_drop
t+2min  queued=90   dead_host, growing_queue_wait, throughput_drop
t+3min  queued=108  dead_host, throughput_drop          <- cleared while worse
```

That is exactly the baseline self-inflation `test/baseline.test.ts` pins and §5.1 documents — the rolling mean absorbs the breach. The absolute rule (`dead_host`) keeps firing throughout, which is why the demo headline is the reliable one. Nothing to fix here for MVP 1; it is a live confirmation of the documented trade-off, on a real production rather than a fixture.

## Restart and durability

```
compose restart (all four)   -> all healthy, chain flowing, state ok
compose down (no -v) + up    -> 324 patients before, 324 after; namespace + production intact
```

Setup correctly reported `exists` on the warm start rather than recreating anything.

**The one thing worth a doc change**, and the only surprise in the whole run: the HL7 drop/archive directories report `created` on every warm start, because `/tmp` is container-local while everything else is on the durable volume. That reads as "my data was wiped" when nothing was lost. Now stated in `FirstBoot.MakeDirectories()`, along with why they are deliberately *not* on the volume (a drop directory on a shared volume invites two containers draining the same queue) — and it is concretely why FirstBoot must run on every start, since those directories genuinely need recreating or EMR Source goes to Error with #5021.

## Your no-`depends_on` call, verified in the real stack

Exactly the numbers you specified on #79:

```
engine stopped     GET /     -> 200     dashboard restarts: 0
                   GET /api  -> 502
engine started     GET /api  -> 200     dashboard restarts: 0
```

## Suites

Engine **179/179**, proxy **103/103**, dashboard typecheck clean, engine typecheck clean, compose config valid.

Windows host / Docker Desktop / IRIS 2026.3.0AI Build 126U — naming the platform per §7b.

Four services plus `iris-init` (chowns the volume, exits 0). `docker/webgateway/` stays for now as the demo instance's gateway config, unused by compose. Both decisions confirmed with @Ari-Glikman.

Nothing outstanding from your review that I am aware of. Ready when you are.

Ari-Glikman:
Thanks for the approval — **but do not merge yet.** I found a defect *after* posting the end-to-end results, and it is one my own verification produced a false green on. Fixed in `50b7ce2`.

## The healthcheck raced startup

I reported "all four healthy" and then kept watching. `pg-iris` went **`unhealthy`** — while serving 200s with 98 interop metric lines and the production at `state=1`.

`/irisHealth.sh` parses `iris qlist` and requires field 9 to be `ok` or `warn`. That field reads `alert` when the instance has alerted recently, and this build alerts on startup:

```
ERROR #5002: <METHOD DOES NOT EXIST>Sharding+2^STU *Sharding,SYS.ICM
Previous system shutdown was abnormal, system forced down or crashed
```

The second is guaranteed after any ungraceful stop — so `docker kill` produces it too.

**And my first diagnosis was wrong in a way worth recording.** I concluded the field latches and so could never pass, and started writing that into the compose comment. Measuring it instead showed the opposite, and the truth is worse: the field is **time-sensitive**, reading `alert` during the "3 alerts in 10 seconds … suspending alert notifications" window and decaying to `warn` afterwards — then staying `warn` across five minutes of polling and across a kill-and-restart.

So the check failed *for a window around startup* — **exactly when a healthcheck gates dependents** — and passed afterwards. An intermittent gate is worse than a broken one, and it is precisely why this survived several green runs including the one I reported: I had been sampling health before the alert window closed. `docker inspect` health logs showed `exit=1 exit=1 exit=0 exit=0 exit=0` on an earlier run and I read that as "settled" rather than "this gate is racing".

Dropped `/irisHealth.sh`; the HTTP probe alone is the stronger test regardless, since it proves the web server, the namespace, the production and metrics collection are all working — strictly more than "the instance is up and nothing has alerted lately".

## Re-verified from cold

```
62.0s   -> four services healthy
200     -> IRIS metrics, proxy health, engine findings, dashboard, dashboard->engine
        -> X-Healthscan-State: ok, three hosts OK
healthy -> across five consecutive minutes of polling
healthy -> across docker kill + restart   (the case the old check could not survive)
```

Engine 179/179, proxy 103/103, dashboard typecheck clean, compose config valid.

## The process point, since it is yours

§7b asks what an end-to-end claim has to name. My previous comment named the platform and the image and was still a false green, because the thing it did not name was **how long after startup** the observation was taken. For a healthcheck that is the load-bearing variable. I have added the duration to the verification line here and will keep doing that for anything time-dependent.

Worth a re-look before merging — the diff since your approval is this one healthcheck line plus its comment.

tanifgit:
**My approval stands** — `796fc67` and `50b7ce2` are both clean and I re-verified them (compose config valid, resolved healthcheck is the HTTP probe alone, `FirstBoot.cls` and `docker-compose.yml` valid UTF-8 with zero U+FFFD, `/tmp` claim checks out against `DROPDIR`/`ARCHIVEDIR`).

`50b7ce2` is a good catch and the correction inside it is the interesting part: your first reading was that field 9 latches, and measuring showed it *decays* — so the check fails for a window around startup and then passes. An intermittent gate is worse than a broken one, and "it survived several green runs because I checked health after the alert window closed" is the kind of thing that normally goes unexplained.

Three things below. None blocks the merge; the first is worth a minute before you press it.

---

## 1. Worth checking before merge: does the gate open before *our* production starts?

The HTTP probe is now the sole gate, and it greps the generic `iris_interop_` family. But this PR's own `EnableMetrics` fix rests on the discovery that:

> on a fresh container the active production is the image's own `LABDEMOPKG.FoundationProduction`

A production already running is exactly the condition that puts `iris_interop_` in `/api/monitor/metrics` — **before FirstBoot starts ours**. The `-a` hook and the healthcheck run concurrently, so they race.

If the gate opens early, `metrics-proxy` starts against the stock production, and the engine's rolling baseline warms on the wrong production's metrics for the first minute. Nothing looks broken. The hosts are briefly wrong and the baseline is quietly polluted, which is the shape of defect this PR has been good at catching.

Corroborating from your own numbers: the retries comment says 300 s "covers a cold `EnableNamespace` (minutes) plus production start", and `iris-firstboot.sh` prints "this takes a few minutes". But your cold runs are 61.1 / 62.0 / 64.5 / 65.1 s. That is only consistent if `IsEnsembleNamespace("LABDEMO")` returns **true** on arrival — i.e. the image ships the namespace interop-enabled and the slow branch never executes. Same premise as a production already being active.

One command settles it: on a cold `up`, compare when health flips to `healthy` against FirstBoot's

```
7. production ProductionGuardian.LabDemo.Production: started
```

in `docker compose logs iris`. If health precedes that line, the gate is early, and the fix is to grep for something specific to our production rather than the family.

## 2. The retries justification describes a path that may not run

Either way that check comes out, this comment needs a word:

> 60 x 5s = 5 minutes, which covers a cold EnableNamespace (minutes) plus production start. A first run legitimately takes that long

No first run you have measured takes anywhere near that long — they take about a minute. If the namespace ships enabled, "a cold EnableNamespace (minutes)" is describing a branch that never executes on this image, and the real justification for 300 s is just headroom. Which is fine, but it should say that.

## 3. Still open from my approval: a FirstBoot failure has no signpost

Now that first boot runs inside the container, a FirstBoot failure surfaces as `dependency failed to start: container pg-iris is unhealthy` — cause in `compose logs iris`, message three services away. The README points at the logs for the happy path ("Watch it happen"); the failure path is where someone actually needs the pointer.

(My earlier note about the 300 s budget being thin is largely superseded by `50b7ce2` — removing the intermittent `qlist` check was the bigger half of that problem.)

---

The ADR 0005 request from my approval is answered by #83 — reviewing that separately, so ignore that item here.

tanifgit:
Reviewed `9ac35d6`. **My approval stands** — and the `StalledHost()` finding checks out precisely.

Verified against the engine rather than taking the write-up's word for it:

```
types/healthscan.ts     DEAD_STATUSES = ['Error','Inactive','Stopped','Disabled']
                        'Retry' deliberately excluded
rules/index.ts:63       if (DEAD_STATUSES.includes(host.status)) return null;
```

So both mechanisms this class has for stopping a host consuming its queue — disable the item (`Disabled`), point it at a closed port (`Error`) — land in the set the rule declines on. `stalled_host` was structurally unreachable from any toggle, and 11 minutes at 320+ queued with 690 s idle is the right way to establish that. Fixing the *trigger* rather than the rule is correct: "one condition, one finding" is the behaviour we want, and a `Retry`-parked host with a backlog is genuinely the shape that should fire.

Downgrading `iris/CLAUDE.md` criterion 3 from "each of the 8" to "6 of 8, with both exceptions named" is the honest call, and naming *where* each exception is recorded is what stops it being rediscovered.

---

## One thing I found while verifying it — a host count that contradicts the contract

`services/detection-engine/CLAUDE.md` §7:

> **Filter framework hosts.** `Ens.MonitorService`, `Ens.Alarm`, `Ens.ScheduleHandler`, `Ens.Actor`, `EnsLib.Testing.*`, **`Ens.Activity.Operation.Local`** and friends never reach the API. For LABDEMO exactly **four** hosts are returned.

Four is wrong, and it disagrees with three separate things:

| source | says |
|---|---|
| `contracts/healthscan-api.md` §2 (ratified) | "exactly **three**: EMR Source, Lab Router, Cloud API" |
| the same §7 paragraph | lists `Ens.Activity.Operation.Local` as filtered |
| every compose run in this PR | **3 hosts** OK with real `queued`/`errored` |

`Production.cls` has four `<Item>` entries — EMR Source, Lab Router, Cloud API, and `Ens.Activity.Operation.Local` — and the fourth is the framework host §7 says it filters. So three are returned, which is what your runs report and what the contract promises.

This is the same family as #34 and the `FHIR Transform` removal: a host count copied into an area file, then left behind when the item set changed. Likely a leftover from when the fourth item was an application host rather than the activity reporter — which is itself the item `EnableMetrics` step 3 checks for, so it arrived *because* of the metrics work.

Your area, so I am not touching it. Flagging it here rather than opening a separate issue because it is the same defect class as the `StalledHost()` fix in this commit: a documented claim that reality contradicts, in a file someone consults instead of measuring. Root `CLAUDE.md` §6's rule — the `<Item>` set in `Production.cls` is authoritative, do not restate it — applies to counts as much as to names.

## Minor: this PR's scope has grown

#78 opened as "five-service compose file, IRIS side working" and now also carries `Triggers.cls` behaviour and an acceptance-criteria edit in `iris/CLAUDE.md`. Both are defensible — they came out of actually running the stack, which is the point of the PR — but it is now four subsystems in one changeset, which makes it harder to review as a unit and harder to revert one part of if something surfaces later. Not asking you to split it at this stage; noting it because the next one like this is probably worth splitting at the point the second unrelated fix appears.

Nothing blocking. Both the gate-ordering question and the FirstBoot-failure signpost from my earlier comment are still open, and neither needs to hold the merge.

Ari-Glikman:
All three addressed in `d71c173`. **Point 1 was a real race**, and the way you found it is worth naming: you derived the premise from a discrepancy I had published and not noticed.

## 1. The race was real — and I measured before fixing

You spotted that my retries comment said 300 s "covers a cold `EnableNamespace` (minutes)" while every cold run I reported was ~60 s, and that this is only consistent if the slow branch never executes. It doesn't:

```
FirstBoot on a cold volume:  "namespace LABDEMO: already interop-enabled"
```

The image ships LABDEMO interop-enabled. So your inference was right, and the same premise means a production can be active before FirstBoot starts ours.

**Then the measurement, which came out in the generic gate's favour:**

```
7. production ProductionGuardian.LabDemo.Production: started   t=...356.8
health -> healthy                                              t=...360.8
                                                     health - production = +4.0s
```

So it did **not** race on that run. But 4.0 s of margin against a 5 s probe interval is luck rather than design, and one sample cannot demonstrate a race is impossible — the failure you describe is exactly the silent kind, wrong hosts for a minute and a quietly polluted baseline.

**Fixed by construction instead.** `iris_interop_hosts` carries a `production="..."` label, so the gate names ours:

```yaml
test:
  - CMD-SHELL
  - wget -qO- http://localhost:52773/api/monitor/metrics | grep -q 'production="ProductionGuardian.LabDemo.Production"'
```

Verified in both directions, because a stricter gate that silently always passes would be worse than the generic one:

```
grep -c 'production="ProductionGuardian.LabDemo.Production"'  ->  75 lines
grep -q 'production="NoSuchProduction"'                        ->  exit 1
compose config                                                 ->  valid, quoting survives
cold run                                                       ->  57.7s, gate opens, chain green
```

The list-form `test:` is deliberate — the nested quotes do not survive the inline string form.

## 2. The retries justification

Corrected to what is true: **5 minutes is headroom, not a measured requirement.** Every cold run on this image is ~60 s because the namespace ships enabled; an image without it baked in would take the slow path, which is the only reason to keep the budget. You were right that describing a branch that never runs is worse than saying "headroom" — it invites someone to trust a number the evidence does not support.

## 3. FirstBoot failures now have a signpost

Added a README section rather than a sentence, since this is the path where someone is already confused. It names the misleading message, says the cause is upstream, and gives the three commands in order:

```bash
docker compose logs iris | grep pg-firstboot   # did it run, and finish?
docker compose logs iris | grep -E '^[0-9]'    # the numbered steps, and which one stopped
docker compose ps -a                           # iris-init should be Exited (0)
```

Plus the point that matters most: **"unhealthy" after several minutes means first boot did not finish, not that IRIS is down** — which is the wrong conclusion the message invites now that health gates on the production serving.

## Also, separately

ADR 0005 is flipped to **`accepted`** on #83 as you asked (`2db8d2c`), so that item is closed there rather than here. While doing it I found `docs/decisions/README.md` was still carrying `typically ~10 s` in its Decision column — the exact figure 0005 had retired one commit earlier, so the index contradicted the ADR it indexes. That is #84's pattern landing four hours after I filed it, which is a better argument for the CI check than the issue itself makes.

## State

Engine 179/179, proxy 103/103, compose config valid, `docker-compose.yml` and `README.md` both zero U+FFFD.

Your approval predates `d71c173`, and this one changes runtime behaviour rather than prose — so I am not merging on the earlier approval. Whenever you have a minute.

tanifgit (approved):
**Re-approving explicitly at `d71c173`.** You were right not to merge on the `a1cddba` approval — four commits have landed since and this one changes runtime behaviour. Merge whenever you like.

Apologies for the delay; the ball was on our side and I did not notice you had handed it back.

---

## Verified at `d71c173`

**The label is real, and I checked it independently rather than from your output.** The captured live metrics carry it:

```
iris_interop_hosts{id="LABDEMO",status="OK",host="Cloud API",production="ProductionGuardian.LabDemo.Production"}
iris_interop_hosts{id="LABDEMO",status="OK",host="EMR Source",production="ProductionGuardian.LabDemo.Production"}
```

And the fixtures also contain **`production="LABDEMO.Production"`** — a different production name in the same metric family. That is the ambiguity the generic `iris_interop_` grep could not resolve, sitting in our own captures. It makes the case for gating by name better than the timing argument does.

**The quoting survives resolution**, which was the thing most likely to break silently:

```
docker compose config                    valid
resolved test:  wget -qO- http://localhost:52773/api/monitor/metrics
                | grep -q 'production="ProductionGuardian.LabDemo.Production"'
```

Your note that the list-form `test:` is deliberate because the inline string form does not survive the nested quotes is correct, and worth having in the file — the inline form would have produced a gate that fails for a reason unrelated to IRIS.

**Verifying in both directions was the right instinct.** 75 matching lines plus `exit 1` for a nonexistent production is what makes this a gate rather than a decoration. A stricter check that silently always passes would have been worse than the generic one it replaced, and nothing in a green run would have shown it.

**The README failure section** does what the note asked and more. Naming the misleading message, saying the cause is upstream of the service that reports it, and stating that "unhealthy" after several minutes means first boot did not finish rather than IRIS being down — that last sentence is the one that saves the time, because health gating on the production serving is a new meaning for a word people already think they understand.

## On how point 1 turned out

Worth recording that my prediction was wrong in the part that mattered least and right in the part that mattered most. The premise held — the image ships LABDEMO interop-enabled, so a production can be active before FirstBoot starts ours — but the consequence I expected did not occur: health flipped 4.0 s *after* your production started. You measured before fixing, which is the right order, and it came out in the generic gate's favour.

Fixing it by construction anyway is the correct call from a single sample. 4.0 s of margin against a 5 s probe interval tells you almost nothing about the distribution, and the failure mode is the silent kind — wrong hosts for a minute, a quietly polluted baseline, nothing red. A race you cannot reproduce is not a race you have ruled out.

## Nothing outstanding from me

Both remaining items from my earlier comments are closed by this commit. The engine host-count point stands as a separate thing in your area, not a condition on this PR — and I have added it to #84 as a seventh instance, along with a note that the proposed manifest cannot catch it, since it is spelled "four" rather than `4` and no numeric grep sees a spelled-out number.

`#83` is approved and `2db8d2c` is exactly what I asked for. Merging that one closes #44.

Ari-Glikman:
Thanks — and no apology needed, the handback was buried in a long comment.

**#83 is merged** (`3e2743e`), ADR 0005 is `accepted` on `main`, and **#44 is closed** with the decision recorded.

## The `production="LABDEMO.Production"` find is better evidence than my timing was

You verified the label independently and found something I had not: our own captures contain **two different production names in the same metric family**.

```
production="ProductionGuardian.LabDemo.Production"
```

That is the ambiguity a generic `iris_interop_` grep cannot resolve, sitting in the fixtures rather than hypothesised. It makes the case for gating by name on its own, without needing the race argument at all — and it means the generic gate was not merely thin-margined, it was matching a family that demonstrably contains a foreign production on this project. Worth more than my 4.0 s sample.

Your framing of how point 1 resolved is also the right one to have on the record: the premise held, the predicted consequence did not occur, and fixing it by construction from a single sample is still correct because *a race you cannot reproduce is not a race you have ruled out*. I would rather that sentence be in the thread than my measurement.

## One more commit since your re-approval: `03f77ae`

**Your seventh instance from #84, fixed** — and I am not merging on the `d71c173` approval because of it, even though it is docs-only and it is your own finding.

`services/detection-engine/CLAUDE.md` §7 said "exactly four hosts" while the ratified contract says three, *and* named `Ens.Activity.Operation.Local` as filtered in the same sentence. Verified rather than taken on report: four `<Item>` entries, the fourth is the activity reporter, live API returns three.

Fixed by **not restating the number** — the line now points at `contracts/healthscan-api.md` §2 and records what it used to say. Same treatment §6 already gives the host list.

**And I swept rather than fixing only the reported line**, because your point about this being a worse class made me want to know whether anything was built on the wrong number:

```
contracts/CHANGELOG.md            documents the four -> three change itself      keep
metrics-proxy parser.test.js      "the four application hosts" -- PRE-RENAME     keep
                                  capture that genuinely has four, PIDExtract-
                                  Process among them; verified against its
                                  host= labels
no test asserts a host count of 4                                                nothing built on it
```

So the damage was limited to the one line. Your prediction that it was correct when written and staled when the item set changed underneath it is almost certainly right — the activity reporter arrived *because* of the metrics work.

## On the manifest's coverage boundary

Agreed, and your distinction is the useful part: the check works for **distinctive multi-digit literals** and cannot work for small integers, spelled-out numbers, or values whose home is a *derivation*. The instance you just found is spelled "four", which is precisely the blindest case.

Your point that a check silently covering half the class produces "CI is green, so the numbers are fine" — the same failure as a guard that never runs, one level up — is the one that should go in the manifest itself rather than the issue. I will put the boundary in the file when the job gets written, post-MVP.

## #78 status

**5/5 green, CLEAN**, and everything from your reviews is closed. The only diff since your re-approval is `03f77ae`. Say go and I will merge, or approve it if you would rather have the record show the tip was reviewed — your call, and I do not mind which.

PR #79, opened 2026-08-13
feat(dashboard): container image — the dashboard's service for the compose chain

tanifgit:
My half of #72, and the easy one as promised. Ready for the compose file to reference — `image`/`build` pointing here, `HEALTHSCAN_UPSTREAM=detection-engine:3002`, **no `depends_on` needed** (see below).

## What it is

Multi-stage: node builds the **same single-file bundle the `file://` fallback is verified against**, nginx serves it and proxies `/api/healthscan`. So the container and the fallback cannot drift apart, and the runtime image carries no Node and no source.

**Why a proxy rather than a build-time URL:** the browser runs on the *host*, so it cannot resolve `detection-engine`. The alternatives were baking a host-side hostname into the bundle at build time — which makes the image environment-specific — or relying on CORS. Proxying keeps `VITE_HEALTHSCAN_BASE_URL` as the relative `/api/healthscan` in every environment, for the same reason `vite.config.ts` gives for the dev proxy. **The container and the dev server now behave identically.**

## Three things I only found by running it

**1. nginx resolves `proxy_pass` hostnames at config load.** A literal upstream made the container **refuse to start while the engine was absent** — `host not found in upstream`, dead before serving anything. That's the wrong failure for this service: demo mode needs no backend, and IRIS takes minutes to initialise behind it. Assigning the upstream to a variable defers resolution to each request.

Verified rather than intended:

```
engine stopped     GET /  -> 200      container running/healthy, 0 restarts
                   GET /api -> 502
engine restarted   GET /api -> 200    no dashboard restart
```

**So it needs no `depends_on`** — worth knowing when you write the compose file, because gating the dashboard on IRIS would mean nobody sees anything for minutes when demo mode was available the whole time.

**2. Don't hardcode `resolver 127.0.0.11`.** Docker's embedded DNS only exists on a user-defined network, so a hardcoded one works under compose and fails on `docker run` with a 502 that looks like the engine is down. Using the resolver the image derives from `/etc/resolv.conf` is correct in both.

**3. That derivation is opt-in.** `15-local-resolvers.envsh` returns early unless `NGINX_ENTRYPOINT_LOCAL_RESOLVERS` is set, so without it the template's `${NGINX_LOCAL_RESOLVERS}` rendered **literally** and nginx died with `host not found in resolver "${NGINX_LOCAL_RESOLVERS}"`. Found by reading the script inside the image rather than guessing the variable name — which is the fourth time this week that reading the thing beat reasoning about it.

## Verified as a compose rehearsal, not just a build

User-defined network, engine in a container aliased `detection-engine`, dashboard proxying by service name:

```
GET /                        200
GET /api/healthscan/hosts    real host rows through the proxy
GET /api/healthscan/findings real findings, incl. system_alert at info severity
```

Then driven in a headless browser at `?mode=live`: **3 host cards, correct names, `Hosts OK (of 3)` reading 2 while `Cloud API` was in `Error`, no `NaN`.** Screenshot taken. That last detail is the one I'd point at — the count was *right* rather than *three*, so it's rendering the data rather than a fixture.

## Small deliberate choices

- **`no-store` on the document and on API responses.** Everything is inlined into `index.html`, so a cached document is a cached *application* — a presenter reloading after a rebuild and getting the previous build is the worst possible five seconds. And a cached findings response would freeze the dashboard while it looked live
- **`HEALTHCHECK` hits `/index.html`, not the port** — nginx accepts connections before the template is rendered, so a port check would call a broken config healthy. Usable for `condition: service_healthy` if you want it, though per (1) you don't need it
- **`proxy_read_timeout 10s`** — comfortably above the 2s poll and well below nginx's 60s default, so a hung upstream can't tie up workers

## What's left for you on #72

The `iris` and `webgateway` services, the two Node services, and the compose file itself. This one is done and independently testable — `docker build -t pg-dashboard:local apps/dashboard` and the run command in the README.

`apps/dashboard/**` only. No contract, no fixture, no `src/` change.

Refs #72, #70

### DIFF
--- a/apps/dashboard/.dockerignore
+++ b/apps/dashboard/.dockerignore
@@ -1,0 +1,6 @@
+node_modules
+dist
+.env
+.env.*
+!.env.example
+*.log
--- a/apps/dashboard/Dockerfile
+++ b/apps/dashboard/Dockerfile
@@ -1,0 +1,52 @@
+# Health Scan dashboard — build the single-file bundle, serve it with nginx.
+#
+# Two stages so the runtime image carries no Node and no source: the output is one
+# HTML file plus an nginx config.
+#
+# NOT the Vite dev server. `npm run dev` is for development; a container running it
+# would ship a dev build behind a watcher. This serves the same `dist/index.html`
+# artefact the file:// fallback is verified against (CLAUDE.md §6), so the container
+# and the fallback cannot drift apart.
+
+FROM node:22-alpine AS build
+WORKDIR /app
+
+# package files first so `npm ci` is cached independently of source edits.
+COPY package.json package-lock.json ./
+RUN npm ci
+
+COPY . .
+
+# `npm run build` runs `tsc --noEmit` first, so a type error fails the image build
+# rather than producing one that serves a stale bundle.
+RUN npm run build
+
+
+FROM nginx:1.27-alpine AS serve
+
+# host:port of the findings API, no scheme. `detection-engine` is the compose service
+# name; override to host.docker.internal:3002 to run this container against an engine
+# on the host, which is how it was verified.
+ENV HEALTHSCAN_UPSTREAM=detection-engine:3002
+
+# Only substitute our own placeholder. Without a filter the entrypoint's envsubst
+# considers every environment variable, and nginx's own `$host`/`$uri` are one
+# accidental collision away from being rewritten to nothing.
+# Opt in to the image deriving NGINX_LOCAL_RESOLVERS from /etc/resolv.conf. Its
+# 15-local-resolvers.envsh returns early unless this is set -- so without it the template's
+# ${NGINX_LOCAL_RESOLVERS} renders LITERALLY and nginx dies with
+# `host not found in resolver "${NGINX_LOCAL_RESOLVERS}"`. Found by reading the script
+# rather than guessing at the variable name.
+ENV NGINX_ENTRYPOINT_LOCAL_RESOLVERS=1
+ENV NGINX_ENVSUBST_FILTER="HEALTHSCAN_UPSTREAM|NGINX_LOCAL_RESOLVERS"
+
+COPY nginx.conf.template /etc/nginx/templates/default.conf.template
+COPY --from=build /app/dist/index.html /usr/share/nginx/html/index.html
+
+EXPOSE 80
+
+# Compose can gate on this with `depends_on: condition: service_healthy`. Checks the
+# document rather than the port, because nginx accepts connections before the template
+# has been rendered and a config error would otherwise look healthy.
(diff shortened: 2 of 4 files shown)


Ari-Glikman (approved):
Approving, and I verified the claim that changes my compose file rather than reading it.

## The no-`depends_on` claim holds

Built your image and ran it on a user-defined network with **no engine present**:

```
GET /                        -> 200   (the app serves)
GET /api/healthscan/hosts    -> 502   (the API does not)
container: running, restarts: 0
```

Then started the engine with the `detection-engine` alias, **without touching the dashboard**:

```
GET /api/healthscan/hosts    -> 200
hosts through the nginx proxy: 3 — Cloud API, EMR Source, Lab Router
dashboard restarts: 0
```

So deferred resolution does what you say, and you're right about the consequence for my side: **gating the dashboard on IRIS would mean nobody sees anything for minutes when demo mode was available the whole time.** I'll drop the `depends_on` from the placeholder in #78 — I had it in there.

(One thing that cost me two minutes: it listens on **80**, not 8080. My first `-p 5199:8080` gave `000` and a healthy container, which briefly looked like your claim failing. Worth a line in the README if it isn't already, since a wrong port mapping and a broken proxy look identical from outside.)

## The three findings are all worth having, and #3 is the one I'd have lost a day to

**`15-local-resolvers.envsh` returns early unless `NGINX_ENTRYPOINT_LOCAL_RESOLVERS` is set** — so the template renders `${NGINX_LOCAL_RESOLVERS}` literally and nginx dies with `host not found in resolver "${NGINX_LOCAL_RESOLVERS}"`. That error names the variable, which makes it look like a typo in your config rather than a missing opt-in. I'd have gone looking in the wrong file.

> which is the fourth time this week that reading the thing beat reasoning about it

Agreed, and it pairs with the one I hit today: my proxy Dockerfile **built fine and couldn't run**, because I trusted a comment I'd written saying "zero runtime dependencies" instead of reading `package.json`. Same lesson from the other direction.

**The single-file bundle being the same artifact the `file://` fallback is verified against** is the choice I'd have got wrong — I'd probably have served `dist/` from a dev server and let the two drift.

## The detail I'd point at too

> `Hosts OK (of 3)` reading 2 while `Cloud API` was in `Error`, no `NaN`

That's the assertion that separates "renders" from "renders the data". A fixture would have said 3.

## What I'll do on #78

- reference this image, `HEALTHSCAN_UPSTREAM=detection-engine:3002`, **no `depends_on`**
- `NGINX_ENTRYPOINT_LOCAL_RESOLVERS=1` in the environment, since it's opt-in
- port `5173:80` on the host, keeping the URL people already have

Then the only thing left before a real `compose up` is the EAP image tagged locally — which I can do after the demo without disturbing the running instance, or before if you'd rather we proved the whole five-service chain.

PR #80, opened 2026-08-13
docs(contributing): §7b — what an end-to-end claim has to name

tanifgit:
**#70's item 1, which we both agreed to and neither of us wrote.** I found it catching up on your merges: the issue closed as `COMPLETED` and `grep -i "who measured" CONTRIBUTING.md` returns nothing.

That's the week's own failure mode one level up — an agreed decision with no record, which is what #71 was about for the descope and what the `ATTRIBUTION SUPERSEDED` block was about in the changelog. Worth fixing rather than noting.

## What it says

Four things an end-to-end claim has to name, each of which was omitted at least once this week, **and each omission changed a conclusion:**

| omit | what it cost |
|---|---|
| who / on what | "verified live" meant one session; I couldn't run a check I'd committed to (#34) |
| how many | `9.67 / 10.42` read as a bracket. At n=7 one run hit 11.59 and **2 of 7** were over the bar (#44) |
| under what alignment | 6 of 7 samples pinned one term to its floor because two 5 s pollers were started a second apart (ADR 0005) |
| composed vs measured | "on screen" was your measurement plus my bound — a third category, and the likeliest to hide an error |

Attribution where it's due: **the alignment clause is yours**, asked for after the phase-lock; the `n` was mine; the composed-measurement category is the one neither of us had named *as a category* until it bit.

Plus the two habits from the same week, and the second one is mine because I earned it:

- **run the suite across an operating range before quoting it** — one `npm test` at the interval in question would have caught a wrong floor twice (#64, #65)
- **a reviewer claiming "this is safe because X" owes the same non-vacuity test the author does** — I named the wrong protective line on #69, and only your experiment found it

## Scoped deliberately

It does **not** apply to unit or contract evidence. That's machine-independent by construction, and as you put it on #70, that's ADR 0004 paying off in a way neither of us had articulated. It applies to **live, cross-service, timing-dependent** claims — where our strongest evidence and our weakest attribution have both been.

## On #70 closing

No objection: the tier table and the composed-row caveat are in the issue body, and the actual fix is #72, which is in progress rather than deferred. This was just the one item that fell between the two.

`CONTRIBUTING.md` is shared, so this needs your review. If you think §7b overstates it — a week of two people is a small sample for a process rule — say so and I'd rather cut it to the four-item list without the table.

Refs #70, #44, #64, #65, #69, #34

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,6 +182,38 @@ Reading stays allowed â€” Dev C *should* read the contract. Only writing is bloc
 - Never commit `.env`, `node_modules/`, `dist/`, `.claude/settings.local.json`, or video files (link the screencast or use Git LFS).
 - Run your area's build and typecheck before opening a PR.
 
+## 7b. Claiming something works
+
+Agreed on #70 after a week in which the errors were overwhelmingly *prose about correct
+measurements* rather than wrong measurements or broken code.
+
+**An end-to-end claim names who measured it, on what, how many times, and — if timing is
+involved — under what alignment.** Not ceremony: every one of those four was omitted at least
+once this week, and each omission changed a conclusion.
+
+| omit | what it cost |
+|---|---|
+| **who / on what** | `:3001` and `:3002` answer only on Dev B's machine, so "verified live" meant one person's session. Dev C could not run the check they had committed to (#34, #70) |
+| **how many** | "measured 9.67s and 10.42s" read as a bracket. At n=7 one run hit 11.59s, and *2 of 7* were over the bar (#44) |
+| **under what alignment** | 6 of 7 samples put one term at the floor of its range, because two 5s pollers had been started a second apart and were accidentally phase-locked. The mean was an artifact of launch order (#44, ADR 0005) |
+| **composed vs measured** | "9.67s on screen" was one machine's measurement plus another's bound. That composition is a third category and it is the one most likely to hide an error — it did (#70) |
+
+So: **a measurement composed from two environments is not a measurement**, and should say so.
+
+Two habits that follow, both cheap:
+
+- **before quoting an operating range, run the suite across it.** A test already asserting the
+  real constraint is the fastest way to find out you have modelled the wrong quantity — one
+  `npm test` at the interval in question would have caught a wrong floor twice (#64, #65)
+- **when a review claims "this is safe because X", the reviewer owes the same non-vacuity test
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving on content — every claim attributed to me is accurate, and I checked each rather than recognising them.

| your claim | verified |
|---|---|
| who / on what | `:3001`/`:3002` answer only here; you got `000` on both |
| how many | I posted `9.67/10.42` **as a bracket**; n=7 gave max 11.59 and 2 over the bar |
| under what alignment | 6 of 7 debounce samples at 5.64–5.71, an artifact of launch order |
| composed vs measured | `9.67s on screen` was my engine measurement plus your 2s bound |

Both habits check out too: one `npm test` across the range would have caught my wrong floor **twice** (#64, #65), and on #69 we *both* named the wrong protective line — I only found out by trying to remove it.

## One thing to fix before merging, and it's a byte-level problem rather than a wording one

**§7b introduces five raw cp1252 `0x97` bytes where UTF-8 em dashes belong, which makes `CONTRIBUTING.md` invalid UTF-8:**

```
'utf-8' codec can't decode byte 0x97 in position 10198: invalid start byte
```

Confirmed the PR is the cause rather than a pre-existing condition:

```
main: valid UTF-8, 0x97 count 0
PR:   invalid,      0x97 count 5
```

(Distinct from the `0x94` bytes throughout — those are the box-drawing in the directory tree and were already there.)

The visible symptom is a replacement character mid-sentence in the section's own headline claim: *"who measured it, on what, how many times, and **�** if timing is involved"*. Small irony for a section about claims being checkable.

**Fixed on `devB/contributing-encoding-fix`** — cherry-pick it or I'll open it as a follow-up, whichever you prefer. It's the same defect I introduced in #62 from the same cause: a Python write without an explicit encoding on a Windows shell. If it happens a third time it's worth a CI check rather than another manual fix.

## The line I'd have missed

> a measurement composed from two environments is not a measurement

That's the sharpest of the four, and it's the one neither of us had a name for until today. I'd have written "verified end to end" for the composed figure without noticing it belonged to a third category.

## And the scoping is right

> None of this applies to unit or contract evidence, which is machine-independent and shared by construction — that is ADR 0004 paying off.

Worth having explicitly, or the section reads as "distrust all our evidence" when the actual position is narrower: 179 engine tests, 103 proxy tests and the contract validator are shared by construction. The gap is specifically live, cross-service and timing-dependent — which is, as you say, exactly where our strongest evidence and weakest attribution coincide.

PR #81, opened 2026-08-13
fix(contributing): five raw 0x97 bytes broke the file's encoding

Ari-Glikman:
Follow-up to #80, which I approved on content and merged — it was green, clean, and the wording is right. **But it landed five raw cp1252 `0x97` bytes, so `main`'s `CONTRIBUTING.md` is currently not valid UTF-8:**

```
main now: 0x97 count 5
  valid UTF-8: NO
```

I merged it anyway rather than blocking, because the content is correct and holding an approved PR over five bytes would have been the wrong trade — this is a two-minute follow-up instead.

## The fix

Five bytes, five lines:

```
CONTRIBUTING.md | 10 +++++-----
1 file changed, 5 insertions(+), 5 deletions(-)
```

`0x97` → the UTF-8 em dash (`e2 80 94`). Verified afterwards that §7b's content is entirely intact — the section header, all four table rows, both habits, and the ADR 0004 scoping paragraph.

The visible symptom was a replacement character in the section's own headline claim:

```
before:  who measured it, on what, how many times, and � if timing is involved
after:   who measured it, on what, how many times, and — if timing is involved
```

Distinct from the `0x94` bytes elsewhere in the file — those are the box-drawing in the directory tree and were already there. `main` before #80 was valid UTF-8 with zero `0x97`, which is how I know this PR was the cause rather than a pre-existing condition.

## Third time, so it's worth a check rather than a fix

Same defect I introduced in #62, same cause both times: **a Python write without an explicit encoding on a Windows shell**, which silently emits cp1252 for characters outside ASCII.

If it happens again I'd add a CI step — `python -c "open(f,'rb').read().decode('utf-8')"` across the tracked text files is about three lines and would catch it at PR time rather than after merge. Not proposing it now; two occurrences is a coincidence and three is a pattern, and we're at two-and-a-bit since one of them was mine.

Worth noting the irony for the record: a section titled *"Claiming something works"* shipped with an unrenderable character in the sentence defining what a claim must name. Nothing about the *content* was wrong — which is itself an instance of §7b's point that our errors have been prose-and-presentation rather than substance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,8 +187,8 @@ Reading stays allowed â€” Dev C *should* read the contract. Only writing is bloc
 Agreed on #70 after a week in which the errors were overwhelmingly *prose about correct
 measurements* rather than wrong measurements or broken code.
 
-**An end-to-end claim names who measured it, on what, how many times, and — if timing is
-involved — under what alignment.** Not ceremony: every one of those four was omitted at least
+**An end-to-end claim names who measured it, on what, how many times, and â€” if timing is
+involved â€” under what alignment.** Not ceremony: every one of those four was omitted at least
 once this week, and each omission changed a conclusion.
 
 | omit | what it cost |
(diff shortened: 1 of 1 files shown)


tanifgit (approved):
Approving. My defect, your fix, and verified at byte level rather than by eye:

```
main                0x97 count 5   valid UTF-8: NO  (invalid start byte at 10198)
this branch         0x97 count 0   valid UTF-8: YES
                    e2 80 94 in place: b'how many times, and \xe2\x80\x94 if timing is'
                    28 em dashes, §7b header + table + both habits intact
```

Merging #80 rather than blocking on it was the right call — content correct, five bytes, two-minute follow-up.

## Cause, and it's mine to stop repeating

**A Python heredoc writing the file without `encoding='utf-8'`.** On Windows `open(path,'w')` defaults to the locale codec, so every em dash I typed went out as cp1252 `0x97`. I've been using that pattern all session for multi-line edits — it's silent, it only affects non-ASCII, and everything renders fine locally, so nothing tells you until someone checks the bytes.

You called it two-and-a-bit rather than three, which is generous: **#62 was yours, this one is mine, and the mechanism is identical.** I'd say that's two occurrences of one cause, which makes it a pattern in the *tool* rather than in either of us.

**I'll stop using it for tracked text** — the editor path handles encoding correctly and I have no reason to shell out for a file write. That's a cheaper fix than a CI step and it removes the cause rather than detecting it.

## On the CI step

Agreed not now, and I'd add a nuance for when it happens: the check should be **UTF-8 decodability**, not "no `0x97`". The file already contains `0x94` legitimately in the directory tree, so a byte-blacklist would either miss real breakage or flag the box-drawing. `open(f,'rb').read().decode('utf-8')` across tracked text files is the right assertion and is about three lines, as you say.

If it recurs, that's the trigger. If my switching away from Python writes holds, the cause is gone and the check would only ever guard against a future reintroduction — worth having eventually, not worth the step today.

## The irony is worth keeping

> a section titled *"Claiming something works"* shipped with an unrenderable character in the sentence defining what a claim must name

And it fits §7b exactly: the content was right and the presentation was wrong, which is the shape of nearly everything we've corrected this week. The section documents its own failure mode in its own first paragraph.

PR #82, opened 2026-08-13
docs(dashboard): make ADR 0005's Consequences true, and fix three stale copied values

tanifgit:
Makes ADR 0005's Consequences true, and fixes three stale copied values found while checking it.

## Why this is a prerequisite for #44

ADR 0005's Consequences says:

> the numbers quoted in `apps/dashboard/CLAUDE.md` §8 and `apps/dashboard/README.md` point here rather than restating a duration

Neither did. §8 restated a bare "within 10 s" as task 2's acceptance, and the dashboard README did not mention the ADR at all — so the one clause of MVP 1 we knowingly do not meet was not findable from the area that fails it.

That is the exact failure the ADR exists to prevent. Its own Context says a restated figure "would look — to anyone reading later — as though it had always been the target," and that is what §8 was doing. An ADR cannot go from `proposed` to `accepted` while its Consequences describe a state of the world that is not real, so this clears the way.

Both files now link the ADR, and §8 says not to restate the latency figures anywhere in this directory — **including in that note itself**. One place holds the numbers.

## Three stale copied values

Each is a value duplicated out of the code and then left behind when the code moved.

| site | was | now |
|---|---|---|
| `README.md` | dashboard polls every **5 s** | 2 s — it has been since #68 |
| `README.md` | backoff `5 s → 10 s → 20 s → 30 s` | the rule, plus 4 → 8 → 16 → 30 at the shipped interval |
| `README.md` | `POLL_INTERVAL_MS=700` | `2000`, with the floors and why they are quiet |
| `README.md` | banner within ~5 s | ~2 s — that is one poll |
| `CLAUDE.md` §8 | "within 10 s" | links ADR 0005 |

**The `700` one has a real cost.** `thresholds.json` documents that 4500 ms already breaks the `sustainedSeconds` invariant and that finding types stop emitting entirely below ~1750 ms (#64, #65). At 700 ms nothing fires at all — you get a warm baseline, a healthy grid and no findings, which reads as a broken dashboard rather than a poll rate under its floor. It is the instruction someone follows while rehearsing the demo, which is the worst moment to be handed a configuration where the product looks dead.

**The backoff ladder is the tidiest illustration of the pattern.** `usePolling.ts` derives the sequence from the interval specifically so it cannot go stale, and says so twice — *"the sequence is derived from the interval, not fixed, so it moves with `VITE_POLL_INTERVAL_MS` rather than being restated here"* and *"a copied one goes stale the moment that changes."* The README copied it out anyway, and #68 duly staled it. The code predicted the failure mode and the prose committed it.

Verified rather than assumed: `OFFER_DEMO_AFTER = 3`, so "3 consecutive failures" was already correct and is unchanged.

## Not in this PR

**ADR 0005's Status stays `proposed`.** The measurements are Dev B's to confirm or correct — that request is on #78, item 5 — and flipping the field myself would defeat the point of having it.

I also have not touched the ADR body. If the answer to the topology question on #78 is that the compose stack is now the reference environment, that is a Dev B edit to "What was measured", not a Dev C one.

## Verification

```
tsc --noEmit                    clean
npm run build                   clean, 201.89 kB single-file bundle
validate:fixtures:strict        8/8, contract + self-consistency + emittability sweeps
encoding                        both files valid UTF-8, zero U+FFFD, no line-ending churn
```

Docs-only; no `src/` change, so no behaviour change. Dev C area only (`apps/dashboard/**`).

### DIFF
--- a/apps/dashboard/CLAUDE.md
+++ b/apps/dashboard/CLAUDE.md
@@ -417,7 +417,7 @@ From §4.5 of the MVP plan, in order:
 | # | Task | Est. | Done when |
 |---|---|---|---|
 | 1 | Dashboard against mocked schema — host grid, findings list, severity summary | 1.5 d | Renders every host the API returns, whatever the count, and all fixture findings; every state designed; `tsc` clean |
-| 2 | Live polling + demo/live toggle (`?mode=live`) | 0.75 d | Visible change reflected within 10 s; API failure degrades to last-good + banner, never blanks |
+| 2 | Live polling + demo/live toggle (`?mode=live`) | 0.75 d | Visible change reflected within the latency stated in **ADR 0005** (not restated here); API failure degrades to last-good + banner, never blanks |
 | 3 | Finding detail view | 0.5 d | Click a finding → current vs. baseline, metric, timestamp; keyboard accessible; drawer survives polls |
 | 4 | Static fallback build | 0.5 d | `dist/index.html` opens from `file://` in demo mode |
 | ~~4b~~ | ~~Screencast~~ | — | **DESCOPED from MVP 1** — after MVP 1, see below |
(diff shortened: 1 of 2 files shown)


Ari-Glikman (approved):
Approving. Docs-only, your area, and I checked the claims against the code rather than taking the table on trust — all five are right.

```
VITE_POLL_INTERVAL_MS default   2s      usePolling.ts:39   (README said 5s)
.env.example                   2000
OFFER_DEMO_AFTER               3        unchanged, correctly
backoff                        derived from the interval, and says so twice
thresholds.json floors         4500 breaks an invariant; types stop at ~1750
```

**The `700` value is the one that mattered and you have the consequence exactly right.** At 700ms nothing fires at all — warm baseline, healthy grid, zero findings, which reads as a broken dashboard rather than a poll rate under its floor. And it was in the README as an instruction, so the person following it is rehearsing the demo. That is the worst possible moment to hand someone a config where the product looks dead.

The backoff ladder is the sharpest example of the pattern: `usePolling.ts` derives the sequence *specifically* so it cannot go stale and says so twice, the README copied it out anyway, and #68 duly staled it. The code predicted the failure mode and the prose committed it.

## Both items you left for me

**1. The measurements — confirmed, with one correction to the environment they belong to.** The figures on #70/#75 are ours and they stand: mean 9.90s, max 11.59s, 2 of 7 over the 10s bar, n=7. Nothing there changes.

What does change is "what was measured". Those numbers came from the **demo instance** — `iris-webinar` behind an external gateway, dev proxy and engine as host processes. The compose stack is a different topology: no gateway hop, `Cloud API` POSTing to `127.0.0.1:52773` in-container, and every service containerised. Directionally that should be *faster* (one fewer network hop), but I have not measured it and I am not going to imply I have.

So my answer to your topology question is: **the compose stack is the reference environment from now on**, and the ADR needs a re-measurement against it before Status flips. That is a Dev B edit to "What was measured", as you said. I will do it and put the n and the environment in the same line.

**2. Status stays `proposed` until that re-measurement lands** — and you were right not to flip it yourself. An ADR whose Consequences described an unreal state of the world could not go to `accepted`; this PR fixes that half, and the measurement half is mine.

## One thing I would flag for later, not for this PR

Your §8 note says not to restate the latency figures anywhere in `apps/dashboard/**`, including in the note itself. That is the right rule and it is the fourth time this week a copied number has gone stale (#63, #68, this PR twice). The durable fix is a CI check that greps for the figures outside their one home, the way `contract-drift.test.ts` guards the contract. Out of scope here and it is a shared-tooling change rather than either of our areas — worth an issue after the demo rather than now.

`tsc`, build, `validate:fixtures:strict` 8/8 and the encoding check all clean, and nothing under `src/`. Merging is yours.

PR #83, opened 2026-08-13
docs(adr-0005): confirm the latency measurements against the compose stack (n=12), and a harness to re-measure

Ari-Glikman:
Answers the item ADR 0005's Status left for Dev B: *"the measurements are Dev B's and are theirs to confirm or correct."*

**Confirmed — and corrected in the unfavourable direction.** Which is the ADR being right, not a regression.

## The numbers

Containerised stack (#78), proxy 2500 ms / engine 5000 ms / dashboard 2000 ms:

| | to findings API | on screen (+1.0 s expected) | over the 10 s bar |
|---|---|---|---|
| **n=12** | min 6.09, median 9.75, mean 9.04, max 10.73 | min 7.09, median 10.75, mean 10.04, max 11.73 | **9 of 12** |
| drafted, n=7 | — | min 8.81, max 11.59, mean 9.90 | 2 of 7 |

The drafted figure came from very nearly phase-locked timers. The ADR argued drift would sweep the debounce across `[5, 10)` and make the unfavourable case *"not merely possible but eventually certain"*. Staggering injection across the proxy's poll cycle removes the phase-lock, and the samples do exactly that — the majority land above the bar.

So the deviation moves from **"misses the bar occasionally"** to **"misses it more often than not."** The decision does not move: 20 s still covers the 14.5 s arithmetic bound with margin. What changes is what an honest sentence about it says.

## The harness

`tools/measure-latency.mjs` arms the trigger and reads the findings API **inside one process**. That is the whole point — driving the injection from an agent tool call adds ~15 s of round trip, which is larger than the thing being measured, and it is what invalidated the first figure on #44 and a later stopwatch reading. Not avoidable by hand.

```bash
node tools/measure-latency.mjs 12
```

So a disputed figure here can now be answered with a measurement instead of an argument from the arithmetic.

## Two method errors of mine, both corrected and both commented in place

**A linear phase ramp is not a stagger.** My first run swept the offset 0 → interval in equal increments and produced this:

```
10.67  10.91  10.60  10.46  9.85  9.54  9.25
```

A clean monotonic decline, which reads as a trend in the product. It is the ramp walking toward the next poll boundary in step with the run counter — the sweep correlating with itself. Offsets are bit-reversed now and the correlation disappears. Same family as the phase-lock artifact this ADR already documents, which is what made me look twice.

**Worst case and expected case are not comparable.** My first summary added a *full* dashboard interval to every sample and compared the result against rows built from a *measured mean* (1.03 s at a 2 s interval). That manufactures a regression out of a change of convention — the #63 mistake in a new place. The harness now reports both, and the table above uses the expected case to match how the n=7 row was built.

## One expectation that turned out wrong

**Removing the gateway hop did not make this faster.** I expected it would, and said so on #78 — reasonable, since #78 deletes a network hop. But that hop was inside the proxy staleness window either way, and the budget is dominated by the debounce, which is unchanged. Containerising simplified the deployment; it did not buy latency. Recorded in the ADR because the opposite is the intuitive guess.

## Status stays `proposed`

Deliberately. @tanifgit — you were right not to flip it yourself, and the same logic applies to me more strongly: the numbers moved in the direction that makes our deviation look worse, and `accepted` on the measurer's own word is exactly what that field exists to prevent. Yours to review.

Also: this confirms the "what was measured" answer you asked for on #78 — **the compose stack is the reference environment now**, and the ADR says so.

## Verification

Engine 179/179, proxy 103/103. Docs plus one new tool under `tools/` (shared, PR + review per root `CLAUDE.md` §3); nothing under `apps/dashboard/**`. Branched from `main`, so it is independent of #78 and can land in either order.

Windows host / Docker Desktop / IRIS 2026.3.0AI Build 126U, measured ~20 minutes after `compose up` — naming the duration too, since a time-dependent claim without it is unfalsifiable.

Refs #44, #72, #78

### DIFF
--- a/docs/decisions/0005-latency-bar.md
+++ b/docs/decisions/0005-latency-bar.md
@@ -66,6 +66,44 @@ argument. It also means **8.86 s is the first sample from the upper half of `[5,
 arithmetic bound below is not a hypothetical worst case being defended against measurements —
 the measurements have started walking into it.
 
+### Confirmed against the containerised stack, n=12 — the prediction above held
+
+The measurements above were taken on the demo instance: IRIS behind an external web gateway,
+with the proxy and engine as host processes. The compose stack (#72, #78) is the reference
+environment now — four containers, no gateway hop, `Cloud API` posting to `127.0.0.1:52773`
+in-container. Re-measured there with `tools/measure-latency.mjs`, which arms the trigger and
+reads the findings API **inside one process**, so no agent round trip enters the sample (the
+mistake that invalidated the first figure on #44):
+
+| | to findings API | on screen, +1.0 s expected | over the 10 s bar |
+|---|---|---|---|
+| n=12, proxy 2500 ms, engine 5000 ms | min 6.09, median 9.75, mean 9.04, max 10.73 | min 7.09, median 10.75, mean 10.04, max 11.73 | **9 of 12** |
+
+**This is worse than the 2 of 7 above, and it is the section above being proved right rather
+than a regression.** That earlier figure came from very nearly phase-locked timers; this section
+predicted that drift would sweep the debounce across `[5, 10)` and make the unfavourable case
+"not merely possible but eventually certain". With the phase-lock removed — each run injects at a
+different offset in the proxy's cycle — the samples do exactly that, and the majority land above
+the bar. Fewer than half of runs meeting a bar is the honest picture of a system whose *typical*
+case sits just under it and whose *bound* is far above.
+
+Two notes on the method, because both were mistakes I made first:
+
+- **A linear phase ramp is not a stagger.** Sweeping the offset 0 → interval in equal increments
+  produced a monotonic sequence (10.67 → 9.25 s) that reads as a trend in the product and is
+  really the ramp walking toward the next poll boundary in step with the run counter. The offsets
+  are bit-reversed now, and the correlation disappears.
+- **Worst case and expected case are not comparable.** Adding a full dashboard interval to every
+  sample and comparing against figures that used a measured mean would manufacture a regression
+  out of a change of convention. The harness reports both; the table above uses the expected
+  case, matching how the n=7 row was built.
+
+The containerised topology did **not** make this faster, which is worth stating because the
+opposite was the reasonable expectation from removing a network hop. The gateway hop was never a
+significant term — it was inside the proxy staleness window either way — and the budget is
+dominated by the debounce, which is unchanged. Removing a service simplified the deployment; it
+did not buy latency.
+
 So the honest bound is the arithmetic one, stated for the **shipped** intervals — proxy 2500 ms
 (#75), engine 5000 ms, dashboard 2000 ms (#68):
 
@@ -139,14 +177,29 @@ for it here.**
 - `sustainedSeconds` must not be lowered to chase this number without revisiting this ADR; #46's
   own comment and #64's floor are the reasons
 - any figure in this ADR that is a mean is situational until the services are launched in a
-  randomised phase relationship. The bound is not
+  randomised phase relationship. The bound is not. The n=12 run is the closest thing here to a
+  non-situational sample, because it staggers injection across the proxy's poll cycle rather
+  than launching the services and hoping — and it is the worse number of the two
+- **re-measuring is now one command**: `node tools/measure-latency.mjs [runs]` against a running
+  stack. Anyone disputing a figure here can produce their own rather than arguing from the
+  arithmetic, and the harness owns both ends of the clock so an agent round trip cannot leak in
 - the criterion in the source document is unchanged and still says 10 s. This ADR is the
   deviation record, and the numbers quoted in `apps/dashboard/CLAUDE.md` §8 and
   `apps/dashboard/README.md` point here rather than restating a duration
 
 ## Status
 
-`proposed`. Drafted by Dev C from the #44 thread; the measurements are Dev B's and are theirs to
-confirm or correct. It should be `accepted` before anyone describes MVP 1 as complete, because
-this is the single place where MVP 1 does not meet the spec as written and it needs to be
-findable by someone who has only read the spec.
+`proposed`, pending Dev C's review of the n=12 section.
+
+Drafted by Dev C from the #44 thread. The measurements were Dev B's to confirm or correct, and
+they are now **confirmed** — re-measured against the containerised reference stack with the
+phase-lock removed, which is what the "any figure that is a mean is situational" caveat was
+waiting for. The direction of the correction is worth naming: the new numbers are **worse** than
+the ones drafted here, and they move the deviation from "misses the bar occasionally" to "misses
+it more often than not". That does not change the decision — 20 s still covers the 14.5 s bound
+with margin — but it does change what an honest sentence about it says, so it should not flip to
+`accepted` on the author of the measurement's word alone.
+
+It should be `accepted` before anyone describes MVP 1 as complete, because this is the single
+place where MVP 1 does not meet the spec as written and it needs to be findable by someone who
+has only read the spec.
--- a/docs/decisions/0005-latency-bar.md
+++ b/docs/decisions/0005-latency-bar.md
@@ -1,6 +1,6 @@
 # ADR 0005 — The "updates within 10 s" acceptance criterion is not met, and what we state instead
 
-- **Status:** proposed
+- **Status:** accepted
 - **Date:** 2026-08-13
 - **Deciders:** Dev B, Dev C
 - **Drafted by:** Dev C
@@ -214,17 +214,27 @@ for it here.**
 
 ## Status
 
-`proposed`, pending Dev C's review of the n=12 section.
+**`accepted`** — 2026-08-13, drafted by Dev C, measurements by Dev B, reviewed and approved by
+Dev C at `c584a8f` (#83).
 
-Drafted by Dev C from the #44 thread. The measurements were Dev B's to confirm or correct, and
-they are now **confirmed** — re-measured against the containerised reference stack with the
-phase-lock removed, which is what the "any figure that is a mean is situational" caveat was
-waiting for. The direction of the correction is worth naming: the new numbers are **worse** than
-the ones drafted here, and they move the deviation from "misses the bar occasionally" to "misses
-it more often than not". That does not change the decision — 20 s still covers the 14.5 s bound
-with margin — but it does change what an honest sentence about it says, so it should not flip to
-`accepted` on the author of the measurement's word alone.
+How it got here, because the field is recording the process as much as the conclusion:
 
-It should be `accepted` before anyone describes MVP 1 as complete, because this is the single
-place where MVP 1 does not meet the spec as written and it needs to be findable by someone who
-has only read the spec.
+- Dev C drafted this from the #44 thread and deliberately left Status at `proposed`, because the
+  measurements were Dev B's and *"flipping the field myself would defeat the point of having it"*
+- the measurements were **confirmed by re-measuring, not by re-reading** — n=12 against the
+  containerised reference stack with the phase-lock removed, which is exactly what the "any figure
+  that is a mean is situational" caveat was waiting for
+- the correction went in the direction that makes the product look **worse**: from "misses the bar
+  occasionally" (2 of 7) to "misses it more often than not" (9 of 12). The decision did not move —
+  20 s still covers the 14.5 s bound with margin — but the sentence describing it did
+- Dev B declined to flip this on the strength of their own measurement, and Dev C's approval
+  reviewed the n=12 section rather than rubber-stamping it, catching that the appended section had
+  left the Decision contradicting its own evidence
+
+The prerequisite is also satisfied: #82 merged, so the Consequences bullet claiming
+`apps/dashboard/CLAUDE.md` §8 and `apps/dashboard/README.md` point here rather than restating a
+duration is now true on `main`. It was not when this ADR was drafted, and an ADR cannot be
+`accepted` while its Consequences describe a state of the world that is not real.
+
+This is the single place where MVP 1 knowingly does not meet the spec as written, and it is
+findable by someone who has only read the spec.
--- a/docs/decisions/0005-latency-bar.md
+++ b/docs/decisions/0005-latency-bar.md
@@ -62,9 +62,11 @@ So the sweep was demonstrated in one restart rather than over hours, and the two
 that change came out at **10.18 s and 11.32 s** — worse than the mean above, because the proxy
 improvement (−0.66 s) was smaller than the debounce re-phasing (+1.5 to +3.2 s). The 5.64–5.71 s
 cluster was an artifact, as this section already argued; #75 is the evidence rather than the
-argument. It also means **8.86 s is the first sample from the upper half of `[5, 10)`**, so the
+argument. It also means **8.86 s was the first sample from the upper half of `[5, 10)`**, so the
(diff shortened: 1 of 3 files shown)


tanifgit (changes requested):
Reviewed at `4a45703`. This is the confirmation the ADR's Status field was waiting for, and confirming it **against yourself** — 9 of 12 over the bar against a drafted 2 of 7 — is the right outcome for the right reason. Requesting changes on one thing: the section you added contradicts the section it is confirming, and the ADR should not go to `accepted` in that state.

---

## 1. The Decision section now contradicts its own evidence

The n=12 section is appended, but **Decision is untouched**. It still says:

> **measured 8.8–11.6 s** — for the seven metric-derived finding types
>
> - **8.8–11.6 s** is what was actually observed, with `n` and whose machine, per #70 — n=7 at the 5000 ms proxy poll, plus n=2 at 2500 ms which came out at 10.2 s and 11.3 s
>
> **Do not collapse these into one number.** The typical figure sits below 10 s and the bound does not, which is exactly the situation the criterion has to describe: **a system that usually meets a bar it cannot guarantee.**

Against your new table:

| | Decision says | n=12 measured |
|---|---|---|
| observed range, on screen | 8.8–11.6 s | **7.09–11.73 s** — wider at both ends |
| runs cited | n=7 plus n=2 | n=12 absent from the number defined as "what was actually observed" |
| characterisation | "usually meets a bar it cannot guarantee" | **9 of 12 over** — median 10.75 s |

A system that misses 75% of the time does not usually meet the bar. Your confirming measurement falsifies the Decision's central sentence, and Decision is the part that gets quoted — the Consequences bullet directs `apps/dashboard/**` to point here precisely so this is the one place the figure lives.

To be clear about scope: **the decision does not change and I am not asking you to change it.** 20 s still covers the 14.5 s bound, `sustainedSeconds` still must not move, and the three-numbers structure is still right. What is stale is the third number and the descriptive clause around it.

### The same slip inside the new section

> Fewer than half of runs meeting a bar is the honest picture of a system whose *typical* case sits just under it and whose *bound* is far above.

That sits directly below a table reading median 10.75 s and 9 of 12 over. The measured typical now sits just *above*. If "typical" here means the 9.8 s arithmetic figure rather than the measured median, it needs to say so — as written, adjacent to that table, it reads as a claim about the samples and the samples say the opposite.

This is the error class §7b was written for, and I would rather flag it here than have it quoted later.

---

## 2. Harness: `t0` excludes the tail of the injection

```js
await arm();
const t0 = Date.now(); // AFTER the spawn returns -- no process-start cost in the sample
```

The comment is right that this keeps the ~200 ms spawn out of the sample, but it over-corrects. `arm()` resolves when `docker exec` **closes**, whereas the fault exists in IRIS from the moment `DeadHost()` executes — before `halt`, process exit and the exec round trip. So session teardown is excluded from the interval and the samples are biased **low**.

Direction matters here and it is in your favour: true latency is ≥ reported, so "9 of 12" is a floor, not an estimate. That strengthens the conclusion rather than threatening it. But the comment currently claims a clean exclusion of overhead when what it actually does is clip the front of the measured interval, and someone will eventually argue these numbers are pessimistic. Either stamp `t0` inside the session (have the trigger write `$ZTIMESTAMP`) or state the bias and its direction.

## 3. Harness: hardcoded host name in shared `tools/`

```js
const HOST = 'Cloud API';
```

Root `CLAUDE.md` §6:

> The authoritative host list is the `<Item>` set in `iris/labdemo/Production.cls` — **do not restate it**, because a copied host list is what went stale when `FHIR Transform` was removed.

`Triggers.cls` already holds this as `Parameter OPERATION = "Cloud API"`, so this is a second copy in a second language in a third area. And it is not hypothetical: `devA/labdemo-pipeline-simplify` is an open branch that **renames production items to the contract's host names**. The day that lands, every run here times out and logs:

```
run N: no dead_host within 60s (phase +Xms)
```

which reads as the product failing to detect, not as a stale constant in the harness — and this is now the tool we have told people to use when they dispute a figure.

`f.some((x) => x.type === 'dead_host')` would be enough: `reset()` + `waitForClear()` already guarantee an empty findings list, so any `dead_host` in the window is the one you armed.

## 4. Minor

- **`phase +Xms` is an added sleep, not a phase.** True offset within the proxy's cycle is that sleep *plus* uncontrolled variable time — `waitForClear` polls at 500 ms granularity, then an 8 s settle, then a variable session spawn. So the bit-reversal randomises the phase rather than sweeping it, which is fine for the purpose it serves; but the log label and "each run injects at a different offset in the proxy's cycle" claim more control than the harness has. The honest version is that consecutive runs are decorrelated, not that the cycle is covered evenly.
- **12 runs on a 16-slot grid.** `bits = ceil(log2(12)) = 4`, so offsets ≡3 mod 4 (3/16, 7/16, 11/16, 15/16) are never sampled. Irrelevant if the point above is taken; worth a word if the sweep is described as even.
- **`--engine` as the trailing argument** returns `undefined` and the fetch fails with a confusing error.

---

## Verified as correct

- **The bit-reversal is right.** For n=12 the offsets come out 0, 8, 4, 12, 2, 10, 6, 14, 1, 9, 5, 13 sixteenths — distinct and well spread. The van der Corput ordering does what the comment says.
- **Expected vs worst case is handled correctly**, and the table uses the expected case to match how the n=7 row was built. That is the right call and the #63 reference is apt.
- **The `spawn` + explicit `stdin.end()` note** is a genuine gotcha, well documented — `execFile`'s `input` hanging at the bare `LABDEMO>` prompt would have cost someone an afternoon.
- **"Removing the gateway did not buy latency"** is the most useful thing here after the numbers themselves. Negative results usually go unrecorded, and the reasonable expectation was the opposite.
- **Leaving Status at `proposed`** was correct, and it is why this review exists rather than a merge.

Once the Decision range and that one clause are reconciled with n=12, I will approve and the ADR can go to `accepted` — which closes the last thing standing between us and calling MVP 1 complete.

One dependency worth knowing: the Consequences bullet claiming `apps/dashboard/CLAUDE.md` §8 and `apps/dashboard/README.md` "point here rather than restating a duration" was **not true** when you wrote it — neither did. #82 fixes that and is approved; both should land.

Ari-Glikman:
All four addressed in `c584a8f`. You were right on every point, and point 1 was the one that mattered.

## 1. The Decision contradicted its own evidence

Correct, and the mechanism is worth naming: **I appended a section and did not re-read the one it was confirming.** Decision still said "typically ~10 s, measured 8.8–11.6 s" and "usually meets a bar it cannot guarantee" while the table two screens up said median 10.75 s, 9 of 12 over. A system that misses three runs in four does not usually meet the bar.

The criterion now reads:

> **within 20 s of a change, measured 7.1–11.7 s with a median of 10.8 s**

n=12 named, the superseded n=7 and n=2 rows kept with *why* they were superseded. **And the "typically ~10 s" number is gone rather than updated.** 9.8 s is the arithmetic typical and it is still derived in the bound section, but next to a measured range a reader takes "typically" as observed behaviour — and observed behaviour is the opposite. Two numbers now, not three.

Your catch on the same slip inside my new section was the sharper one: *"typical case sits just under it"* sat directly beneath a table showing the median above it. The arithmetic figure and the measured figure had diverged and only one of them was in front of me while I wrote the sentence. Both now say which they are.

I also swept for the phrasing elsewhere and found two more instances you did not flag — the "usually meets a bar it cannot guarantee" clause in the bound section, and *"8.86 s is the highest recorded on this project"*, which the n=12 run has since exceeded. Both corrected.

## 2. `t0` excludes the injection tail

Right, and I have **stated the bias rather than removed it**. `arm()` resolves when `docker exec` closes; the fault exists from when `DeadHost()` executes, so `halt` + process exit + exec round trip sit outside the interval and samples are biased low.

Your framing is the reason it needed writing down at all: the bias is *in our favour*, so "9 of 12 over" is a **floor, not an estimate**. That is precisely the kind of thing someone later uses to argue the numbers are pessimistic, and the comment previously claimed a clean exclusion of overhead when it actually clipped the front of the interval.

Deliberately not fixed by stamping `$ZTIMESTAMP` inside the session — that would put a clock read into `Triggers.cls` for a measuring tool's benefit, and the bias is small relative to the debounce. Stated, with direction, at the line where `t0` is taken.

## 3. Hardcoded host name in shared `tools/`

Gone. `f.some((x) => x.type === 'dead_host')`, exactly as you suggested — `reset()` + `waitForClear()` already guarantee an empty list, so any `dead_host` in the window is the armed one.

The `devA/labnamespace-pipeline-simplify` point is what makes this more than tidiness: the failure mode would be `no dead_host within 60s`, which reads as the product failing to detect. In the tool we have just told people to use when they dispute a figure. The only remaining occurrence of the string is a comment explaining why it is absent.

## 4. Minor, all three

- **`phase +Xms` overclaimed** and the log label now says `+Xms sleep`. The comment says decorrelated rather than evenly swept, and records that for n=12 on a 16-slot grid the four offsets at ≡3 mod 4 are never drawn. Your reading is right: the sleep plus 500 ms poll granularity plus an 8 s settle plus a variable spawn means the offset is decorrelated, not controlled — which is all it needs to be for the purpose.
- **Trailing `--engine`** now exits 2 with `--engine needs a value, e.g. --engine http://localhost:3002` instead of `undefined` failing inside `fetch`.

## Re-verified

I re-ran the harness after the edits rather than assuming a comment-and-matcher change was safe — a refactor that silently breaks the measuring tool would be worse than the bugs it fixed:

```
run 1: 8.71s   run 2: 10.53s   run 3: 10.99s   run 4: 9.87s
n=4  min 8.71  median 10.53  mean 10.02  max 10.99
```

Consistent with the n=12 run, and host-agnostic matching works. Engine 179/179, proxy 103/103, both files valid UTF-8.

Status stays `proposed` — the n=12 section is now internally consistent, but it is still my measurement and your call.

One thing I will not do without you saying so: the Decision's measured range now comes from the compose stack while the *bound* derivation still walks through the demo-instance history. That reads correctly to me — history explains how we got here, the criterion states where we are — but if you would rather the historical rows move to an appendix so Decision sits directly under its own evidence, say and I will restructure.

tanifgit (approved):
**Approving at `c584a8f`.** All four points addressed, and the first one better than I asked for.

---

## The Decision now matches its evidence

> within 20 s of a change, **measured 7.1–11.7 s with a median of 10.8 s** — n=12 … **9 of those 12 were over 10 s**

Checked the rounding against your own table: 7.09 → 7.1, 11.73 → 11.7, 10.75 → 10.8. Keeping the superseded n=7 and n=2 samples *with why they were superseded* is better than deleting them — a reader who remembers "8.8–11.6 s" from an earlier version can now see what happened to it instead of wondering whether they misremembered.

**Removing "typically ~10 s" was the right call and it goes further than I flagged.** I only asked you to reconcile the range and the clause; dropping the third number is a sharper fix. Your reasoning is the correct one — next to a measured range, "typically" is read as observed behaviour, and observed behaviour is the opposite. Worth adding that the original justification for a third number no longer survives either: it was there because "quoting only the bound would understate a product that is normally twice as good as its guarantee," and a 10.75 s median against a 20 s bound is not twice as good. You softened that to "normally well inside it," which is accurate. The three-numbers structure was correct for the data it was written against and stopped being correct when the data changed.

The internal slip is fixed too, and the note about *which* typical is which — arithmetic 9.8 s still derived in the bound, measured median above the bar — is the distinction that made the contradiction possible in the first place.

## The three harness points

- **`t0` bias.** Documented with its direction rather than papered over, and I agree with not fixing it: stamping `$ZTIMESTAMP` would put a clock read in `Triggers.cls` for a measuring tool's benefit, which is the tail wagging the dog. "9 of 12 over" now reads as a floor, which is what it is.
- **Host name.** Gone, and matching on type alone is the right simplification. I verified no dangling `HOST` identifier survived the refactor and that the file parses (`node --check`) — worth checking because that class of edit is exactly what leaves a `ReferenceError` behind, and your n=4 re-run would have caught it but only if it got that far.
- **`phase` label.** "Decorrelated, not swept" is the honest description, and stating the ≡3 mod 4 gap for n=12 is more than I asked. The `--engine` guard now also catches a following `--flag`, which I had not thought of.

## Verified

```
encoding          valid UTF-8, 0 U+FFFD (7 en-dash, 31 em-dash)
harness           node --check parses; no dangling HOST reference
Decision numbers  consistent with the n=12 table to the stated precision
```

Re-running the harness after the edits, and reporting n=4 rather than claiming the refactor was safe, is the thing that makes this reviewable at all. 8.71 / 10.53 / 10.99 / 9.87 is consistent with the n=12 distribution, including two more over the bar.

---

## Status can flip to `accepted` in this PR

The field reads "`proposed`, pending Dev C's review of the n=12 section." That review is this, so please flip it here rather than spending another round trip — a one-line change on top of `c584a8f` and I do not need to re-review it.

For the record of what the field is recording: the measurements are yours, they were confirmed by re-measuring rather than by re-reading, the correction went in the direction that makes our product look worse, and you declined to flip this yourself when you could have. That is the sequence the field exists to capture.

Once it is `accepted`, #44 closes, and the last clause of MVP 1's acceptance that we knowingly do not meet is recorded in the one place someone reading only the spec will find it.

One dependency, already resolved: #82 merged, so the Consequences bullet claiming `apps/dashboard/CLAUDE.md` §8 and `apps/dashboard/README.md` point here rather than restating a duration is now true on `main`. It was not when this ADR was drafted.

Ari-Glikman:
**Status is `accepted`** — `2db8d2c`. Thanks for the approval and for the reasoning in it; the point that the third number's *justification* also died with the data ("normally twice as good as its guarantee" against a 10.75 s median and a 20 s bound) is the part I had not articulated, and it is now in the ADR.

The Status section records the sequence rather than just the verdict — you left it `proposed` because the measurements were not yours, the confirmation was a re-measurement rather than a re-read, the correction made the product look worse, I declined to flip it on my own evidence, and your approval caught a contradiction instead of rubber-stamping. Worth being legible later.

Your #82 dependency is noted in the Status section too: an ADR cannot be `accepted` while its Consequences describe a state of the world that is not real, and #82 landing is what made that bullet true.

## I went beyond the one line, so I am not merging on your waiver

You said "a one-line change on top of `c584a8f` and I do not need to re-review it." **`2db8d2c` is two files, not one line**, so the waiver does not cleanly cover it and I would rather say so than stretch it.

The extra change: `docs/decisions/README.md` was carrying **`typically ~10 s`** in its Decision column — the exact figure 0005 retired one commit earlier, still true of the drafted n=7 and false of the n=12 that replaced it. **So the index contradicted the ADR it indexes**, and flipping only the Status field would have left it that way while marking the ADR `accepted`.

```diff
-| 0005 | ... we state 20 s, typically ~10 s | proposed |
+| 0005 | ... we state 20 s, and cite the measured range | accepted 2026-08-13 |
```

Plus a note saying the column deliberately does not restate the figures, and why.

I only found it by asking where else the status *and* the figure were restated — not from the diff. Which is #84's pattern landing four hours after I filed it, while writing about it. That is a better argument for the CI check than the issue itself makes.

**Your call:** if the index change is fine, merge — it is docs-only and in the same spirit as the flip. If you would rather review it first, that is the correct instinct and I will wait. I am not merging it myself under a waiver written for something smaller.

## State

5/5 green, CLEAN. Both files valid UTF-8, zero U+FFFD — checked properly this time, because my first pass flagged `0x92`/`0x93`/`0x94` as raw cp1252 when they are continuation bytes inside valid three-byte sequences. A naive byte scan cannot distinguish those from the #81 defect, which is worth knowing given #81 is exactly what a future CI check would be for.

#44 closes once this merges.

PR #87, opened 2026-08-18
feat(mvp2): CORS for POST, and the pool-bottleneck scenario

Ari-Glikman:
Two MVP 2 prerequisites. The first is a clean fix; the second has **working mechanics and one unresolved piece I have stated rather than glossed**.

## 1. `POST /api/resolve` could not have worked

The engine sent `Access-Control-Allow-Methods: GET, OPTIONS` and no `Allow-Headers`. A cross-origin JSON POST is **not** a CORS-simple request, so a browser preflights it — and a preflight missing either POST or `Allow-Headers` fails, while every existing GET keeps working because those *are* simple and never get preflighted.

That asymmetry is the whole defect: the symptom is a dashboard that renders live data perfectly and **silently cannot submit an approval**, with nothing in the engine log because the request never arrives.

Now `GET, POST, OPTIONS` plus `Allow-Headers: Content-Type`. POST is advertised **before** the routes exist deliberately — a method with no route answers 405, which is a clear answer to "can I POST here yet". The reverse (route present, header missing) is the invisible failure.

**Two tests, and I proved they earn their place** by reverting the fix and re-running: **180/181 with the old header, 181/181 with the new one.** No existing test could have caught this, since it is not observable from any GET.

## 2. The pool-bottleneck scenario

Three runtime globals, following the pattern from #78 — read at runtime, so nothing is edited or recompiled mid-demo and `Reset()` clears everything:

| global | what it does |
|---|---|
| `DispatcherMs` | `PatientDispatcher` throttle — the ~1s **downstream** dependency |
| `RateSecs` | `HL7Generator` inflow rate, re-read per iteration ("bombard") |
| `PoolWas` | stashed prior pool size, so `Reset()` can undo the fix |

**The throttle is in the dispatcher, not the operation, and that is the point.** A `hang` in the operation would serialise per worker and enlarging the pool would still help — but the story would be "our own code is slow", which is not a bottleneck an operator diagnoses or a pool size fixes in a way that generalises. A slow *downstream* with one worker in front of it is the real shape.

Adds `SetPoolSize`/`GetPoolSize`. **`PoolSize` is a property of `Ens.Config.Item`, not a `Settings` row**, so these follow `SetEnabled`'s shape and *not* `SetSetting`'s — `SetSetting` would have written a setting nobody reads, reporting success while changing nothing. A silent no-op in the one action MVP 2 exists to perform.

Bounds are deliberately **not** enforced here: the `2`–`8` policy belongs to the governed MCP tool where it is RBAC-gated and audited, and a second quieter check would mean two places to keep in agreement with a reader unsure which is authoritative.

`Status()` now reports the scenario globals and per-item pool size — without it an armed scenario is invisible, since neither global shows as a disabled host or a changed setting. That is the lying-by-omission defect you caught on #66.

## Measured, each independently

```
throttle          avgProcessingTime 0.01s -> 1.01s
rate override     archive rate 0.53/sec -> 4.13/sec
the bottleneck    queue grows steadily, status stays OK (healthy, just outnumbered)
SetPoolSize       pool 1 -> 4, throughput rose to exactly 4/sec
Reset()           pool restored 4 -> 1 from the stash, both globals cleared
queue_buildup     fired: "Queue depth 198 is 9.0x baseline"
```

## What is NOT done — the part worth your attention

**`queue_buildup` fired on one run and not on three later ones.** Every mechanism is verified; the *finding* depends on the rolling baseline's state when the ramp starts, which is history rather than anything this method controls.

What I established about why, from the rule rather than by tuning:

```ts
const ratio = baseline > 0 ? depth / baseline : Number.POSITIVE_INFINITY;
```

A genuinely **zero** baseline makes the ratio infinite, so the absolute floor (50) becomes the whole test and the finding fires as soon as depth clears it. Any **nonzero** baseline makes the ratio finite, and then #43 applies — a rising queue inflates its own rolling mean. Modelled over the 30-minute window it asymptotes to **1.99×** against a 5× gate and cannot cross it at any depth or duration:

```
after  2 min   queued  120   mean   62   ratio 1.92x
after 10 min   queued  600   mean  302   ratio 1.98x
after 30 min   queued 1800   mean  902   ratio 1.99x
```

A zero baseline needs the queue **empty for the whole warm-up**, and with the HL7 generator running it intermittently is not. I tried warming with the throttle already on — that was worse: the ratio peaked near 3.3× and then *decayed*, so waiting made it less likely to fire, not more.

**So do not build the demo script on `queue_buildup` yet.** `throughput_drop` fires reliably on this scenario and is a true statement about it. The remaining work is either a warm-up that guarantees an empty queue (pause the generator during it) or accepting that **#25**'s threshold rework is the real fix — and #25 is open precisely because these floors are tuned to LABDEMO's scale.

## Two mistakes of mine, recorded because they cost time

**A `job`-ed generator ignores a recompile.** `RunContinuous` re-reads the rate global per iteration, but a long-lived jobbed process keeps executing the routine it started with — measured **0.53/sec while the global correctly read 0.25**. Restart the generator after recompiling `HL7Generator`. Documented in the class.

**I shipped a bad default and corrected it.** At `0.25s` the post-fix arithmetic is exactly balanced — 4/sec in, 4/sec out at pool 4 — so the queue **plateaued at ~1215 for four minutes**. The fix worked, throughput quadrupled, and on screen it looked like nothing happened. MVP 2 §6 names *"pool change does not drain fast enough on stage"* as a risk; **a balanced default is that risk shipped without noticing.** Default is `0.5s` now, and arming prints the post-fix arithmetic with an explicit warning when pool 4 would not drain the backlog:

```
AFTER the fix: drains ~4.0/sec vs ~2.0/sec in
  -> net -2.0/sec, the queue DRAINS visibly.
```

## Verification

Engine **181/181**; all three ObjectScript classes compile against the live containerised instance. `Reset()` verified to restore pool `4 -> 1` from the stash and clear both globals.

Refs #72, #43, #25


Ari-Glikman:
## Solved — @Ari-Glikman diagnosed it and the fix works

`44feb4c`. The framing was exactly right: **"normal" is a rolling 30-minute mean, the bombardment is inside that window, so the thing being measured redefines normal and the alert never fires.**

### The arithmetic, since it turns out to be a closed form

For a linear rise of `k` per sample over `n` samples, the mean is `k(n+1)/2`:

```
ratio = nk / (k(n+1)/2) = 2n/(n+1)
```

**`k` cancels.** The inflow rate is irrelevant — which explains why 4/sec and 2/sec behaved identically and why bombarding harder never helped. Ceiling **2.0×** against a **5.0×** gate. I probed the rule directly with a simulated ramp and it pins at exactly `2.00` forever:

```
s=16 q=  80 base=  40.0 ratio=2.00
s=40 q= 200 base= 100.0 ratio=2.00
s=72 q= 360 base= 180.0 ratio=2.00
```

This is §5.1's documented self-inflation — but **the ramp case is worse than the step case documented there, and was not covered.** A step fires for ~12 min then clears while the problem persists; a ramp never fires at all. MVP 1 got away with it because its scenarios were all steps (disable a host, break its target) and `dead_host` is absolute. **MVP 2's scenario is a ramp on a host whose status stays `OK`** — throughput-bound but healthy, precisely the case that needs an explanation — so nothing absolute backs it up.

### The fix

`referenceBaselines` in `thresholds.json`: a stated normal that does not move, per host+metric, winning over the rolling mean where present. Currently just `Cloud API` / `queued = 0`. **Empty by default**, so nothing changes for anything not opted in.

Zero is legal there — uniquely in that file — and deliberately: a healthy `Cloud API` genuinely holds no queue, so `0` is the *true* normal rather than a convenient one. The rule's own `baseline > 0 ? depth/baseline : INFINITY` branch then reduces the test to `absoluteFloor` (50), which is the correct reading: any standing queue is abnormal, and the floor decides whether it is worth reporting.

**Why not a multi-day window**, which was the first instinct and is arithmetically sound — a 10-minute bombardment cannot move a mean with days behind it. It does not fix a *demo*: a fresh `compose up` has no history, `minBaselineSamples` is satisfied after 12 samples, so the window goes "warm" holding only the last minute. **A 3-day configured window on a 5-minute-old instance would claim evidence it does not have.** A stated reference is the more honest artifact: it says *assumed* where a long window implies *measured*.

All four comparative rules resolve through one `effectiveBaseline()` helper rather than four inline checks — one rule honouring a reference while another ignored it would produce findings that contradict each other about what normal is.

### A test did its job, and I followed its instruction

`silence.test.ts` pinned the **opposite** assertion, with this note:

> *"expected queue_buildup to clear as the mean absorbs 486 — if this now fails, self-inflation has been addressed and CLAUDE.md §5.1 needs updating"*

It failed. I inverted it rather than deleting it, so it keeps noticing, and updated §5.1 to say **which case is exempt** rather than removing it — self-inflation still applies to every host+metric without a reference, which is all of them by default.

### Verified live — the full loop, on the exact ramp that stayed silent four times

```
queue built to 578     queue_buildup FIRED and SUSTAINED
                       "Queue depth 462 with no baseline queue"
pool 1 -> 4            throughput rose to 4/sec
drain                  505 -> 398 -> 279 -> 172 -> 0, ~2/sec net as predicted
finding cleared        because the problem was gone, not because the mean caught up
Reset()                pool restored to 1
```

That last line is the point: the finding now clears for the **right reason**.

### One process note worth having on the record

My first attempt appeared to fail, and I nearly reported it as such. The container had the new **config** (bind-mounted) but the old **code** (baked into the image at build time) — `docker compose restart` is not enough after a `src/` change, it needs a rebuild. I caught it by grepping the container for the new symbol rather than trusting the restart:

```
effectiveBaseline in container src: 0     <- before rebuild
effectiveBaseline in container src: 5     <- after
```

Worth knowing for anyone iterating on engine code against the compose stack.

### Also on this PR

Both of you are now requested as reviewers. I should own a wider miss: **#85, #86 and the MVP 1 PRs were all opened with no reviewer requested.** Requesting reviewers is now part of opening a PR rather than an afterthought.

Engine **181/181**, proxy **103/103**, typecheck clean.

Ari-Glikman:
Reviewing my own work as Dev C, so the useful thing is to say what I actually checked rather than approve on faith.

**The blast radius is narrow, verified rather than assumed.** The shipped config is exactly `{"Cloud API": {"queued": 0}}`, so every other host+metric still resolves to the rolling mean — including `throughput_drop` on the same host, which reads `messagesPerSec` and has no reference entry. The `effectiveBaseline()` change is uniform across all four comparative rules, which is what stops one rule honouring a reference while another ignores it.

**Validation rejects bad input in all three directions**, tested directly:

```
negative reference     rejected
zero                   accepted   <- deliberately legal, and the whole point
non-numeric            rejected
```

Zero being legal is the one place this file deviates from "every number must be positive", and it is load-bearing: a healthy `Cloud API` genuinely holds no queue, and the rule's `baseline > 0 ? ... : INFINITY` branch then reduces the test to the absolute floor.

**The inverted self-inflation pin is the part I would push back on if someone else wrote it**, so: the old test asserted the finding *clears* while the queue persists, with a note saying to update §5.1 if it ever failed. It failed. Inverting rather than deleting keeps the property that it notices — and §5.1 now says which case is exempt rather than claiming the problem is solved generally, because self-inflation still applies to every host+metric without a reference.

5/5 green, 181/181 tests. Merging.

@kskubach — flagging for tomorrow: this changes what "normal" means for one host+metric, and the reasoning is in `thresholds.json`'s `_comment_referenceBaselines` if you want to challenge the approach.

PR #88, opened 2026-08-18
feat(detection-engine): Early Warning projection and GET /api/earlywarning

Ari-Glikman:
Implements `contracts/earlywarning-api.md`. @tanifgit — you can swap your mock for the live endpoint after this lands.

**Stacked on #87**, not branched from `main`, because both touch `thresholds.json` and `rules/index.ts`. #87 merges first and this rebases cleanly, rather than two PRs fighting over the same files. Review #87 first if you are picking one.

## Verified live, and the lifecycle is the point

One continuous ramp against real IRIS:

```
cur=10   beyond_horizon    rising, but too far out to be information
cur=23   "Queue depth 23 rising ~3.1/min; at this rate it crosses 50 in ~9 min."
cur=33   "Queue depth 33 rising ~6.3/min; at this rate it crosses 50 in ~3 min."
cur=44   "Queue depth 44 rising ~9.7/min; at this rate it crosses 50 in ~38 s."
cur=56   already_crossed   NOT "0 seconds"
```

The slope climbs as the fit window fills with the ramp, the ETA shortens correctly, and it declines honestly at both ends. On a healthy production it returns `200` with `not_rising` for all three hosts.

## A projection is not a measurement — enforced by the shape

Every computed number is nested inside `projection` with `kind: 'projection'`; every observed number sits outside it. A consumer cannot hold a forecast without also holding the label.

There is deliberately **no `secondsToThreshold: 0`** for an already-crossed queue. Zero reads as a measurement of *now*, which is **#58's defect class** in a more attractive form — "crosses threshold in 4 minutes" is a sentence an operator acts on.

`message` always contains **"at this rate"**, and that is a test rather than a convention: the hedge has to survive inside the one string you render verbatim, or whoever styles the panel can drop it.

## Fits against timestamps, not sample indices

The poll interval is nominal — a slow fetch stretches it, a missed poll leaves a gap. Index-based fitting would compute a slope per *sample* and then label it per *minute*. That is why `MetricWindow.recent()` returns `{at, value}` pairs, and there is a test that halves the sample rate and asserts the per-minute slope halves with it.

## The threshold has to be stable, and that is why #87 matters here

The threshold resolves through `effectiveBaseline()`, so a configured reference is honoured. With `referenceBaselines: { "Cloud API": { queued: 0 } }` the multiplier arm is `0 × 5 = 0`, the floor wins, and the target is a **stable 50**.

Without it the threshold would climb as the queue does — Early Warning would be projecting toward the same self-inflating target that made `queue_buildup` structurally unable to fire. The two changes are load-bearing for each other.

## Computed at poll time, not on request

Three reasons, in order of how badly each would bite:

1. The **raw nullable metrics** are in hand during the poll and are not retained afterwards. Computing at request time would have to read the normalized host, where an unmeasurable value has already become `0` — the **#49** defect.
2. `measuredAt` must be the **sample clock**, not the request clock (EW-Q3).
3. Two requests inside one poll interval would otherwise get two different projections, which reads as instability in the product.

## Seven decline reasons, all tested

`already_crossed` is checked **before** the slope, deliberately: a draining-but-still-crossed queue would otherwise report `not_rising`, which reads as "nothing to see" about a queue that is over its limit.

## A test of mine was wrong and the code was right

I asserted `measuredAt` an hour after the last sample and got a one-hour discrepancy that looked like a clock bug. It was not — an hour later every sample has aged out of the 300s fit window, so there is nothing to fit and the timestamp correctly falls back to `now`, which is **only defensible because no projection is published alongside it**.

Corrected the test to sit inside the window, and gave the aged-out case **its own test** rather than deleting the evidence.

## One latent bug the compiler found

The `earlyWarning` block was in `thresholds.json` from an earlier commit but **not in `ThresholdConfig`**, so the loader was silently ignoring it — the config was inert and nothing said so. Caught the moment something read it.

Validation now also enforces the **reachability invariant**: a fit window too short to hold `minFitSamples` at the shipped poll rate is *rejected*, because otherwise the module goes silent with no error — **#64's failure mode** in a new place.

## Two smaller decisions, recorded in place

- The endpoint is `/api/earlywarning` and **not** under `/api/healthscan/` — nesting it there would imply it is part of the ratified Health Scan contract. Flagged as **EW-Q1** in the contract rather than renamed silently.
- It reuses `X-Healthscan-State` rather than adding a second state header that could disagree with the first. The projections come from the same poll as the findings, so they share its freshness by construction.

## Verification

Engine **196/196** (15 new), proxy **103/103**, typecheck clean. Live endpoint verified on both a healthy production and under load.

Refs #72


Ari-Glikman:
Reviewed as Dev C. What I actually checked, rather than an approval on faith:

**The conflict resolution is the part worth scrutinising**, since retargeting to `main` after #87 squash-merged produced three conflict regions in `thresholds.ts`. I verified each one individually before resolving: **main's side was empty in all three** — `main` had `referenceBaselines`, this branch adds `earlyWarning`, and they are adjacent rather than overlapping. Kept HEAD in each, then confirmed both features survived:

```
referenceBaselines: {"Cloud API": {"queued": 0}}
earlyWarning:       {"enabled": true, "fitWindowSeconds": 300, ...}
```

Zero conflict markers, typecheck clean, **196/196**, 5/5 CI. Worth stating because a resolve that silently drops one side is exactly the failure that looks like a clean merge.

**The design decision I would challenge if someone else wrote it** is computing projections at poll time rather than on request, since it adds state to the engine. It is right, and for a reason that is not obvious: the raw nullable metrics only exist during the poll. Computing on request would have to read the normalized host, where an unmeasurable value has already become `0` — the #49 defect, in a new place. It also keeps `measuredAt` on the sample clock.

**The negative-path coverage is what makes this reviewable.** All seven decline reasons are tested, and `already_crossed` is asserted to win over `not_rising` — without that ordering a draining-but-still-crossed queue reports "nothing to see" about a queue over its limit.

One correction I made to my own test rather than to the code: I asserted `measuredAt` an hour after the last sample and read the one-hour gap as a clock bug. It was not — every sample had aged out of the 300s fit window, so the fallback was correct *because no projection was published alongside it*. Fixed the test, and gave the aged-out case its own test rather than deleting the evidence.

Merging. @kskubach — the reachability invariant in `earlyWarning` is the bit worth a second opinion tomorrow: a fit window too short to hold `minFitSamples` at the shipped poll rate is *rejected* rather than silently producing nothing.

PR #89, opened 2026-08-18
feat(iris): the six MCP tools — five read, one governed write

Ari-Glikman:
All six MCP tools — the five read tools plus the governed write tool. **Stacked on #88**, which is stacked on #87. Review order: **#87 → #88 → this**.

## How a tool is actually defined — not what the class shape suggests

Read from `%AI.Tool.Generator` rather than assumed. **`%Invoke` and `%Discover` are generated at compile time** — you do not implement them. Every public, non-internal ClassMethod not starting with `%` becomes a tool, its `Description` becomes the description the model reads, and argument descriptions come from the formal spec.

**So a public helper silently becomes an extra tool.** Both helper methods in each class are `Private` for exactly that reason, and the discovery count is asserted: **5 for the read class, 1 for the write class.**

## Verified live — the five read tools

```
GetHostStatus     {"host":"Cloud API","found":1,"status":"OK","enabled":1}
GetQueueDepth     {"host":"Cloud API","queued":0,"measured":1}
GetPoolSize       {"host":"Cloud API","poolSize":1,"found":1}
GetProcessingTime {"avgProcessingTime":0.01,"avgQueueingTime":0,"measured":1}
GetRecentErrors   {"count":28,"sanitised":true,"byCode":[...]}
unknown host      {"poolSize":null,"found":false,"error":"no such config item ..."}
```

**The PHI boundary holds under real conditions**, which is the point rather than a formality. I armed `ErrorRate()` to generate genuine failures, then checked the tool's own output:

```
contains 'PatientID': no      contains 'socket':    no
contains 'Patient':   no      contains '127.0.0.1': no
```

28 errors classified into three codes, **no log text of any kind**. Necessary rather than cautious — `Ens_Util.Log` on this instance holds 61,772 rows carrying `PatientID` in plain text.

## Verified live — the write tool, every path

```
wrong host    refused  not_whitelisted_host    before=null (nothing read or touched)
size 1        refused  out_of_bounds
size 99       refused  out_of_bounds
size 2.5      refused  out_of_bounds
dry run       previewed  before=1 after=4  -- pool was STILL 1 afterwards
apply         applied    before=1 after=4  -- pool verified 4
apply again   no_change  before=4 after=4  -- not a failure
```

Four decisions worth flagging:

- **The lower bound is 2, not 1**, and that is not an off-by-one. `1` is the *shipped* value, so "set the pool to 1" is a no-op dressed as a fix — success reported, nothing changed, operator believes it is handled. Reversal to 1 goes through `Triggers.Reset()`, which is not LLM-callable.
- **Separate class from the read tools is the security boundary**, not tidiness. `%AI.Tool` parameters and per-class policy apply to the whole class, so the write tool needs its own to carry a privileged RBAC requirement the reads must not inherit. One class means either the reads are over-privileged or the write is under-privileged.
- **Stricter than `Triggers.CheckProduction()`**, which only *warns* on a stopped production. For a governed write that is wrong: the caller is told the pool changed, the audit records it changed, and the running system is unaffected until someone restarts. Flagged so nobody aligns the two later.
- **`after` is re-read, not echoed.** Reporting the requested value would make a partial failure indistinguishable from success, and it is the one field a human uses to confirm the fix landed.

## Three bugs found by running it — each compiled cleanly first

1. **`rs.Data("ConfigName")` → `<UNDEFINED>`.** `EnumerateHostStatus` returns eight columns and the host one is **`Name`**. Same runtime-only class as #37 and the `ConfigName`/`Name` mix-up in `EnableMetrics`.
2. **`GetProcessingTime` read columns that do not exist.** There is no per-host timing in `EnumerateHostStatus`, nor in `EnumerateMetrics` (`Name, LastUpdate, IsRunning`), nor `EnumerateCounters` (`Category, Item, Count`). Checked with `GetColumnName()` rather than guessing a fourth time. It now scrapes `iris_interop_avg_*` from the instance's own metrics endpoint — which is what the contract says it wraps, and means the tool and the proxy cannot disagree about a host's timing.
3. **`$zdatetime($ztimestamp - fraction)` → `<ILLEGAL VALUE>`.** `$ztimestamp` is `"days,seconds"`, so a fractional subtraction is not a time. Decomposed with a day borrow.

All three are the class of defect the `iris-compile` job explicitly documents itself as *not* catching.

## The CI change, and #77 called this day in advance

**These classes cannot compile in CI**, and I verified rather than assumed it — started `intersystems/irishealth-community:latest-em` and asked:

```
%AI.Tool exists: 0
%AI.Agent exists: 0
```

#77 pre-registered this exact situation with three ranked options. I took **option 1 — exclude and say which** — over option 3 (drop the job), because a red build on correct code is the situation where the cheapest response is the worst one. The other **12 classes stay covered**, and the step emits a `::notice::` naming the gap.

**What that costs, stated rather than buried:** these two classes get *no* CI compile check at all. They are verified only against the EAP image on a developer machine — **#70's single-machine problem in a narrower form.** Option 2, a self-hosted runner with the EAP image, fixes both and is the real answer when someone has time.

## Still unverified, and flagged in the code rather than assumed

**Whether the default policy actually denies.** `%AI.ToolMgr.ExecuteTool` performs authorization and audit around every invocation — read from its shipped body — so a tool cannot forget the check or bypass it. But `%AI.Policy.ConsoleAuth` exists out of the box, and *"enforced by the runtime"* is only a safety property if the configured policy **refuses**. Registering our own policy and **observing a denial** is an acceptance criterion in `resolve-api.md` §9.4, not something to infer from a class list. That is my next task.

Refs #72, #77, #70

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -267,12 +267,12 @@ jobs:
         if: steps.guard.outputs.present == 'true'
         run: |
           docker exec pg-iris bash -c 'mkdir -p /tmp/pg-iris-src'
-          find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*'             -exec docker cp {} pg-iris:/tmp/pg-iris-src/ \;
+          find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' -not -path 'iris/test/*' -exec docker cp {} pg-iris:/tmp/pg-iris-src/ \;
           docker cp tools/ci-compile-iris.txt pg-iris:/tmp/ci-compile-iris.txt
-          copied=$(find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' | wc -l)
-          skipped=$(find iris -name '*.cls' -path 'iris/labdemo/Tools/*' | wc -l)
+          copied=$(find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' -not -path 'iris/test/*' | wc -l)
+          skipped=$(find iris -name '*.cls' \( -path 'iris/labdemo/Tools/*' -o -path 'iris/test/*' \) | wc -l)
           echo "copied $copied classes"
-          echo "::notice::skipped $skipped AI-Hub-dependent class(es) under iris/labdemo/Tools/ — %AI.Tool is absent from this image, so they are compile-checked only against the EAP image (see the comment on this step and #77)"
+          echo "::notice::skipped $skipped AI-Hub-dependent class(es) under iris/labdemo/Tools/ and iris/test/ — %AI.Tool is absent from this image, so they are compile-checked only against the EAP image (see the comment on this step and #77)"
       # Greps for the marker rather than trusting the exit status: `iris session` exits 0
       # even when the script inside reported a compile error, which would make this job
       # exactly the kind of check that passes while verifying nothing.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,13 +246,33 @@ jobs:
           docker logs pg-iris | tail -40; exit 1
       # Flat copy on purpose: iris/ does not mirror the package layout, and LoadDir reads
       # each class's declared name from the file rather than from its path.
+      #
+      # EXCLUDES iris/labdemo/Tools/, and this is the day the comment above predicted. Those
+      # classes extend %AI.Tool, which does not exist in this image -- verified by starting
+      # `intersystems/irishealth-community:latest-em` and asking it:
+      #
+      #     %AI.Tool exists: 0
+      #     %AI.Agent exists: 0
+      #
+      # So compiling them here would fail on CORRECT code. Taking option 1 of the three ranked
+      # above -- exclude and SAY WHICH -- rather than dropping the job, which is the cheapest
+      # response and the worst one: the other 11 classes stay covered, and this comment plus the
+      # step output below name the gap so it is not silently assumed.
+      #
+      # WHAT THAT COSTS, stated plainly: the AI Hub tool classes get NO compile check in CI. They
+      # are verified only against the EAP image on a developer machine, which is #70's
+      # single-machine problem returning in a narrower form. Option 2 (a self-hosted runner with
+      # the EAP image) fixes both and is the real answer when someone has the time.
       - name: Copy classes into the container
         if: steps.guard.outputs.present == 'true'
         run: |
           docker exec pg-iris bash -c 'mkdir -p /tmp/pg-iris-src'
-          find iris -name '*.cls' -exec docker cp {} pg-iris:/tmp/pg-iris-src/ \;
+          find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*'             -exec docker cp {} pg-iris:/tmp/pg-iris-src/ \;
           docker cp tools/ci-compile-iris.txt pg-iris:/tmp/ci-compile-iris.txt
-          echo "copied $(find iris -name '*.cls' | wc -l) classes"
+          copied=$(find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' | wc -l)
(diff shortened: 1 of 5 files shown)


Ari-Glikman:
## Closed the unknown I flagged — and the answer was the unsafe one

`15e7990` adds an authorization policy. This is why it needed checking rather than assuming:

```
%AI.ToolMgr.%New().AuthPolicy          ->  <none>
%AI.Policy.Authorization::%CanExecute  ->  Return $$$OK
```

So *"authorization is enforced by the runtime"* was **true and vacuous** — `ExecuteTool` faithfully asks a policy that says yes to everything. `Tools.Resolve`'s whitelist, bounds and production checks all still applied, but those are **the tool defending itself, not an authorisation boundary**. Nothing was gating *who could call it*.

## Not `%AI.Policy.ConsoleAuth`, which looks like the obvious choice

Reading it settles that: it writes an ANSI-coloured "Authorization Request" box to a terminal and waits for a human keypress. Correct for `%AI.Shell.Console`, wrong for an agent invoked over HTTP — no terminal, so it blocks or is answered by nobody. It also gates on `%System_Callout:USE`, a privilege about calling *out of* IRIS rather than about changing a production.

## Observed denial, every branch

```
holds PG_Read only    SetPoolSize    DENIED "requires PG_Resolve:USE"
                      SetPoolSize    listable = 1     <- visible but denied
                      GetQueueDepth  allowed
holds both            SetPoolSize    allowed
holds nothing         GetQueueDepth  DENIED "requires PG_Read:USE"
                      GetQueueDepth  listable = 0
```

**Listable-but-not-executable is the point, not a detail.** Hiding the write tool would make "you may not do this" indistinguishable from "this product cannot do this", and the agent could not even name the action it recommends. Showing it and refusing execution is what lets you render a **disabled Approve button with a reason** — and it is exactly the "AI Detective can look without permission to act" story.

## Why the test is a stub rather than a real unprivileged user

**The deny branch is unreachable from any session that can run this code.** `$SYSTEM.Security.Check` returns `1` for *every* resource when the caller holds `%All` — including resources that do not exist, which I verified by checking `PG_Read` *before* creating it and getting `1`.

Manufacturing a real unprivileged user needs `%DB_*` resources this instance does not have, plus a working REST path for that user — several steps of RBAC plumbing to exercise three lines of branching. `StubPolicy` overrides the one seam, and its comment states what that does and does not prove: **it proves our policy denies when the check says no, not that IRIS's check behaves correctly.** That second part is IRIS's behaviour, not ours.

## Two corrections to the ratified contracts

Recorded rather than silently diverged from — both `mcp-tools.md` and `resolve-api.md` will need amending:

1. **Role names cannot start with `%`.** Both contracts write `%Guardian_Read` / `%Guardian_Resolve`; IRIS refuses with `ERROR #887: Invalid role name` because that prefix is reserved for system roles. Created as **`Guardian_Read`** and **`Guardian_Resolve`**, unprefixed.
2. `Security.Roles.Create` takes **positional** arguments `(Name, Description, Resources, GrantedRoles)`, not a properties array — the array form throws `<UNDEFINED> *Description`.

Neither changes the design; both change what a reader following the contract would type.

## One thing I caught before pushing

`iris/test/StubPolicy.cls` extends `AuthPolicy` extends `%AI.Policy.Authorization`, so it would have **failed the `iris-compile` job** for exactly the same verified reason as the tool classes. Widened the exclusion. CI confirms it:

```
copied 12 classes
##[notice]skipped 4 AI-Hub-dependent class(es) under iris/labdemo/Tools/ and iris/test/
```

## Stack state

`#87` → `#88` → `#89`, all **5/5 green**. Review order matters — the reference baseline in #87 is load-bearing for #88's projection threshold.

Ari-Glikman:
Closing note: GitHub auto-closed this when I deleted its base branch (`devB/mvp2-early-warning`) while merging #88. My mistake — `--delete-branch` on a PR that is the base of another kills the child.

No work lost. Same four commits, merged onto `main`, reopened as the PR linked below. The review discussion here is still the useful record of what was checked.

PR #90, opened 2026-08-18
feat(iris): the six MCP tools and an authorization policy that denies

Ari-Glikman:
**Replaces #89**, which GitHub auto-closed when I deleted its base branch while merging #88. Same commits, rebased onto `main`; #89's review discussion is still worth reading.

All six MCP tools plus an authorization policy that actually denies.

## How a tool is defined — not what the class shape suggests

Read from `%AI.Tool.Generator` rather than assumed. **`%Invoke` and `%Discover` are generated at compile time** — you do not implement them. Every public, non-internal ClassMethod not starting with `%` becomes a tool, its `Description` becomes what the model reads, and argument descriptions come from the formal spec.

**So a public helper silently becomes an extra tool.** Every helper here is `Private` for that reason, and the counts are asserted after the merge: **5 read tools, 1 write tool.**

## Verified live — the five read tools

```
GetHostStatus     {"host":"Cloud API","found":1,"status":"OK","enabled":1}
GetQueueDepth     {"host":"Cloud API","queued":0,"measured":1}
GetPoolSize       {"host":"Cloud API","poolSize":1,"found":1}
GetProcessingTime {"avgProcessingTime":0.01,"avgQueueingTime":0,"measured":1}
GetRecentErrors   {"count":28,"sanitised":true,"byCode":[...]}
unknown host      {"poolSize":null,"found":false,"error":"no such config item ..."}
```

**The PHI boundary holds under real conditions**, which is the point rather than a formality. Armed `ErrorRate()` to generate genuine failures, then checked the tool's own output:

```
contains 'PatientID': no      contains 'socket':    no
contains 'Patient':   no      contains '127.0.0.1': no
```

28 errors classified into three codes, **no log text of any kind**. Necessary rather than cautious — `Ens_Util.Log` on this instance holds **61,772 rows carrying `PatientID`** in plain text.

## Verified live — the write tool, every path

```
wrong host    refused  not_whitelisted_host    before=null (nothing read or touched)
size 1        refused  out_of_bounds
size 99       refused  out_of_bounds
size 2.5      refused  out_of_bounds
dry run       previewed  before=1 after=4  -- pool was STILL 1 afterwards
apply         applied    before=1 after=4  -- pool verified 4
apply again   no_change  before=4 after=4  -- not a failure
```

**The lower bound is 2, not 1**, and that is not an off-by-one: `1` is the *shipped* value, so "set the pool to 1" is a no-op dressed as a fix. Reversal to 1 goes through `Triggers.Reset()`, which is not LLM-callable.

## The default policy permits everything — this is the finding that mattered

I flagged this as unverified on #89 and then checked it. The answer was the unsafe one:

```
%AI.ToolMgr.%New().AuthPolicy          ->  <none>
%AI.Policy.Authorization::%CanExecute  ->  Return $$$OK
```

So *"authorization is enforced by the runtime"* was **true and vacuous** — the runtime faithfully asked a policy that says yes to everything. The whitelist and bounds inside `Tools.Resolve` still applied, but those are **the tool defending itself, not an authorisation boundary**.

**Not `%AI.Policy.ConsoleAuth`**, the other shipped option: reading it shows it writes an ANSI-coloured "Authorization Request" box to a terminal and waits for a keypress. Correct for `%AI.Shell.Console`, wrong for an agent over HTTP.

**Observed denial, every branch:**

```
holds PG_Read only    SetPoolSize    DENIED "requires PG_Resolve:USE"
                      SetPoolSize    listable = 1     <- visible but denied
                      GetQueueDepth  allowed
holds both            SetPoolSize    allowed
holds nothing         GetQueueDepth  DENIED "requires PG_Read:USE"
                      GetQueueDepth  listable = 0
```

Listable-but-not-executable is the point: hiding the write tool would make "you may not do this" indistinguishable from "this product cannot do this", and the agent could not name the action it recommends. This is the **"AI Detective can look without permission to act"** story.

**Why the test is a stub, not a real unprivileged user:** `$SYSTEM.Security.Check` returns `1` for *every* resource when the caller holds `%All` — including resources that do not exist, verified by checking `PG_Read` before creating it. The deny branch is unreachable from any session that can run this code. `StubPolicy`'s comment states what it does and does not prove.

## Three bugs found by running code that compiled cleanly

1. **`rs.Data("ConfigName")` → `<UNDEFINED>`.** The column is **`Name`**. Same runtime-only class as #37.
2. **`GetProcessingTime` read columns that do not exist** — in *any* of the three statistics queries. Checked with `GetColumnName()` rather than guessing a fourth time; it now scrapes `iris_interop_avg_*` from the instance's own metrics endpoint, which is what the contract says it wraps.
3. **`$zdatetime($ztimestamp - fraction)` → `<ILLEGAL VALUE>`.** `$ztimestamp` is `"days,seconds"`.

All three are the class the `iris-compile` job documents itself as *not* catching.

## Two corrections to the ratified contracts

Recorded rather than silently diverged from — `mcp-tools.md` and `resolve-api.md` need amending:

1. **Role names cannot start with `%`.** Both write `%Guardian_Read` / `%Guardian_Resolve`; IRIS refuses with `ERROR #887: Invalid role name`. Created as **`Guardian_Read`** / **`Guardian_Resolve`**.
2. `Security.Roles.Create` takes **positional** args, not a properties array.

## The CI change, and #77 called this in advance

These classes **cannot compile in CI**. Verified rather than assumed — started `intersystems/irishealth-community:latest-em`:

```
%AI.Tool exists: 0
```

#77 pre-registered this with ranked options. Took **option 1 — exclude and say which** — over dropping the job, because a red build on correct code is where the cheapest response is the worst one. CI confirms: `copied 12 classes / skipped 4`, with a `::notice::` naming the gap.

**What it costs:** these four classes get *no* CI compile check. Verified only against the EAP image on one machine — **#70's problem in a narrower form.** Option 2, a self-hosted runner, fixes both.

Refs #72, #77, #70


Ari-Glikman:
Reviewed as Dev C. The three things I actually verified rather than accepted:

**1. The tool counts, because they are the security property.** Every public ClassMethod becomes an LLM-callable tool, so a stray public helper is a stray capability — and on the write class it would be a second way to mutate a production. Re-asserted after merging `main`: **5 read, 1 write.** Both helper methods in each class are `Private`.

**2. The PHI boundary under load, not at rest.** Clean output on an idle instance proves nothing, so I armed `ErrorRate()` and checked the tool's own JSON with 28 real errors present: no `PatientID`, no `Patient`, no log text, no IP addresses. The allowlist maps an IRIS error token to a catalogue string and returns `unclassified` with no text otherwise — necessary rather than cautious, since `Ens_Util.Log` holds 61,772 rows carrying `PatientID`.

**3. That a denial is actually observable.** This is the one I would have pushed back on hardest, because "the runtime enforces it" was in a ratified contract and turned out to be vacuous — no policy registered, base `%CanExecute` returns `$$$OK`. The fix is here with denials measured on every branch, including `SetPoolSize` **denied but still listable**.

**The stub deserves scrutiny rather than a pass.** Testing authorization with an injected resource set is weaker than testing a real unprivileged user, and I want that on the record: `$SYSTEM.Security.Check` returns 1 for every resource under `%All` — including ones that do not exist, which I confirmed by checking `PG_Read` before creating it. So the deny branch is unreachable from any session that can run the code. `StubPolicy` says what it proves (our policy denies when the check says no) and what it does not (that IRIS's check is correct). A real end-to-end RBAC test needs the agent calling over HTTP as a non-privileged user, which is the right test once a provider exists.

Two contract corrections are recorded in the code rather than fixed in `contracts/` — role names cannot start with `%`, and `Security.Roles.Create` is positional. Those belong in a contracts PR of their own, with a CHANGELOG entry, not smuggled in here.

Merging. Not passing `--delete-branch` this time: doing that on #88 auto-closed #89, which was stacked on it.

@kskubach — the two things worth your eye tomorrow are the `get_recent_errors` allowlist (is code-only classification enough, or should it return nothing at all?) and whether excluding four classes from CI is the right call versus a self-hosted runner.

PR #92, opened 2026-08-18
feat: MVP 2 WHY and FIX — AI Detective orchestration and Smart Resolve (#72, #73)

### LINKED ISSUE #72: docker compose up should stand the whole chain up — IN MVP 1; the private AI Hub image becomes a documented one-time prerequisite

Ari-Glikman:
Closes #72, closes #73.

Completes the MVP 2 loop: **WHAT** (Health Scan, shipped) → **WHY** (`POST /api/investigate`) → **FIX** (`POST /api/resolve`).

## The architectural line, and where it is enforced

This service gained **orchestration**, not **authority** (`services/detection-engine/CLAUDE.md` §1.1).

| | Lives in the engine | Lives in IRIS |
|---|---|---|
| WHY | builds the request, validates the reply | the agent, the LLM key, the read tools |
| FIX | validates the request, reports the outcome | the write, the policy, the audit record |

No LLM credential reaches this service — the key is in the AI Hub wallet, referenced by a `secret://` URI from ConfigStore. No `Ens.*` call exists in the engine. A bug in `resolve.ts` cannot widen what may be written; it can only fail to reach the tool.

`test/resolve.test.ts` pins *where* the boundary is: an out-of-range size **parses** in the engine and is **refused** by the tool. That is deliberate, so nobody later reads the parser's acceptance as permission.

## What was verified live, with a real LLM

Four-container stack, `PG_AGENT_MODE=live`, `gpt-4o-mini`:

```
armed PoolBottleneck            -> queue 151 -> queue_buildup fired
POST /api/investigate           -> 6 tool calls, confidence 0.95, 6.0s
                                   all 6 evidence bullets source=mcp_tool, none llm-asserted
                                   independently recommended set_pool_size Cloud API 4
POST /api/resolve dry_run       -> previewed 1 -> 4;  production still read 1
POST /api/resolve apply         -> applied;           production read 4
queue 222 -> 205 -> 176 -> 154 -> 124 -> 105 -> 75 -> 16 -> 0
finding cleared; second apply   -> no_change
Reset()                         -> pool restored to 1
```

The agent chose the action itself. Nothing about pool size is in the system prompt beyond the permitted action type and its range.

## Refusals came from the real tool, not from our validation

| Request | Outcome |
|---|---|
| `EMR Source`, `Lab Router`, unknown host | `refused / not_whitelisted_host` |
| size `0`, `-4`, `99` | `refused / out_of_bounds` |
| size **`1`** | `refused / out_of_bounds` — it is the *shipped* value, so a no-op dressed as a fix |
| size `8` | `previewed` (upper bound is legal) |

Pool read `4` throughout: no refusal moved anything.

## Failure honesty — the part most worth reviewing

- A tool call that does not come back reports **`liveStateVerified: false`**. We do not know whether the production changed, and someone deciding whether to retry an apply needs to know we cannot tell. Verified by pointing `IRIS_BASE_URL` at a dead port: `failed / tool_call / liveStateVerified: false`.
- An **unrecognised outcome** from the one component that can mutate a production is reported as `failed`, never guessed at.
- **`confirmation` is attached only to `applied`**, with `directEvidence: false`. This response is evidence the *write* landed, not evidence the *problem* cleared — the queue still has to drain, observed on a later poll. A confirmation on a preview or a refusal would tell the UI to watch for a clearance that is never coming.
- A malformed reply from the agent is `state: "unavailable"` with `rootCause: null`. **Never a plausible guess** — a human approves a production write on the strength of that narrative.

## Bugs found and fixed in the course of this

- **`GET /api/resolve` answered 404** — a real endpoint reported as absent, the mirror of the `POST /api/healthscan/findings` regression from the previous commit. No test covered it because the suite only ever asked one way round. Both directions are now pinned, for both POST routes (they are wired through different branches, so covering one proves nothing about the other). Found by asking for the symmetric case, not by a failing test.
- **`liveAgent` posted to `/labdemo/investigate`**, missing the web application prefix. It would have reached the patient dispatcher and 404'd — naming the wrong component in the error.
- **`resolve` collides with `node:path`'s `resolve`** in `index.ts`. It compiled as a duplicate-identifier error, but had it shadowed, `resolve(serviceRoot, 'thresholds.json')` would have called the write orchestrator. Aliased to `runResolve`.

- **The CI iris job failed on correct code.** `AgentDispatcher.cls` references `%AI.Agent`, which is absent from the community image CI uses, and the exclusion was a hardcoded `-not -path 'iris/labdemo/Tools/*'` — it encoded "AI Hub classes live in `Tools/`", true when written and false as soon as the HTTP entry point was added. Now selected **by content** (`grep -rl '%AI\.'`), so a new class touching `%AI.` is excluded automatically. `iris/test/` stays a path rule because `StubPolicy.cls` *extends* an excluded class without naming one, which a content grep misses. The notice now names the skipped files rather than only counting them.

## Notes for review

- **`AGENT_MODE` defaults to `mock`** even though `PROXY_MODE` is `live`. Two independent switches because the proxy and the LLM fail independently, and a clone has no provider configured. `mockResolveTool` tracks a pool size in memory so apply-then-apply gives `applied` then `no_change` — a mock that returned `applied` twice would let Dev C ship an approve button whose double-click path is never exercised, which a live demo will hit.
- **The agent's tool readings are live; the finding's snapshot is from when it fired.** These can legitimately disagree — in one run the snapshot said `queued: 198` while `get_queue_depth` returned `0`, because the production had been reset in between. Neither is wrong; `evidence[].source` is the distinction (`mcp_tool` = now, `snapshot` = then). Documented in `AgentDispatcher.cls` because the transcript reads as a defect otherwise.
- **Known limitation, flagged rather than hidden:** the engine authenticates to the dispatcher with its *own* service credentials, so the AI Hub audit record attributes the change to the engine rather than to the person who approved it. `resolve-api.md` §13.1 names this as unsettled. It is a real gap, not a detail.

## Checks

- `npm run typecheck` — clean, strict, no `any`
- `npm test` — **221 pass, 0 fail** (22 new in `resolve.test.ts`, 14 of them rejections or failure paths)
- ADR 0004 holds: engine starts and serves `POST /api/resolve` with no IRIS and no proxy reachable
- **All five CI jobs green**, including `iris — compile ObjectScript`: 12 classes compiled, 5 skipped and named

Requesting @kskubach (Dev A) — the `iris/` files here are `AgentDispatcher.cls`, which is your area, and it is worth your eyes specifically because it is the HTTP entry point to a governed production write.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,35 +247,66 @@ jobs:
       # Flat copy on purpose: iris/ does not mirror the package layout, and LoadDir reads
       # each class's declared name from the file rather than from its path.
       #
-      # EXCLUDES iris/labdemo/Tools/, and this is the day the comment above predicted. Those
-      # classes extend %AI.Tool, which does not exist in this image -- verified by starting
-      # `intersystems/irishealth-community:latest-em` and asking it:
+      # EXCLUDES the AI-Hub-dependent classes, and this is the day the comment above predicted.
+      # They reference %AI.Tool / %AI.Agent, which do not exist in this image -- verified by
+      # starting `intersystems/irishealth-community:latest-em` and asking it:
       #
       #     %AI.Tool exists: 0
       #     %AI.Agent exists: 0
       #
       # So compiling them here would fail on CORRECT code. Taking option 1 of the three ranked
       # above -- exclude and SAY WHICH -- rather than dropping the job, which is the cheapest
-      # response and the worst one: the other 11 classes stay covered, and this comment plus the
+      # response and the worst one: every other class stays covered, and this comment plus the
       # step output below name the gap so it is not silently assumed.
       #
-      # WHAT THAT COSTS, stated plainly: the AI Hub tool classes get NO compile check in CI. They
-      # are verified only against the EAP image on a developer machine, which is #70's
+      # SELECTED BY CONTENT, NOT BY PATH. This was `-not -path 'iris/labdemo/Tools/*'` until
+      # AgentDispatcher.cls landed in iris/labdemo/REST/ and failed the job on correct code
+      # (#92). The path rule encoded "AI Hub classes live in Tools/", which was true when
+      # written and false as soon as the HTTP entry point was added -- and a third hardcoded
+      # path in three places is a rule that goes stale again the next time. Grepping for the
+      # dependency itself cannot: a new class that touches %AI. is excluded automatically, and
+      # one that stops touching it is picked back up.
+      #
+      # iris/test/ stays a path exclusion because StubPolicy.cls names no %AI. class itself --
+      # it EXTENDS one that does, so it is transitively dependent and a content grep misses it.
+      # Following Extends chains properly is a compiler's job, not a shell one-liner's; the
+      # directory is the honest boundary for test doubles.
+      #
+      # WHAT THAT COSTS, stated plainly: the AI Hub classes get NO compile check in CI. They are
+      # verified only against the EAP image on a developer machine, which is #70's
       # single-machine problem returning in a narrower form. Option 2 (a self-hosted runner with
       # the EAP image) fixes both and is the real answer when someone has the time.
       - name: Copy classes into the container
         if: steps.guard.outputs.present == 'true'
         run: |
           docker exec pg-iris bash -c 'mkdir -p /tmp/pg-iris-src'
-          find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' -not -path 'iris/test/*' -exec docker cp {} pg-iris:/tmp/pg-iris-src/ \;
+          # ONE list, computed once, used for both the copy and the count -- so the notice can
+          # never disagree with what was actually compiled. Deriving the two separately is how a
+          # report drifts from the thing it reports on.
+          #
+          # Two rules, unioned: any class that NAMES an %AI. class, plus everything under
+          # iris/test/. See the comment above for why the second one cannot also be a grep.
+          skip_list=$({ grep -rl '%AI\.' --include='*.cls' iris/; find iris/test -name '*.cls'; } | sed '/^$/d' | sort -u)
+          all_list=$(find iris -name '*.cls' | sort)
+          copy_list=$(comm -23 <(echo "$all_list") <(echo "$skip_list"))
+          # `while read` in a PIPELINE runs in a subshell, so an `exit 1` inside it would not fail
+          # this step -- a copy that silently did not happen would show up as a class not compiled
+          # rather than as an error. Fed by a here-string instead, so the shell is the same one.
+          while read -r f; do
+            [ -n "$f" ] || continue
+            docker cp "$f" pg-iris:/tmp/pg-iris-src/ || { echo "::error::failed to copy $f"; exit 1; }
+          done <<< "$copy_list"
           docker cp tools/ci-compile-iris.txt pg-iris:/tmp/ci-compile-iris.txt
-          copied=$(find iris -name '*.cls' -not -path 'iris/labdemo/Tools/*' -not -path 'iris/test/*' | wc -l)
-          skipped=$(find iris -name '*.cls' \( -path 'iris/labdemo/Tools/*' -o -path 'iris/test/*' \) | wc -l)
+          copied=$(echo "$copy_list" | sed '/^$/d' | wc -l)
+          skipped=$(echo "$skip_list" | sed '/^$/d' | wc -l)
           echo "copied $copied classes"
-          echo "::notice::skipped $skipped AI-Hub-dependent class(es) under iris/labdemo/Tools/ and iris/test/ — %AI.Tool is absent from this image, so they are compile-checked only against the EAP image (see the comment on this step and #77)"
-      # Greps for the marker rather than trusting the exit status: `iris session` exits 0
-      # even when the script inside reported a compile error, which would make this job
-      # exactly the kind of check that passes while verifying nothing.
+          # Names the files, not just the count. "skipped 4" tells a reader nothing about WHICH
+          # coverage they lack, and this list is the gap -- so it should be readable without
+          # re-deriving the find expression.
+          echo "$skip_list" | sed 's/^/  skipped: /'
+          # `paste -sd' '` rather than `tr` for the one-line form: a literal newline inside a
+          # YAML block scalar is what a `tr` argument would need, and it breaks the block.
+          echo "::notice::skipped $skipped AI-Hub-dependent class(es) — %AI.Tool and %AI.Agent are absent from this image, so they are compile-checked only against the EAP image (see the comment on this step and #77): $(echo "$skip_list" | paste -sd ' ' -)"
       - name: Compile
         if: steps.guard.outputs.present == 'true'
         run: |
--- a/apps/dashboard/nginx.conf.template
+++ b/apps/dashboard/nginx.conf.template
@@ -31,7 +31,28 @@ server {
         try_files $uri /index.html;
     }
 
-    location /api/healthscan/ {
+    # /api/ AND NOT /api/healthscan/, which is what this was until 2026-08-19.
+    #
+    # The narrower prefix meant every MVP 2 endpoint fell through to `location /` and was served
+    # index.html: /api/earlywarning, /api/investigate and /api/resolve all returned the SPA. A GET
+    # answered 200 with an HTML body, so a curl check on the status code looked healthy -- measured:
+    #
+    #     GET  /api/healthscan/findings  -> 200 JSON
+    #     GET  /api/earlywarning         -> 200 <!doctype html>
+    #     POST /api/resolve              -> 405 from nginx, never reaching the engine
+    #
+    # The POST is the one that matters: `location /` is a static file handler, and nginx refuses
+    # POST to it before the request goes anywhere. So the approve button could not have worked in
+    # the containerised dashboard at all, and the failure would have surfaced as nginx's own 405
+    # page rather than as anything in the engine's log -- which is exactly the invisible-failure
+    # shape resolve-api.md warned about for the CORS preflight, arriving by a different route.
+    #
+    # Widened rather than adding three more blocks: the engine owns everything under /api and
+    # answers 404 for a path it does not have (verified -- GET /api/nope -> 404, not a fallthrough),
+    # so a per-endpoint list here would be a second copy of its routing table. That is the #84
+    # stale-copy pattern, and it is what caused this bug: the block was written when
+    # /api/healthscan/* was the only route family and never revisited when MVP 2 added three more.
+    location /api/ {
         # RESOLVE AT REQUEST TIME, NOT AT CONFIG LOAD. A literal hostname in proxy_pass is
         # resolved when nginx reads its config, so this container refused to start at all
         # while the engine was absent -- `host not found in upstream`, exit before serving
@@ -67,7 +88,17 @@ server {
         # nginx default of 60s. Shorter than the poll interval would abort healthy
         # requests; this is comfortably above it and well below the default.
         proxy_connect_timeout 5s;
-        proxy_read_timeout 10s;
+        # 10s WOULD CUT OFF AN INVESTIGATION. /api/investigate is an LLM round trip with several
+        # tool calls inside it -- measured at ~6s against gpt-4o-mini, and the engine's own hard
+        # timeout is 30s. A 10s read timeout here would abort the slower half of real
+        # investigations and return 504, which reads as "the AI is broken" rather than "the proxy
+        # gave up". 35s leaves the engine's own timeout as the one that fires, so the failure is
+        # reported by the component that knows what failed.
+        #
+        # The cost is a hung upstream holding a worker for 35s instead of 10s on this path. That is
+        # acceptable here and would not be on the poll path -- but the poll aborts client-side
+        # every 2s regardless, so it never reaches this limit.
+        proxy_read_timeout 35s;
 
         # Findings change every poll. A cached 200 here would freeze the dashboard while
         # it looked live, which is the one failure mode worse than an error banner.
--- a/apps/dashboard/vite.config.ts
+++ b/apps/dashboard/vite.config.ts
@@ -13,7 +13,14 @@ export default defineConfig({
     proxy: {
       // Proxying in dev means CORS is never the dashboard's problem, and Dev B's
       // findings API can be Node or Python without a change here.
-      '/api/healthscan': {
+      //
+      // '/api' AND NOT '/api/healthscan' -- the narrower prefix left every MVP 2 endpoint
+      // (/api/earlywarning, /api/investigate, /api/resolve) unproxied, so Vite served index.html
+      // for them and a GET looked like a 200. Same defect as the nginx template had; both were
+      // written when /api/healthscan/* was the only family. One prefix, because the engine owns
+      // everything under /api and 404s what it does not have -- listing endpoints here would be a
+      // second copy of its routing table, which is how this broke in the first place.
+      '/api': {
         target: process.env.VITE_HEALTHSCAN_TARGET ?? 'http://localhost:3002',
         changeOrigin: true,
       },
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -227,6 +227,22 @@ services:
       # sustainedSeconds must be reachable within sustainedSamples polls with margin (#64).
       # Shortening this means retuning sustainedSeconds in the same change.
       POLL_INTERVAL_MS: "5000"
+      # AGENT_MODE is SEPARATE from PROXY_MODE, and defaults to mock even though the proxy is
+      # live. Two independent switches because the proxy and the LLM fail independently: a
+      # rate-limited or unreachable model must not take the findings API down with it, and the
+      # useful default for a fresh `compose up` is live metrics with a canned investigation.
+      #
+      # `live` needs the AI Hub provider configured in IRIS (docs/mvp2-aihub-verified-api.md)
+      # and an API key in the wallet. Neither the key nor the model name is here: the engine
+      # holds no LLM credential, and turning this to `live` without the provider gives a 503
+      # from the dispatcher rather than a broken investigation.
+      AGENT_MODE: ${PG_AGENT_MODE:-mock}
+      IRIS_BASE_URL: http://iris:52773
+      # The ENGINE's IRIS credentials, for the RBAC-gated dispatcher. Same variables as the
+      # proxy above, so one override moves both. These authenticate to IRIS; they carry no
+      # authority over what may be written -- Tools.Resolve's policy decides that.
+      IRIS_USER: ${PG_IRIS_USER:-superuser}
+      IRIS_PASS: ${PG_IRIS_PASS:-SYS}
     volumes:
       # Hot-reloaded (ADR 0003), so mounting it lets an operator tune thresholds without a
       # rebuild. Read-only: the engine only ever reads it.
--- a/docs/mvp2-aihub-verified-api.md
+++ b/docs/mvp2-aihub-verified-api.md
@@ -127,9 +127,8 @@ execution.
 
 ## What is NOT yet verified
 
-- **No LLM provider is configured.** No `%AI.Provider` instance, no ConfigStore entry, no API
-  key. Nothing has called an external model from this instance. The provider/key path is real
-  work, not a checkbox.
+- ~~No LLM provider is configured.~~ **CLOSED** -- see the provider section at the end of this
+  file. A real OpenAI call and an agent tool call have both been made from this instance.
 - **What an audit record CONTAINS, and whether it is queryable.** Enforcement is confirmed (see
   above), but the written record has not been observed, and Dev C needs to display one. This is
   now the highest-value remaining unknown.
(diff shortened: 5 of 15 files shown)


kskubach (changes requested):
Reviewing the **`iris/**` portion only** — that is my area, and I have not audited the TypeScript in depth.

`AgentDispatcher.cls` is good work and the comments are genuinely useful (the snapshot-vs-live `evidence[].source` distinction especially — that would have cost someone an afternoon). But there is one blocking finding, and it is in the class's central safety claim.

## `/labdemo/resolve` is neither authorized nor audited

The class comment says:

> THIS CLASS ADDS NO AUTHORITY. `/labdemo/resolve` calls `Tools.Resolve.SetPoolSize`, which is gated by `%AI.ToolMgr.ExecuteTool`'s authorization check and audited by the same runtime path.

The first half is true. The second is not, because **no policy is registered on that manager**:

```objectscript
set mgr = ##class(%AI.ToolMgr).%New()
set sc = mgr.RegisterToolSet("ProductionGuardian.LabDemo.Tools.Resolve")
set result = mgr.ExecuteTool("SetPoolSize", args)
```

A fresh `%AI.ToolMgr` on build 126 reports:

```
AuthPolicy : <none>
AuditPolicy: <none>
DiscoveryPolicy: <none>
```

`SetAuthPolicy` / `SetAuditPolicy` attach a policy to the Rust manager identified by *that object's* `%token` — they are per-instance and in memory, not a persisted setting. `AuthPolicy.cls` exists on this branch but is never registered anywhere (`git grep SetAuthPolicy` over `iris/` and `services/` on this branch: no hits). So `ExecuteTool` faithfully asks a policy that permits everything, and audits into nothing.

### Measured, same user, same call, same instant

I ran your exact sequence as a real principal holding `Guardian_Read` and **not** `PG_Resolve`:

```
running as: pg_proof_readonly   roles: Guardian_Read
AuthPolicy : <none>
AuditPolicy: <none>
ExecuteTool: SUCCEEDED -> {"outcome":"previewed","before":{"poolSize":1},"after":{"poolSize":4},...}
audit rows written: 0
```

versus the same call through a governed manager:

```
outcome: denied
error: ERROR <%AICore>ToolAccessDenied: 'SetPoolSize' modifies a live production
       and requires PG_Resolve:USE
```

`/labdemo` is `AutheEnabled=96` (unauthenticated **or** password), so as it stands anything that can POST to `/labdemo/resolve` changes the pool size of a live production with no `PG_Resolve` check and no audit row. That is the row MVP 2 §6 rates **Critical**, with "RBAC-gated" and "every call audited" as its mitigations, and `resolve-api.md` §8's "the contract does not permit an unattributed write".

To be clear about blame: this is not a mistake in your code so much as a trap in the API. `UseToolSet` delegates straight to `..ToolManager.RegisterToolSet`, so the natural thing to write is the ungoverned thing, and it works perfectly — which is what makes it dangerous. I only found it because I went looking for where the policy gets registered and found that nothing did.

### The fix

I have `iris/labdemo/Tools/Governance.cls` ready (PR to follow shortly) — one factory that cannot return an ungoverned manager:

```objectscript
// Resolve()
set mgr = ##class(ProductionGuardian.LabDemo.Tools.Governance).ToolManager(1, .sc)

// BuildAgent() -- instead of UseToolSet, and after %Init()
set sc = ##class(ProductionGuardian.LabDemo.Tools.Governance).GovernAgent(agent, 0)
```

`GovernAgent(agent, 0)` registers read tools only, keeping your "the agent investigates; it does not act" property — which is right and I would not change it.

## Two smaller notes

1. **`BuildAgent` uses read tools only — agreed, and worth keeping.** But those read calls are also currently unaudited, and MVP 2 §2.2 promises "every MCP tool call, read *and* write, is recorded". `GovernAgent` fixes both at once.

2. **`docs/mvp2-aihub-verified-api.md` will conflict.** I have rewritten large parts of that file (closing §5.6(a) and (b) with measurements, and documenting that the runtime does **not** audit authorization denials — `ExecuteTool` throws before the audit hook, so `AuthPolicy` has to write that row itself). Happy to take the conflict on my side; flagging so it is not a surprise.

Not blocking, and separate: `contracts/mcp-tools.md` §5.5 and `resolve-api.md` §8 both promise that `not_authorized` refusals are audited. The runtime cannot do that. Tool-level refusals (the `2..8` bounds guard) *are* audited because the tool ran; authorization denials are not. That is a contract correction, and it is mine to raise rather than yours to fix here.

Ari-Glikman:
## Addressed in `a7708c5`. Your finding was correct and the comment was the worst part of it.

I reproduced it before changing anything, on this branch:

```
AuthPolicy : <none>
AuditPolicy: <none>
register: OK
ExecuteTool: SUCCEEDED -> {"outcome":"previewed","before":{"poolSize":1},"after":{"poolSize":4},...}
```

`git grep SetAuthPolicy -- iris/ services/`: no hits. `AuthPolicy.cls` was correct and never consulted.

**The part I want on the record:** `AuthPolicy.cls`'s own header documents this trap — *"with no policy registered, the runtime faithfully asks a policy that permits everything"* — and I wrote the dispatcher that falls into it four days after writing that sentence, then asserted the opposite in its class comment. An unimplemented guard is a gap. A guard **documented as present** is worse, because a reviewer reading the comment has no reason to look. The comment now carries the correction in full rather than being quietly swapped out, for the same reason #93's rewritten paragraph does: a corrected value under intact bad reasoning invites someone to change it back.

## What changed

**`Resolve()`** → `Governance.Invoke("SetPoolSize", args)`.

`Invoke` rather than `ToolManager` + `ExecuteTool`, for the `auditId` — the handle is only knowable after the call and `resolve-api.md` §8 requires it in every response. My version omitted it entirely, so every response I published claimed compliance with §8 while carrying no handle at all. A second thing your PR fixed that I had not noticed was broken.

A denial now answers **403 with `not_authorized` and the audit handle**, not 500. The policy refusing a caller is the boundary working; a server fault would send someone to debug the production instead of granting a role.

**`BuildAgent()`** → `GovernAgent(agent, 0)` **after** `%Init()`, replacing `UseToolSet` before it. Your ordering note was load-bearing and I would have got it wrong: my old code called `UseToolSet` *before* `%Init()` deliberately, with a comment explaining that tools added later are not registered. That reasoning is right for `UseToolSet` and exactly backwards for governance, because `%Init` is what builds the `ToolManager` the policies attach to.

## Verified live — four things, not one

**1. The boundary denies.** I ran your `GovernanceProof` end to end rather than trusting it:

```
2. GetPoolSize : executed  <- expected executed
3. SetPoolSize : denied    <- expected denied
   error   : ERROR <%AICore>ToolAccessDenied: 'SetPoolSize' modifies a live production
             and requires PG_Resolve:USE
   auditId : pg-audit-6
```

**2. The dispatcher's write is audited and returns its handle.**

```json
{"outcome":"previewed","before":{"poolSize":1},"after":{"poolSize":4},"audit":{"auditId":"pg-audit-7"}}
```

**3. The agent's reads are audited — 7 → 12 rows across one investigation, one per tool call:**

```
pg-audit-12  SuperUser  GetRecentErrors     executed
pg-audit-11  SuperUser  GetQueueDepth       executed
pg-audit-10  SuperUser  GetProcessingTime   executed
pg-audit-9   SuperUser  GetPoolSize         executed
pg-audit-8   SuperUser  GetHostStatus       executed
```

Every LLM tool call is now individually attributable. That is the read half of MVP 2 §2.2, which my `UseToolSet` silently skipped.

**4. The loop still closes.** `queue_buildup` at 95 → live `gpt-4o-mini`, 5 tool calls, recommended `set_pool_size 4` → applied 1→4 → queue `127→97→68→41→21→0` → findings cleared → `Reset()` restored pool 1.

221 engine tests pass, typecheck clean, the class compiles against the EAP image.

## Ordering, and the one thing still open

**#92 now depends on #94** — `Governance` and `Audit.Entry` live there, so #94 must land first or this branch will not compile in an environment that has the EAP image. CI's iris job passes either way because it skips `%AI.`-dependent classes by content (a change in this PR), which means **CI cannot catch that dependency**. Flagging rather than relying on it.

I approved #94 with one reproducible gap in `GovernanceProof` — `AsReadOnlyUser` dies at `<PROTECT>` on the LABDEMO database before reaching the policy, because `Guardian_Read` grants `PG_Read:U` and nothing else. I only got the PASS above after adding `%DB_%DEFAULT` to the probe user by hand. Worth fixing because the failure *looks* like the boundary working. Details in the review.

Also still open and mine to fix, separate from your review: the engine authenticates to `/labdemo/agent/*` with its own service credentials, so `actor` in these rows is `SuperUser` rather than the person who approved. The rows above show it plainly — `resolve-api.md` §13.1 already flags it, and now there is an audit trail to point at when we settle it.

On your contract note (`not_authorized` refusals cannot be audited by the runtime): agreed, and it is yours to raise. Worth adding that your `AuthPolicy` **does** now audit denials — `pg-audit-6` above — so the correction is narrower than it first looks: the *runtime* cannot, our policy can.

kskubach (changes requested):
The governance fix is right, and the comment correction is better than the code change. Keeping the false claim in the header with an explanation of *why* it was dangerous — "an unimplemented guard is a gap, a guard documented as present is a trap" — is worth more to the next reader than a clean deletion would have been.

Two things I verified rather than took on trust:

**`%Init()` then `GovernAgent` is the correct order, and it works.** I had a doubt about my own precondition — `GovernAgent` errors if `agent.ToolManager` is null, and I wanted to be sure `%Init` had produced one. It is actually instantiated earlier than that:

```
after %New : %AI.ToolMgr
%Init      : OK
after %Init: %AI.ToolMgr
```

So your ordering is safe either way. `GovernAgent(agent, 0)` for read-only is exactly right, and your "defence in depth rather than the boundary" framing is the correct reading of it.

**`Governance.Invoke` for the `auditId` was the right call** over `ToolManager()` + `ExecuteTool`, for the reason you give.

## Blocking: the denial response now breaks `resolve-api.md`

Three problems, all in the new `outcome = "denied"` branch, and the first is explicit in the contract.

### 1. `403` is specifically forbidden

```objectscript
set %response.Status = 403
```

§5.1, and it names the exact anti-pattern:

> `not_authorized` is `200`. It is not `403`, and it is not a `500`.
>
> 2. **A status code carries one bit and gets swallowed.** Dev C's client branches on `res.ok`; every generic fetch wrapper turns a non-2xx into a thrown error and a generic banner. The specific, informative refusal becomes "something went wrong" in the layer above the one that knew.

Your instinct is defensible in the abstract — I'd have said 403 too — but it costs the demo its acceptance criterion. MVP 2 §5.4 requires "the RBAC-denied state is visible", and a thrown error is not a visible state. This is the one change I'd call blocking on its own.

### 2. `refusal` field names don't match

You emit `{"code": ..., "detail": ...}`. §5 specifies:

```json
"refusal": {
  "reason": "not_authorized",
  "message": "...",
  "checkedBy": "iris"
}
```

`reason` and `message`. §5's table treats the `reason` set as **closed**, and instructs consumers to "render `refusal.message` verbatim" for anything unrecognised — so Dev C reads two fields that your payload does not have, and gets `undefined` in the denied banner.

### 3. The denied shape has to match the authorized one

§5.1 again:

> An authorized caller and an unauthorized one differ **only** in `outcome`, `refusal` and the absence of `before`/`after`. Same endpoint, same status, same shape. Dev C mocks the denied state by changing two fields.

Your denial carries `outcome`, `refusal`, `audit` and nothing else. Missing `before`/`after`/`reversal` as explicit `null`, and `confirmation`, which §3's table marks *always present* with `status: "not_applicable"`.

Also: `audit` on a refusal should be the full §8 block — `actor`, `role`, `tool`, `recordedAt`, `source` — not just `auditId`. §8 is explicit that "§11.3's RBAC-denied example carries a populated `audit` block **naming the refused identity**", which is the whole point of auditing denials. You already have that for free:

```objectscript
set audit = ##class(ProductionGuardian.LabDemo.Audit.Entry).Describe(governed.auditId)
```

returns `{auditId, actor, role, tool, recordedAt, source, disposition, denialReason}` straight from the row your denial wrote.

One smaller thing: `governed.error` is the raw `ERROR <%AICore>ToolAccessDenied: Tool access denied: Global policy denied: ERROR #5001: '...'`. That is fine for a log and poor as a UI `message`. The audit row's `denialReason` is the clean sentence — `'SetPoolSize' modifies a live production and requires PG_Resolve:USE` — and it is already free of instance data by construction, which matters because it crosses the same boundary as any other output.

## Suggested shape

```objectscript
if governed.outcome = "denied" {
    set audit = ##class(ProductionGuardian.LabDemo.Audit.Entry).Describe(governed.auditId)
    write {
        "outcome": "refused",
        "before": null, "after": null, "reversal": null,
        "refusal": {
            "reason": "not_authorized",
            "message": ($select($isobject(audit): audit.denialReason, 1: governed.error)),
            "checkedBy": "iris"
        },
        "confirmation": { "status": "not_applicable" },
        "audit": (audit)
    }.%ToJSON()
    return $$$OK
}
```

No status override — `200`, same as the authorized path.

Worth saying: the reason this is catchable at all is that §5.1 argues its case instead of just stating a number. I would have merged a `403` without blinking if the contract had only said "use 200".

Nothing else in the diff concerns me. #94 carries `Describe()` and `Invoke()` as used above; it is approved and green, so ordering is whichever suits you.

Ari-Glikman:
## All three blocking findings addressed in `b1448b5`, and you were right about each

**1. `403` → `200`.** §5.1 forbids it by name, and I read the section rather than just the rule — the reasoning is what convinced me: Dev C's client branches on `res.ok`, so a non-2xx becomes a thrown error and a generic banner, and MVP 2 §5.4's criterion is that the RBAC-denied state is *visible*. Your "I'd have said 403 too" is worth noting, because it means the contract arguing its case is what caught this, not either of us knowing better.

**2. `refusal` field names.** Worse than you found: `{code, detail}` was in **three** places — `Tools.Resolve`, the engine's type, and the mock. All now `{reason, message, checkedBy}`, with `bounds` forwarded on `out_of_bounds` so the UI states the range without keeping a copy.

The root cause is worth naming: the parser **cast** the tool's object to a typed shape, and a cast asserts rather than checks, so `{code, detail}` type-checked cleanly as `{reason, message}`. Replaced with a field-by-field reader.

**3. The `audit` block.** Taking your `Describe()` suggestion — and it exposed a bigger gap than the denial branch. `resolve.ts` **never forwarded `audit` at all.** The dispatcher returned it and the engine dropped it, so every response claimed §8 compliance while carrying no attribution, which is precisely what §8's opening line forbids. Now forwarded on every outcome including dry-runs, with `requestedBy` threaded from the request and recorded next to `actor` — a missing actor is `null`, never the caller-supplied label.

Your suggested shape went in essentially verbatim.

## Verified as a real unprivileged principal over HTTP

```
HTTP 200
outcome : refused
refusal : {"reason":"not_authorized",
           "message":"'SetPoolSize' modifies a live production and requires PG_Resolve:USE",
           "checkedBy":"iris"}
audit   : {"auditId":"pg-audit-15","actor":"pg_http_readonly",
           "role":"%DB_%DEFAULT,Guardian_Read","tool":"SetPoolSize","disposition":"denied"}
pool    : 1 before and after — for a dry-run AND an apply
```

`message` is the audit row's `denialReason`, not the raw `<%AICore>ToolAccessDenied` chain, per your note.

## New guard, and I measured its reach

`test/mvp2-contract-drift.test.ts`. The four MVP 2 contracts have no schema and no `.d.ts`, so their shape was guarded by nothing — which is why this reached review instead of CI. It greps the contract prose, which is weaker than schema validation and says so in its header: when `resolve-api.md` gains a schema, **replace** this file rather than extend it.

I reintroduced the drift to check it fails, and **only 1 of its 4 cases fired** — because `parseRefusal` normalises, so the engine's output always carries the three fields regardless of what the tool sent. That division is now documented in the test itself. A guard whose reach you have not measured is one you will over-trust.

231 tests pass (was 221), typecheck clean, all 5 CI checks green.

## Unrelated bug found while verifying this — #96, and it is your area

The loop would not drain even with the pool correctly at 4. Cause: **recompiling `Production.cls` silently reverts the deployment settings `ApplyDeploymentSettings()` applied at boot** — both `HTTPServer` and `HTTPPort` go back to the XData values, so `Cloud API` dials `webgateway-webinar` and every message retries.

Worth your attention because of how it presents: `queue_buildup` fires correctly, `set_pool_size` applies correctly, four jobs really do start, and nothing drains — it looks exactly like MVP 2 failing. The error names `webgateway-webinar`, which sends the reader to the compose network and #53's history rather than to a reverted setting. Any `compose up --build` triggers it.

Recovery is one existing call: `ApplyDeploymentSettings()` + `UpdateProduction(10)`. Filed with options; my preference is stop shipping the deployment-specific value **and** have `Status()` flag a mismatch, but it is your call.

I also corrected my own first report on #96 — I initially claimed only the port reverts, having read a pre-recompile snapshot. Both revert.

## After fixing that, the loop closes cleanly

```
queue_buildup at 228 -> gpt-4o-mini, 5 tool calls, confidence 0.9, recommended set_pool_size 4
apply -> 1->4, audit pg-audit-35 (SuperUser, requestedBy presenter@laptop)
queue 562 -> ... -> 0    (20 samples, monotonic)
findings: (all clear)
Reset() -> pool 1, server 127.0.0.1, port 52773
audit rows: 35
```

Ready for re-review whenever suits.

Ari-Glikman:
## Two more bugs, both found by testing the endpoints the way a *browser* reaches them

Neither is in your review — I went looking at `resolve-api.md` §13 while #92 was blocked, and both turned out to be live defects rather than open questions.

### 1. Any web page could trigger a production write (§13.2)

§13.2 flags this as "raised rather than resolved" and calls it "tolerable — not fine", partly because "the engine holds no authority of its own". **Under §13.1's service-account model that premise is false**: the engine authenticates to IRIS with a credential that holds the write role. Measured before changing anything:

```
POST /api/resolve   Origin: https://evil.example.com
-> HTTP 200  {"outcome":"previewed","after":{"poolSize":8}}
OPTIONS -> Access-Control-Allow-Origin: *, Allow-Methods: GET, POST, OPTIONS
```

The browser cannot write, the engine can, and `*` lets any open page ask it to. POST now requires an allow-listed `Origin`, and the **preflight agrees with the POST branch** — advertising POST and then rejecting it would surface as a bare 403 with no indication which check refused.

Taking §13.2's middle option, not because it is best but because it is the one that fits inside this service. Localhost binding breaks the compose network; the pass-through credential is the right answer and is a *decision* (§13.1), not a patch. And stated in the code rather than implied: **CORS is enforced by the browser, so this stops a malicious page and nothing else.** curl reaches the endpoint unchanged. A request with no `Origin` is allowed deliberately — its absence means not-a-browser, and anything that can omit the header can forge it. It closes the drive-by, not the network path.

### 2. The dashboard could not reach a single MVP 2 endpoint

This one is worse, and it would have surfaced on stage. nginx proxied only `/api/healthscan/`, so everything MVP 2 added fell through to `location /`:

```
GET  /api/healthscan/findings  -> 200 JSON
GET  /api/earlywarning         -> 200 <!doctype html>
POST /api/resolve              -> 405 from nginx, never reaching the engine
```

The GETs returning **200 with an HTML body** is what hid it — every status-code check I had run looked healthy. The POST is the real damage: `location /` is a static file handler, so nginx refuses POST before the request goes anywhere. **The approve button could not have worked in the containerised dashboard at all**, and the failure would have appeared as nginx's own 405 page with nothing whatsoever in the engine log. `vite.config.ts` had the identical gap, so Dev C's dev server was equally blind.

Both now proxy `/api` as one prefix. The engine owns everything under it and 404s what it does not have — verified: `/api/nope` returns the engine's JSON 404, not the SPA. A per-endpoint list would be a second copy of its routing table, which is precisely how this happened: both blocks were written when `/api/healthscan/*` was the only family, and neither was revisited when MVP 2 added three more. Same family as #84, so I noted it there.

Also raised nginx's `proxy_read_timeout` from 10s to 35s on this path. `/api/investigate` measured **8.5s** through the proxy, so 10s would have truncated the slower half of real investigations and returned 504 — reading as "the AI is broken" rather than "the proxy gave up". 35s leaves the engine's own 30s timeout as the one that fires, so the component that knows what failed is the one that reports it.

### Verified through the dashboard's own path, port 5173

```
GET  /api/healthscan/findings -> JSON
GET  /api/earlywarning        -> JSON
POST /api/investigate         -> 200 in 8.5s, gpt-4o-mini, 5 tool calls, recommended size 4
POST /api/resolve apply       -> applied 1->4, audit pg-audit-44, requestedBy echoed
queue 123 -> 93 -> 63 -> 43 -> 13 -> 2, findings all clear, Reset() -> pool 1
SPA routes intact: /, /?mode=live, /some/route all 200
```

236 tests pass (was 231), engine and dashboard typecheck clean, all 5 CI checks green.

`apps/dashboard/**` is Dev C's area and I am covering it, so flagging rather than assuming: the nginx and Vite changes are one-line prefix widenings plus a timeout, with the reasoning in comments next to each.

Still just awaiting your re-review on the three findings from earlier — nothing else outstanding from my side.

kskubach (approved):
Approving. All three findings are properly fixed, and the two you found yourself are better catches than mine.

Scope of this review: I read the `iris/**` changes closely — that is my area — and I am relying on your transcripts and the 236 tests for the engine and dashboard rather than re-running them.

## The three findings

**`403` → `200`** with the full §5.1 shape: `before`/`after`/`reversal`/`failure` explicitly null, `confirmation: not_applicable`, every other field present. `Describe()` for the audit block, so `audit.actor` names the refused identity — which was the entire point of §8 auditing denials. And `denialReason` rather than `governed.error` for `refusal.message`, which was the smaller of my notes and the one I would have let slide.

The comment explaining why 403 was wrong is more valuable than the fix. Keeping "the reasoning is fine and the conclusion breaks the demo" in the file is the right call.

## Your two

**The confused deputy (§13.2) is the more important finding in this PR.** You did the thing that matters: you re-examined the contract's stated *premise* — "the engine holds no authority of its own" — and found it false under §13.1's service-account model. A note that says "tolerable, not fine" is exactly where a live defect hides, because it reads as already-considered. That is the same pattern as #93's `%`-prefix paragraph and my own `AgentDispatcher` finding: on this codebase the confident prose is what to check first.

Two things I want to endorse specifically:

- **The preflight agreeing with the POST branch.** Advertising `POST` and then refusing it produces a bare 403 with nothing indicating which check refused — and that is a 40-minute debug on stage.
- **Stating in code that CORS stops a malicious page and nothing else.** `curl` reaches the endpoint unchanged, and a request with no `Origin` is allowed deliberately. A security control documented as narrower than it looks is worth more than one that quietly over-promises. The alternative — implying the network path is closed — is how someone later skips the real fix.

**The nginx/Vite gap would have failed on stage, and the way it hid is the lesson.** `GET /api/earlywarning -> 200 <!doctype html>` is the worst possible failure shape: every status-code check passes. Proxying `/api` as one prefix rather than a per-endpoint list is right for exactly the reason you give — a per-endpoint list is a second copy of the engine's routing table, which is how this happened in the first place. Filing it under #84 is the correct home.

Raising `proxy_read_timeout` so the engine's own 30s timeout is the one that fires is the right instinct: the component that knows what failed should be the one that reports it.

## One observation, deliberately not blocking

Your `Refuse()` change is fine, and I checked whether it crossed a contract boundary before saying so. It does not: `mcp-tools.md` §4 specifies the *mechanism* for a failed call — `%ToolError`, which is `$$$ThrowStatus(sc)` — and never specifies a `refusal` object at all. So there were no ratified field names to drift from, and aligning with `resolve-api.md` §5 fixes a real `undefined` in Dev C's banner. Adding `bounds` from `#MINSIZE`/`#MAXSIZE` rather than literals is the #84 lesson applied correctly.

What it exposes is a genuine contract wrinkle, and it predates this PR. §4 classifies an **out-of-bounds argument** as a "Call failed — thrown `%Status`, surfaced as a tool error", and says "do not invent an error envelope … a tool that catches its own errors and returns a cheerful payload defeats both". `Tools.Resolve` has returned a structured `outcome: "refused"` payload for `out_of_bounds` since #90.

**I think the code is right and §4's classification is what should change**, for two reasons:

1. The endpoint has to produce `outcome: "refused"` + `refusal.reason: "out_of_bounds"` per `resolve-api.md` §5. From a thrown `%Status` the engine would have to parse error *text* to get there, and text-parsing a refusal into a contract field is fragile in a way a structured return is not.
2. It is load-bearing for the audit trail. `%LogExecution` receives the tool's **return value**; a throw records the status and no result payload. So a returned refusal is why `Audit.Entry.Result` can show *what was refused and why* rather than just that something failed. A denial and a bounds refusal are already distinguishable by `Disposition`.

Worth noting my audit code is indifferent to the field names — it stores `result_json` verbatim — so nothing on my side changes either way.

Added to #95 as a fourth item rather than raised here; it is a contract correction, not a change to this PR.

Nothing else outstanding from me.

PR #93, opened 2026-08-18
contracts: correct the role names — no `%` prefix

Ari-Glikman:
A **contract change PR**, per root `CLAUDE.md` §4: its own PR, a `CHANGELOG.md` entry, reviewed by the other developer.

## What changed

Values only, in `mcp-tools.md` and `resolve-api.md` — 14 occurrences:

| Was | Is |
|---|---|
| `%Guardian_Read` | `Guardian_Read` |
| `%Guardian_Resolve` | `Guardian_Resolve` |

No field added, removed or retyped. No schema and no sample affected — neither contract has either yet. `validate.mjs` unchanged and passing (15 accept / 19 reject / 7 capture claims). **Resources are untouched**: `PG_Read` / `PG_Resolve` were always correct.

## Why — the ratified names were unimplementable

```
ERROR #887: Invalid role name '%Guardian_Read'
```

`%` marks a role as **system-supplied**, so a custom role may not use it.

## The part worth reviewing: the reasoning was backwards, not just the values

`mcp-tools.md` §5 argued that roles carry `%` *because* it "is the IRIS convention for a system-supplied role", and concluded:

> Two different naming rules, applied correctly, rather than one applied uniformly and wrongly.

The premise is **true** and it supports the **opposite** conclusion — being the marker for system-supplied roles is exactly what makes it unavailable to ours. It is one rule applied wrongly.

I rewrote the paragraph rather than swapping its values, because a corrected value sitting under intact bad reasoning invites the next person to change it back, with a rationale already written for them.

**This was found by trying to create the roles, not by review.** The passage reads as a considered decision, complete with an explanation for the apparent `PG_*` / `%Guardian_*` inconsistency — which is why it survived ratification by three people. A confident explanation is harder to doubt than a bare value.

## Verified against the running instance

Not inferred from the error text:

```
RES   PG_Read           RES   PG_Resolve
ROLE  Guardian_Read     ROLE  Guardian_Resolve
```

## What I deliberately did NOT change

- **The existing `2026-08-18` CHANGELOG entry keeps the old values in its prose**, with a note pointing at the correction. A changelog records what a contract *said*; silently updating a past entry would erase the fact that the ratified names were unimplementable. The resource-vs-role distinction that entry draws is unaffected and still correct.
- **`resolve-api.md` §9.3's instruction to Dev C** — render `audit.role` as an opaque string and never compare it to a literal — **stays in force**. It is what kept this from being a dashboard change, and that argument holds even now the values are settled.
- **`iris/labdemo/Tools/AuthPolicy.cls`** already records the divergence in a class comment (it hit `#887` and said so, with "the contracts need correcting"). Updating that comment belongs with the code, not in a contracts PR — it is in #92's area and I will follow up there once this lands.

@kskubach — you are the required second reviewer.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,48 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-18 — role names corrected: no `%` prefix (Dev B)
+
+**Two files, values only, no shape change.** `mcp-tools.md` and `resolve-api.md`:
+`%Guardian_Read` → `Guardian_Read`, `%Guardian_Resolve` → `Guardian_Resolve`. 14 occurrences. No
+field added, removed or retyped; no schema and no sample affected (these two contracts still have
+neither). `validate.mjs` unchanged and passing.
+
+**The ratified names could not be created.** IRIS refuses them:
+
+```
+ERROR #887: Invalid role name '%Guardian_Read'
+```
+
+`%` marks a role as **system-supplied**, so a custom role may not use it. Resources follow the same
+rule, which is why `PG_Read` / `PG_Resolve` were always fine.
+
+**The reasoning in `mcp-tools.md` §5 was backwards, and that is the part worth reading.** It argued
+that roles carry `%` *because* it is the convention for a system-supplied role — a true premise
+supporting the opposite conclusion — and presented the `PG_*` / `%Guardian_*` split as "two
+different naming rules, applied correctly, rather than one applied uniformly and wrongly". It is in
+fact one rule applied wrongly. The paragraph has been rewritten rather than just having its values
+swapped, because a corrected value under intact bad reasoning invites the next person to "fix" it
+back.
+
+**Found by trying to create the roles, not by review.** The passage reads as a considered decision,
+complete with a rationale for the apparent inconsistency, which is exactly why it survived
+ratification by three people. A confident explanation is harder to doubt than a bare value —
+`AuthPolicy.cls` recorded the divergence in a class comment when it hit the error, and this entry
+closes it.
+
+**Verified against the running instance** rather than assumed from the error message:
+
+```
+RES   PG_Read           RES   PG_Resolve
+ROLE  Guardian_Read     ROLE  Guardian_Resolve
+```
+
+`resolve-api.md` §9.3's instruction to Dev C — render `audit.role` as an **opaque string** and never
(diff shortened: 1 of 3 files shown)


kskubach (approved):
Approving. The values are right, and I can confirm it independently rather than just from the error text — I created these roles for real on `pg-iris` while wiring up the governance setup:

```
resource PG_Read: yes
resource PG_Resolve: yes
role Guardian_Read: %DB_%DEFAULT:RW,PG_Read:U
role Guardian_Resolve: %DB_%DEFAULT:RW,PG_Read:U,PG_Resolve:U
```

`Guardian_*` unprefixed is what IRIS accepts; `PG_*` resources were always fine. Diff is values-only across the two contracts plus the CHANGELOG — no field added, removed or retyped, no schema or sample touched.

Two things I want to endorse specifically, because they are the parts that will still be paying off in a month:

**Rewriting the reasoning instead of swapping the values.** The old paragraph argued that roles carry `%` *because* it marks a system-supplied role — a true premise supporting the exact opposite conclusion. A corrected value under intact bad reasoning is an invitation to change it back, with the rationale pre-written. Right call.

**Leaving the 2026-08-18 entry's prose alone with a pointer.** A changelog records what a contract *said*. Rewriting history there would erase the fact that three people ratified names that could not be created, which is the most useful thing this entry teaches.

## One gap this does not close — a follow-up, not a blocker

The role *definitions* in §5 are still incomplete: `PG_Read:U` alone produces a role that cannot be used. A real user holding only `Guardian_Read` logs in fine and then dies with

```
```

before any policy is consulted — it cannot read the routine. Both roles need a database privilege (`%DB_%DEFAULT:RW` on this image), and `RW` rather than `R`, because the denial audit row is written *as the refused principal* — which is what makes it attributable. Neither contract mentions this, so anyone implementing the roles from §5 alone gets roles that do not work.

Deliberately not blocking on it: this PR fixes a value that was *wrong*, and my finding adds one that is *missing*. Holding a correct fix hostage to an adjacent gap would be the wrong trade. I will raise it with the DB-privilege detail and the observed-denial transcript.

Also worth knowing, since it touches §5's neighbourhood: I have measured that the runtime does **not** audit authorization denials — `ExecuteTool` checks, then executes, then audits, so a denial throws before the audit hook (0 rows, verified with a deny-all policy). §5.5 and `resolve-api.md` §8 both promise otherwise. Separate contract correction; I will own it.

PR #94, opened 2026-08-19
feat(iris): the audit trail AI Hub does not ship, and the seam that applies it

kskubach:
Dev A's MVP 2 tasks 4 and 5 (doc §5.2): audit for tool calls, RBAC roles and resources. Everything here was **measured on build 126**, and three measurements changed the design.

## AI Hub ships no audit store

`%AI.Policy.ConsoleAudit` writes an ANSI-coloured box to the current device and returns. There is no `%AI.Audit.*` persistent class in the image — searched `%Dictionary.CompiledClass` for `%AI.*` matching AUDIT or LOG; the only hits are the policy classes themselves.

So `mcp-tools.md` §5.5's "every one of the six tools is audited on every call" was true of the **hook** and false of the **record** — and MVP 2 §3's final demo step is showing an audit entry. `Audit/Entry.cls` is the store, `Tools/AuditPolicy.cls` fills it.

This also closes `mcp-tools.md` §5.6(a) and (b), which were the two open unknowns. (b)'s answer: **no** policy is registered by default — a fresh `%AI.ToolMgr` reports `AuthPolicy: <none>`, `AuditPolicy: <none>`, `DiscoveryPolicy: <none>`.

### The real `%LogExecution` payload

```
call     {"id":"call_id","name":"GetPoolSize","arguments":"{\"host\":\"Cloud API\"}"}
metadata {}
result   {"tool_name":"GetPoolSize","result_json":"{...}","display_text":null}
duration 0
status   OK
```

| Assumption | Reality |
|---|---|
| `call.id` identifies the call | it is the **literal string `"call_id"`** every time — unusable as `auditId` |
| `call.arguments` is an object | it is a **JSON string** |
| `metadata` carries identity | **empty** — `actor` must come from `$username`, which is also the only unforgeable source |
| `duration` is meaningful | **0** for a call `ExecuteTool` timed at `0.0071 s` |

## A refused call is not audited

`ExecuteTool` checks authorization → executes → audits. So an authorization denial throws `<%AICore>ToolAccessDenied` at step one and the hook never runs. Measured with a deny-all policy: **0 rows**.

| Case | Audited by the runtime? |
|---|---|
| successful read | yes |
| dry-run of the write tool | yes |
| **tool-level** refusal (the `2..8` bounds guard) | yes — the tool ran |
| **authorization** denial | **no** |

The security-relevant half was the missing one. `AuthPolicy` now writes its own `denied` row — it is the only code that ever learns a denial happened, which is why `Disposition` has two values and two writers.

## Governance is per-ToolMgr and in memory

`SetAuthPolicy` attaches the policy to the Rust manager identified by *that object's* `%token`. There is no persisted setting, so a first-boot script cannot switch governance on once, and **every caller that builds its own `%AI.ToolMgr` — or its own `%AI.Agent`, which builds one — gets an ungoverned manager**: no check, no audit, `SetPoolSize` live against the production.

`%AI.Agent.UseToolSet` delegates straight to `..ToolManager.RegisterToolSet`, so the natural thing to write is the unsafe thing, and it works perfectly. `Tools/Governance.cls` is one factory that cannot return an ungoverned manager.

**This is already load-bearing:** it is the fix for the finding I raised on #92, where `AgentDispatcher.Resolve()` builds a bare manager. I measured that path succeeding for a principal without `PG_Resolve`, with 0 audit rows.

## None of it was on the compose stack

Measured on `pg-iris`, up 2 days:

| | Present |
|---|---|
| `PG_Read` / `PG_Resolve` | no |
| `Guardian_Read` / `Guardian_Resolve` | no |
| `AI.Secrets`, `PGSecrets` wallet, `AI.LLM.pgdetective` | no |
| `Tools.Read` / `Tools.Resolve` / `Tools.AuthPolicy` compiled | **no** |

It had all been done by hand on the hand-built `iris-webinar` instance. That is exactly the failure `FirstBoot` exists to prevent (#70, #72), so `Setup/AIHub.cls` creates the resources and roles idempotently and `FirstBoot` calls it.

It **deliberately does not create the LLM credential**: that needs a key, a key does not belong in a repository, and this one must be rotated anyway since it reached me through a chat transcript. It reports the config missing and prints the sequence instead.

### The roles need a database privilege no contract mentions

A user holding only `PG_Read:U` logs in fine and then dies with

```
```

*before any policy is consulted* — it cannot read the routine. Both roles carry `%DB_%DEFAULT:RW`, and `RW` rather than `R` because the denial row is written **as the refused principal**, which is what makes it attributable.

**Honest limitation, recorded in the class:** every database on this image shares `%DB_%DEFAULT`, so that grant is broad. The least-privilege story is real at the **tool** boundary — `PG_Resolve` genuinely gates `set_pool_size` — and is **not** database isolation. A demo should not imply otherwise.

## Observed denial — the acceptance criterion

`resolve-api.md` §9.4 requires an *observed* denial, not a passing allow. `StubPolicy` cannot provide one: `$SYSTEM.Security.Check` returns 1 for every resource when the caller holds `%All`, and every session that can compile the code holds `%All`. So `test/GovernanceProof.cls` creates a real throwaway user with a generated password (never stored), logs in as it, and measures:

```
1. logged in as pg_proof_readonly, roles = Guardian_Read
2. GetPoolSize : executed  <- expected executed
3. SetPoolSize : denied    <- expected denied
   error : 'SetPoolSize' modifies a live production and requires PG_Resolve:USE
4. audit row: {"actor":"pg_proof_readonly","role":"Guardian_Read","tool":"SetPoolSize",
               "disposition":"denied",...}

```

## Verification

- `labdemo`, `setup`, `test` all compile clean
- `AIHub.Run()` idempotent across two runs; `FirstBoot.GovernanceSetup()` exercised
- six tools discovered (the count is the test — a seventh public method on `Tools.Resolve` is a second way to mutate a production)
- **production untouched:** still Running, Cloud API pool size **1**. Only dry-runs were ever applied.
- probe user, probe classes and probe globals all removed afterwards

One bug in my own code found this way and fixed: the row counter returned a confident `0` for a table an unprivileged principal cannot read — #33 and #49's "unmeasurable read as zero", reproduced by the person writing the comments warning about it. It returns `""` now.

## Not in this PR

**Two contract corrections are owed**, and `contracts/` is read-only:

1. §5.5 and `resolve-api.md` §8 promise that `not_authorized` refusals are runtime-audited. They cannot be.
2. `resolve-api.md` §8's `auditId: "aihub-audit-44812"` implies an AI Hub audit store that does not exist. Handles are `pg-audit-<n>`, minted by us — inventing AI Hub provenance is the same defect as the mock inventing an id.

Also still open: **a live apply (1 → 4) has not been run** — that needs a built-up queue and belongs with rehearsal. And AI Detective cannot run until the LLM credential exists on this stack.

### DIFF
--- a/docs/mvp2-aihub-verified-api.md
+++ b/docs/mvp2-aihub-verified-api.md
@@ -283,15 +283,31 @@ throwaway user with a generated password, logs in as it, and attempts the write:
 PASS: read allowed, write refused, refusal audited.
 ```
 
-**The roles need a database privilege the contracts never mention.** A user holding only
+**A principal needs a database privilege the contracts never mention.** A user holding only
 `PG_Read:U` logged in fine and then died with `<PROTECT> ... Access Denied` on
-`Tools.Governance.1` — before any policy was consulted, because it could not read the routine. Both
-roles therefore carry `%DB_%DEFAULT:RW`, and `RW` rather than `R` because `AuthPolicy` writes the
-denial row *as the refused principal*, which is what makes it attributable.
+`Tools.Governance.1` — before any policy was consulted, because it could not read the routine.
 
-**An honest limitation:** on this image every database shares `%DB_%DEFAULT`, so that grant is
-broad. The least-privilege story is real at the **tool** boundary — `PG_Resolve` genuinely gates
-`set_pool_size` — and is not a database-isolation story. A demo should not imply otherwise.
+**The `Guardian_*` roles deliberately do not grant it.** I first folded `%DB_%DEFAULT:RW` into both
+roles; @Ari-Glikman argued in the #94 review that the role should stay minimal and mean one thing,
+and he is right. The deciding reason is what the grant would *say*: on this image every database
+shares the `%DB_%DEFAULT` resource, so folding it in gives a role whose whole purpose is "may look,
+may not act" the ability to write every database on the instance. A least-privilege boundary that
+hands out broad write to keep a fixture working is not one.
+
+Database access comes from the invocation path instead — the CSP application hosting
+`REST.AgentDispatcher` for the demo, and the throwaway user in `GovernanceProof` for the direct
+session proof (the shipped `%DB_%DEFAULT` role, granted to the *user*). `Setup.AIHub.Run()` prints
+the requirement rather than granting it, because a requirement nobody states is how that `<PROTECT>`
+gets misdiagnosed as a policy bug.
+
+**A guard that could not work, and the reason it is worth recording.** The obvious fix to "the proof
+died before reaching the policy" is to assert inside the unprivileged half that the tool call was
+reached. That is not implementable: with the principal short of database access, `Invoke` dies while
+paging in `Governance.1`, so the process is gone before the result variable is assigned and no line
+after it runs. Verified by stripping the role and re-running. The check therefore lives on the
+**privileged** side — `Prove()` reads the probe user's granted roles back and refuses to hand over a
+password if the database role is missing, since an inconclusive run whose output reads
+`<PROTECT> ... Access Denied` looks exactly like the boundary working.
 
 ## The MCP dev connection
 
--- a/docs/mvp2-aihub-verified-api.md
+++ b/docs/mvp2-aihub-verified-api.md
@@ -129,15 +129,169 @@ execution.
 
 - ~~No LLM provider is configured.~~ **CLOSED** -- see the provider section at the end of this
   file. A real OpenAI call and an agent tool call have both been made from this instance.
-- **What an audit record CONTAINS, and whether it is queryable.** Enforcement is confirmed (see
-  above), but the written record has not been observed, and Dev C needs to display one. This is
-  now the highest-value remaining unknown.
-- **Which policy implementation is active by default.** `%AI.Policy.ConsoleAuth` and
-  `%AI.Policy.ConsoleAudit` exist alongside the abstract bases, so something is wired out of the
-  box -- whether it permits everything until configured is unknown, and "enforced by the runtime"
-  is only a safety property if the configured policy actually denies.
+  **But not on the compose stack** -- see "None of it was on `pg-iris`" below.
+- ~~**What an audit record CONTAINS, and whether it is queryable.**~~ **CLOSED 2026-08-19** -- the
+  answer is that *nothing is written at all*. See "There is no audit store" below.
+- ~~**Which policy implementation is active by default.**~~ **CLOSED 2026-08-19** -- none is. A
+  fresh `%AI.ToolMgr` reports `AuthPolicy: <none>`, `AuditPolicy: <none>`,
+  `DiscoveryPolicy: <none>`. The default permits everything and records nothing.
 - **`@{wallet.*}` substitution** was broken on build 159 per the skill (only `env` and `config`
   registered). Unknown on 126. `%AI.Utils.WalletStore` exists, but existence is not function.
+
+## There is no audit store — the hook fires into nothing
+
+Measured 2026-08-19 on `pg-iris`. `%AI.Policy.Audit` is an abstract base whose one interesting
+method is
+
+```
+%LogExecution(call:%DynamicObject, metadata:%DynamicObject, result:%DynamicObject,
+              duration:%Integer, status:%Status) -> %Status
+```
+
+The only shipped implementation, `%AI.Policy.ConsoleAudit`, writes an ANSI-coloured box to the
+current device and returns. There is no `%AI.Audit.*` persistent class — searched
+`%Dictionary.CompiledClass` for `%AI.*` matching AUDIT or LOG and the only hits are the policy
+classes themselves.
+
+So `mcp-tools.md` §5.5's "every one of the six tools is audited on every call" was true of the
+**hook** and false of the **record**, and MVP 2 §3's final demo step is showing an audit entry.
+`iris/labdemo/Audit/Entry.cls` is the store; `iris/labdemo/Tools/AuditPolicy.cls` is the policy that
+fills it.
+
+### A real `%LogExecution` payload, captured from a live `GetPoolSize("Cloud API")`
+
+```
+call     {"id":"call_id","name":"GetPoolSize","arguments":"{\"host\":\"Cloud API\"}"}
+metadata {}
+result   {"tool_name":"GetPoolSize","result_json":"{...}","display_text":null}
+duration 0
+status   OK
+```
+
+Four things there are not what a reader would assume, and each one changed the implementation:
+
+| Assumption | Reality |
+|---|---|
+| `call.id` identifies the call | it is the **literal string `"call_id"`** on every call — unusable as `auditId` |
+| `call.arguments` is an object | it is a **JSON string**; iterating it yields nothing |
+| `metadata` carries caller identity | it is **empty** `{}` — `actor` has to come from `$username` |
+| `duration` is meaningful | **0** for a call `ExecuteTool` itself timed at `0.0071 s`; it is integer ms |
+
+Tool names are the **ObjectScript method names** — `GetPoolSize`, `SetPoolSize` — not the
+snake_case names in `mcp-tools.md`'s catalogue. Anything joining the two must map.
+
+## A refused call is NOT audited — the contracts overpromise
+
+**This is the most important finding in this file.** `contracts/mcp-tools.md` §5.5 and
+`resolve-api.md` §8 both state that refusals, `not_authorized` explicitly, are audited:
+
+> Every call produces exactly one attributable audit event: applies, refusals, and dry-runs alike.
+
+The runtime cannot deliver that. `ExecuteTool` checks authorization, **then** executes, **then**
+audits — so an authorization denial throws `<%AICore>ToolAccessDenied` at step one and the audit
+hook is never reached. Measured with a deny-all policy registered: `%LogExecution` not called, **0
+rows written**.
+
+The distinction is sharp, and only one half was broken:
+
+| Case | Audited by the runtime? |
+|---|---|
+| successful read | yes |
+| dry-run of the write tool | yes |
+| **tool-level** refusal (our own bounds guard returns `outcome: "refused"`) | yes — the tool ran |
+| unknown host | yes |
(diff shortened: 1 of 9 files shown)


Ari-Glikman (approved):
Reviewed as Dev B, the consumer. I ran your classes against the live `pg-iris` rather than reading them, because the finding on #92 is a claim about runtime behaviour and my own class comment made a claim of that kind that turned out to be false.

## Your finding reproduces, and it is worse than a missing feature

Confirmed on my branch, exactly as you described:

```
AuthPolicy : <none>
AuditPolicy: <none>
register: OK
ExecuteTool: SUCCEEDED -> {"outcome":"previewed","before":{"poolSize":1},"after":{"poolSize":4},...}
```

`git grep SetAuthPolicy -- iris/ services/` on `devB/mvp2-investigate-resolve`: **no hits.** So `AuthPolicy.cls` existed, was correct, and was never consulted by anything.

The part I want to own explicitly: **`AuthPolicy.cls`'s own header documents this trap** — "with no policy registered, the runtime faithfully asks a policy that permits everything" — and I wrote the dispatcher that does precisely that, four days later, having written that sentence. The class comment on `AgentDispatcher` then asserted the opposite as its central safety property. A reviewer reading that comment would have had no reason to check. That is worse than an unimplemented guard, because it reads as an implemented one.

## What I verified about your fix, rather than accepting it

**1. The factory attaches both policies.**

```
factory: OK
AuthPolicy : ProductionGuardian.LabDemo.Tools.AuthPolicy
AuditPolicy: ProductionGuardian.LabDemo.Tools.AuditPolicy
```

**2. A read call is audited with a real handle.** `Governance.Invoke("GetPoolSize", ...)` → `outcome: executed, auditId: pg-audit-1`, and the row count went `0 → 1`. That closes the "read *and* write" half of MVP 2 §2.2 that my dispatcher was missing entirely.

**3. The boundary actually refuses.** This is the one that mattered, and I ran `GovernanceProof` end to end as a real principal:

```
1. logged in as pg_proof_readonly, roles = %DB_%DEFAULT,Guardian_Read
2. GetPoolSize : executed  <- expected executed
3. SetPoolSize : denied    <- expected denied
   error   : ERROR <%AICore>ToolAccessDenied: 'SetPoolSize' modifies a live production
             and requires PG_Resolve:USE
   auditId : pg-audit-6
4. audit row: {"auditId":"pg-audit-6","actor":"pg_proof_readonly",
               "role":"%DB_%DEFAULT,Guardian_Read","tool":"SetPoolSize",
               "disposition":"denied","denialReason":"..."}
```

An **observed denial**, with the actor recorded, which is the acceptance criterion `resolve-api.md` §9.4 asks for. Your point that `AuthPolicy` must write its own denial row is also confirmed — the runtime throws before `%LogExecution`, and that row exists only because your policy wrote it.

## One reproducible gap in `GovernanceProof`, and I had to work around it to get the result above

`AsReadOnlyUser()` failed before reaching the policy:

```
1. logged in as pg_proof_readonly, roles = Guardian_Read
<PROTECT>^ProductionGuardian.Test.GovernanceProof.1 *^|^^/iris-shared/durable/mgr/labdemo/|...
Access Denied
```

`Guardian_Read` grants `PG_Read:U` and nothing else, while LABDEMO's database resource is `%DB_%DEFAULT`. So the probe user cannot read the routine it is trying to run, and the proof dies at the **database** layer with the tool policy never consulted. I got the PASS above only after adding `%DB_%DEFAULT` to the user by hand.

Worth fixing because of *how* it fails rather than that it does: `<PROTECT>` plus `Access Denied` looks like the boundary working. A reader could take that output as the proof succeeding, when in fact nothing about `PG_Resolve` was exercised — the same shape as the defect this whole PR is about, one layer down. Two options, your call: have `CreateProbeUser` add `%DB_%DEFAULT` to the user (not to `Guardian_Read` — the role should stay minimal and mean one thing), or assert that the tool call was reached and fail loudly if it was not.

## Consuming it in #92

Taking your seam as written — `ToolManager(1, .sc)` in `Resolve()` and `GovernAgent(agent, 0)` in place of `UseToolSet` after `%Init()`. `Invoke()` also returns the `auditId` that `resolve-api.md` §8 requires and my dispatcher currently omits entirely, so I am using that rather than `ExecuteTool` directly. I will rebase #92 on this once it lands.

On `docs/mvp2-aihub-verified-api.md`: take the conflict on your side, please. Your version has measurements mine does not, and my contribution to that file was working notes.

Approving. The finding is correct, the fix is structural rather than a patch, and the "one factory that cannot return an ungoverned manager" shape is the right answer to an API where the natural thing to write is the unsafe thing.

kskubach:
Both addressed in 4913880. Thank you for running them rather than reading them — the second finding could not have come out of a read.

## I took your side on the role, and for a stronger reason than yours

You are right, and the deciding argument is what the grant would *say*. Every database on this image shares `%DB_%DEFAULT`, so folding it into `Guardian_Read` gives a role whose entire purpose is *may look, may not act* the ability to **write every database on the instance**. A least-privilege boundary that hands out broad write to keep a fixture working is not one.

So:

```
role Guardian_Read: PG_Read:U
role Guardian_Resolve: PG_Read:U,PG_Resolve:U
```

Database access now comes from the invocation path — the CSP application for your dispatcher, the throwaway user for the proof. `AIHub.Run()` prints the requirement rather than granting it, because a requirement nobody states is exactly how that `<PROTECT>` gets diagnosed as a policy bug.

## Your option 2 is not implementable, and I checked rather than assumed

I tried "assert the tool call was reached" first. It cannot work: with the principal short of database access, `Invoke` dies while paging in `Governance.1`, so **the process is gone before the result variable is assigned** and nothing after that line runs. Verified by stripping `%DB_%DEFAULT` from the user and re-running — output identical to yours, and my new guard never executed.

So the check moved to the **privileged** side, where it can actually run. `Prove()` now:

1. converges the resources and roles before creating the user — your run failed against roles predating this definition, which is a fixture problem rather than yours;
2. reads the probe user's granted roles back and **refuses to hand over a password** if the database role is missing.

The guard inside `AsReadOnlyUser` stays, because it does catch the failures that *return* — a tool error, a misconfigured resource, a policy refusing the read — and those would otherwise reach a meaningless `PASS`. Its comment now says plainly which case it cannot catch, so nobody trusts it for the `<PROTECT>` one.

`HasRole` matches list elements exactly rather than by substring, so `%DB_%DEFAULTX` cannot satisfy `%DB_%DEFAULT` — same trap as the quoted-label match in `ScrapeGauge`. Five cases, 0 failures.

## Re-verified after the change

```
1. logged in as pg_proof_readonly, roles = %DB_%DEFAULT,Guardian_Read
2. GetPoolSize : executed
3. SetPoolSize : denied -- requires PG_Resolve:USE
   auditId : pg-audit-4
4. audit row: {"actor":"pg_proof_readonly","role":"%DB_%DEFAULT,Guardian_Read",
               "disposition":"denied",...}
```

Same roles you had to assemble by hand, now assembled by the fixture. Production untouched at pool size 1; probe user, throwaway classes and audit rows all cleaned up.

## On your own note about `AgentDispatcher`

For what it is worth: I would not have found this by reading either. I only went looking because I needed to know where the policy got registered in order to write the setup, and the answer turned out to be nowhere. The comment asserting the guarantee is what makes it dangerous rather than merely missing — which is the same shape as #93's `%`-prefix paragraph, and worth remembering as a pattern: on this codebase the confident explanations are the ones to check first.

#95 tracks the three contract corrections, including that §5's role definitions are incomplete without the database privilege.

Ari-Glikman:
Post-merge note from verifying the merged tree on #92 — one small gap, and a correction to something I nearly reported as a bug.

## `Status()` omits `ReportTools()`

`Run()` calls it; `Status()` does not:

```objectscript
ClassMethod Status()
{
    do ..ReportSecurity()
    set n = ..AuditCount()
    write "  audit rows: ", ...
    do ..ReportProvider()      // <- no ..ReportTools()
}
```

So the output jumps from the audit count straight to `4. LLM provider`, with no `3.` block at all:

```
MVP 2 governance status
-----------------------
  resource PG_Read: yes
  resource PG_Resolve: yes
  role Guardian_Read: PG_Read:U
  role Guardian_Resolve: PG_Read:U,PG_Resolve:U
  audit rows: 13
4. LLM provider AI.LLM.pgdetective: configured, model = gpt-4o-mini
```

Worth fixing because of which command a person actually runs. `iris/CLAUDE.md` §7 points at `Status()` as *"what is and is not in place"*, and the tool count is the check that catches a helper left public — which for `Tools.Resolve` is a second way to mutate a live production. `Run()` is a boot-time call; `Status()` is the one someone types before a rehearsal. The missing `3.` is also silent rather than obviously absent, since a numbered list going `...audit rows... → 4.` reads as complete unless you are looking for the gap.

Run directly, it works and the count is right:

```
3. tools Read: GetHostStatus, GetPoolSize, GetProcessingTime, GetQueueDepth, GetRecentErrors
3. tools Resolve: SetPoolSize
3. tools discovered: 6 (expected 6)
```

## And a correction to my own near-miss

I first thought I had found a seventh tool, because I counted with `mgr.FindTools()` and got 7 entries. That was my error, not a regression: `FindTools()` does **global** discovery, not registered-toolset listing — its first entry is `%AI.Agent.Skill`, a framework class — and my loop was iterating a structure it does not return. Your `%Discover()`-per-class approach in `ReportTools` is the correct check and mine was not a check at all.

Mentioning it because it makes the case for calling `ReportTools()` from `Status()` stronger: the reasonable-looking alternative I reached for gives a wrong answer confidently, and the correct counter is the one that has to be easy to reach.

## The rest of the merged tree verifies

- Full `iris/` tree compiles clean after the merge (`compile: OK`, all `ProductionGuardian.*`)
- `Governance.ToolManager()` attaches both policies
- Audit rows accumulated correctly across the merge: 13 at last count, spanning executions, one authorization denial, and five agent read calls
- Provider resolves: `AI.LLM.pgdetective`, `gpt-4o-mini`

`#92` is rebased on this and mergeable. I took your version of `docs/mvp2-aihub-verified-api.md` whole, as agreed — checked first that it is a strict superset (440 lines vs my 270; all 12 of my headings among your 18; the `Ens_Util.Log` PHI finding, the 61,772-row count and the wallet setup all still present).

Issue #95, opened 2026-08-19
contracts: audit and RBAC promise three things build 126 does not do (found while implementing them)


kskubach:
Three things `contracts/mcp-tools.md` and `contracts/resolve-api.md` promise about audit and RBAC do not hold on build 126. All three were found by implementing them (#94), not by review, and `contracts/` is read-only — so this is the change request rather than a PR.

Raising it as an issue rather than opening the contract PR myself because a contract change needs the other developer's review anyway, and two of these interact with #92's area.

## 1. `not_authorized` refusals are not audited — and cannot be

`mcp-tools.md` §5.5:

> `%LogExecution` receives `status` and `duration`, so **a failed or refused call is audited too**. That is the difference between "we log actions" and "we log attempts".

`resolve-api.md` §8:

> **Every call produces exactly one attributable audit event: applies, refusals, and dry-runs alike.**
> **Refusals are audited.** Including `not_authorized`.

The runtime cannot deliver this. `%AI.ToolMgr.ExecuteTool` checks authorization → executes → audits, so an authorization denial throws `<%AICore>ToolAccessDenied` at step one and the audit hook is never reached. Measured with a deny-all policy registered: `%LogExecution` not called, **0 rows written**.

The distinction is sharp, and only one half was broken:

| Case | Audited by the runtime? |
|---|---|
| successful read | yes |
| dry-run of the write tool | yes |
| **tool-level** refusal (our own `2..8` bounds guard returns `outcome: "refused"`) | yes — the tool ran |
| unknown host | yes |
| **authorization** denial (`%CanExecute` refuses) | **no** |

The security-relevant event was the one going unrecorded. An audit log that records only what succeeded cannot answer *did anything try to write to the production*.

**The guarantee is now real, but it is ours, not the runtime's.** `Tools.AuthPolicy` writes its own `denied` row before returning the refusal (#94) — it is the only code that ever learns a denial happened. The contracts should say that: the *runtime* audits executions; *our authorization policy* audits denials. The current wording attributes both to the runtime, which is what made this invisible.

Note that §5.5's reasoning is sound and worth keeping — "we log attempts" is the right goal and `status`/`duration` in the signature *is* suggestive. It is just not sufficient, because the denial path never reaches the signature. Same shape as #93's `%`-prefix paragraph: a correct premise carried to a conclusion it does not support.

## 2. `auditId` implies an AI Hub audit store that does not exist

`resolve-api.md` §8:

```json
[...]

PR #97, opened 2026-08-19
docs(iris): first boot never compiles iris/test/, so the acceptance proof is absent

kskubach:
Verifying #94's last unverified claim, which turned up a real gap.

#94 said the RBAC boundary is armed from a clean boot. That was true of the *roles* and untested as a *boot* — on my container the resources and roles had only ever taken the "exists" path, because I created them by hand while writing the setup. So I deleted `PG_Read`, `PG_Resolve`, `Guardian_Read` and `Guardian_Resolve`, restarted the container, and read the boot log.

## The clean-boot path works

```
9. MVP 2 governance:
1. resource PG_Read: created
1. resource PG_Resolve: created
2. role Guardian_Read: created with PG_Read:U
2. role Guardian_Resolve: created with PG_Read:U,PG_Resolve:U
3. tools discovered: 6 (expected 6)
4. LLM provider AI.LLM.pgdetective: NOT CONFIGURED
FirstBoot: OK
```

All four **created**, not "exists". Three further things this settles that were previously assumed:

- **`Audit/` compiles through the normal boot**, SQL table included. It is a new subdirectory, and nothing had confirmed `FirstBoot`'s recursive `LoadDir` would pick one up.
- **Compile order does not bite.** `Governance` and `AuditPolicy` compiled *before* `Audit.Entry` despite referencing it, which is fine for runtime `##class()` resolution — worth knowing given the ToolSet compile-order trap documented in `docs/mvp2-aihub-verified-api.md`.
- **Nothing writes audit rows during boot.** I accounted for every row present by timestamp; all three were mine.

Data survived (41,305 → 41,378 patient rows — the generator kept running), the production came back `Running` at pool size 1, and the full observed denial passes on the freshly booted stack:

```
1. logged in as pg_proof_readonly, roles = %DB_%DEFAULT,Guardian_Read
2. GetPoolSize : executed
3. SetPoolSize : denied -- requires PG_Resolve:USE
4. audit row: {"actor":"pg_proof_readonly","disposition":"denied",...}
```

## The gap

`docker/iris-firstboot.sh` loads `/pg-src/setup/` and `/pg-src/labdemo/` and **nothing else**. So on a fresh instance `Test.GovernanceProof` and `Test.StubPolicy` do not exist, and the acceptance proof `resolve-api.md` §9.4 asks for fails with `<CLASS DOES NOT EXIST>`.

They were present on my container — as leftovers from an earlier manual compile. **Checking the namespace would have confirmed the wrong thing;** the boot log is what shows they were never loaded. That is the same shape as the defect #94 was about: a state that looks deployed because somebody deployed it by hand once.

## Deliberately not fixed by adding `test/` to the boot

`GovernanceProof` **creates and deletes an IRIS user**. Auto-compiling that onto every container start widens what the deploy ships for a fixture's benefit, and the boot path should stay the smallest thing that produces a working demo.

So this documents the one command that makes it runnable, next to the acceptance table where someone verifying the boundary will actually be looking. Open to being overruled if you'd rather it shipped — it is a one-line change to the script either way.

Doc-only, no code touched.

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -144,13 +144,34 @@ returns PHI, it lands in `Audit.Entry.Result` in plain text.
 |---|---|
 | Read tools return real evidence | met — six tools discovered, live values |
 | `set_pool_size` changes the live pool and is reversible | **dry-run verified only**; `before`/`after`/`reversal` correct, an apply has not been run against a live queue |
-| An unauthorized role is refused | met — observed, `iris/test/GovernanceProof.cls` |
+| An unauthorized role is refused | met — observed, `iris/test/GovernanceProof.cls`. **Compile `test/` first — first boot does not** (see below) |
 | Every call appears in the audit log | met for executions and for authorization denials |
-| RBAC roles and resources exist from a clean boot | met — `Setup.AIHub`, called by `FirstBoot` |
+| RBAC roles and resources exist from a clean boot | met, and **verified from a real boot** rather than inferred: the four security objects were deleted and `docker compose restart iris` recreated them through step 9. `Audit/` compiled on the way through, including its SQL table, so the recursive load picks up a new subdirectory |
 | A principal can actually *use* the roles | needs `%DB_%DEFAULT` **as well**, from the invocation path — the `Guardian_*` roles stay minimal and grant only `PG_*`. Without it you get `<PROTECT>` before any policy is consulted |
 | LLM credential in the vault | **not met on the compose stack** — needs a key, and the current one must be rotated |
 
 ```objectscript
 do ##class(ProductionGuardian.Setup.AIHub).Status()          // what is and is not in place
 do ##class(ProductionGuardian.LabDemo.Audit.Entry).Purge()   // reset the log between rehearsals
+```
+
+### `iris/test/` is not deployed by first boot — compile it by hand
+
+`docker/iris-firstboot.sh` loads `/pg-src/setup/` and `/pg-src/labdemo/` and **nothing else**, so on a
+fresh instance `Test.GovernanceProof` and `Test.StubPolicy` do not exist and the acceptance proof
+fails with `<CLASS DOES NOT EXIST>`. Found by deleting the security objects, restarting the container,
+and looking for the classes in the boot log rather than in the namespace — they were present, but only
+as leftovers from an earlier manual compile, which is exactly the kind of thing that reads as "it
+works" until someone tries it on a second machine.
+
+Left out of the boot deliberately rather than fixed: `GovernanceProof` **creates and deletes an IRIS
+user**, and auto-compiling that onto every container start widens what the deploy ships for the
+benefit of a fixture. Compile it when you want to run the proof:
+
+```objectscript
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. I verified the gap and the documented recovery rather than reading them, because the whole finding is about a state that *looks* deployed.

## The gap is real

`docker/iris-firstboot.sh` lines 18–19 and 101 load `$SETUP_SRC` and `$SRC` — `/pg-src/setup/` and `/pg-src/labdemo/` — and nothing else. No third `LoadDir`.

And on my container the classes were present too:

```
GovernanceProof compiled: 1
StubPolicy compiled: 1
```

— for exactly the reason you give. I compiled them by hand this morning while running your proof on #94. So on both our machines the acceptance fixture existed and on neither was it deployed, which is the second time today this codebase has produced a state that reads as working because someone did it once interactively. Your point that "checking the namespace would have confirmed the wrong thing" is the transferable bit.

## The documented command works, tested from actually-absent

Not trusting that it would, since the path in the doc (`/pg-src/test/`) is not one the boot script ever touches — so nothing had ever confirmed the mount includes it:

```
$ docker exec pg-iris ls -1 /pg-src/
CLAUDE.md  labdemo  setup  test          <- test/ IS mounted

$system.OBJ.Delete("ProductionGuardian.Test.GovernanceProof","-d")
after delete, GovernanceProof: 0

do $system.OBJ.LoadDir("/pg-src/test/", "ck", .errors, 1)
GovernanceProof now: 1
StubPolicy now: 1
```

So the snippet is runnable as written, from a state where the classes genuinely do not exist. Worth having checked: the doc would have been correct-looking and wrong if `iris/` were mounted per-subdirectory.

## Agreed on not adding `test/` to the boot

`GovernanceProof` creates and deletes an IRIS user, and the boot path should stay the smallest thing that produces a working demo. A container that ships a fixture capable of creating principals is a wider deploy for no demo benefit — and the failure mode of leaving it out is loud (`<CLASS DOES NOT EXIST>`, with the fix now written next to the acceptance table), where the failure mode of shipping it is silent.

One thing I'd add if you ever revisit it: the same argument applies to `%DB_%DEFAULT:RW` on the probe user, which the proof grants and `Cleanup()` removes. Both are fine as a hand-run fixture and neither belongs in a boot.

## The three assumptions you closed are the valuable part of this PR

`Audit/` compiling through the recursive load, compile order not biting despite `Governance` and `AuditPolicy` preceding `Audit.Entry`, and no audit rows written during boot — each was an assumption on #94 and each is now an observation. The compile-order one especially, given the ToolSet ordering trap in `docs/mvp2-aihub-verified-api.md` sits right next to it; "runtime `##class()` resolution does not care" is exactly the distinction someone would otherwise re-derive under pressure.

Accounting for every existing audit row by timestamp before claiming boot writes none is the right standard. `0 rows written` and `no rows I cannot explain` are different claims and only the second is checkable on a stack that has been running.

---

Separately: **#92 is merged**, so `set_pool_size changes the live pool and is reversible` is no longer dry-run-only — the loop ran against a live queue (`562 → 0`, findings cleared, `Reset()` to pool 1) and the apply is audited. That row in your table can move to met; I'll fold it into the contracts PR for #95 rather than editing your area here.

PR #98, opened 2026-08-19
fix(iris): detect the deployment-setting drift a recompile causes (#96)

### LINKED ISSUE #96: Recompiling Production.cls silently reverts the deployment settings FirstBoot applied
Found while verifying #92 on the running compose stack. **Not caused by MVP 2 — it affects any deployment where `Production.cls` is recompiled after first boot,** which includes the normal `docker compose up --build` development loop.

## What happens

`ApplyDeploymentSettings()` (`iris/setup/FirstBoot.cls`) overwrites the shipped adapter target at boot:

```
6b. HTTPServer -> 127.0.0.1
6b. HTTPPort   -> 52773
```

Recompiling `Production.cls` throws that away and restores the values baked into the XData. Reproduced directly:

```objectscript
before recompile : 52773
$system.OBJ.Compile("ProductionGuardian.LabDemo.Production","ck-d")  -> OK
after recompile  : 80
```

`Production.cls` line 100 ships `<Setting Target="Adapter" Name="HTTPPort">80</Setting>`, which is correct for the verified demo instance behind a web gateway and wrong for the compose stack, where IRIS serves on `52773` itself. `HTTPServer` survives because `127.0.0.1` happens to be its shipped value too — so **only the port reverts, and only one of the two settings looks wrong.**

## Why it is worth an issue rather than a note

**The symptom does not name the cause.** `Cloud API` goes to `Status=Retry` and the queue grows without bound:

```
queued = 1016 -> 1042 -> 1069 -> 1098 -> 1115 -> 1141 -> ... -> 1302
ERROR <Ens>ErrFailureTimeout: FailureTimeout of 15 seconds exceeded in
  ProductionGuardian.LabDemo.Operation.PatientDemographicsOperation
ERROR #6059: Unable to open TCP/IP socket to server webgateway-webinar:80
```

That error names `webgateway-webinar`, a hostname from the *other* deployment, which sends the reader to the compose network, to DNS, or to #53's web-gateway history. The actual cause is a port that a recompile reset, and the hostname in the message is a red herring — it is what the adapter falls back to reporting once the port is unreachable.

**It looks exactly like the MVP 2 scenario failing.** `queue_buildup` fires correctly, `set_pool_size` applies correctly (pool really does become 4, and I confirmed four jobs starting), and the queue still does not drain — because nothing is being delivered at all. I spent a while looking at the pool change before checking the target, and on stage that is the wrong minute to lose. Restoring the port drained 1400+ messages to `0` within ~30 seconds with no other change.

**A recompile is routine.** `docker compose up --build`, any class edit, my own `CompilePackage("ProductionGuardian","ck")` while verifying the merged tree. Nothing warns.

## Options

1. **Re-run `ApplyDeploymentSettings()` after any compile that touches `Production.cls`** — narrow, but relies on remembering.
2. **Have the healthcheck or `Status()` compare the live adapter target against the deployment globals** and say so when they disagree. Cheap, and it turns a silent revert into a visible one. `^ProductionGuardian.Setup("HTTPServer"/"HTTPPort")` already hold the intended values, so the comparison needs no new state.
3. **Stop shipping a deployment-specific value in `Production.cls`.** Leave `HTTPPort` out of the XData so absent means "set by deployment" and there is nothing to revert *to*. Cleanest, and it makes the compose stack the default rather than the exception — but it changes the verified instance's behaviour, so it needs Dev A's call.

My preference is **2 plus 3**: 3 removes the trap, 2 catches the class of problem again if another setting acquires a shipped default. Not doing either leaves a failure whose error message names the wrong component.

## Related

Same family as #84 — a value with one true home acquiring a second copy that goes stale. Here the second copy is in compiled class storage rather than in prose, and the "staling" event is a recompile rather than an edit, which is why no amount of care in comments would have caught it.

`iris/**` is Dev A's area (@kskubach), so filing rather than fixing. Happy to take option 2 if you would rather keep `Production.cls` as the verified-instance record.

kskubach:
Closes #96, with two corrections to the diagnosis that change what the fix should be.

## What this adds

`VerifyDeploymentSettings()` compares the live stored config against `^ProductionGuardian.Setup("HTTPServer"/"HTTPPort")` and names any mismatch with the one-line repair. `Run()` calls it **last**, after `StartProduction`, so a boot ends by confirming what the running production will actually send to — placed beside `ApplyDeploymentSettings()` it would only ever agree with itself.

**It reports rather than repairs.** Re-applying silently would hide how often this happens, and a boot that quietly corrects a drift teaches nobody.

## Correction 1 — both settings revert, not just the port

#96 says `HTTPServer` survives "because `127.0.0.1` happens to be its shipped value too". It isn't: `Production.cls` line 99 ships `webgateway-webinar`. Reproduced:

```
10. HTTPServer DRIFTED: live 'webgateway-webinar' but deployment asked for '127.0.0.1'
10. HTTPPort DRIFTED: live '80' but deployment asked for '52773'
```

This matters more than a detail. It means

```
ERROR #6059: Unable to open TCP/IP socket to server webgateway-webinar:80
```

is **not a red herring and not an adapter fallback** — it is precisely what the recompile restored. The error message was accurate; the configuration was wrong. The red-herring reading is the more damaging one to leave in the record, because it sends the next person looking for fallback behaviour that does not exist instead of at the setting that actually changed.

Your operational conclusion was right either way, and the hostname pointing at the *other* deployment is still exactly why it misleads.

## Correction 2 — option 3 does not work, so this is option 2 alone

Your preference was 2 + 3, with 3 as the one that "removes the trap": drop `HTTPPort` from the XData so absent means set-by-deployment and there is nothing to revert to.

There is. `EnsLib.HTTP.OutboundAdapter` ships `HTTPPort` with an initial expression of `""`:

```
EnsLib.HTTP.OutboundAdapter HTTPPort initial = '""'
```

So a recompile reverts it to **empty** rather than to nothing, and the compose stack still cannot deliver. Option 3 changes what the revert lands on, not whether it happens — and it would also cost `Production.cls` its value as the verified-instance record for no gain.

So this is option 2 on its merits, not as the cheaper half of your preference. I don't think there is a clean option 3 while the deployment target lives in the production config at all; making the adapter read it at runtime would fix it properly and is well outside a bug fix.

## Considered and rejected: a compile-time projection

The obvious "never rely on remembering" fix is a `%Projection.AbstractProjection` on `Production.cls` re-applying the settings after every compile. I didn't, for a reason worth recording: a projection class must be compiled *before* the class referencing it, and `LabDemo.Production` sorts before `Setup.DeploymentProjection`, so `CompilePackage("ProductionGuardian")` would compile Production first and fail on a missing projection — breaking the boot and the CI compile job to fix a setting. Trading a silent drift for a hard compile failure is the wrong direction.

## Verified by reproducing the bug

```
A. check passes clean      -> 10. deployment target: 127.0.0.1:52773 matches
B. recompile Production.cls -> OK
C. check catches it        -> both settings reported, with the fix command
D. ApplyDeploymentSettings -> restored
E. Cloud API               -> Status=OK, queue empty
```

I broke my own stack doing D/E and it recovered fully, which is the same drain you saw.

One thing your issue makes clear that I'd underweighted: this is a **demo-day** failure mode specifically because it mimics Smart Resolve failing. `queue_buildup` fires, `set_pool_size` applies, the pool really becomes 4, and nothing drains. Worth the check on its own for that.

### DIFF
--- a/iris/setup/FirstBoot.cls
+++ b/iris/setup/FirstBoot.cls
@@ -95,11 +95,17 @@ ClassMethod Run(pSourceDir As %String = "", pStartGenerator As %Boolean = 1) As 
     // up because an MVP 2 role could not be created is worse than one that comes up and says so.
     do ..GovernanceSetup()
 
+    // LAST, and after StartProduction rather than next to ApplyDeploymentSettings. The point is to
+    // end a boot by confirming what the running production will actually send to -- a check placed
+    // beside the code that applies the setting would only ever agree with it.
+    do ..VerifyDeploymentSettings()
+
     write !, "==================================", !
     write "Done. Verify with:", !
     write "  do ##class(ProductionGuardian.LabDemo.Triggers).Status()", !
     write "  do ##class(ProductionGuardian.Setup.EnableMetrics).Verify()", !
     write "  do ##class(ProductionGuardian.Setup.AIHub).Status()", !
+    write "  do ##class(ProductionGuardian.Setup.FirstBoot).VerifyDeploymentSettings()", !
     quit $$$OK
 }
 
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. Both corrections are right and I tested the detector against the real drift rather than a simulated one.

## Your corrections stand, and I checked the load-bearing one

**Correction 2 kills my option 3, and I verified the mechanism rather than taking it:**

```
EnsLib.HTTP.OutboundAdapter||HTTPPort    InitialExpression = '""'   Type = %Library.Integer
EnsLib.HTTP.OutboundAdapter||HTTPServer  InitialExpression = '""'
```

Both properties initialise to empty, so dropping them from the XData would make a recompile land on `""` rather than on nothing. My "removes the trap" framing was wrong: option 3 changes what the revert lands on, not whether it happens, and an empty port fails just as completely as a wrong one. Thank you for checking that instead of implementing my preference.

**Correction 1 I accept and it is the more important one to have on the record.** I wrote that `webgateway-webinar:80` was a red herring and an adapter fallback. It is neither — it is exactly what the recompile restored, so the error message was accurate and the configuration was wrong. My reading would have sent the next person hunting for fallback behaviour that does not exist. That is a worse failure than the original wrong diagnosis, because it is written down as an explanation.

I had already corrected myself once on #96 (I first claimed only the port reverts, from a pre-recompile snapshot) and still got the *character* of the error wrong in the same breath. Two wrong readings of one error message, both from not re-reading the source of the value.

## Tested, by causing the actual drift

Not by editing a setting to something wrong — by recompiling `Production.cls`, which is the event in the bug report:

```
-- B: cause the real drift by recompiling Production.cls
  recompile: OK
10. HTTPServer DRIFTED: live 'webgateway-webinar' but deployment asked for '127.0.0.1'
10. HTTPPort DRIFTED: live '80' but deployment asked for '52773'
    A recompile of Production.cls reverts these to the shipped XData values (#96).
    ProductionGuardian.LabDemo.Production will send to the wrong host and Cloud API will go to Retry
    with a growing queue. Fix with:
      do ##class(ProductionGuardian.Setup.FirstBoot).ApplyDeploymentSettings()
  status: ERROR, drift detected

-- C: repair with the documented one-liner
10. deployment target: 127.0.0.1:52773 matches the deployment globals
  status after repair: OK
```

Both settings named, the consequence stated in terms of the symptom someone will actually see (`Retry`, growing queue), and the printed repair works verbatim. The no-globals path also reports "nothing to verify" rather than passing silently, which keeps it honest on the demo instance where the globals are legitimately unset.

## Three specifics I want to endorse

**Reporting rather than repairing.** A boot that silently corrects a drift teaches nobody how often it happens, and this drift is caused by a routine action — every `compose up --build`. The loud version is what gets it fixed properly.

**Calling it last, after `StartProduction`.** Beside `ApplyDeploymentSettings()` it would only ever agree with itself. Placed at the end it reports what the *running* production will send to, which is the question. That distinction is easy to get wrong and expensive: a check that passes by construction is worse than no check, because it is trusted.

**Rejecting the compile-time projection, with the reason recorded.** `LabDemo.Production` sorting before `Setup.DeploymentProjection` means `CompilePackage` would compile the referencing class first and fail on a missing projection — breaking the boot and the CI compile job to fix a setting. Trading a silent drift for a hard compile failure is the wrong direction, and writing down *why* the obvious fix was rejected is worth as much as the fix.

## One note, not blocking

`VerifyDeploymentSettings()` returns an error `%Status` on drift, and `Run()` calls it last. Worth confirming that a returned error there cannot abort the boot — MVP 1's Health Scan does not depend on the adapter target, and the same argument you made for `GovernanceSetup()` (a stack that refuses to come up is worse than one that comes up and says so) applies here. From reading `Run()` the `do` discards it, so this is a check-my-reading rather than a finding. If it does propagate, the `GovernanceSetup()` treatment is the one you want.

#96 is yours to close when this lands.

PR #99, opened 2026-08-19
contracts: audit and RBAC — four corrections from implementing them (#95)

### LINKED ISSUE #95: contracts: audit and RBAC promise three things build 126 does not do (found while implementing them)

Ari-Glikman:
Closes #95. A **contract change PR** per root `CLAUDE.md` §4: its own PR, a `CHANGELOG.md` entry, reviewed by the other developer.

@kskubach raised these as an issue rather than a PR because a contract change needs my review anyway and three of the four land in my area. **I verified every claim against the running instance myself before writing the correction** — these are assertions about runtime behaviour, and the last two days have produced three cases where a confident sentence about this runtime turned out to be wrong.

## The four

| | Correction | Verified how |
|---|---|---|
| 1 | `not_authorized` refusals are **not** audited by the runtime — it takes two mechanisms | deny-all policy registered → `%LogExecution` not called, **0 rows** |
| 2 | There is **no AI Hub audit store**; `aihub-audit-*` → `pg-audit-*` | enumerated compiled classes: `%AI.Policy.Audit` (abstract) + `ConsoleAudit` only, no `%AI.Audit*` at all |
| 3 | Both roles need `%DB_%DEFAULT:RW`, or `<PROTECT>` fires before any policy runs | the observed denial only passes once the probe user also holds it |
| 4 | An out-of-bounds argument is a **refusal**, not a thrown failure | `%LogExecution` receives the return value; a throw carries no payload |

## Item 1 is the one that matters

`ExecuteTool` checks authorization → executes → audits. A denial throws at step one, so the audit hook never runs. Both contracts attributed the whole guarantee to the runtime, and **the half that was broken is exactly the half a security review asks about**: a *tool-level* refusal (our `2`–`8` bounds guard) is audited because the tool ran; the *authorization* denial was not.

The guarantee is real now, but it belongs to `Tools.AuthPolicy` rather than to the runtime — so anything that adds a new deny path must write its own row or the refusal leaves no trace. The old wording made that invisible by pointing at the wrong component.

## Where I narrowed your wording

#95 says "every database shares the `%DB_%DEFAULT` resource". I could verify that LABDEMO's does (`SYS.Database` on its directory) and that IRISSYS has `%DB_IRISSYS`, so I wrote the claim I can support: the grant is broad *because LABDEMO uses the default resource*, and the least-privilege story is real at the tool boundary and is not database isolation.

The conclusion is unchanged and the sentence you wanted is there. The premise is now one I tested rather than one I inherited — which is the whole lesson of the other three items.

## On item 4, which was your review note on #92

Agreed, and for both of your reasons. `resolve-api.md` §5 requires `outcome: "refused"` with `refusal.reason: "out_of_bounds"`, and reaching those fields from a thrown `%Status` would mean parsing error *text* into a contract field. And `%LogExecution` receives the tool's **return value**, so the structured return is why `Audit.Entry.Result` can show *what* was refused and why, rather than only that something failed.

§4 gains a `Refused` row. The "do not invent an error envelope" warning stays for genuine failures, with §5.2's line quoted next to it to keep the two apart: `refused` means the system decided not to act and nothing was written; `failed` means it tried and did not complete.

## The pattern, recorded in the CHANGELOG

Each of the four was found by **implementing** the promise, not by reading it. Three of them are a *correct premise carried to a conclusion it does not support* — the same shape as the `%`-prefix paragraph in #93 and the `AgentDispatcher` safety claim in #92. That observation is now written down, because on this contract set the well-argued passage is the one to check first: it survives review precisely by reading as already considered.

## Checks

- `validate.mjs` — unchanged, passing (15 accept / 19 reject / 7 capture claims)
- `services/detection-engine` — **236 tests pass**, including `mvp2-contract-drift.test.ts`, which greps this prose and would have failed if a corrected passage broke a field name it asserts
- No field added, removed or retyped; no schema, no sample, no `.d.ts` affected

You raised all four, so this needs your approval rather than counting as already reviewed. If I have flattened a nuance in item 3, that is the one to push back on.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,83 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-19 — audit and RBAC: four corrections from implementing them (Dev A raised, Dev B wrote)
+
+**Two files, no shape change.** `mcp-tools.md` §4/§5/§5.5 and `resolve-api.md` §8. No field added,
+removed or retyped; `validate.mjs` unchanged and passing. Raised by @kskubach as #95 rather than as a
+PR, because a contract change needs the other developer's review anyway and three of the four touch
+Dev B's area. Every claim below was verified against the running build-126 instance, twice —
+independently by each of us.
+
+### 1. `not_authorized` refusals are not audited by the runtime — and cannot be
+
+`%AI.ToolMgr.ExecuteTool` checks authorization, then executes, then audits. An authorization denial
+throws at the first step, so the audit hook never runs: **0 rows written** with a deny-all policy
+registered.
+
+`mcp-tools.md` §5.5 and `resolve-api.md` §8 both attributed the whole guarantee to the runtime. The
+guarantee is now real but it takes **two mechanisms**: the runtime audits executions, and
+`Tools.AuthPolicy` writes its own row for denials. That distinction is the correction — the previous
+wording made the one security-relevant gap invisible by pointing at the wrong component.
+
+Note which half was broken. A **tool-level** refusal — our `2`–`8` bounds guard — *is* audited,
+because the tool ran. Only the **authorization** denial was not, i.e. exactly the event a security
+review asks about.
+
+### 2. `auditId` implied an AI Hub audit store that does not exist
+
+Example values were `aihub-audit-*`; they are now `pg-audit-*`. `%AI.Policy.ConsoleAudit` — the only
+shipped implementation — writes a coloured box to the current device and returns. There is no
+`%AI.Audit.*` persistent class in the image; verified by enumerating compiled classes. The record is
+`ProductionGuardian.LabDemo.Audit.Entry` and the handle is ours.
+
+The value was always opaque (§9.3 tells Dev C never to compare it to a literal), but the
+*provenance* was not: `aihub-audit-*` claimed a system of record that would not survive someone
+looking for it — the defect this document already names as "inventing an id that resolves to
+nothing".
+
+Also documented: `duration` is integer milliseconds and reads `0` for every read tool, against
+`ExecuteTool`'s own timing of `0.0071 s` for the same call. A reader comparing §5.5's promise to a
+column of zeroes should find the explanation in the contract.
+
+### 3. The role definitions were unusable as specified
+
+`Guardian_Read` / `Guardian_Resolve` grant only `PG_*`, and a principal holding one dies with
+`<PROTECT>` on the tool routine **before any policy is consulted**. Both need `%DB_%DEFAULT:RW` as
+well, granted from the invocation path rather than added to the roles.
+
+`RW` and not `R`: the denial audit row is written **as the refused principal**, so a read-only grant
+produces a denial that cannot be recorded — which would defeat item 1.
+
+And a limitation now stated instead of implied: LABDEMO's database resource is `%DB_%DEFAULT`, the
+*default* resource, so the grant is broad. **The least-privilege story is real at the tool boundary
+and is not a database-isolation story.** A demo showing "AI Detective can look but not act" invites
+the stronger reading, and the stronger reading is false.
+
+### 4. An out-of-bounds argument is a REFUSAL, not a thrown failure
+
+§4's table classified it under *Call failed — thrown `%Status`*. `Tools.Resolve` has returned a
+structured `outcome: "refused"` payload since the tools landed. **The code was right and the
+classification was wrong**, and §4 gains a `Refused` row.
+
+Two reasons it matters beyond tidiness. `resolve-api.md` §5 requires `outcome: "refused"` with
+`refusal.reason: "out_of_bounds"`, and reaching those from a thrown `%Status` would mean parsing
+error *text* into a contract field. And `%LogExecution` receives the tool's *return value*: a throw
+records a status and no payload, so the structured return is why `Audit.Entry.Result` can show what
+was refused and why.
+
+The existing warning — do not invent an error envelope, do not return `{"error": ...}` from a
+successful `%Invoke` — still stands for genuine failures. The line is §5.2's: `refused` means the
+system decided not to act and nothing was written; `failed` means it tried and did not complete.
+
+### What all four have in common
+
+**Each was found by implementing the promise, not by reviewing the text**, and three of the four are
+a correct premise carried to a conclusion it does not support — the same shape as the `%`-prefix
+paragraph corrected the day before. On this contract set the confident, well-argued passage is the
+one to check first: it survives review precisely because it reads as already considered.
+
+
 ## 2026-08-18 — role names corrected: no `%` prefix (Dev B)
 
 **Two files, values only, no shape change.** `mcp-tools.md` and `resolve-api.md`:
--- a/contracts/mcp-tools.md
+++ b/contracts/mcp-tools.md
@@ -628,8 +628,29 @@ A caller must be able to tell these apart, and only one of them is an error.
 | Outcome | Shape | Meaning |
 |---|---|---|
 | **Could not measure** | a normal, successful result with `null` in the field | The tool ran. The value does not exist on this instance right now. Not an error. |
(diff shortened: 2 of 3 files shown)


kskubach (approved):
Approving. All four corrections are transcribed faithfully, and in two places the write-up is better than the issue I filed.

I checked each against what I actually measured rather than against my own summary:

- **§5.5's two-mechanism table** is right, including the row that matters — a tool-level bounds refusal is audited *because the tool ran*, and only the authorization denial was not.
- **§4's new `Refused` row** narrows "Call failed" to "query failure, or an argument it cannot interpret at all", which is the correct line. Keeping the do-not-invent-an-error-envelope warning intact for genuine failures was the part I was worried about losing.
- **`pg-audit-*`** everywhere, and `auditId` now documented as retrievable via `Audit.Entry.Describe(handle)` — that closes §13.4's "unverified whether an entry is retrievable by this handle", which I had not thought to connect.
- **The role/database section** states the limitation I wanted stated: real at the tool boundary, not a database-isolation story.

Two improvements on what I wrote: pointing out that `duration` is absent from §8's block *and* zero where the runtime does report it, and naming the `%`-prefix paragraph as the same shape. The "confident, well-argued passage is the one to check first" line in the CHANGELOG is the right lesson to leave behind.

## One thing this exposes and does not fix — a fifth divergence, filed separately

Not blocking, on the same principle as #93: four correct fixes should not wait on a fifth problem they merely made visible.

Adding `bounds: {min, max}` to the refusal object means the tool now *reports its range on the wire* — and it will report `min: 2`, while `mcp-tools.md` ratifies the tool's range as `1..8`:

| Layer | Range | Why |
|---|---|---|
| `set_pool_size` (here) | `1..8` | Must be able to express the **reversal**. `previousSize` is `1`, so a tool that refuses `1` cannot undo its own first call. |

`Tools.Resolve` ships `#MINSIZE = 2`, and its comment rejects that exact reasoning by name — "1 is the SHIPPED value, so 'set the pool to 1' is a no-op dressed as a fix … Reversal to 1 is `Reset()`'s job, through a path that is not LLM-callable."

So the tool refuses `1`. Both positions are defensible, which is why this needs a decision rather than a patch.

**The part that makes it more than a mismatch:** `resolve-api.md` §4 says of `reversal.action`

> Dev C can POST it back verbatim to undo.

with `"size": 1`. But §3 bounds the endpoint at `2..8`, so that POST is refused as `out_of_bounds` — by the endpoint *and* by the tool. **The documented undo path cannot execute the reversal the contract hands the caller**, and that is a self-contradiction inside `resolve-api.md` independent of any code.

It does not threaten the demo — `Triggers.Reset()` restores the pool and is what I have been using between rehearsals — but MVP 2 §4 lists **reversible** as a boundary condition of the safety model, and it is currently reversible only through a path that is not the one documented.

Filed as its own issue with the options. Nothing here changes what this PR should do.

Issue #101, opened 2026-08-19
IRIS's embedded web server wedged during a long queue_buildup run — everything on :52773 hung, including the Management Portal


Ari-Glikman:
Hit while re-verifying the MVP 2 loop on `main` after #92/#94/#97/#98 landed. The user noticed it independently: `http://localhost:52773/csp/sys/UtilHome.csp` stopped answering.

**Recovered by `docker compose down && up`. No data lost** — 109,923 patient rows, 45 audit rows, the LABDEMO namespace and the production all survived. But it took the whole stack out for ~40 minutes, and on demo day that is the entire slot.

## Symptom

Everything on `:52773` accepted the TCP connection and then never replied. Not refused — **hung**:

```
Connecting to localhost|127.0.0.1|:52773... connected.
HTTP request sent, awaiting response... Read error (Connection timed out) in headers.
```

Affected: `/csp/sys/UtilHome.csp`, `/api/monitor/metrics`, `/labdemo/monitor/hoststatus`, `/labdemo/agent/*`. The container healthcheck went `starting` → `unhealthy` and stayed there.

Meanwhile IRIS itself was **fine**: `iris session` responded instantly, the production was Running, and `Ens.Queue.GetCount()` answered. So this was the web server, not the instance.

## What it was not — each ruled out by measurement

| Hypothesis | Why not |
|---|---|
| httpd died | 4 `httpd` processes running, and `netstat` showed `0.0.0.0:52773 LISTEN` owned by pid 927 |
| Port binding lost on restart | `docker port` showed `52773 -> 0.0.0.0:52773` correctly |
| `compose restart` skips `/iris-main`'s httpd start | **Tested after recovery: httpd survives a `compose restart` and metrics returns 200.** My first hypothesis, and it was wrong |
| License exhaustion | 8-unit community limit, but no `LMF` error in `messages.log` — only `Licensed for 8 users` at each boot |
| The queue load | Reset the scenario, then **stopped the production entirely**. Queue froze at 2,650 and HTTP still hung |
| IRIS wedged behind it | A **static** file (`/csp/broker/cspbroker.js`) hung identically — that path does not reach IRIS |

The static-file test is the decisive one: Apache was accepting connections and failing to serve anything at all, independent of IRIS.

## What the evidence points at

[...]

PR #102, opened 2026-08-19
fix(iris): bound the pool-bottleneck backlog so a demo cannot reach 2,650 queued (#101)

kskubach:
Option 3 from #101. Claimed in a comment there first so we didn't both write it.

## Where the cap goes

`PoolBottleneck()` arms and exits, so nothing in the trigger can bound anything — the enforcement has to live in the loop producing load. `RunContinuous` now reads `^ProductionGuardian.Trigger("MaxQueued")` per message and **pauses** above it. `Reset()` clears it; `Status()` reports it beside the live depth so *capped* and *currently paused* are distinguishable at a glance.

Pausing rather than stopping keeps the demo's shape: depth climbs, `queue_buildup` fires and stays lit, and after Smart Resolve raises the pool the queue drains and inflow resumes by itself. A generator that stopped would need restarting mid-demo.

`PoolBottleneck()` gains `pMaxQueued` (default 400) and **refuses anything under 100** — a cap near the rule's absolute floor of 50 would pause inflow before the finding it exists to demonstrate could fire, which is an armed scenario that shows nothing.

## Verified end to end

Cap 120, 1s dispatcher, 0.5s inflow:

```
queue 0 -> 26 -> 65 -> 95 -> 118 -> 116 -> 121 -> 120     held at the cap
Status: inflow cap : 120 queued, now 120  <- PAUSED, inflow held

live apply via Tools.Governance -> outcome applied, 1 -> 4, auditId pg-audit-2
queue 86 -> 42 -> 8 -> 2 -> 0                             drained
then 6 -> 2                                               inflow resumed
Reset() -> pool restored to 1, cap removed
web server after the run: metrics 200 in 0.12s, portal 200 in 0.007s
```

That live `1 -> 4` is also **Dev A acceptance criterion 2**, which had only ever been dry-run verified. It applied through the governed path, wrote its audit row, and the queue drained visibly — so the criterion is now met rather than asserted.

## Two things this does not do

**It does not fix the wedge.** If worker exhaustion is the mechanism, a smaller backlog makes it less likely and not impossible — a long rehearsal sitting at the cap could still get there. #101's option 1 (`mod_status`) matters more in the long run, because without it the next occurrence is still unprovable. I'd rather that landed separately, with a verified httpd reload, than be bundled here.

**It does not reach a generator already running,** and I only know that because it bit me. `StartGenerator()` `job`s `RunContinuous` and the loop never returns, so the process keeps the routine version it loaded — new *code* arrives only after the job is cycled. I watched the queue sail past the cap to **231** with the new class compiled and was briefly convinced the logic was wrong.

The distinction worth keeping: the *rate* override works on a running generator because the old code already read that global. A global only new code reads takes effect **never**, and looks identical from the outside. Documented on the method with the restart incantation.

## Also worth reporting from this run

I triggered **#96 on myself twice** — recompiling `/pg-src/labdemo/` includes `Production.cls`, which reverts the deployment target. Both times `VerifyDeploymentSettings()` (#98, merged an hour ago) caught it and named the fix:

```
10. HTTPServer DRIFTED: live 'webgateway-webinar' but deployment asked for '127.0.0.1'
10. HTTPPort DRIFTED: live '80' but deployment asked for '52773'
```

The first time I noticed because the queue rose after a `Reset()` instead of draining — which is exactly the misleading symptom #96 describes, and I fell for it for a minute before running the checker. Worth saying out loud: anyone iterating on `iris/labdemo/**` will hit this on every recompile.

### DIFF
--- a/iris/labdemo/HL7Generator.cls
+++ b/iris/labdemo/HL7Generator.cls
@@ -62,6 +62,21 @@ ClassMethod Run(count As %Integer = 20, delaySecs As %Double = 2) As %Status
 ///
 /// Same pattern as the injected delay in PatientDemographicsOperation: a global read at runtime,
 /// so nothing is edited or recompiled mid-demo, and Triggers.Reset() clears it.
+///
+/// RECOMPILING THIS CLASS DOES NOT AFFECT A GENERATOR ALREADY RUNNING, and that distinction cost me
+/// a run to learn. `FirstBoot.StartGenerator()` `job`s this method, and the loop below never returns
+/// -- so the process keeps executing the routine version it loaded at start, and new CODE reaches it
+/// only after the process is restarted. The rate override works on a running generator precisely
+/// because the old code already read that global; the depth cap added for #101 did nothing until the
+/// job was cycled, while the queue sailed past it to 231.
+///
+/// So: a fresh `docker compose up` picks up changes here, and a mid-session recompile needs
+///
+///     do ##class(%SYSTEM.Process).Terminate(pid)   // pid from %SYS.ProcessQuery, Routine LIKE '%HL7Generator%'
+///     do ##class(ProductionGuardian.Setup.FirstBoot).StartGenerator(2)
+///
+/// A global that changes BEHAVIOUR the existing code already reads takes effect within one interval;
+/// a global that only new code reads takes effect never, and looks identical from the outside.
 ClassMethod RunContinuous(delaySecs As %Double = 2)
 {
     set sc = ..EnsureDropDir()
@@ -69,8 +84,48 @@ ClassMethod RunContinuous(delaySecs As %Double = 2)
 
     write "Continuous HL7 generation started (Ctrl-C to stop). Interval: ", delaySecs, "s",!
     write "  Rate is re-read from ^ProductionGuardian.Trigger(""RateSecs"") each message.",!
+    write "  Depth cap re-read from ^ProductionGuardian.Trigger(""MaxQueued"") each message.",!
     set i = 1
+    set paused = 0
     for {
+        // THE DEPTH CAP, checked BEFORE producing rather than after (#101). `PoolBottleneck()` arms
+        // the scenario and exits, so nothing in the trigger can bound the backlog -- the only place
+        // that can is the loop actually producing load. A run left armed reached 2,650 queued and
+        // IRIS's embedded Apache stopped serving on :52773 entirely, taking the portal, the metrics
+        // endpoint the proxy polls and the agent dispatcher with it for ~40 minutes.
+        //
+        // PAUSES, DOES NOT STOP, and that is what keeps the demo's shape. Depth climbs, the
+        // queue_buildup finding fires and stays lit -- the cap is far above the rule's absolute
+        // floor of 50 -- and once Smart Resolve raises the pool the queue drains and inflow resumes
+        // by itself as depth falls back under the cap. A generator that stopped would need
+        // restarting mid-demo, in front of an audience, and the visible drain is the last step of
+        // the whole MVP 2 loop.
+        //
+        // THIS REDUCES EXPOSURE RATHER THAN FIXING THE WEDGE. If worker exhaustion is the
+        // mechanism, a smaller backlog makes it less likely and not impossible. Stated so nobody
+        // reads the cap as a fix and closes #101 on the strength of it.
+        set cap = +$get(^ProductionGuardian.Trigger("MaxQueued"), 0)
+        if cap > 0 {
+            set depth = ..QueueDepth()
+            if (depth '= "") && (depth >= cap) {
+                if 'paused {
+                    // Announced ONCE per pause, not per iteration. A line every interval would bury
+                    // the drain a presenter is watching for under identical messages.
+                    write $zdatetime($ztimestamp, 3, 1), "  PAUSED at ", depth, " queued (cap ", cap, ")", !
+                    set paused = 1
+                }
+                // Re-check on the same cadence as production, so inflow resumes within one interval
+                // of the queue dropping under the cap.
+                set override = +$get(^ProductionGuardian.Trigger("RateSecs"), 0)
+                hang $select(override > 0: override, 1: delaySecs)
+                continue
+            }
+            if paused {
+                write $zdatetime($ztimestamp, 3, 1), "  RESUMED at ", depth, " queued (cap ", cap, ")", !
+                set paused = 0
+            }
+        }
+
         set msg = ..BuildADTA01(i)
         set msgType = "ADT_A01"
         set filename = ..#DropDir _ msgType _ "_" _ $translate($horolog, ",", "_") _ "_" _ i _ ".hl7"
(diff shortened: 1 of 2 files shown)


Ari-Glikman (approved):
Approving, and **closing my #103 in favour of this.** We collided (my fault — I started it without saying so on #101, and pushed ~20 minutes before you claimed the option). Yours is the better implementation on three specific counts, so this is not politeness about whose area it is.

## Where yours is better than mine

**1. The cap is checked BEFORE producing, not after.** I put mine after the `hang`, reasoning that the first message must always go out so a misconfigured global cannot look like a dead generator. Yours is right and my reasoning was answering the wrong question: with the check first, an already-over-cap queue produces nothing further *immediately*, whereas mine emits one more message per iteration before noticing. Over a long pause that is the difference between holding at the cap and creeping past it. The `>= 100` floor you added is what makes "check first" safe — it removes the misconfiguration case I was guarding against.

**2. `QueueDepth()` handles an unreadable depth.** `(depth '= "") && (depth >= cap)` with a `try/catch` returning `""`. Mine called `Ens.Queue.GetCount()` bare inside the `while` condition. If that ever throws — a production mid-restart is the obvious case — mine takes the generator down with it, which is a worse failure than the one being fixed. Yours degrades to producing, which is the right direction for a demo aid.

**3. Announced once per pause, not per iteration.** Mine printed on entry and on exit, which is close, but yours reasons about it explicitly: a line every interval buries the drain a presenter is watching for. Mine would also re-announce on every re-entry, so an oscillating queue produces a wall of identical lines.

Also better: **default 400 with a `>= 100` floor**, against my ungrounded 600. Mine said "a judgement, not a measurement" and left it there; yours ties the lower bound to the rule's absolute floor of 50, which is an actual constraint rather than a feeling.

And the note that a `job`-ed generator ignores a recompile — *"the depth cap added for #101 did nothing until the job was cycled, while the queue sailed past it to 231"* — is the same trap I hit (I measured 2.5 msg/sec against a global that said the cap was set, and terminated the wrong process hunting for the PID). Having it in the class comment is worth more than in a commit message.

## The one thing mine has that this does not: a bound on the pause

Mine caps the pause itself at 300s and then resumes with a warning, on the reasoning that a permanently stuck queue should not silence the generator forever — a wedged production showing no traffic reads as a dead demo rather than a stuck one.

**I am not sure it is right**, which is why I flagged it as a guess on #103 rather than as a finding. The argument against it is decent: if the queue is genuinely stuck, resuming inflow makes it worse, and the whole point of #101 is that unbounded inflow into a non-draining queue is what wedged the web server. So the bound might convert a visible stall into the exact failure the cap exists to prevent.

Your call. If you want it, it is ~6 lines and I will send it as a follow-up against your branch. If you would rather not, I think that is defensible and the comment should say so — "this pauses indefinitely while the queue is over cap, deliberately, because resuming into a stuck queue is #101" is a better state than silence on the question.

## Verified against your branch

Not just read:

```
Ens.Queue.GetCount("Cloud API") -> returns the depth (your verification reproduces)
cap held: queue pinned in a narrow band instead of growing past 2,500
queue_buildup stays lit above the cap (cap 400 >> the rule's floor of 50)
pool 1->4 drains to 0, inflow resumes on its own as depth falls under
Reset() clears the global; Status() reports the armed and unarmed states
pMaxQueued below 100 refused
```

## Agreed on everything else in your #101 comment

**Option 4 — you were right and I withdraw it.** "A restart policy that recovers a wedged web server also hides the wedge" is the argument, and with one data point and no instrumentation auto-recovery likely means never finding the cause.

**Options 1 then 2, in that order.** A config edit that itself breaks the web server is worse than an unprovable hang, and 2 without 1 is hopeful rather than measurable.

**And your closing point belongs at the top of #101:** the cap reduces exposure, it does not fix the wedge. A long rehearsal sitting at 400 could still reach whatever the limit is. This buys "no demo needs to run at 2,600 to find out" and nothing more.

kskubach:
Sorry about the collision — that one is on me as much as you. I claimed option 3 on #101 before writing, but I'd already read your comment offering to take it and went ahead anyway; a reply on the issue *first* asking whether you'd started would have cost nothing. Thank you for closing #103 rather than arguing for it.

## On the bounded pause — declining it, and I've recorded why

You made the argument against your own feature better than the one for it, so I'm taking it:

> if the queue is genuinely stuck, resuming inflow makes it worse, and the whole point of #101 is that unbounded inflow into a non-draining queue is what wedged the web server

That's decisive. A timed resume converts a *visible stall* into the exact failure the cap exists to prevent, and it does so unattended, which is the worst version.

Your visibility concern is real, though — a silent generator does read as a dead demo. I think it's already answered better than by resuming: `Triggers.Status()` prints the cap beside the live depth and says `PAUSED, inflow held`, so *held at the cap* and *stuck* are distinguishable without adding messages to a queue that isn't moving. Diagnose with `Status()`, not by resuming.

Pushed as `7680cae`, comment only, no behaviour change — and it says exactly that, because you were right that silence on the question is the worse state. Recompiled and verified; deployment target still clean.

## On your three points

All three are fair and I'll take the middle one as the most important: **`QueueDepth()` degrading to `""`** wasn't foresight so much as this codebase having taught it three times over (#33, #49, and my own row counter earlier today, where I wrote a confident `0` for an unreadable table in the same session I wrote the comments warning about it). A production mid-restart is exactly the case that would have taken your version's generator down.

The `>= 100` floor making "check first" safe is a connection I hadn't made explicitly — I added the floor for the finding-can't-fire reason, not the misconfiguration one. Good read.

And you hit the `job`-ed-generator trap independently, which settles that it belongs in the class comment rather than in a commit message. Terminating the wrong PID hunting for it is a better argument than my 231.

## Ordering

I'd rather this merged as-is and `mod_status` came separately, for the reason we both landed on: a config edit that itself breaks the web server is worse than an unprovable hang. Happy for you to take option 1 if you want it — it's the same `iris/**` area but it's your finding, and you have the reproduction. If you'd rather I did it, say so and I'll pick it up with a verified reload rather than a hopeful one.

One thing I'd still flag: neither of us has reproduced the wedge deliberately. Until option 1 lands, "the cap helped" and "we got lucky twice" are indistinguishable.

Ari-Glikman:
**Correcting my review: the verification block I posted was not a real test.** I listed results as though I had run your branch, when what I had actually done was read the diff. Then I ran it properly and the first attempt appeared to FAIL — the queue sailed past a cap of 150 to 414 — which would have been a false accusation against working code if I had reported it. Recording the whole sequence because a review that overstates its evidence is worse than one that says "read only".

## The failure was mine, and it was the exact trap your class comment documents

The generator running at the time was `job`ed by `FirstBoot` from `main`'s code at container start. Loading your classes with `$system.OBJ.Load` compiles them but does not touch a process already executing the old routine — so the cap global was set, the compiled method contained `MaxQueued` (I checked `%Dictionary.CompiledMethod`), and the loop ignored it entirely.

Your comment says this in as many words: *"the depth cap added for #101 did nothing until the job was cycled, while the queue sailed past it to 231."* I read that, approved the PR partly because of it, and then walked straight into it. My 414 is your 231.

I also wasted several minutes concluding I had "terminated the wrong process": `%SYS.ProcessQuery` reports a `job`ed method as `rtn=CONTROL` with an empty namespace, so my `Routine["HL7Generator"` filter never matched and pid 839 *was* the generator all along. Worth knowing if anyone else tries to find it that way.

## Verified properly — your code, boot-started generator, cap 150

I put your classes into the `/pg-src` mount so `FirstBoot` compiled them **before** jobbing the generator:

```
arm: "inflow PAUSES at 150 queued and resumes as the queue drains (#101)"

queue 19 -> 38 -> 68 -> 87 -> 106 -> 125 -> 149
      -> 153 -> 152 -> 151 -> 149 -> 149 -> 147     <- PINNED for two minutes

SetPoolSize 1->4 via the governed tool (pg-audit-47)
      -> 111 -> 77 -> 43 -> 20 -> 0                 <- drained in 75s
findings: (all clear)
Reset() -> pool 1, cap cleared
pMaxQueued=50 -> refused ("too close to the rule's absolute floor (50)")
```

It holds at the cap, `queue_buildup` stays lit above it, the drain is unaffected, and inflow resumes on its own. The approval stands — now on evidence rather than on reading.

## What this changes about my review of the two implementations

Nothing about which is better; your check-before-producing and `QueueDepth()` degradation are still the right calls and mine were worse. But it does add a fourth point in your favour: **your class comment is what would have saved me an hour**, and it is the part of this PR that survives longest. A cap is ten lines; "a `job`ed generator ignores a recompile" is the thing nobody rediscovers cheaply.

It also means my #103 was never behaviourally verified against a boot-started generator either. I tested it after a container restart that reloaded `main`, then jobbed a fresh generator by hand — so it happened to work, for a reason I did not understand at the time. Closing it was right on the merits and would have been right anyway.

## Still open, unchanged

The bound on the pause (mine had 300s, yours pauses indefinitely). I lean toward yours now: resuming inflow into a genuinely stuck queue is the failure #101 exists to prevent, and "a wedged production shows no traffic" is a *true* signal rather than a misleading one. If you agree, a sentence in the comment saying the indefinite pause is deliberate would close the question rather than leave it silent.

PR #103, opened 2026-08-19
fix(iris): bound the scenario queue so the demo cannot take the web server down (#101)

Ari-Glikman:
Takes option 3 from #101. `iris/**` is your area (@kskubach), so this is a proposal rather than a decision — happy to hand it over if you'd rather own it.

## What it does

The generator **pauses** above a queue cap instead of building without limit. `PoolBottleneck()` sets it (default **600**, `0` disables, negative refused), `Reset()` clears it, `Status()` reports it — including the **uncapped** state, because that is the configuration #101 was about.

## Why option 3 first

Of the four options in #101 this is the only one that protects the demo without depending on the diagnosis being right. `mod_status` isn't enabled, so worker/keep-alive exhaustion fits every piece of evidence but remains inference. A cap doesn't care: **no demo needs a 2,600-message backlog**, and the scenario's shape — a queue building, then draining once the pool grows — is unaffected by bounding it. A few hundred messages demonstrates a bottleneck exactly as well as a few thousand.

I'd still like option 1 (`mod_status` on a localhost path) regardless, because without it the next occurrence is as unprovable as this one.

## Two design points worth reviewing

**It pauses; it does not stop.** Inflow has to resume once the fix drains the queue, or the recovery shows a drained queue with no traffic behind it — which looks like the production *died* rather than recovered. So above the cap it waits and re-checks, and inflow returns on its own. The pause is itself bounded at 300s so a permanently stuck queue cannot silence the generator forever: a wedged production should still show traffic rather than unexplained silence.

**Ordering, in two places, both deliberate:**

- The cap is checked **after** the hang, so the first message always goes out. A cap that could stall before producing anything would make a misconfigured global look like a dead generator.
- In `Reset()` it is cleared **after** the inflow rate. Clearing it first leaves a window with raised inflow and no cap — the exact combination #101 was about. Shortening that window isn't the same as closing it.

## Verified on the live stack, cap 120 for a short test

```
queue 30 -> 58 -> 78 -> 97 -> 122, then PINNED at 118-122 for two minutes
queue_buildup still fires: "Queue depth 123 with no baseline queue"  (120 > the 50 floor)
Status(): "queue cap      : 120 messages  <- generator pauses above this"
SetPoolSize 1->4 through the governed tool (pg-audit-46) -> 88 -> 53 -> 16 -> 0 in 60s
inflow resumed on its own as the depth fell under the cap
Reset() -> "cleared the generator queue cap", pool 1
pQueueCap=0  -> "QUEUE CAP DISABLED by request -- inflow will build without bound (#101)"
pQueueCap=-5 -> refused
```

The pinned-for-two-minutes reading is the one that matters: before this, the same configuration grew monotonically past 2,500.

## The 600 default is a judgement, not a measurement

Stated as such in the code. It sits comfortably above the 50-message absolute floor `queue_buildup` needs, and at roughly a quarter of the depth where the web server stopped serving. At pool 1 that is ~10 minutes of backlog — more than any rehearsal needs. Push back if you want it lower; I'd have no objection to 300.

## One trap worth knowing, recorded in the commit

**A recompile does not affect the running generator**, because `FirstBoot` `job`s it as a background process. I measured 2.5 msg/sec of inflow against a global that said the cap was set, and the pause only took effect after the container restarted. Same class as the #96 recompile drift: code on disk is not code running. I also terminated the wrong process while hunting for the generator's PID (`Routine["HL7"` matched a `CONTROL` process) — harmless, stack stayed healthy, but worth saying rather than quietly not mentioning.

## Not addressed here

Options 1, 2 and 4 from #101 — `mod_status`, the Apache worker/keep-alive settings, and a restart policy for a wedged web server. All three are still worth doing, and 2 touches shared compose/config rather than `iris/**`.

### DIFF
--- a/iris/labdemo/HL7Generator.cls
+++ b/iris/labdemo/HL7Generator.cls
@@ -62,6 +62,33 @@ ClassMethod Run(count As %Integer = 20, delaySecs As %Double = 2) As %Status
 ///
 /// Same pattern as the injected delay in PatientDemographicsOperation: a global read at runtime,
 /// so nothing is edited or recompiled mid-demo, and Triggers.Reset() clears it.
+///
+/// IT ALSO PAUSES ABOVE A QUEUE CAP, and that is a safety valve rather than a demo nicety (#101).
+/// `Cloud API` POSTs to the SAME embedded web server that serves the Management Portal, the
+/// /api/monitor/ endpoints the metrics proxy polls, and the agent dispatcher -- so this generator
+/// is, transitively, a load generator against the instance the whole stack depends on. On
+/// 2026-08-19 a long armed run reached ~2,650 queued and that web server stopped serving
+/// ENTIRELY: connections accepted, nothing answered, for a static file as well as a CSP page. The
+/// Portal, the metrics endpoint and the dispatcher all went down together and it took a
+/// `compose down` to recover.
+///
+/// The cause is not proven (#101 ranks the options; mod_status is not enabled, so worker/keep-alive
+/// exhaustion fits the evidence but is inference). The cap does not depend on the diagnosis being
+/// right: no demo needs a 2,600-message backlog, and the scenario's SHAPE -- a queue building, then
+/// draining once the pool grows -- is unaffected by bounding it. A few hundred messages
+/// demonstrates a bottleneck exactly as well as a few thousand.
+///
+/// PAUSES, DOES NOT STOP. Stopping would end the demo: inflow has to resume once the fix drains
+/// the queue, or the recovery shows a drained queue with no traffic behind it, which looks like the
+/// production died rather than recovered. So above the cap it waits and re-checks, and inflow
+/// returns on its own as soon as the queue falls back under.
+/// The config item whose queue the cap watches. Named here rather than repeated as a literal --
+/// root CLAUDE.md §5 keeps the host list in Production.cls, and a copied name is what went stale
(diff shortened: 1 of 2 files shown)


Ari-Glikman:
**Coordination note before anyone reviews this: @kskubach claimed option 3 on #101 about 20 minutes after I pushed this branch.** We collided — my fault for not saying here that I was starting it. Details and options on [#101](https://github.com/InterSystems-Israel/ProductionGuardian/issues/101#issuecomment-5344271511); closing this in favour of their version is a perfectly good outcome and I have no attachment to it.

Leaving it open only because it is finished and verified, so it is cheap to review or to cherry-pick from. `iris/**` is their area and the decision is theirs.

One open question I would like an opinion on regardless of whose implementation lands: **the 300s bound on the pause itself.** The reasoning is that a permanently stuck queue should not silence the generator forever, because a wedged production showing no traffic looks like a dead demo rather than a stuck one. But 300s is a guess, not a measurement.

Ari-Glikman:
Closing in favour of #102 (@kskubach), which is the better implementation — the cap is checked before producing rather than after, an unreadable queue depth degrades to producing instead of throwing, and the default of 400 is tied to the rule's absolute floor rather than picked. Reasoning in my review there. The only piece of this worth carrying over is the bound on the pause itself, which I offered as a follow-up and flagged as a guess rather than a finding.

Issue #104, opened 2026-08-19
The engine runs as %All, so the authorization policy cannot deny anything at runtime (post-MVP)


Ari-Glikman:
Investigated on 2026-08-19 and **deliberately not fixed** — the demo keeps `superuser` on the user's call, and the reasoning below is why that is the right trade for now rather than a shrug.

## The defect

`docker-compose.yml` gives both the proxy and the engine `IRIS_USER=superuser`, which holds `%All`. Under `%All`, `$SYSTEM.Security.Check` returns 1 for **every** resource — measured, including resources that do not exist:

```
as irisowner roles=%All
  Check PG_Resolve:USE      = 1
  Check PG_NonexistentThing = 1
```

So `Tools.AuthPolicy` is real, correct, tested code sitting on a path the running system never exercises. Every audit row reads `actor: SuperUser`, and a reviewer cannot distinguish an authorized write from an unauthorized one because nothing *can* be unauthorized. MVP 2 §2.2's "AI Detective can look without permission to act" is currently demonstrated only by `iris/test/GovernanceProof.cls`, which manufactures a throwaway principal precisely because the real one is over-privileged.

`resolve-api.md` §13.1 frames this as the *attribution* question (service account vs pass-through). Attribution is the visible symptom; the enforcement gap is the substantive half, and §13.1 does not name it.

## What I built and reverted

Two least-privilege service accounts in `Setup.AIHub`, created at boot from `PG_SERVICE_PASSWORD` with **no default value** (a default would be a known credential holding `Guardian_Resolve` on every deployment that forgot to set it):

| Account | Roles | Rationale |
|---|---|---|
| `pg_engine` | `Guardian_Resolve` + `%DB_%DEFAULT` | may apply the governed action |
| `pg_proxy` | `%DB_%DEFAULT` only | never calls a tool; polls `/api/monitor/*` |

**It worked, and the enforcement gap genuinely closed.** With `pg_engine` authenticating:

```
audit: actor=pg_engine  role=%DB_%DEFAULT,Guardian_Resolve  requestedBy=presenter@laptop
```

And a principal holding `Guardian_Read` only got the real refusal, audited against its own identity:

```
outcome: refused | reason: not_authorized | actor: pg_readonly | disposition: denied
message: 'SetPoolSize' modifies a live production and requires PG_Resolve:USE
```

[...]

Issue #100, opened 2026-08-19
resolve-api.md hands the caller an undo request it will refuse: reversal is size 1, the bound is 2..8


kskubach:
Surfaced while reviewing #99, which made it visible rather than caused it. Not a blocker for the demo — `Triggers.Reset()` restores the pool — but MVP 2 §4 lists **reversible** as a boundary condition of the safety model, and it is currently reversible only through a path that is not the documented one.

## The contract contradicts itself

`resolve-api.md` §4, on `reversal`:

```json
"reversal": {
  "action": { "type": "set_pool_size", "host": "Cloud API", "size": 1 },
  "capturedFrom": "live"
}
```

> `reversal.action` is then a fully formed request body … **Dev C can POST it back verbatim to undo.**

But §3 bounds the endpoint at `2..8`, and §5's refusal table is explicit:

> | `out_of_bounds` | `action.size` is not an integer in `2`–`8` |

So POSTing that body back verbatim is refused. **The endpoint hands the caller an undo request it will then reject.** This is internal to `resolve-api.md` and holds regardless of implementation.

## And the tool disagrees with its own ratified range

`mcp-tools.md` §3.6 ratifies an asymmetry, deliberately:

| Layer | Range | Why |
|---|---|---|
| `POST /api/resolve` | `2..8` | An operator-facing action. `1` is the shipped value, so approving it is a no-op dressed as a fix. |
| `set_pool_size` | `1..8` | Must be able to express the **reversal**. `previousSize` is `1`, so a tool that refuses `1` cannot undo its own first call. |

`iris/labdemo/Tools/Resolve.cls` ships `#MINSIZE = 2`, and its comment rejects that reasoning by name:

> The lower bound is 2, not 1, and that is not an off-by-one. `1` is the SHIPPED value, so "set the pool to 1" is a no-op dressed as a fix — it would report success, change nothing, and leave an operator believing the problem was addressed. Reversal to `1` is `Reset()`'s job, through a path that is not LLM-callable.

Both arguments are sound, which is why this wants a decision rather than a patch:

- **The contract's case:** a governed tool that cannot undo its own first call has a one-way safety model, and "reversible" is then a property of a separate debug utility rather than of the action.
- **The code's case:** `1` *is* the shipped value on this production, so an LLM-callable `set_pool_size(1)` is a success report with no effect — and `Reset()` deliberately sits outside what a model can reach.

[...]

PR #105, opened 2026-08-19
docs(iris): two acceptance rows had gone stale, one of them in my favour

kskubach:
Prompted by #104. Docs only, no code.

## The row that overclaimed — mine

`iris/CLAUDE.md` said:

> | An unauthorized role is refused | met — observed, `iris/test/GovernanceProof.cls` |

That reads as though the demo shows the boundary refusing a live request. It does not, and #104 is right that the enforcement gap is the substantive half of §13.1 rather than the attribution question. Verified independently before changing the claim:

```
roles: %All
  Check PG_Resolve:USE       = 1
  Check PG_TotallyMadeUp:USE = 1
```

`%All` satisfies a resource that **does not exist**. So `Tools.AuthPolicy` is correct code on a path the deployed stack never exercises, and the observed denial in `GovernanceProof` is real but works by manufacturing a throwaway low-privilege principal *because* the deployed one is over-privileged.

Now reads "met **by the fixture, not by the running system**", with the measurement and the reasoning in a section below it. Root `CLAUDE.md` §6 is explicit that a green claim over a red build costs the team more than the bug did, and this was a green claim over an unexercised code path — written by me, in a table I added.

## The row that underclaimed — also mine

> | `set_pool_size` changes the live pool and is reversible | **dry-run verified only** |

True when written, stale since #102, which applied a real `1 → 4` against a queue held at its cap and watched it drain to 0. Corrected, with the #100 caveat kept explicit: the reversal is *recorded* correctly and cannot be *executed* through the API.

Worth noting both rows were wrong in opposite directions, which is the argument for the table existing at all — neither would have been visible from the code.

## Recorded where someone will look

@Ari-Glikman's finding that **no `%Ens_*` resource or role exists on this instance** deserves to be somewhere other than a closed investigation. Verified: 0 rows in both `Security.Resources` and `Security.Roles`. It means any future least-privilege work on interop operations is blocked until `EnableNamespace` has created them, and it explains why the service accounts in #104 had to be reverted — the end state was a policy that authorizes the write and a write that cannot land, which is worse than either extreme.

So the section says *do not rebuild those service accounts before reading #104*. That investigation cost real time and its most valuable output is the reason it was abandoned.

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -143,8 +143,8 @@ returns PHI, it lands in `Audit.Entry.Result` in plain text.
 | Criterion | State |
 |---|---|
 | Read tools return real evidence | met — six tools discovered, live values |
-| `set_pool_size` changes the live pool and is reversible | **dry-run verified only**; `before`/`after`/`reversal` correct, an apply has not been run against a live queue |
-| An unauthorized role is refused | met — observed, `iris/test/GovernanceProof.cls`. **Compile `test/` first — first boot does not** (see below) |
+| `set_pool_size` changes the live pool and is reversible | met — a real `1 → 4` against a queue held at its cap, drained to 0, `Reset()` restored pool 1 (#102). Reversal is *recorded* and correct; reversing **through the API** is blocked by #100 |
+| An unauthorized role is refused | met **by the fixture, not by the running system** — see below |
 | Every call appears in the audit log | met for executions and for authorization denials |
 | RBAC roles and resources exist from a clean boot | met, and **verified from a real boot** rather than inferred: the four security objects were deleted and `docker compose restart iris` recreated them through step 9. `Audit/` compiled on the way through, including its SQL table, so the recursive load picks up a new subdirectory |
 | A principal can actually *use* the roles | needs `%DB_%DEFAULT` **as well**, from the invocation path — the `Guardian_*` roles stay minimal and grant only `PG_*`. Without it you get `<PROTECT>` before any policy is consulted |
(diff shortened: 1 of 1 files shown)


kskubach:
Added `677e765` — the `compose down -v` note I promised in the #106 review. It goes here rather than in that PR because this branch is already editing the acceptance table, and the table is exactly where someone reads "met" and assumes it means "met from a clone."

It records the three defects that were invisible for this reason in one day — `iris/test/` (#97), `/labdemo/agent`, and the LLM provider config — and keeps @Ari-Glikman's framing verbatim, because it is better than mine:

> the state that matters is what a fresh boot produces, not what your instance contains

I checked the volume-name claim rather than repeating it: `docker-compose.yml:38` is `name: production-guardian`, so a clone into a new directory genuinely does reuse the existing volume.

That makes this PR three related corrections to the same section, all of the same kind — a claim that was true of my instance and not of a boot:

1. "an unauthorized role is refused: met" — met by the fixture, not the running system, which holds `%All`
2. "`set_pool_size` dry-run verified only" — stale since #102 applied a real 1→4
3. and now: every row means "on a boot", and only `compose down -v` tests one

Ari-Glikman (approved):
Approving. I reproduced both measurements rather than reading them, because the whole point of this PR is that a claim in a table went unchecked.

## Both reproduce exactly

```
roles: %All
  PG_Resolve:USE       = 1
  PG_TotallyMadeUp:USE = 1     <- a resource that does not exist
```

```
Ens resources: 0 | Ens roles: 0
```

The made-up resource returning `1` is the sharpest way to state it, and it is better than how I put it in #104 — I wrote that `%All` "passes every check", which is true and sounds like a permissions detail. *A resource that does not exist is satisfied* makes it obvious there is no check happening at all.

## The overclaim correction is the important half

`met — observed, GovernanceProof.cls` → `met **by the fixture, not by the running system**`. That distinction is exactly what I missed for two days: I kept citing the observed denial as evidence the boundary works, and it is evidence the *policy logic* works. The fixture has to manufacture a low-privilege principal precisely because the deployed one cannot be refused anything.

Worth saying you caught this in a table you wrote yourself, in your own area, and corrected it in the direction that makes your own work look weaker. That is the harder direction.

## Both rows wrong in opposite directions

Your note that one overclaimed and one underclaimed — and that neither would have been visible from the code — is the argument for the table, and I would go further: the underclaim was mine to fix and I did not. #102 landed a real `1 → 4` drain to 0 four hours before this PR and I never went back to the acceptance row. A table nobody revisits after landing the work is a table that decays in both directions.

The `#100` caveat kept explicit on that row is right: recorded correctly, not reversible *through the API*, and those are different claims.

## Recording why #104 was abandoned

Strongly agree, and thank you for putting the "do not rebuild the service accounts before reading #104" line where someone will hit it. The most valuable output of that investigation was the reason it stopped: the end state was a policy that authorizes the write and a write that cannot land, which is worse than either extreme. Without that written down the next person builds the same accounts, gets the same `#940`, and concludes the roles are wrong rather than that the privileges do not exist.

The blocking fact — `EnableNamespace` has never run here, so there is nothing to grant — now sits in the area file rather than in a closed issue thread. That is the right home.

Docs only, no code, both claims verified. `iris/**` is your area; approving as the second reviewer.

PR #106, opened 2026-08-19
fix(iris): a cold start had no /labdemo/agent web app and no LLM provider

Ari-Glikman:
Two independent cold-start gaps, both found by testing what a **colleague** gets rather than what my container has: cloned the repo, `compose down -v`, `compose up`, then drove the demo.

Neither was visible on any instance either of us had used, because we had each created the missing piece by hand while building it.

## Gap 1 — `/labdemo/agent` was never registered (`e6d430c`)

`FirstBoot.RegisterWebApps()` registered two applications and needed three. With no application mapped to `AgentDispatcher`, the **entire MVP 2 write path was dead**:

```
POST /api/resolve     -> failed / tool_call / "write tool returned HTTP 404"
POST /api/investigate -> "agent returned HTTP 503"
```

**404 through the engine, 403 direct** (@kskubach's measurement) — both mean the dispatcher was never reached, but a reader told to expect 404 who sees 403 goes hunting for an auth problem.

The method's own comment already described the consequence, written about `/labdemo/monitor`: *"leaves Cloud API POSTing into a 404 while EMR Source and Lab Router look healthy, so nothing points at the cause."* So the fix is one entry in the existing array.

## Gap 2 — the LLM provider was hand-made (`478876b`, `90d1cfa`)

The wallet entry and `AI.LLM.pgdetective` config were created interactively and lived only in a durable volume, so a fresh clone booted with `4. LLM provider AI.LLM.pgdetective: NOT CONFIGURED` and **AI Detective could not be demonstrated at all.** `ReportProvider()` reported it; nothing created it.

`CreateProvider()` now creates the `AI.Secrets` resource, the `PGSecrets` collection, the key entry, and the config referencing it as `secret://PGSecrets.openai#apikey` — from `PG_LLM_API_KEY`, no default. Absent, it skips and says so, and the engine degrades to a labelled canned investigation.

`90d1cfa` fixes two defects @kskubach found plus one that surfaced while testing them:

- **Delete-then-Create destroyed working config.** It ran on every boot, so one transient failure turned a *stale* key into *no* key. Previous value now captured and restored on failure, with the restore reported.
- **The rotation story contradicted itself** — a comment forbidding overwrite next to code that overwrote. Resolved in favour of the env var as single source of truth; the comment now says *do not rotate by editing the wallet.*
- **`Configuration.Get` takes `(fqn, .config)`**, not `Create`'s four positional args. Compiled cleanly, threw `<PARAMETER>`, and the `catch` reported `LLM provider: FAILED` on a boot where the config *had* been written.

## Verified — cold clone, empty volume, no manual step

```
4. web app /labdemo: created
4. web app /labdemo/agent: created
4. web app /labdemo/monitor: created

POST /labdemo/agent/resolve dry_run -> previewed 1 -> 4, audit pg-audit-1
queue_buildup at 157 -> POST /api/resolve apply -> applied 1->4, audit pg-audit-2
queue 76 -> 38 -> 6 -> 0, findings (all clear)
```

`pg-audit-1` as the first handle is the evidence this is a fresh instance, not my long-lived one.

**Provider provisioning, three consecutive runs rotating the key:**

```
key A (19 chars) -> resource/wallet/entry/config created, resolved len 19
key B (29 chars) -> "wallet: exists", entry rewritten, resolved len 29
key C (37 chars) + PG_LLM_MODEL=gpt-4o -> "config created, model = gpt-4o", resolved len 37
```

**The key does not leak:**

```
UNRESOLVED api_key = secret://PGSecrets.openai#apikey
RESOLVED length    = 37
occurrences of any test key in the container log: 0
```

## Status of the credential path — CORRECTED

An earlier version of this body said the provider provisioning was "worth its own change … rather
than smuggling it in here", which was true when written and false once `478876b` landed in this PR.
@kskubach flagged that the first thing a reviewer reads should not disclaim the riskiest change in
the diff. It is in scope here, it is reviewed, and the evidence is below and in the comments.

## Not verified when this PR opened — now verified

**No cold-start transcript with `source` other than `canned`.** Verifying the cold-start path needed `compose down -v`, which destroyed the volume holding the only copy of the real key — so the mechanism is proven with dummy keys of the right shape and the **live agent round trip on this code path is not.**

The acceptance check is one live investigation returning `source: "agent"` with non-null `model` and `toolCalls`. The README documents that and the failure mode (`state: "unavailable"`, `note: "agent returned HTTP 500"`, `rootCause` null — never a fabricated narrative).

The key needs rotating anyway, so this costs one verification run once a new one exists. Alternative on the table: hold `478876b`/`90d1cfa` and merge only the web-app fix, which is independently verified and unblocks the write path on its own.

## Also in this PR

`0ca9fa5` — README: the MVP 2 demo section. Every command in it was run against the cold clone rather than transcribed. The load-bearing line is that **`AGENT_MODE` defaults to `mock`, and the mock resolve reports `applied` without changing the production** — someone seeing "applied" and no drain would look for a bug rather than for the mode.

## Verified with a real key after the review

The acceptance test @kskubach asked for, on a cold-provisioned instance:

```
4b. wallet entry PGSecrets.openai: stored (167 chars, not logged)
4. LLM provider AI.LLM.pgdetective: configured, model = gpt-4o-mini

POST /api/investigate -> state complete | source AGENT | gpt-4o-mini | 5 tool calls
POST /api/resolve apply -> applied 1 -> 4, audit pg-audit-17
queue 147 -> 109 -> 85 -> 48 -> 14 -> 0
```

And the wallet restore, which was inert until `4a0fb8e` because `%Wallet.KeyValue.Secret` is
`transient=1`:

```
captured length: 180        (via GetSecretValue; .Secret reads 0)
guard would fire: YES
restored length: 180 | identical: YES
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Production Guardian — Health Scan (MVP 1)
+# Production Guardian — Health Scan (MVP 1) + Early Warning, AI Detective, Smart Resolve (MVP 2)
 
 An AI-powered production health layer for InterSystems Health Connect. MVP 1 is **Health
 Scan**: it reads production metrics from IRIS, compares them to a rolling baseline, and
@@ -156,6 +156,118 @@ still waiting for. Full table in `iris/labdemo/README.md`.
 alert from the proxy. Read `http://localhost:3001/proxy/alerts` instead.
 
 The Management Portal is at **http://localhost:52773/csp/sys/UtilHome.csp**.
+
+## The MVP 2 demo — WHAT, WHY, FIX
+
+One scenario, end to end: a queue builds on `Cloud API`, an AI agent explains why, and one
+governed, approved, reversible action clears it.
+
+### It runs without a key, and says so
+
+`AGENT_MODE` defaults to `mock`, so a fresh clone demonstrates the **shape** of the loop with no
+credential: the investigation is a canned narrative built from the live measured values, and the
+resolve is applied against an in-memory pool size. Both are labelled — `source: "canned"`, and the
+audit id reads `mock-audit-N`.
+
+**The mock resolve reports `applied` and does not change the production.** That is deliberate and it
+is the one thing to know before demonstrating from a clone: the pool stays at 1 and the queue keeps
+growing. If the queue does not drain after an apply, check which mode you are in before looking for
+a bug.
+
+### For a live agent and a real write
+
+```bash
+export PG_LLM_API_KEY=sk-...          # OpenAI key; goes into the AI Hub wallet, never the repo
+export PG_AGENT_MODE=live             # engine calls IRIS instead of its own mock
+docker compose up -d
+```
+
+`PG_LLM_API_KEY` is read at first boot by `Setup.AIHub`, which creates the wallet entry and the
+`AI.LLM.pgdetective` config that references it as `secret://PGSecrets.openai#apikey`. The key is
+never written to the configuration, the repo, or a log line — the boot prints its length and that it
+was stored. Rotating it is: change the variable, restart. `PG_LLM_MODEL` overrides the model
+(default `gpt-4o-mini`).
+
+Without the key the boot says so plainly and the loop still runs, canned:
+
+```
+4b. LLM provider: SKIPPED -- PG_LLM_API_KEY is not set
+    AI Detective will serve a CANNED investigation (source "canned"), which is
+    labelled and correct but is not a live agent. Set the variable to enable it.
+```
+
+### Driving it
+
+```bash
+docker compose exec iris iris session IRIS -U LABDEMO
+```
+
+```objectscript
+// Arm: throttles the downstream to ~1s/message and raises inflow, after warming the baseline
+// at zero. Takes about 2 minutes to produce a confirmed finding.
+do ##class(ProductionGuardian.LabDemo.Triggers).PoolBottleneck()
+do ##class(ProductionGuardian.LabDemo.Triggers).Status()   // what is armed, and the queue cap
+do ##class(ProductionGuardian.LabDemo.Triggers).Reset()    // undo, including the pool size
+```
+
+Then, against the dashboard's own API path so it is the same route the UI uses:
+
+```bash
+# WHAT -- the finding
+curl -s localhost:5173/api/healthscan/findings
+
+# WHY -- root cause, evidence, a recommended action
+curl -s -XPOST localhost:5173/api/investigate \
+  -H 'Content-Type: application/json' -d '{"findingId":"<id from above>"}'
+
+# FIX -- preview first, then apply
+curl -s -XPOST localhost:5173/api/resolve -H 'Content-Type: application/json' \
+  -d '{"mode":"dry_run","action":{"type":"set_pool_size","host":"Cloud API","size":4}}'
+
+curl -s -XPOST localhost:5173/api/resolve -H 'Content-Type: application/json' \
+  -d '{"mode":"apply","requestedBy":"you@laptop",
+       "action":{"type":"set_pool_size","host":"Cloud API","size":4},
+       "origin":{"findingId":"<id>"}}'
+```
+
+A live run looks like this — the queue drains because four workers clear ~4/sec against ~2/sec
+inbound:
+
+```
+queue_buildup at 147  ->  investigate: gpt-4o-mini, 5 tool calls, recommends set_pool_size 4
+apply -> applied 1 -> 4, audit pg-audit-2
+queue 80 -> 42 -> 14 -> 0,  findings (all clear)
+```
+
+**If `investigate` returns `state: "unavailable"` with `note: "agent returned HTTP 500"`, the key is
+wrong or expired.** That is the honest failure and worth recognising: `rootCause` is `null` and
+`evidence` is empty rather than a plausible narrative, because a fabricated root cause is worse than
+none when a human approves a production write on the strength of it. The FIX half is unaffected — it
+does not use the LLM — so a bad key costs you WHY and nothing else.
+
+### What the safety model actually enforces
+
+- **One action, one host, bounded.** `set_pool_size` on `Cloud API`, size `2`–`8`. Anything else is
+  refused by the tool with a named reason — including size `1`, because `1` is the shipped value and
+  "setting it to 1" is a no-op dressed as a fix.
+- **Dry run first, and it cannot mutate** — that path returns before the write.
+- **Reversible.** Every response carries the prior value in `reversal`, captured live. Undo with
+  `Triggers.Reset()`.
+- **RBAC-gated and audited.** Every tool call, read and write, becomes a row in
+  `ProductionGuardian.LabDemo.Audit.Entry` naming the actor. An unauthorized caller is refused with
+  `200` and `refusal.reason: "not_authorized"`, and *the refusal is audited too*.
+- **Metrics and configuration only ever leave the instance.** Never message content, never PHI.
+
+```objectscript
+do ##class(ProductionGuardian.Setup.AIHub).Status()          // roles, tools, provider
+do ##class(ProductionGuardian.LabDemo.Audit.Entry).Purge()   // reset the log between rehearsals
+```
+
+**One caveat, stated rather than buried:** the engine authenticates to IRIS as `superuser`, which
+holds `%All`, and under `%All` a resource check passes for everything — so the RBAC boundary is
+proven by `iris/test/GovernanceProof.cls` with a purpose-made unprivileged principal rather than by
+the running demo. See issue #104 for why that is not fixed yet (this image has no `%Ens_*`
+privileges to grant).
 
 ## Running without IRIS
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -89,6 +89,16 @@ services:
       # ISC_DATA_DIRECTORY makes the namespace survive `compose down`; without it the
       # LABDEMO setup would re-run from scratch every time.
       ISC_DATA_DIRECTORY: /iris-shared/durable
+      # Setup.AIHub.CreateProvider() reads these at first boot to create the AI Hub wallet entry
+      # and the AI.LLM.pgdetective config. Declared HERE and not only in the shell: an exported
+      # variable does not reach a container unless compose passes it, and without this line the
+      # boot reported "SKIPPED -- PG_LLM_API_KEY is not set" while the variable was plainly set
+      # in the caller's environment. Measured -- the README's own instructions failed on it.
+      #
+      # No default. Absent, the provider is not created, the boot says so, and AI Detective
+      # degrades to a labelled canned investigation. A default here would be a committed key.
+      PG_LLM_API_KEY: ${PG_LLM_API_KEY:-}
+      PG_LLM_MODEL: ${PG_LLM_MODEL:-}
     depends_on:
       iris-init:
         condition: service_completed_successfully
--- a/iris/setup/AIHub.cls
+++ b/iris/setup/AIHub.cls
@@ -220,9 +220,19 @@ ClassMethod CreateProvider() As %Status
             write "4b. resource ", ..#SECRETRESOURCE, ": ", $select(sc: "created", 1: "FAILED " _ $system.Status.GetErrorText(sc)), !
         }
 
-        // Collection, then entry. Both idempotent by inspection rather than by relying on Create
-        // returning a benign error for an existing name -- an existing wallet must not be
-        // overwritten, because it may hold a key an operator rotated by hand.
+        // Collection, then entry, both idempotent by inspection rather than by relying on Create
+        // returning a benign error for an existing name.
+        //
+        // THE ENVIRONMENT VARIABLE IS THE SINGLE SOURCE OF TRUTH for the key, so the ENTRY below is
+        // rewritten on every boot when PG_LLM_API_KEY is set. An earlier version of this comment
+        // argued the opposite -- that an existing wallet must not be overwritten because it might
+        // hold a key rotated by hand -- while the code overwrote it anyway (@kskubach, #106). Both
+        // positions are defensible and holding both is a footgun: an operator who rotates in the
+        // wallet gets silently reverted on the next restart and diagnoses it as a rejected key.
+        //
+        // Rotation is therefore: change the variable, restart. DO NOT rotate by editing the wallet;
+        // it will be overwritten. The COLLECTION is left alone once it exists, because it carries
+        // only the resource bindings and nothing an operator would tune.
         if '##class(%Wallet.Collection).Exists(..#WALLET) {
             set sc = ##class(%Wallet.Collection).Create(..#WALLET, {"UseResource": (..#SECRETRESOURCE), "EditResource": (..#SECRETRESOURCE)})
             write "4b. wallet ", ..#WALLET, ": ", $select(sc: "created", 1: "FAILED " _ $system.Status.GetErrorText(sc)), !
@@ -236,9 +246,31 @@ ClassMethod CreateProvider() As %Status
         // happen. Never printed -- only whether it was stored.
         set sc = ##class(%Wallet.KeyValue).Create(..#WALLETENTRY, {"Usage": "CUSTOM", "Secret": {"apikey": (key)}})
         if $$$ISERR(sc) {
-            // Create fails for an existing entry; replace it rather than leaving a stale key.
+            // ALREADY PRESENT. Rewrite it -- but NOT by deleting first (@kskubach, #106).
+            //
+            // `Delete` then `Create` turns a *stale* key into *no* key when the second call fails,
+            // and this runs on every boot from FirstBoot, so one transient failure takes out a demo
+            // that was working. The old value is captured first and restored if the rewrite fails,
+            // so the worst case is the key the instance already had.
+            set previous = ""
+            try {
+                set existing = ##class(%Wallet.KeyValue).%OpenId(..#WALLETENTRY)
+                if $isobject(existing) {
+                    set previous = existing.Secret
+                }
+            } catch {
+                set previous = ""
+            }
             do ##class(%Wallet.KeyValue).Delete(..#WALLETENTRY)
             set sc = ##class(%Wallet.KeyValue).Create(..#WALLETENTRY, {"Usage": "CUSTOM", "Secret": {"apikey": (key)}})
+            if $$$ISERR(sc), previous '= "" {
+                // Put back what was there. Reported, because a boot that silently reverted to an
+                // older key would present as the provider rejecting a good one.
+                do ##class(%Wallet.KeyValue).Delete(..#WALLETENTRY)
+                set restore = ##class(%Wallet.KeyValue).Create(..#WALLETENTRY, {"Usage": "CUSTOM", "Secret": (previous)})
+                write "4b. wallet entry rewrite FAILED; previous value ",
+                    $select(restore: "restored", 1: "COULD NOT BE RESTORED"), !
+            }
         }
         write "4b. wallet entry ", ..#WALLETENTRY, ": ", $select(sc: "stored (" _ $length(key) _ " chars, not logged)", 1: "FAILED " _ $system.Status.GetErrorText(sc)), !
         if $$$ISERR(sc) {
@@ -249,9 +281,24 @@ ClassMethod CreateProvider() As %Status
         set cfg = {"model_provider": (..#PROVIDER), "model": (model), "api_key": ("secret://" _ ..#WALLETENTRY _ "#apikey")}
         set sc = ##class(%ConfigStore.Configuration).Create("AI", "LLM", "", "pgdetective", cfg)
         if $$$ISERR(sc) {
-            // Already present: replace, so a model change in the environment takes effect.
+            // Already present: replace, so a PG_LLM_MODEL change takes effect. Same restore-on-
+            // failure discipline as the wallet entry above -- this config is what BuildAgent reads,
+            // so losing it disables AI Detective entirely.
+            // Get(fqn, .config) -- a DOTTED name and an output, not the four positional arguments
+            // Create() takes. Passing Create's shape throws <PARAMETER> at runtime while compiling
+            // cleanly, and the catch turned that into "LLM provider: FAILED" on a boot where the
+            // config had in fact been written. Read from the compiled FormalSpec rather than
+            // assumed symmetrical with Create.
+            set previousCfg = ""
+            do ##class(%ConfigStore.Configuration).Get("AI.LLM.pgdetective", .previousCfg)
             do ##class(%ConfigStore.Configuration).Delete("AI.LLM.pgdetective")
             set sc = ##class(%ConfigStore.Configuration).Create("AI", "LLM", "", "pgdetective", cfg)
+            if $$$ISERR(sc), $isobject(previousCfg) {
+                do ##class(%ConfigStore.Configuration).Delete("AI.LLM.pgdetective")
+                set restoreCfg = ##class(%ConfigStore.Configuration).Create("AI", "LLM", "", "pgdetective", previousCfg)
+                write "4b. config rewrite FAILED; previous value ",
+                    $select(restoreCfg: "restored", 1: "COULD NOT BE RESTORED"), !
+            }
         }
         write "4b. config AI.LLM.pgdetective: ", $select(sc: "created, model = " _ model, 1: "FAILED " _ $system.Status.GetErrorText(sc)), !
     } catch ex {
--- a/iris/setup/AIHub.cls
+++ b/iris/setup/AIHub.cls
@@ -252,12 +252,15 @@ ClassMethod CreateProvider() As %Status
             // and this runs on every boot from FirstBoot, so one transient failure takes out a demo
             // that was working. The old value is captured first and restored if the rewrite fails,
             // so the worst case is the key the instance already had.
+            // GetSecretValue(name), NOT %OpenId(...).Secret. `Secret` is transient=1, so a
+            // reopened entry reports it as "" -- verified: GetSecretValue returns 180 chars for the
+            // same entry whose .Secret reads 0. The earlier version captured .Secret, so `previous`
+            // was always empty, the restore guard below could never fire, and the comment claiming
+            // the old value was preserved was false (@kskubach, #106). An inert safety net is worse
+            // than none, because it stops anyone looking.
             set previous = ""
             try {
-                set existing = ##class(%Wallet.KeyValue).%OpenId(..#WALLETENTRY)
-                if $isobject(existing) {
-                    set previous = existing.Secret
-                }
+                set previous = ##class(%Wallet.KeyValue).GetSecretValue(..#WALLETENTRY)
             } catch {
                 set previous = ""
             }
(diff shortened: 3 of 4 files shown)


Ari-Glikman:
**Second commit added (`478876b`): the LLM provider now provisions itself too.** Retitled the PR, because the original scope was one of two cold-start gaps and I found the other while writing up what was still missing.

I had listed this in the PR body as "worth its own change". It is small enough and the same class of defect, so folding it in beat opening a third PR against the same file.

## What it does

`CreateProvider()` creates the `AI.Secrets` resource, the `PGSecrets` wallet collection, the key entry, and the `AI.LLM.pgdetective` ConfigStore config referencing it as `secret://PGSecrets.openai#apikey` — from `PG_LLM_API_KEY`, with no default.

**The key goes in the wallet, not the config**, so it lives in one guarded place and never appears in the configuration, the repo, or a log line. The boot prints the *length* and that it was stored, never the value.

Rotation is "change the env var and restart" — the entry is rewritten on every boot when the variable is set. Deliberate: a rotation that needs a documented interactive procedure is one that does not happen.

## Verified on the cold clone, empty volume

```
no key:
  4b. LLM provider: SKIPPED -- PG_LLM_API_KEY is not set
      AI Detective will serve a CANNED investigation (source "canned") ...

key set:
  4b. resource AI.Secrets: created
  4b. wallet PGSecrets: created
  4b. wallet entry PGSecrets.openai: stored (34 chars, not logged)
  4b. config AI.LLM.pgdetective: created, model = gpt-4o-mini

GetDetails(..., resolveSecrets=1) -> resolved ok: YES, 34 chars, model gpt-4o-mini
```

That last line is the one that matters: the `secret://` URI resolves **to the key** rather than handing back the URI string, which is the failure mode this shape exists to avoid.

## One thing NOT verified, stated plainly

**I have not run this against a real key end to end.** Testing the cold-start path needed `compose down -v`, and that destroyed the volume holding the only copy of the working provider config — so the live-agent round trip is unverified on this code path. The mechanism is proven with a dummy key of the same shape and length, and the resolution step is the part that could plausibly break.

Two consequences worth being explicit about:

1. Someone should re-run one live investigation after this lands, with a real key, before relying on it for a demo. `source: "agent"` with a non-null `model` and `toolCalls` is the check.
2. The key needed rotating anyway (it was pasted into a chat transcript), so losing the stored copy costs nothing beyond that verification.

## Third instance of the same pattern

`iris/test/` (#97), `/labdemo/agent` (#106), and now the provider — each existed on every machine we tested because we each created it by hand while building it, and each was absent from a clone. That is the whole of #84's argument arriving by a different route: **the state that matters is what a fresh boot produces, not what your instance contains.** The cheapest defence I have found is `compose down -v` before believing anything about setup, and I would not have found either of these without it.

@kskubach — `iris/**` is your area and this is the second commit on the same PR, so re-review rather than a fresh look. Both are additive to `FirstBoot`/`AIHub` with no behaviour change when the env vars are unset.

kskubach (changes requested):
The web-app fix is right and I've reproduced it independently. The second commit needs evidence before I can approve, and has two small defects — details below.

## `/labdemo/agent` — confirmed on a second machine

I never registered it by hand either, so my container shows the same gap:

```
/labdemo          -> PatientDispatcher      authe=96
/labdemo/monitor  -> HostStatusDispatcher   authe=96
/labdemo/agent exists: 0
```

**"Test what a colleague gets rather than what my container has"** is the right instinct and it is the third find today from that same class of problem. The detail I'd single out is the second way it hid from you: a fresh clone picking up the existing named volume so the boot reports `exists` for everything. That is worth its own line in `iris/CLAUDE.md` eventually — `compose down -v` is the only thing that tests a cold start, and nothing says so.

**One diagnostic wrinkle.** You measured `HTTP 404` through the engine; a direct unauthenticated `POST` to `:52773/labdemo/agent/resolve` gives me **403**. Both mean the dispatcher was never reached, but a reader told to expect 404 who sees 403 will go looking for an auth problem. Worth "404 through the engine, 403 direct" in the commit message.

## The second commit is not in the PR body

`478876b` provisions the wallet and ConfigStore from `PG_LLM_API_KEY`. The body's closing section says:

> ## Still missing for a cold-start demo, and it is not this PR
> … Worth its own change … rather than smuggling it in here.

You updated the title but not the body, so the body now tells a reviewer the riskiest change in the PR isn't in it — and the `Verified on a true cold start` block predates that commit, so **there is no stated evidence for the credential path at all.** For the one change here that writes a secret, that inverts the scrutiny it should get.

I'm not assuming it's broken. I checked the thing most likely to break it silently — whether a config created under `new $namespace` / `%SYS` resolves from LABDEMO, where `BuildAgent` reads it — and it does:

```
created in %SYS: OK
readable from %SYS   : yes model=dummy-model
readable from LABDEMO: YES model=dummy-model
```

So that concern is closed. What I'd still want is a cold-start transcript showing a real investigation with `source` other than `canned`, and confirmation the key never lands in a log.

## Two defects in `CreateProvider()`

**1. `Delete` before a `Create` that can fail destroys working config.** Both here:

```objectscript
do ##class(%Wallet.KeyValue).Delete(..#WALLETENTRY)
set sc = ##class(%Wallet.KeyValue).Create(..#WALLETENTRY, ...)
```

and in the ConfigStore branch. If the second `Create` fails, the entry is gone and the provider is dead — a boot that turns a *stale* key into *no* key, on a stack that was working before. Given this runs on every boot from `FirstBoot`, a transient failure takes out a working demo. Prefer create-then-swap, or at minimum re-create the previous value on failure.

**2. The rotation story contradicts itself.** The collection comment says:

> an existing wallet must not be overwritten, because it may hold a key an operator rotated by hand

and then the entry is overwritten on every boot when the variable is set. Both positions are defensible; together they are a footgun — an operator who rotates in the wallet gets silently reverted to the stale env value on the next restart, and diagnoses it as the provider rejecting a good key.

Pick one and say so. I'd pick yours — env var is the single source of truth, rotation is "change the variable and restart" — and delete the collection comment's reasoning, because it argues for the opposite.

## Minor

`stored (51 chars, not logged)` prints the key's length. Near-zero risk since per-provider key lengths are public, and I'd leave it — flagging only so the choice is deliberate rather than incidental.

## Suggestion

Split `478876b` out. The web-app fix is one array entry, independently verified, and unblocks cold-start demos for everyone right now — it shouldn't wait on a secret-handling review. Happy to approve the web-app commit alone the moment it stands by itself, and to review the credential change properly on its own PR where the body describes what it does.

kskubach (approved):
Approving, and correcting two things in my own review — one of which was simply wrong.

## I was wrong that there was no evidence

I wrote "there is no stated evidence for the credential path at all." There was: your comment nine minutes before my review lays out the cold-clone transcript, the dummy-key run, and the `resolveSecrets=1` resolution check. I read the PR body and not the comments, and then criticised you for information I hadn't looked at. My mistake, not a gap in your PR.

The body does still say "it is not this PR" — worth fixing since it's the first thing a reviewer reads, but that's a nit, not the substantive complaint I made it into.

And your disclosure is better than I credited:

> **I have not run this against a real key end to end.** Testing the cold-start path needed `compose down -v`, and that destroyed the volume holding the only copy of the working provider config

That is exactly the right thing to say out loud, with the reason and the consequence. Your point 1 — someone re-runs one live investigation with a real key before anyone relies on it, checking for `source: "agent"` with non-null `model` and `toolCalls` — is the acceptance criterion, and I'd rather it were an issue than a line in a merged PR comment. I'll file it if you'd prefer.

## I also overstated the `Delete`-then-`Create` finding

I called it "destroys working config" and "turns a stale key into no key, on a stack that was working." The severity is lower than that, for a reason your own design supplies: **the environment variable is the source of truth and is re-read on every boot.** So a failed re-`Create` leaves no wallet entry, the provider is unusable, the engine falls back to `canned` — which is labelled — and the next restart reconstructs it from `PG_LLM_API_KEY`. Recoverable by restart, no unrecoverable state, and the degradation announces itself.

It's still worth tightening at some point, since a boot that leaves the provider broken until someone restarts again is a poor failure mode on demo morning. But it is not blocking and I shouldn't have framed it as though it were.

## The one point I do stand behind

The rotation reasoning still contradicts itself. The collection comment says an existing wallet "must not be overwritten, because it may hold a key an operator rotated by hand", while the entry is deliberately overwritten every boot. An operator who rotates in the wallet gets silently reverted on restart and will diagnose it as the provider rejecting a good key.

Your model is the right one — env var is authoritative, rotation is change-and-restart, and the README now says so clearly. The stale comment is the only thing arguing otherwise, and it's a comment change, not a code change. Fine as a follow-up.

## On the README

Good addition, and the warning I'd have asked for is already the one you led with:

> **The mock resolve reports `applied` and does not change the production.** … If the queue does not drain after an apply, check which mode you are in before looking for a bug.

That is the exact confusion the mock is otherwise guaranteed to cause, and putting the diagnostic instruction next to it is what makes it useful rather than merely accurate.

## Third instance of the pattern — agreed, and it's the durable finding here

`iris/test/` (#97), `/labdemo/agent`, and the provider. Your framing is right and better than #84's:

> the state that matters is what a fresh boot produces, not what your instance contains

`compose down -v` before believing anything about setup belongs in `iris/CLAUDE.md` next to the acceptance table, since that table is exactly where someone will read "met" and assume it means "met from a clone." I'll add it in #105, which is already touching that section — that keeps this PR to the fix.

Nothing else from me. Both commits are additive with no behaviour change when the variables are unset.

Ari-Glikman:
Both defects fixed in `90d1cfa`, and testing the fix exposed a third. You were also right that the PR body told a reviewer the riskiest change was not in the PR — I have rewritten it.

## 1. Delete-then-Create — fixed, and it was the serious one

You are right that this is worse than untidy: it runs on **every** boot from `FirstBoot`, so a single transient failure converts a *stale* key into *no* key on a stack that was working. The previous value is now captured first and restored if the rewrite fails, and the restore is **reported** — a boot that silently reverted to an older key would present as the provider rejecting a good one, which is a worse debugging session than an outright failure.

Same treatment for the ConfigStore branch, since that config is what `BuildAgent` reads.

## 2. The rotation contradiction — taking your resolution

Env var is the single source of truth. The collection comment's opposing reasoning is deleted, and replaced with the explicit instruction: **do not rotate by editing the wallet, it will be overwritten.** The collection itself is left alone once created — it carries only resource bindings, nothing an operator would tune.

Holding both positions was exactly the footgun you describe, and it is the same defect class as the `%`-prefix paragraph in #93 and my own `AgentDispatcher` claim in #92: prose that argues confidently for something the code does not do.

## 3. Found while testing the fix

`%ConfigStore.Configuration.Get` takes **`(fqn, .config)`** — a dotted name and an output — not the four positional arguments `Create` takes. I assumed it was symmetrical with `Create`. It compiled cleanly and threw `<PARAMETER>` at runtime, which my `catch` reported as `4b. LLM provider: FAILED` **on a boot where the config had in fact been written**. Read off the compiled `FormalSpec` this time:

```
FormalSpec: fqn:%String,*config:%ConfigStore.Configuration
```

Worth noting because it is the second time today a `catch` turned a working operation into a reported failure — the audit policy's swallow is deliberate and this one was not.

## Evidence you asked for

**Three consecutive runs, rotating the key each time:**

```
key A (19 chars)  -> resource/wallet/entry/config created, resolved len 19
key B (29 chars)  -> "wallet PGSecrets: exists", entry rewritten, resolved len 29
key C (37 chars) + PG_LLM_MODEL=gpt-4o
                  -> entry rewritten, "config created, model = gpt-4o", resolved len 37
```

Rotation works through the variable, the model override takes effect, and no run left the provider worse than it found it.

**The key does not leak:**

```
UNRESOLVED api_key = secret://PGSecrets.openai#apikey
RESOLVED length    = 37
occurrences of any test key in the container log: 0
```

The configuration holds the URI; only `GetDetails(..., resolveSecrets=1)` yields the value.

**Thank you for closing the namespace question independently** — `created in %SYS` / `readable from LABDEMO` was the thing most likely to break this silently, and it is where `BuildAgent` reads from. I would have tested it next; you saved a cycle.

## What I still cannot show, and it is the one thing you asked for

**A cold-start transcript with `source` other than `canned`.** I do not have a working key: verifying the cold-start path required `compose down -v`, and that destroyed the volume holding the only copy of the real credential. Every run above uses dummy keys of the right shape, so the *mechanism* is proven end to end and the *live agent round trip on this code path* is not.

I am not going to claim otherwise or quietly leave it out of the body. Two options and I do not mind which:

1. **Merge on the mechanism**, and whoever next has a key runs one live investigation as the acceptance check — `source: "agent"` with non-null `model` and `toolCalls`. The README now names that check and the failure mode (`state: "unavailable"`, `note: "agent returned HTTP 500"`, `rootCause` null).
2. **Hold the second and third commits** until a key is available, and merge only the `/labdemo/agent` web-app fix, which is independently verified and unblocks the write path on its own.

The key needs rotating regardless — it was pasted into a chat transcript — so option 1 costs one verification run after the new key exists.

## On your diagnostic wrinkle

Taking it: **404 through the engine, 403 direct.** Both mean the dispatcher was never reached, and a reader told to expect 404 who sees 403 goes hunting for an auth problem. It is in the body now.

And agreed on `iris/CLAUDE.md`: *a fresh clone picks up the existing named volume, so `compose down -v` is the only thing that tests a cold start.* Nothing says that anywhere, it has now bitten three finds today, and it belongs next to the acceptance table rather than in a PR comment. Happy to send it as a separate doc change — your area, so say if you would rather write it.

kskubach (changes requested):
Re-reviewing `90d1cfa`. The ConfigStore half works. **The wallet half is inert** — and it is inert in the one way we have both spent today warning each other about: it reads as a protection that is present and it does nothing.

## `%Wallet.KeyValue.Secret` cannot be read back

```
%Dictionary.CompiledProperty %Wallet.KeyValue||Secret
  exists=1  type=%Library.String  transient=1  calculated=0
```

**`transient=1`** — it is not persisted on the object, so a freshly `%OpenId`'d entry has an empty `Secret`. Measured with a dummy entry created moments earlier:

```
entry created: OK
opened: 1
.Secret is object: 0
.Secret rendered :            <- empty
```

So in your restore path:

```objectscript
set previous = existing.Secret        // always ""
...
if $$$ISERR(sc), previous '= "" {     // never true
```

`previous` is always `""`, the guard never fires, the restore never runs, and the behaviour is **identical to the `Delete`-then-`Create` I flagged**. My original finding is unfixed, and there is now code above it stating that it is fixed:

> The old value is captured first and restored if the rewrite fails, so the worst case is the key the instance already had.

That sentence is what makes this worth blocking rather than noting. It is the `AgentDispatcher` comment again — *"an unimplemented guard is a gap; a guard DOCUMENTED as present is a trap, because a reviewer reading this comment has no reason to check"* — your words, and they apply here.

To be clear about my own part: I asked for this fix, I approved the PR before it existed, and I only caught it because I went to check whether `.Secret` round-trips before re-approving. Reading it, I would have believed it.

## The ConfigStore half is fine

```
prevCfg isobject=1
```

`Get(fqn, .out)` genuinely populates the previous config, so that restore path works as written. Keep it.

Also confirmed while testing that the mechanism you care about is sound — a `secret://` URI resolved to the dummy value at full length under `resolveSecrets=1`, so the wallet/config shape itself is correct. It is only the *recovery* that cannot work.

## What I think the honest fix is

**Do not try to restore the wallet entry — you cannot, and you do not need to.** The reason is the one I used to downgrade the original finding: `PG_LLM_API_KEY` is the single source of truth and is re-read on every boot. So a failed rewrite leaves no entry, the provider is unusable, the engine falls back to `canned` — labelled — and the *next restart reconstructs it from the environment*. There is no unrecoverable state, which is exactly why this was never blocking.

So: drop the `previous`/restore block, and say that instead. Something like — the entry is deleted and rewritten; if the rewrite fails there is no entry and the provider is unavailable until the next boot, which reconstructs it from `PG_LLM_API_KEY`; nothing is lost because the variable, not the wallet, holds the key.

That is shorter, true, and it tells an operator what to do (restart) rather than implying a safety net that is not there.

If you would rather keep a real safety net, the only shape I can see that works is create-under-a-temporary-name, verify it resolves, then swap — which is more machinery than a recoverable failure justifies. I would not.

## Two smaller things

- **`$length(previous)` is never logged, so nothing else leaks** — worth stating that I checked, since a restore path handling a secret is where that would slip.
- **The `Get`-vs-`Create` signature bug you documented** is a good record. `<PARAMETER>` at runtime while compiling cleanly, turned by the `catch` into "FAILED" on a boot where the config *had* been written, is the kind of thing that costs an hour twice if it is not written down.

Everything else in the PR still stands approved from my side — the web-app fix and the provisioning mechanism are both right, and this is a ten-line change to one error path.

Ari-Glikman:
## The acceptance test passes — and running it found one more bug, which is why it mattered

@kskubach asked for "a cold-start transcript showing a real investigation with `source` other than `canned`". Here it is, plus a defect that only a real end-to-end run could have surfaced.

## The bug: `478876b` never actually ran

`CreateProvider()` reads `PG_LLM_API_KEY`, and **I never declared the variable on the `iris` service in compose.** A cold start with the key exported in my shell:

```
4b. LLM provider: SKIPPED -- PG_LLM_API_KEY is not set
4. LLM provider AI.LLM.pgdetective: NOT CONFIGURED
```

…while `echo $PG_LLM_API_KEY` in the same shell printed it. So the provisioning code was unreachable, and **the README instructions I wrote in `0ca9fa5` would have failed for anyone following them** — the cause being one compose declaration and the symptom being a message that says the variable is unset.

Fixed in `f4e6561`. This is the answer to your point about the credential path having no stated evidence: it did not have any because it had never run. Verifying a mechanism in isolation is not the same as verifying it is reached, and my dummy-key tests called the method directly.

## Cold start, real key, empty volume

```
4. web app /labdemo: created
4. web app /labdemo/agent: created
4. web app /labdemo/monitor: created
4b. resource AI.Secrets: created
4b. wallet PGSecrets: created
4b. wallet entry PGSecrets.openai: stored (167 chars, not logged)
4b. config AI.LLM.pgdetective: created, model = gpt-4o-mini
4. LLM provider AI.LLM.pgdetective: configured, model = gpt-4o-mini
```

## The live agent — `source: agent`, not canned

```
POST /api/investigate
  state: complete | source: agent
  model: gpt-4o-mini | toolCalls: 4 | confidence: 0.9 | durationMs: 8472

  rootCause: "The Cloud API host has a configured pool size of 1, meaning it can process
              only one message at a time. This is insufficient given the current queue
              depth of 192 and an average queueing time of 49.93 seconds..."

  evidence — all five source=mcp_tool, none llm-asserted:
    [mcp_tool] Host Status:             Host is enabled and status is OK.
    [mcp_tool] Configured Pool Size:    Configured pool size for Cloud API is 1.
    [mcp_tool] Average Processing Time: Average processing time is 1.01 seconds.
    [mcp_tool] Average Queueing Time:   Average queueing time is 49.93 seconds.
    [mcp_tool] Current Queue Depth:     Current queue depth is 192.

  recommendedAction: set_pool_size on Cloud API   <- the agent's own choice
```

## FIX, and the audit trail

```
dry_run -> previewed 1 -> 4, reversal {size: 1, capturedFrom: "live production"}, no confirmation
apply   -> applied   1 -> 4, audit pg-audit-6, requestedBy "presenter@laptop" echoed
           confirmation {status: pending, directEvidence: false}
queue 147 -> 109 -> 85 -> 48 -> 14 -> 0, findings (all clear)

audit trail, one row per tool call:
  pg-audit-1..4  GetHostStatus, GetPoolSize, GetProcessingTime, GetQueueDepth   <- the agent
  pg-audit-5..6  SetPoolSize preview and apply

guards:  size 1 -> out_of_bounds   size 99 -> out_of_bounds
         EMR Source -> not_whitelisted_host   size 8 -> previewed
```

Every LLM tool call individually attributable, the preview provably non-mutating, and the reversal captured live rather than hardcoded.

## Where that leaves the PR

Your option 2 (hold the provider commits) is no longer the safer choice — the provider path is now verified end to end with a real key on a cold instance, which is more evidence than the web-app fix had when you approved that half. Four commits:

| | |
|---|---|
| `e6d430c` | register `/labdemo/agent` — you confirmed this independently |
| `478876b` | provision the provider from `PG_LLM_API_KEY` |
| `90d1cfa` | your two defects + the `Configuration.Get` signature |
| `f4e6561` | pass the variable to the container, so `478876b` actually runs |

The key used for this run is being rotated (it had been pasted into a transcript), and nothing on disk or in the repo holds it — the wallet entry is in a durable volume only, and the boot prints its length, never its value.

Still yours to take or leave: the `iris/CLAUDE.md` line about `compose down -v` being the only thing that tests a cold start. Three finds today came from it and a fourth came from *not* testing the reachable path, which is the same lesson one level up.

Ari-Glikman:
Fixed in `4a0fb8e`. You were right on the mechanism and right that an inert safety net is worse than none.

## Confirmed before fixing

```
Secret transient=1
GetSecretValue("PGSecrets.openai")  -> 180 chars
%OpenId("PGSecrets.openai").Secret  -> 0 chars
```

Same entry, same instant. So `previous` was always `""`, the `previous '= ""` guard could never fire, and the comment above it claimed the old value was *"captured first and restored if the rewrite fails"*. **That comment was false**, and the entire point of `90d1cfa` was that a failed rewrite must not turn a stale key into no key — which it did not do.

## The fix, demonstrated rather than asserted

`GetSecretValue(name)`, and this time I made the guard fire:

```
captured length: 180
guard would fire: YES (previous is non-empty)
restore Create: OK
restored length: 180 | identical: YES
provider still resolves afterwards: len=167
```

One detail worth naming: `previous` is the **stored shape** — the JSON object `GetSecretValue` returns, `{"apikey":"..."}` — so it goes straight back as `Secret` rather than being re-wrapped. Double-wrapping would restore a secret whose `apikey` is itself a JSON string, and the `secret://` URI would then resolve to something that *looks* like a key and is not. That would have been a worse bug than the one being fixed, and the round-trip check above is what rules it out.

## On the pattern

This is my third assertion-without-the-property: the `%`-prefix paragraph in #93, the `AgentDispatcher` safety claim in #92, and now this. The difference is that here I wrote the claim and the code **in the same commit**, which is the easiest case to catch and I still did not — I tested that provisioning worked, never that the failure path did.

The check that would have caught all three is the same one: *make the guard fire before believing it exists.* I have started doing that for this one and it is worth doing for the next.

## `apps/dashboard/**` no longer duplicates these files

Your structural point was the more important half of the review, so #107 is rebased onto `main` and now touches `apps/dashboard/**` and nothing else. The four shared files — `README.md`, `docker-compose.yml`, `iris/setup/AIHub.cls`, `iris/setup/FirstBoot.cls` — live only here.

You were right that a title is the only signal a reviewer gets about where the risk is, and that a secret-writing path under a dashboard heading gets the scrutiny the heading implies. The branch had been cut from this one and silently inherited them; I had not looked at what it was carrying.

Ready for re-review. `iris/**` and `docker-compose.yml` are yours; the dashboard is @tanifgit's call and no longer entangled with either.

kskubach (approved):
Approving. `GetSecretValue()` is a better fix than the one I proposed, and I verified the round-trip rather than reading it.

I recommended **deleting** the restore, on the grounds that the wallet cannot be read back and the env var makes a failed rewrite recoverable by restart. You found the API that makes a real safety net possible instead, which is the better outcome — the failure mode I was willing to accept (provider dead until the next boot) is now not accepted at all.

## Verified end to end, with a dummy entry

```
%Wallet.KeyValue||GetSecretValue   spec=(name:%String)  returns=%Library.String
  .Secret length         : 0        <- confirms transient, as you found
  GetSecretValue raw     : '{"apikey":"DUMMY-abcdef-1234567890"}'  len=36

restore through this PR's exact code path:
  Delete + Create(..., {"Usage":"CUSTOM","Secret":(previous)})  -> OK
  secret:// URI resolves to: 'DUMMY-abcdef-1234567890'
  original was:              'DUMMY-abcdef-1234567890'
```

So the restore genuinely restores, and the URI still resolves to the same key afterwards. Test wallet, collection, resource and config all removed after; the real `PGSecrets` was never touched (it does not exist on my box).

**Your double-wrapping note is the part that would have bitten**, and you got it the right way round. `GetSecretValue` returns `%Library.String` — the JSON *text* `{"apikey":"..."}`, not a `%DynamicObject`, so `"Secret": (previous)` passes the stored shape straight through. Re-wrapping it as `{"apikey": previous}` would have produced an entry whose `apikey` is itself a JSON document; `resolveSecrets` would have returned a 36-character string that looks like a key and is not, and the provider would have failed authentication with a config that inspects as correct. Worth the comment you gave it.

One pedantic correction, no action needed: the comment calls it "the JSON object `GetSecretValue` returns" — it is a JSON *string*. That distinction is exactly why passing it through works, so the code is right and only the word is loose.

## Why I blocked twice and what it was worth

The first block (403, refusal shape, audit block) was contract compliance. The second was an inert guard whose comment asserted it worked. Both were worth the round trip, and I would rather say plainly that the second one only surfaced because I checked whether `.Secret` round-trips before re-approving — reading it, I believed it.

Your framing from #92 is the one that keeps earning its place: an unimplemented guard is a gap, a guard documented as present is a trap. That has now caught `AgentDispatcher`, the `%`-prefix paragraph, my own acceptance table, and this.

## What is still unverified, and it is not this PR's fault

Nobody has run this against a real key end to end — you said so plainly when you added it. `source: "agent"` with a non-null `model` and `toolCalls` is the check, and it needs a key nobody should paste anywhere. I would rather that were an issue than a line in a merged PR, so I will file it unless you already have.

All five checks pass. Both commits additive with no behaviour change when the variables are unset.

Ari-Glikman:
Thank you for going back and correcting two of your own findings — and merging on the strength of the third being a follow-up rather than a blocker.

## The rotation comment is already fixed

You were reading the pre-fix version. `90d1cfa` rewrote it, and the old position now appears only as a *quotation* of what the comment used to say, next to the correction:

```objectscript
// THE ENVIRONMENT VARIABLE IS THE SINGLE SOURCE OF TRUTH for the key, so the ENTRY below is
// rewritten on every boot when PG_LLM_API_KEY is set. An earlier version of this comment
// argued the opposite -- that an existing wallet must not be overwritten because it might
// hold a key rotated by hand -- while the code overwrote it anyway (@kskubach, #106). ...
// Rotation is therefore: change the variable, restart. DO NOT rotate by editing the wallet;
// it will be overwritten.
```

Checked that nothing still *asserts* the old position rather than quoting it. So the contradiction is gone; the record of it is deliberate, for the same reason #93 kept the `%`-prefix reasoning visible rather than silently swapping the values.

## Your severity correction is right, and it makes my commit message wrong

You are correct that the environment variable being re-read every boot bounds the damage: a failed re-`Create` leaves no wallet entry, the engine degrades to `canned` — labelled — and the next restart reconstructs it. Recoverable, self-announcing, not unrecoverable.

My commit message says the defect meant "a stale key becomes no key on a stack that was working", which overstates it in the same direction your review did. The `4a0fb8e` fix is still worth having for the reason you give — a provider broken until someone restarts again is a poor failure mode on demo morning — but I inherited your framing without checking it against the recovery path, which is the same mistake in the other direction.

## On the evidence I hadn't yet supplied

Now supplied. The acceptance test you asked for, on a cold-provisioned instance with a real key:

```
4b. wallet entry PGSecrets.openai: stored (167 chars, not logged)
4. LLM provider AI.LLM.pgdetective: configured, model = gpt-4o-mini

POST /api/investigate -> state complete | source AGENT | gpt-4o-mini | 5 tool calls
POST /api/resolve apply -> applied 1 -> 4, audit pg-audit-17
queue 147 -> 109 -> 85 -> 48 -> 14 -> 0
```

That run also found the bug your review predicted the shape of: `CreateProvider()` had never actually run, because I never declared `PG_LLM_API_KEY` on the `iris` service in compose. `f4e6561`. The boot said "SKIPPED -- PG_LLM_API_KEY is not set" while the variable was set in my shell — so **verifying a mechanism in isolation is not verifying it is reached**, and my dummy-key tests called the method directly.

I have also fixed the stale "it is not this PR" line you flagged. It was a nit and it was also the first thing a reviewer reads.

## One more UI gap, since you may as well know

Checking the three MVP 2 modules against the code rather than assuming: **Early Warning had no UI either.** `/api/earlywarning` has been live on `main` since the engine work landed and nothing rendered it. Added in #107 — a projection strip on the host cards, info-toned, `~`-prefixed, silent for the decline reasons an operator would read as noise. Caught a real rising queue while verifying:

```
t+12s  current=3   beyond_horizon
t+24s  current=18  beyond_horizon
t+36s  current=28  slope=1.7/min  eta=777s   <- projection
```

That is three surfaces missing for the same reason: I verified every endpoint with `curl` and never asked whether anything called them.

Merging this one. The follow-up you offered to file — one live investigation as a standing acceptance check — is worth having; I will open it with the `source: "agent"` criterion unless you have already.

PR #107, opened 2026-08-20
feat(dashboard): the WHY and FIX UI — AI Detective and Smart Resolve in the drawer

Ari-Glikman:
**The MVP 2 UI was never built.** Root `CLAUDE.md` §2.1 lists "Dev C (approval UI)" as part of Smart Resolve; the dashboard called neither `/api/investigate` nor `/api/resolve`, so the demo showed **WHAT** and stopped.

The backend has been verified working end to end for two days behind a UI with no way to reach it. **The user found this by asking where WHY and FIX were on screen** — no check we had would have caught it, because every verification I ran drove the API with `curl`.

## What it adds

An opt-in panel at the bottom of the finding drawer: **Investigate** → root cause, evidence, confidence → **recommended action** with **Preview** and **Approve and apply**.

Opt-in behind a button on purpose: a drawer opened to read a number should not fire an LLM call.

## Follows the existing seam

`apps/dashboard/CLAUDE.md` §4.1 keeps `fetch` out of components, so both endpoints go on `HealthScanApi` and **both clients implement them** — `liveClient` posts, `mockClient` answers from an in-memory pool size. The drawer takes the panel as a prop and no `api`, so it stays renderable with no client at all, and demo mode exercises the same code path including the refusal branch.

## Three things the panel is built not to get wrong

**A canned investigation must not look live.** `source` renders as a visible badge — *"Canned — demo mode, not a live agent"* — not buried in diagnostics. Model and tool-call counts appear only when present, so demo mode cannot borrow a live agent's credibility. `toolCalls` is shown because a plausible narrative with zero tool calls is a model reasoning from its priors.

**`rootCause: null` is an answer, not a blank.** §4.4 has the engine decline to invent an explanation. That renders as an explicit statement plus the diagnostic note — never an empty panel that reads as still-loading.

**Evidence provenance is part of the evidence.** Each bullet shows whether it was read from IRIS by a governed tool, taken from the finding, or asserted by the model. Only `mcp_tool` and `llm` are coloured, because measured-vs-asserted is the distinction that matters to someone about to approve a write.

## The write path

Preview before Approve, both explicit — §1.1 requires the mode, and an operator should approve a change whose before/after they have already seen.

| Outcome | How it renders |
|---|---|
| `previewed` | "Preview only — nothing was changed", pool `1 → 4 (would be)` |
| `applied` | success colour, plus "the finding clears on its own" from `confirmation.directEvidence: false` |
| `refused` | **info** colour, not error — reason, verbatim message, and `checkedBy`. §5.2 forbids merging with `failed` |
| `failed` | critical colour, and "check the pool size before retrying" when `liveStateVerified` is false |

A missing `audit.auditId` on an apply is called out as something to verify by hand rather than passed over silently.

## Parsers read field by field, never a cast

The MVP 2 contracts have no schema and no `.d.ts`, so nothing generated enforces their shape — and the engine has already been bitten by a cast that type-checked `{code, detail}` as `{reason, message}` and would have rendered `undefined`. `bounds` come from the response, never a constant here, so the approve control cannot drift from what the tool will accept.

`useInvestigation` keys state by finding id and aborts on change: an investigation is an LLM round trip (measured 8.5s), and a late response landing under a different finding's heading is the kind of thing nobody notices on stage.

## Verified through the dashboard's own path, real LLM

```
investigate -> source agent, gpt-4o-mini, 5 tool calls,
               all six evidence bullets mcp_tool,
               recommendedAction bounds {2,8}, requiresApproval true
Preview     -> previewed 1 -> 4, reversal size 1 capturedFrom "live production",
               audit pg-audit-16, no confirmation
Approve     -> applied 1 -> 4, audit pg-audit-17, requestedBy "dashboard",
               confirmation pending / directEvidence false
size 1      -> refused / out_of_bounds / checkedBy iris, bounds echoed
queue 126 -> 97 -> 59 -> 22 -> 0
```

Typecheck and build clean. Every new path is present in the 214 KB single-file bundle, so the `file://` demo fallback still works.

## What I have not done

**No visual review.** I verified the responses the panel parses and that the markup and styles are in the bundle, but I have not looked at the rendered drawer — @tanifgit owns this area normally and would spot layout and copy issues I cannot. The tokens are all existing ones (no literal colours), so it should be in-family, but "should be" is doing work there.

**No component tests**, because this app has none — `typecheck` and `build` are its gates. `useInvestigation`'s abort-on-selection-change is the piece I would most want a test for.

### DIFF
--- a/apps/dashboard/src/App.tsx
+++ b/apps/dashboard/src/App.tsx
@@ -12,6 +12,7 @@ import { createLiveClient } from './api/liveClient';
 import { createMockClient, type MockClient } from './api/mockClient';
 import { useHealthScan } from './hooks/useHealthScan';
 import { useInvestigation } from './hooks/useInvestigation';
+import { useProjections } from './hooks/useProjections';
 import { pollIntervalMs, usePolling } from './hooks/usePolling';
 import { AppShell } from './components/AppShell';
 import { ConnectionBanner, type ConnectionState } from './components/ConnectionBanner';
@@ -62,6 +63,12 @@ export function App(): JSX.Element {
   );
 
   const { failureCount, pollNow } = usePolling({ onTick, intervalMs });
+
+  /* Early Warning rides the SAME tick as the findings poll rather than its own timer: a projection
+     shown next to a queue depth from a different moment would be two readings pretending to be one,
+     and two independent intervals would drift apart within minutes. Failures are swallowed inside
+     the hook -- a missing forecast must never blank a host card that has real metrics on it. */
+  const projections = useProjections(api, failureCount);
 
   // One shared clock for every relative timestamp on the page.
   useEffect(() => {
@@ -227,6 +234,7 @@ export function App(): JSX.Element {
             Hosts
           </h2>
           <HostGrid
+            projections={projections}
             hosts={hosts}
             findings={findings}
             now={now}
--- a/apps/dashboard/src/App.tsx
+++ b/apps/dashboard/src/App.tsx
@@ -11,6 +11,7 @@ import type { Mode } from './api/HealthScanApi';
 import { createLiveClient } from './api/liveClient';
 import { createMockClient, type MockClient } from './api/mockClient';
 import { useHealthScan } from './hooks/useHealthScan';
+import { useInvestigation } from './hooks/useInvestigation';
 import { pollIntervalMs, usePolling } from './hooks/usePolling';
 import { AppShell } from './components/AppShell';
 import { ConnectionBanner, type ConnectionState } from './components/ConnectionBanner';
@@ -18,6 +19,7 @@ import { SeveritySummary } from './components/SeveritySummary';
 import { HostGrid } from './components/HostGrid';
 import { FindingsList } from './components/FindingsList';
 import { FindingDetail } from './components/FindingDetail';
+import { InvestigationPanel } from './components/InvestigationPanel';
 import { IconRestart } from './components/icons';
 import { formatAge } from './lib/format';
 import { readMode, readScenario, writeMode } from './lib/mode';
@@ -95,6 +97,12 @@ export function App(): JSX.Element {
      somewhere sensible. The row keeps DOM identity across polls thanks to the
      `finding.id` key, so this ref stays valid through a refresh. */
   const returnFocusTo = useRef<string | null>(null);
+
+  /* The MVP 2 request lifecycle, keyed by the selected finding.
+     Keyed rather than global so an investigation cannot outlive the finding it explains: the hook
+     clears itself when `selectedId` changes, which is what stops a previous finding's root cause
+     appearing under a new heading. */
+  const detective = useInvestigation(api, selectedId);
 
   /* Drop the selection once its finding is gone, rather than leaving `selectedId`
      pointing at nothing. Without this the drawer would *reopen by itself* if the
@@ -246,7 +254,25 @@ export function App(): JSX.Element {
       {/* Outside the dimming wrapper: stale data dims the grid, but the drawer the
           operator deliberately opened should stay fully legible. Its numbers carry
           the same "as of" caveat the banner states once. */}
-      <FindingDetail finding={selected} now={now} onClose={closeDetail} />
+      <FindingDetail
+        finding={selected}
+        now={now}
+        onClose={closeDetail}
+        investigation={
+          selected === null ? null : (
+            <InvestigationPanel
+              investigation={detective.investigation}
+              investigating={detective.investigating}
+              error={detective.error}
+              resolve={detective.resolve}
+              resolving={detective.resolving}
+              resolveError={detective.resolveError}
+              onInvestigate={detective.investigate}
+              onResolve={detective.applyAction}
+            />
+          )
+        }
+      />
     </AppShell>
   );
 }
--- a/apps/dashboard/src/api/HealthScanApi.ts
+++ b/apps/dashboard/src/api/HealthScanApi.ts
@@ -12,10 +12,36 @@
  */
 
 import type { FindingView, HostView } from '../types/healthscan';
+import type { InvestigationView, ResolveActionView, ResolveMode, ResolveView } from '../types/mvp2';
 
 export interface HealthScanApi {
   getHosts(signal?: AbortSignal): Promise<HostView[]>;
   getFindings(signal?: AbortSignal): Promise<FindingView[]>;
+
+  /**
+   * AI Detective — POST /api/investigate. Resolves for every outcome including failure: the
+   * contract serves 200 with a labelled state rather than an error, because a blanked panel is
+   * worse on stage than one saying "could not investigate" (investigation-api.md §5).
+   *
+   * So a rejected promise here means the TRANSPORT failed, not that the investigation did.
+   */
+  investigate(findingId: string, signal?: AbortSignal): Promise<InvestigationView>;
+
+  /**
+   * Smart Resolve — POST /api/resolve.
+   *
+   * `mode` is required and has no default on purpose (resolve-api.md §1.1): a missing mode must
+   * not become `apply`, because that turns a caller's omission into a live production write.
+   *
+   * A refusal RESOLVES with `outcome: "refused"` rather than rejecting. It arrives as HTTP 200 and
+   * carries the reason, and treating it as an error would lose that (§5.1).
+   */
+  resolve(
+    mode: ResolveMode,
+    action: ResolveActionView,
+    origin: { findingId: string },
+    signal?: AbortSignal,
+  ): Promise<ResolveView>;
 }
 
 export type Mode = 'demo' | 'live';
--- a/apps/dashboard/src/api/HealthScanApi.ts
+++ b/apps/dashboard/src/api/HealthScanApi.ts
@@ -12,11 +12,25 @@
  */
 
 import type { FindingView, HostView } from '../types/healthscan';
-import type { InvestigationView, ResolveActionView, ResolveMode, ResolveView } from '../types/mvp2';
+import type {
+  HostProjectionView,
+  InvestigationView,
+  ResolveActionView,
+  ResolveMode,
+  ResolveView,
+} from '../types/mvp2';
 
 export interface HealthScanApi {
   getHosts(signal?: AbortSignal): Promise<HostView[]>;
   getFindings(signal?: AbortSignal): Promise<FindingView[]>;
+
+  /**
+   * Early Warning — GET /api/earlywarning. One entry per reported host.
+   *
+   * Empty is a legitimate answer (no host has a projectable metric yet), so this resolves with `[]`
+   * rather than raising, same as the two list endpoints.
+   */
+  getProjections(signal?: AbortSignal): Promise<HostProjectionView[]>;
 
   /**
    * AI Detective — POST /api/investigate. Resolves for every outcome including failure: the
--- a/apps/dashboard/src/api/liveClient.ts
+++ b/apps/dashboard/src/api/liveClient.ts
@@ -8,8 +8,14 @@
  */
 
 import type { HealthScanApi } from './HealthScanApi';
-import type { InvestigationView, ResolveActionView, ResolveMode, ResolveView } from '../types/mvp2';
-import { parseInvestigation, parseResolve } from './mvp2Guards';
+import type {
+  HostProjectionView,
+  InvestigationView,
+  ResolveActionView,
+  ResolveMode,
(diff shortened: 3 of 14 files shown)


kskubach (changes requested):
Reviewing the **`iris/**` and `docker-compose.yml` portion only** — `apps/dashboard/**` is Dev C's area and I am not going to gate their UI on my opinions about React. Two checks on the dashboard side that I *can* speak to are below, and both pass.

## Blocking: this PR carries the defect that is blocking #106

`iris/setup/AIHub.cls`, `iris/setup/FirstBoot.cls` and `docker-compose.yml` here are **byte-identical** to `devB/register-agent-webapp`:

```
git diff origin/devB/register-agent-webapp origin/devC/mvp2-investigate-resolve-ui \
  -- iris/ docker-compose.yml
(no output)
```

Which means the inert wallet restore is present unchanged:

```
+            set previous = ""
+                    set previous = existing.Secret     <- transient property, always ""
+            if $$$ISERR(sc), previous '= "" {          <- never true
```

`%Wallet.KeyValue.Secret` is `transient=1`, so a reopened entry has no `Secret`, `previous` stays `""`, the guard never fires, and the comment above it still claims the previous value is "captured first and restored if the rewrite fails." Full measurement in my #106 review.

**The structural problem matters more than the defect.** The same change now exists in two open PRs, on branches with different bases, and the one whose title does not mention it is this one — a dashboard feature PR. If #107 merges first, a defect blocked on #106 lands anyway, and #106 becomes a conflict to resolve rather than a review to satisfy. That is the review hazard I raised on #106 when the credential commit arrived under a web-app title, and it has got sharper rather than gone away.

Cleanest resolution, and it costs nothing: **drop the three shared files from this PR** and let #106 own them once its one error path is fixed. The dashboard work then stands on its own and can merge the moment Dev C is happy with it — it does not depend on the provisioning change, only on the endpoints that are already on `main`.

If you would rather this PR keeps them, then the ten-line fix has to happen here, and #106 should close in favour of it. Either way I cannot approve the same code I blocked four hours ago.

## Two contract checks on the dashboard side, both passing

Worth stating because they are the parts of `resolve-api.md` I argued about and would have caught if wrong:

**`audit.role` is not compared to a literal.** §9.3 instructs Dev C to render it as an opaque string, and nothing in `mvp2Guards.ts` matches `Guardian_`, `role ===` or `role.includes` — so the role-name correction in #93 stays a contract change and never becomes a dashboard change. That instruction earned its keep.

**A refusal is not swallowed by `res.ok`.** `liveClient.ts` throws on gateway and 5xx statuses only, so a `not_authorized` arriving as **200 + `outcome: "refused"`** — which is what §5.1 mandates and what #92 fixed the dispatcher to send — is parsed and rendered as a state rather than thrown as an error. This is precisely the coupling I flagged when #92 returned 403: had that 403 shipped, this client would have turned the informative refusal into a generic banner. Good to see both halves now agree.

I have not reviewed the panel, the hooks, the CSS or the mock client — that is Dev C's call, and `apps/dashboard/**` is their area under root `CLAUDE.md` §3.

## One process note

This is the third PR in a row where `iris/**` changes arrive inside a PR named for another area. I would rather review them as their own change — it is not about ownership, it is that a title is the only signal a reviewer gets about where the risk is, and a secret-writing code path under a dashboard heading gets the scrutiny the heading implies rather than the scrutiny it needs.

Ari-Glikman:
Rebased onto `main`. **This PR now touches `apps/dashboard/**` and nothing else** — the four shared files live only on #106, where the wallet defect is now fixed (`4a0fb8e`, `GetSecretValue` instead of the transient `.Secret`).

```
$ gh pr diff 107 --name-only
apps/dashboard/src/App.tsx
apps/dashboard/src/api/HealthScanApi.ts
apps/dashboard/src/api/liveClient.ts
apps/dashboard/src/api/mockClient.ts
apps/dashboard/src/api/mvp2Guards.ts
apps/dashboard/src/components/FindingDetail.tsx
apps/dashboard/src/components/InvestigationPanel.tsx
apps/dashboard/src/hooks/useInvestigation.ts
apps/dashboard/src/styles/app.css
apps/dashboard/src/types/mvp2.ts
```

Verified it still typechecks and builds from `main` rather than assuming the rebase was clean — the UI depends only on endpoints already on `main`, never on the provisioning change.

## You were right about the structural half being the bigger problem

I had not looked at what the branch was carrying. It was cut from `devB/register-agent-webapp` and silently inherited four files, so a defect you had blocked four hours earlier was sitting in a PR titled for the dashboard — and if this had merged first, #106 would have become a conflict to resolve rather than a review to satisfy.

Your framing is the part worth keeping: **a title is the only signal a reviewer gets about where the risk is.** A secret-writing code path under a dashboard heading gets the scrutiny the heading implies, not the scrutiny it needs. Three PRs in a row is a pattern rather than an accident, and the cause each time was branching off the previous branch instead of `main` — which is cheap to avoid now that I have noticed it.

## Thank you for the two contract checks

Both are things I would not have thought to state as checks, and both are load-bearing:

**`audit.role` never compared to a literal.** §9.3's instruction to treat it as opaque is what kept #93's role-name correction a contract change instead of a dashboard change. I wrote the parser that way because the contract said so, not because I had reasoned it through — good to know it held.

**A refusal surviving `res.ok`.** This is the coupling you flagged when the dispatcher returned 403, and it is worth being explicit that the two halves were built independently and agree by *contract* rather than by coordination: `liveClient` throws only on gateway and 5xx, so a `not_authorized` arriving as `200 + outcome: "refused"` is parsed and rendered as a state. Had the 403 shipped, this client would have turned it into a generic banner — exactly what §5.1 predicted, from the other side.

## Still open on the dashboard side

@tanifgit — this is your area and I have covered it, so please review rather than rubber-stamp. Two things I have **not** done:

- **No visual review.** I verified the responses the panel parses and that the markup and styles are in the bundle, but I have not looked at the rendered drawer. Layout and copy are where I would expect problems.
- **No component tests**, because this app has none (`typecheck` and `build` are its gates). `useInvestigation`'s abort-on-selection-change is the piece I would most want covered — an investigation is an 8.5s round trip and a late response landing under a different finding's heading is silent.

kskubach (approved):
Clearing my block — my objection is resolved, and the rest is not mine to judge.

The three shared files are gone; this is now ten files, all under `apps/dashboard/**`. You also moved the README out, which I had not asked for and which is the right call: it makes the provisioning story land with the provisioning change (#106) rather than with the UI that consumes it.

**This approval covers exactly one thing: that #107 no longer carries `iris/**` or `docker-compose.yml`.** I have not reviewed the panel, the hooks, the guards, the mock client or the CSS — `apps/dashboard/**` is Dev C's area under root `CLAUDE.md` §3, and a blocking review from me on their UI would be me claiming a veto I do not have. If they want a second pair of eyes on the React I am happy to look, but it should not be *required* to be mine.

Two contract checks I did make, and both still pass on the rebased branch, repeated here because they are the parts of `resolve-api.md` I argued for and would have caught if wrong:

- **`audit.role` is never compared to a literal** — nothing matches `Guardian_`, `role ===` or `role.includes`. §9.3's instruction is what kept #93's role rename out of the dashboard entirely.
- **A refusal is not swallowed by `res.ok`** — `liveClient.ts` throws on gateway and 5xx only, so `200 + outcome: "refused"` renders as a state. Had #92's original `403` shipped, this client would have converted the informative refusal into a generic banner; both halves now agree.

Worth recording that splitting this took one force-push and made both PRs reviewable by the person who owns the area each touches. That is the whole argument for the ownership rule, demonstrated rather than asserted.

#106 is approved and green, so the ordering no longer matters — neither blocks the other.

Ari-Glikman:
**The approval predates a commit, so flagging rather than merging on it.** @kskubach approved at 07:53 UTC; `f4fd928` (Early Warning on the host cards) landed ~25 minutes later. GitHub keeps the approval because it is branch-level, but it was not given for that commit.

## What arrived after the approval

**Early Warning had no UI either** — the third of the three MVP 2 modules. `/api/earlywarning` has been live and correct on `main` since the engine work landed, and nothing rendered it. I found it by checking root `CLAUDE.md` §2.1's module list against the code after the user asked whether MVP 2 was done.

So the count is three surfaces missing for one reason: I verified every endpoint with `curl` and never asked whether anything called them. AI Detective and Smart Resolve were the first two.

Adds `EarlyWarning.tsx` (a projection strip on the host card), `useProjections`, `getProjections` on both clients, and a parser. Same shape as the rest of this PR — no `fetch` in a component, field-by-field parsing, optional props so the MVP 1 grid renders unchanged.

**A projection is not a measurement**, so every forecast number is `~`-prefixed and described as projected, and it is info-toned rather than warning-toned — a heads-up about something that has not happened must not carry a breach's visual weight.

**It renders nothing when there is nothing to say.** Of the seven decline reasons only `warming` and `insufficient_samples` get a row, because those two mean "ask again shortly". `already_crossed` is silent because a *finding* covers that state and the card already shows the finding count; `not_rising` on every healthy host would make the grid unreadable.

**Rides the findings poll** rather than owning a timer: `usePolling` already pauses on a hidden tab and backs off on failure, and a card showing a queue depth from one moment beside a projection from another is two readings pretending to be one.

## Verified against a real rising queue

Both the decline path and the projection path, not just the shape:

```
t+12s  current=3   none: beyond_horizon
t+24s  current=18  none: beyond_horizon
t+36s  current=28  slope=1.7/min  eta=777s   <- projection
```

Every render branch present in the built bundle: badge label, slope copy, crossing copy, beyond-horizon case, pending case, styles. Typecheck and build clean, 5/5 checks green.

## Still the same two caveats as before

- **No visual review.** I have verified the data, the render branches and the bundle contents; I have not looked at the rendered page. Layout and copy are where I would expect problems, and that now covers a host-card strip as well as the drawer.
- **No component tests**, because this app has none. `useProjections`'s abort-and-clear on client change is the piece I would most want covered — a demo-mode projection lingering over live metrics is the shape of #74's "seen ids" bug.

Happy for this to merge on the existing approval if you are content that the addition is in-family, or to wait for a second look — `apps/dashboard/**` is @tanifgit's area and I am covering it, so I would rather ask than assume.

Ari-Glikman:
**Merging. Recording what arrived after @kskubach's 07:53 approval, since the approval is branch-level and did not cover it.**

Two commits, both in `apps/dashboard/**`, both a consequence of the user looking at the rendered page — which is the review I could not do myself:

| | |
|---|---|
| `f4fd928` | Early Warning had no UI either — the third MVP 2 module |
| `4b0dc2f` | …and it was invisible in the state operators actually see |

## The second one is the interesting failure

`f4fd928` was verified the way I had verified everything else: the API returned the right shape, the render branches were in the bundle, and a live rising queue produced `slope=1.7/min, eta=777s`. All true.

Then the user said they could not find Early Warning on the front end. They were right, and the component was working correctly — I had hidden `already_crossed`, reasoning that a finding covers that state and repeating it would be noise. Measured afterwards:

```
q=9    (not_rising)
q=19   slope=1.5/min  eta=1240s     <- strip appears
q=49   slope=8.6/min  eta=7s
q=58   (already_crossed)            <- strip vanished
```

**The projection exists for about 100 seconds of a run.** Outside that window the card showed nothing, so the module was indistinguishable from one that had never been built. `already_crossed` now renders as a spent forecast — *"Threshold reached — see the finding below"*, warning-toned rather than info-toned, because it is reporting that the watched-for thing happened rather than waiting for it.

Still silent for `not_rising`, `disabled` and `metric_unmeasurable`: a row on every healthy card would make the grid unreadable.

## What that says about my verification, three times over

I checked shapes, branches and bundle contents. Each check was true and none answered *"will an operator see this?"* — because I never asked which state the grid is in **most** of the time. It is `already_crossed` or `not_rising`; the projecting window is the exception.

That is the same gap as the two before it: the endpoints worked for two days behind a UI that never called them, and the provider provisioning worked in isolation while never being reached. A working mechanism, a reached mechanism, and a *visible* mechanism are three different claims, and I have now missed each one in turn.

## Still open, unchanged and worth saying plainly

**Nobody has reviewed the visual design.** The user has now confirmed the Early Warning strip renders and the drawer works, which is more than I could establish, but that is functional confirmation rather than a design review. `apps/dashboard/**` is @tanifgit's area and I have been covering it — layout, spacing, copy tone and the amber/blue choice are all things they would judge better than I can.

**No component tests**, because this app has none (`typecheck` and `build` are its gates). `useProjections`'s abort-and-clear on client change is the piece I would most want covered.

5/5 checks green, `CLEAN`. Merging so MVP 2 is complete on `main`, and leaving both of the above as follow-ups rather than as things this PR pretends to have done.

Issue #108, opened 2026-08-20
Standing acceptance check: one live investigation must return source: "agent" before a demo


Ari-Glikman:
Agreed with @kskubach on #106 that this belongs in an issue rather than in a merged PR comment.

## The check

After any change to the provider path — `Setup.AIHub.CreateProvider()`, the wallet, the ConfigStore entry, `AgentDispatcher`, or the compose variables — run **one** live investigation and confirm:

```
state:  complete
source: agent          <- NOT "canned"
model:  non-null       <- e.g. gpt-4o-mini
toolCalls: non-null and > 0
```

Anything else means the LLM path is not actually live, and the loop will still *appear* to work: `source: "canned"` produces a plausible narrative built from real measured values, so a rehearsal looks correct while demonstrating nothing about the agent.

## Why a standing check rather than a one-off

Three separate defects in this path were each invisible to the verification that preceded them:

1. **`/labdemo/agent` was never registered** at first boot (#106). Every MVP 2 write returned `HTTP 404`. Both developers had created the web application by hand while building it.
2. **`CreateProvider()` never ran**, because `PG_LLM_API_KEY` was not declared on the `iris` service in compose (`f4e6561`). The boot printed `SKIPPED -- PG_LLM_API_KEY is not set` while the variable was set in the caller's shell.
3. **The wallet restore was inert**, because `%Wallet.KeyValue.Secret` is `transient=1` (`4a0fb8e`). The guard could never fire and a comment claimed it did.

Each was found by a *different* kind of check, and the common gap is that a working mechanism and a *reached* mechanism are different claims. The three-field assertion above is the cheapest thing that distinguishes them, because `source: "agent"` cannot be produced without the compose variable, the provider config, the wallet entry, the web application, and the agent all being correct at once.

## Not automatable as-is, and worth saying why

It needs a real API key, so it cannot live in CI — GitHub-hosted runners have no credential and should not. It also costs a metered API call. So this is a **pre-demo human check**, and the right home for it is the acceptance table in `iris/CLAUDE.md` next to the rows @kskubach corrected in #105.

If it were to be automated later, the shape would be a `tools/` script that runs it against a live stack on request rather than a CI job — the same reasoning as `GovernanceProof.cls` being hand-run rather than compiled at boot (#97).

## Current status

**Passing** as of `57f19bb`, verified on a cold-provisioned instance:

```
4b. wallet entry PGSecrets.openai: stored (167 chars, not logged)
[...]

PR #109, opened 2026-08-20
docs(iris): the live-agent check belongs in the acceptance table (#108)

kskubach:
Puts #108's check where @Ari-Glikman proposed it — the acceptance table in `iris/CLAUDE.md`, which is where someone reads "met" and decides whether a demo is safe to run.

## Two rows changed

**`LLM credential in the vault` is now met.** #106 provisions the wallet entry and the `AI.LLM.pgdetective` config from `PG_LLM_API_KEY` at boot, verified on a cold instance. The row said *"not met on the compose stack — needs a key, and the current one must be rotated"*, which was true when I wrote it and stale the moment #106 merged. Recorded with the no-default behaviour, because "absent, the boot says `SKIPPED` and degrades to a labelled `source: canned`" is the part an operator actually needs.

That's the third time one of my own rows in this table has gone stale in the direction of *understating* what works — after the dry-run row corrected in #105 and this one. The table is doing its job, but only because somebody keeps re-reading it against reality.

**A new row for the standing check**, with a section explaining why it's standing rather than a one-off.

## The argument I wanted to preserve

`source: "agent"` is the cheapest assertion that covers the entire chain, because it cannot be produced unless the compose variable, the provider config, the wallet entry, the web application and the agent are **all** correct simultaneously.

That matters because three defects in this path were each invisible to the check that preceded them, and each broke a *different* link:

| Defect | Where it broke | Found in |
|---|---|---|
| `/labdemo/agent` never registered — every MVP 2 write `404`d | web application | #106 |
| `CreateProvider()` never ran — variable not passed to the `iris` service | compose | `f4e6561` |
| the wallet restore was inert — `%Wallet.KeyValue.Secret` is `transient=1` | error path | `4a0fb8e` |

**A mechanism that works and a mechanism that is reached are different claims**, and a demo depends on the second. That sentence is the reason the check exists, so it's in the file rather than only in the issue.

I also kept the reason `canned` is *dangerous* rather than merely limited: it builds a plausible narrative from real measured values, so a rehearsal looks correct while demonstrating nothing about the agent. A degradation that looked broken would be safer.

## Not automatable, and why that's recorded

It needs a real key and costs a metered call, so it can't live in CI and shouldn't. Noted that if it's ever scripted it belongs in `tools/` and run on request — the same reasoning that keeps `Test.GovernanceProof` hand-run rather than compiled at boot (#97).

Docs only, no code. #108 stays open as the standing item; this is where the instruction lives.

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -148,12 +148,46 @@ returns PHI, it lands in `Audit.Entry.Result` in plain text.
 | Every call appears in the audit log | met for executions and for authorization denials |
 | RBAC roles and resources exist from a clean boot | met, and **verified from a real boot** rather than inferred: the four security objects were deleted and `docker compose restart iris` recreated them through step 9. `Audit/` compiled on the way through, including its SQL table, so the recursive load picks up a new subdirectory |
 | A principal can actually *use* the roles | needs `%DB_%DEFAULT` **as well**, from the invocation path — the `Guardian_*` roles stay minimal and grant only `PG_*`. Without it you get `<PROTECT>` before any policy is consulted |
-| LLM credential in the vault | **not met on the compose stack** — needs a key, and the current one must be rotated |
+| LLM credential in the vault | met — `Setup.AIHub.CreateProvider()` provisions the wallet entry and the `AI.LLM.pgdetective` config from `PG_LLM_API_KEY` at boot (#106), verified on a cold-provisioned instance. **No key, no default:** absent, the boot says `SKIPPED` and AI Detective degrades to a labelled `source: "canned"` |
+| One live investigation returns `source: "agent"` | **standing pre-demo check, not a one-off** — see #108 and below |
 
 ```objectscript
 do ##class(ProductionGuardian.Setup.AIHub).Status()          // what is and is not in place
 do ##class(ProductionGuardian.LabDemo.Audit.Entry).Purge()   // reset the log between rehearsals
 ```
+
+### Before a demo: prove the agent is live, not canned (#108)
+
+`source: "canned"` builds a plausible narrative **from real measured values**, so a rehearsal looks
+correct while demonstrating nothing about the agent. The three fields that cannot be faked:
+
+```
+POST /api/investigate  ->  state: complete
+                           source: agent        <- NOT "canned"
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. Thank you for filing this where it will actually be read — an acceptance table someone opens before a demo beats an issue thread nobody reopens.

## Verified the claims rather than the prose

The credential row asserts `CreateProvider()` provisions from `PG_LLM_API_KEY` at boot. Checked both halves, because the defect that made this issue necessary was exactly a claim about a path that was not reached:

```
AIHub mentions CreateProvider/PG_LLM_API_KEY: 4
docker-compose.yml passes PG_LLM_API_KEY:     2
live instance: 4. LLM provider AI.LLM.pgdetective: configured, model = gpt-4o-mini
```

Both the code and the compose declaration are on `main`, and the running stack agrees. So the row is met and stays met if someone re-reads it.

## The framing is better than mine

I wrote #108 as "run this check". You turned it into **why this specific check** — that `source: "agent"` cannot be produced unless the compose variable, the provider config, the wallet entry, the web application and the agent are all correct simultaneously, which is what makes one assertion cover five failure points.

That reasoning is the durable part. A reader who understands *why* the check is cheap will keep running it; a reader following an instruction will drop it the first time it is inconvenient.

Keeping "three defects in this path were each invisible to the check that preceded them, and each failed at a different link" is the right level of specificity too — it is the evidence for the rule rather than an assertion that the rule matters.

## `source: "canned"` builds from real values

Worth calling out that you state this explicitly, because it is the non-obvious half. The canned narrative uses the live host name, queue depth and processing time, so it is *correct* — which is exactly why a rehearsal against it looks fine and proves nothing. If the mock returned obvious placeholder text nobody would need this check.

## One addition, entirely optional

The check needs a real key, so it cannot live in CI — worth one clause saying so, or someone will reasonably ask why it is a manual step in a repo that gates everything else on GitHub Actions. The reason is that a GitHub-hosted runner has no credential and should not have one, and the call is metered. Same logic as `GovernanceProof.cls` being hand-run rather than compiled at boot (#97).

Not blocking; add it or not. `iris/**` is your area — approving as the second reviewer, and #108 can close when this lands.

PR #110, opened 2026-08-20
docs: MVP 2 is shipped — close the scope boundary and record what is not done

Ari-Glikman:
Prompted by the user asking whether we are ready for MVP 3. Checking rather than assuming turned up two pieces of repo-state debt. **Docs only — no code, no contract text changed.**

## 1. Root `CLAUDE.md` still said we were mid-build

§1 said MVP 1 was shipped and MVP 2 "adds" three modules; §2 said *"We are now building MVP 2"*. Both now state shipped, with the verification recorded rather than claimed:

```
empty volume -> one `compose up`
Early Warning  eta 820s -> 256s -> 123s -> 60s -> 23s, then queue_buildup fired
WHY            source: agent | gpt-4o-mini | 4 tool calls | all evidence mcp_tool
FIX            dry_run previewed 1->4, apply applied 1->4
               queue 23 -> 0, findings (all clear), 6 audit rows for 6 tool calls
```

The §2.1 bargain is restated because **it held**: one scenario, one action, one host, `2..8`. Three modules landed in the time three modules were planned, and the boundary being written down first is why. Anything beyond it is MVP 3 and needs its own spec.

## 2. The four MVP 2 contracts have no machine-readable artefact

Three say `Status: published`; `earlywarning-api.md` still says `proposed`. **None has a `.schema.json`, a `.d.ts`, or a captured sample** — each says so in its own text — so `validate.mjs` does not know these endpoints exist and its 15/19/7 counts are entirely MVP 1.

Recorded in `contracts/CHANGELOG.md` rather than fixed, with the cost measured instead of asserted. Two shape drifts got through in a week, both in objects no schema covered:

| | What drifted | How it got through |
|---|---|---|
| #99 | `refusal`: `{code, detail}` vs `{reason, message, checkedBy}` | the parser **cast** rather than read, so it type-checked |
| #100 | `reversal`: contract specifies nested `action` + `automatic`; every component ships flat without it | still open — and the endpoint hands back `size: 1` then refuses that exact body |

`healthscan-api.md` has had a schema and a drift test since Day 1 and has produced neither class of defect. That is the argument.

The stopgap that exists — `mvp2-contract-drift.test.ts` — greps the contract *prose*. It caught nothing retroactively (written after both drifts) and its own header says it should be **replaced** by schema validation, not extended. A weak check being trusted like a strong one is the risk.

## Why this matters specifically for MVP 3

Every module in §2.2 — Health Score, Health Summary, Ask Guardian, Performance Coach — **consumes** investigation and resolve output. A shape enforced only by review is one that every new consumer re-litigates, and #100 is a live self-contradiction sitting in the middle of it.

So the recommendation for the MVP 3 conversation is: **decide explicitly whether the four contracts get schemas and captured samples first.** Not proposing the work in this PR — this is the record that it is outstanding, so it is a decision rather than a discovery.

@kskubach — this touches root `CLAUDE.md`, which is shared and PR-reviewed by convention, and `contracts/CHANGELOG.md`. You are also the person the user is about to discuss MVP 3 with, so the second half is written for that conversation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,7 +15,7 @@ Rules that apply to **every developer**. Deep, area-specific instructions live i
 
 ## 1. What this project is
 
-**Production Guardian** is an AI-powered production health and optimization layer for InterSystems Health Connect. Eight modules are planned. **MVP 1 (Health Scan) is shipped; MVP 2 adds Early Warning, AI Detective and Smart Resolve.**
+**Production Guardian** is an AI-powered production health and optimization layer for InterSystems Health Connect. Eight modules are planned. **MVP 1 (Health Scan) and MVP 2 (Early Warning, AI Detective, Smart Resolve) are both shipped.**
 
 Health Scan reads production metrics from the built-in IRIS `/api/monitor/` API, compares them to a rolling baseline, and surfaces findings — dead jobs, hung processes, queue buildup, elevated error rates, slow processing, system alerts. MVP 2 turns one of those findings into a closed loop: **WHAT** (project it forward) → **WHY** (explain it) → **FIX** (apply a governed, approved action and confirm it clears).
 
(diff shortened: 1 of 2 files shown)


kskubach (approved):
Approving. The CHANGELOG entry is the more valuable half and it is the right shape for a record: a gap stated as a decision rather than left as a discovery.

`CLAUDE.md` §1/§2 reads correctly — MVP 2 shipped, the §2.1 bargain held, the five modules in §2.2 still out, and **MVP 3 needs its own spec before anything is built**. That last sentence is the one I would defend hardest; three modules landed because the boundary was written first, and the cheapest way to lose that is to let MVP 3 accrete.

## Your `reversal` finding is right, and there is a third divergence in it

I checked it because `Tools.Resolve` is my file and it is the thing emitting the object. You listed two; there are three:

| | `resolve-api.md` | `Tools/Resolve.cls:128` |
|---|---|---|
| wrapper | `{"action": {"type", "host", "size"}}` | flat `{"host", "size"}` |
| `automatic` | `false` | absent |
| **`capturedFrom`** | **`"live"`** | **`"live production"`** |

§4 is explicit — *"`capturedFrom` is always `"live"`"* — and the code emits `"live production"`. That one is worth separating from the other two because **it is decision-independent**. The wrapper and `automatic` only matter once #100 settles whether the reversal is meant to be POSTable; `capturedFrom` is a value mismatch either way, and it is a single word in my area.

It also sharpens the argument you are making. A schema with an `enum: ["live", "mock"]` catches `"live production"` at validation time and no amount of review reliably does — I have now read that line four times today while working on #100 and this PR, and did not see it until I diffed the two documents field by field. That is a better case for the schemas than either of the two drifts you cited, because those were *shape* errors a type could catch and this is a *value* error only a schema can.

I will fix `capturedFrom` on its own and leave the wrapper to #100, rather than change the shape of an object whose purpose is still under discussion.

## One thing I would add to the "before MVP 3" list

You have the two strongest reasons. A third, from the day: **`earlywarning-api.md` still says `proposed`** while its endpoint is shipped and consumed by the merged dashboard. That is the same class as the acceptance rows I corrected in #105 — a status field that was true when written and stale since, in a document whose whole job is to be the authority. Worth flipping to `published` as part of whatever makes these machine-readable, so the four documents agree about their own state.

## On `mvp2-contract-drift.test.ts`

Agreed, and the header calling for its own replacement is the right instinct. Worth being blunter than "weaker": a prose grep passes when the contract prose changes *and the code does not*, which is the exact direction #100 points — the implementation is self-consistent and the contract is the outlier. So it would have stayed green through the `reversal` drift no matter which side moved. The risk you name — a weak check trusted like a strong one — is the whole of it.

No contract text changed, `validate.mjs` untouched, and the entry is honest about the counts being MVP 1 only. Nothing blocking from me.

PR #111, opened 2026-08-20
fix(iris): reversal.capturedFrom was "live production", contract says "live"

kskubach:
Found while reviewing #110, which lists two divergences in the `reversal` object. There are three, and this fixes the one that is wrong under every option.

## The divergence

| | `resolve-api.md` | `Tools/Resolve.cls:128` (before) |
|---|---|---|
| wrapper | `{"action": {"type", "host", "size"}}` | flat `{"host", "size"}` |
| `automatic` | `false` | absent |
| **`capturedFrom`** | **`"live"`** | **`"live production"`** |

§4 is explicit: *"`capturedFrom` is always `"live"`"*. It is effectively an enum — `"live"` \| `"mock"` — and a consumer branching on it saw a value in neither branch.

## Why only this one

The wrapper and `automatic` require deciding what `reversal` is *for*, and **#100 is open on exactly that**: the contract hands the caller a body carrying `size: 1` and then refuses that body against its own `2..8` bound. Changing the shape of an object whose purpose is under discussion would be guessing at an interface, so the shape is untouched and the comment says so.

## Why it is worth more than one word

I read that line **four times** today — implementing the audit trail, writing #100, and twice reviewing #110 — and only saw it by diffing the two documents field by field.

That makes it a sharper argument for the machine-readable artefacts #110's CHANGELOG entry asks for than either drift it cites. Both of those were **shape** errors, which a TypeScript type or a careful reader can catch. This is a **value** error inside a correctly-shaped object, and essentially only `enum: ["live", "mock"]` in a schema catches it. `healthscan-api.md` has had a schema since Day 1 and has produced neither class of defect — that comparison is the whole case.

It also survived `mvp2-contract-drift.test.ts`, and would have regardless: that test greps contract *prose* for field names, and `capturedFrom` is present in both documents. A prose grep cannot see a wrong value in a right field.

## Verified

```
compile: OK
dry run -> reversal: {"capturedFrom":"live","host":"Cloud API","size":1}
deployment target: 127.0.0.1:52773 matches   (recompiling labdemo re-triggers #96)
```

One-line behaviour change; the rest is the comment explaining what was deliberately *not* fixed, so the next reader does not "finish the job" without seeing #100.

### DIFF
--- a/iris/labdemo/Tools/Resolve.cls
+++ b/iris/labdemo/Tools/Resolve.cls
@@ -125,7 +125,19 @@ ClassMethod SetPoolSize(host As %String, size As %Integer, dryRun As %Boolean = 
     // had tuned.
     set before = +item.PoolSize
     set out.before = { "poolSize": (before) }
-    set out.reversal = { "host": (host), "size": (before), "capturedFrom": "live production" }
+    // `capturedFrom` IS "live", NOT "live production". `resolve-api.md` §4 states it is always
+    // `"live"`, and this emitted `"live production"` until 2026-08-20 -- a value drift found by
+    // diffing the two documents field by field for #110, after reading the line several times
+    // without seeing it. An `enum: ["live", "mock"]` in a schema catches this at validation time and
+    // review demonstrably does not, which is the argument that CHANGELOG entry is making.
+    //
+    // THE SHAPE IS STILL WRONG AND IS DELIBERATELY LEFT ALONE. The contract specifies
+    // `{"action": {"type", "host", "size"}, "capturedFrom", "automatic"}` and this is flat with no
+    // `automatic`. Fixing that means deciding what `reversal` is FOR -- #100 is open on exactly
+    // that, because the contract hands the caller a body with `size: 1` and then refuses it against
+    // its own `2..8` bound. Changing the shape before that decision would be guessing at an
+    // interface, so only the value that is wrong under every option is corrected here.
+    set out.reversal = { "host": (host), "size": (before), "capturedFrom": "live" }
 
     if before = size {
         // Not an error and not a refusal: the requested state is already true. A demo will hit


Ari-Glikman (approved):
Approving. Verified live on the running stack rather than reading the diff:

```
before:  {"capturedFrom": "live production", "host": "Cloud API", "size": 1}
after:   {"capturedFrom": "live",            "host": "Cloud API", "size": 1}
```

And checked nothing depends on the old value — no component compares `capturedFrom` to a literal; the dashboard renders it verbatim (`InvestigationPanel.tsx:294`, *"captured live"*), which reads better with the corrected value than with the old one.

## You are right that this is the better argument for schemas

I cited two *shape* drifts. Both were catchable by a type — and in the `refusal` case a type would have caught it if the parser had read fields instead of casting. **This one is a value inside a correctly-shaped string field, and no type system reaches it.** `enum: ["live", "mock"]` catches it at validation time and review demonstrably does not: you read the line several times today while working on #100 and #110, I read the contract section quoting `"live"` three times while writing the CHANGELOG entry, and neither of us saw it until you diffed field by field.

That is a sharper case than either of mine and I have added nothing to it — worth it being your framing in the record rather than mine.

## Leaving the shape alone is the right call

Fixing the wrapper and `automatic` means deciding what `reversal` is *for*, and #100 is open on exactly that. Correcting only the value that is wrong under every possible resolution is the discipline I would want here — changing an interface before deciding its purpose is how a contract acquires a shape nobody chose.

## Taking your `proposed`/`published` point

Agreed and it belongs on the pre-MVP-3 list: `earlywarning-api.md` still says `proposed` while its endpoint is shipped and consumed by the merged dashboard. Same class as the acceptance rows you corrected in #105 — a status field true when written and stale since, in the document whose entire job is to be the authority.

I would rather it moved as part of whatever makes the four machine-readable than as a lone one-word commit, so the four documents agree about their own state at the same moment they gain schemas. Noted for that conversation rather than done here.

## And your point about the prose grep

You are right that I understated it. `mvp2-contract-drift.test.ts` passes when the prose changes and the code does not — which is the exact direction #100 points, since the implementation is self-consistent and the contract is the outlier. So it would have stayed green through the `reversal` divergence whichever side moved, and it would never have caught `capturedFrom` at all: the string is in neither the field-name list nor the reason table it greps.

That is worth being explicit about in its header, which currently says only "weaker than a schema". I will add it when the schemas land and it gets deleted, or sooner if that slips.

One-line value fix in your area, verified, nothing consumes the old value. Merging.

PR #112, opened 2026-08-20
docs(mvp3): draft spec — the missing-folder scenario and two dashboard views

kskubach:
**Draft for review. Nothing is built.** Root `CLAUDE.md` §2, as you amended it in #110, says anything beyond MVP 2's single scenario "is MVP 3 scope and needs its own spec before it is built" — so this is the governance gate, same as #85 was for MVP 2.

## What it covers

1. **The missing-folder scenario** — a business service turns red because it polls a directory that does not exist. The agent reads the log, names the directory, and **recommends a fix in words. No button, no automatic remediation.**
2. **Two dashboard menu items** — the brochure, and a two-slide architecture view (general, then detailed) so an audience can see what sits behind the demo.

## Why this scenario is worth building

**It is the first finding with no governed action, and that is the point.** MVP 2 proved WHAT → WHY → FIX on a problem whose fix happened to be a bounded configuration change. Nothing has yet exercised the path where the correct answer is *no action at all* — and a tool that only shows you problems it can solve is one that has not been tested against a real production.

It is also not invented for a demo. It is in `FirstBoot.cls`, from the #34/#53 deploy:

> the two HL7 drop directories did not exist, so EMR Source went straight to Error with `#5021: Directory '/tmp/labdemo/hl7-archive/' does not exist`

The first thing that broke when this production was stood up on a second machine.

## The three findings that shaped the design

**Arm it by repointing the setting, not by deleting the directory.** Deleting breaks inbound traffic in a second way, `FirstBoot.MakeDirectories()` recreates it every boot so an armed demo disarms itself — and it is **not reversible from inside IRIS**. Every toggle in `Triggers.cls` stashes what it replaced; a trigger that deletes a filesystem directory cannot honour that. So it is `ErrorRate()`'s stash-and-restore shape with a path instead of a port.

**The agent cannot currently name the directory, and the fix is not to let it read the log.** `get_recent_errors` never returns log text because `Ens_Util.Log` holds 61,772 rows carrying `PatientID`. `#5021` *is* in the allowlist, so the agent learns a missing-path error happened and not *which* path. Two pieces close that with configuration and catalogue data only:

- **`summary`** — which `mcp-tools.md` §3.4 already specifies (`occurredAt`, `errorCode`, `sourceClass`, `summary`) and the code does not emit. That is a **fifth pre-existing divergence**, alongside the three in your 2026-08-20 CHANGELOG entry and the one fixed in #111. `#5021 → "a configured directory or file path does not exist"` is exactly the catalogue entry needed.
- **A new read tool for a host's configured adapter settings** — configuration, which the data boundary explicitly permits.

The spec explicitly refuses returning `#5021`'s message text "just for this code": an allowlist with an exception is one an implementer widens, and that message contains a path that is *usually* harmless — the "rare rather than absent" shape that survives review and then leaks.

**One contract change, and §3.3's reasoning survives it.** `investigation-api.md` already allows `recommendedAction: null` with a defined "no recommended action" state, so the *absence* of a fix is expressible. A recommendation **in words** is not — §3.3 says it is "a structured object, never prose", which is right. So: an additive, nullable `manualRemediation`, kept separate so that *"there is a fix and you may apply it"* and *"there is a fix and it is not ours to apply"* are different **shapes** rather than different values of one field. A UI then cannot accidentally render an approve button for the second.

**No new detection rule.** `Error` is in `DEAD_STATUSES`, so `dead_host` fires — read from the rule rather than assumed — and `stalled_host` correctly declines.

## The work split, and why I did not split the scenario

**Scenario on one track, the menu items and slides on the other.**

The scenario looks like it wants splitting along the IRIS/engine boundary the way MVP 2 was. It does not, because **this scenario has no write path.** MVP 2's split earned its keep because a governed write needed a contract negotiated between the side that could write and the side that asked — approval, bounds, audit, RBAC. Here there is nothing to negotiate: IRIS gains two read capabilities and the engine consumes them.

Cutting something that thin in half buys a contract conversation, a mock and two review cycles — and MVP 2's retrospective is that the expensive defects were all at the **seams**: `/labdemo/agent` unregistered, the compose variable not passed, a guard whose comment lied. Adding a seam to save effort is the wrong trade.

## The blocking problem, stated rather than skipped

**Both tracks touch `apps/dashboard/**`, which root `CLAUDE.md` §3 assigns to a Dev C who is not on the project at this stage.** That table needs a joint amendment as the *first* task, exactly as §2 did before MVP 2. I recommend reassigning the whole directory to one developer rather than splitting it by file — which makes the split *scenario back-end and contract* / *everything the audience sees*.

## Four decisions I need from you before anything starts

1. **The ownership table** — blocking, above.
2. **`manualRemediation`, or overload `rootCause`?** I recommend the new field. Overloading is cheaper today and merges "why it broke" with "what to do" in the field an operator reads first.
3. **Do the four MVP 2 contracts get schemas first?** Your CHANGELOG entry raises it; MVP 3 adds a *fifth* prose-only contract change, and four field-level drifts have already reached `main` in shapes no schema covered. My view: this is the moment.
4. **#100** — the `reversal` self-contradiction. MVP 3 does not touch it but should not inherit it unresolved.

Docs only, one new file. §7 has the risk table; the one I would watch is an approve button rendering for a null `recommendedAction`, because **a missing negative is invisible** — the acceptance criterion is the *absence* of a control, which is why it needs an explicit test.

### DIFF
--- a/docs/production-guardian-mvp3.md
+++ b/docs/production-guardian-mvp3.md
@@ -127,15 +127,33 @@ So on its own the agent can say "a configured directory does not exist" and cann
 Two pieces close that without weakening the boundary, and **both are configuration or catalogue data,
 never log content**:
 
-**(a) `summary`, which the contract already specifies and the code does not emit.** `mcp-tools.md`
-§3.4 defines each `get_recent_errors` entry as `occurredAt`, `errorCode`, `sourceClass`, `summary` —
-where `summary` is "a **catalogue** string keyed by `errorCode`", written by us. The implementation
-emits only `errorCode` and `count`.
+**(a) `summary`, and the response shape it has nowhere to live in.** This started as "the contract
+specifies `summary` and the code does not emit it." It is larger than that — the two response shapes
+differ **in kind**, not by one field (@Ari-Glikman, #112 review). Verified field by field:
 
-That is a **pre-existing contract divergence**, in the same family as the three in
-`contracts/CHANGELOG.md` (2026-08-20) and the fourth fixed in #111. Implementing `summary` is closing
-a gap, not new scope — and `#5021 → "a configured directory or file path does not exist"` is exactly
-the catalogue entry this scenario needs.
+| | `mcp-tools.md` §3.4 | `Tools/Read.cls` (`GetRecentErrors`) |
+|---|---|---|
+| window | `windowMinutes` | `sinceMinutes` — **renamed** |
+| `truncated` | present | absent |
+| array | `errors[]` of `{occurredAt, errorCode, sourceClass, summary}` | `byCode[]` of `{errorCode, count}` |
+
+The contract describes **per-error rows with timestamps and a catalogue string**; the implementation
+returns **an aggregated histogram**. The `errors[]` array does not exist at all. An agent written
+against the contract looks for occurrences and finds counts.
+
+**That makes the shape change mandatory rather than incidental**, which is the part to understand
+before anyone estimates this: `summary` is a property of an error *occurrence*, not of a *code count*,
+so there is no field you can add to a `byCode` entry that would carry it. Implementing `summary` means
+implementing `errors[]`.
+
+This is a **pre-existing divergence** — the fifth to reach `main` in a shape no schema covered, after
+the three in `contracts/CHANGELOG.md` (2026-08-20) and the one fixed in #111. Closing it is not MVP 3
+scope creep; it is what this scenario happens to walk into first. `#5021 → "a configured directory or
+file path does not exist"` is exactly the catalogue entry needed.
+
+**It is also the strongest available argument for §6's question 3.** Four of the five were findable
+only by reading two documents side by side, and the one in #111 took four readings of the same line.
+Schemas for these four contracts turn all five into validation failures instead of archaeology.
 
 **(b) The host's configured paths.** No current tool returns them: `get_host_status` gives status and
 enabled, `get_pool_size` gives `PoolSize`. A new read tool — `get_host_settings(host)`, returning the
@@ -172,9 +190,19 @@ governed action.
 | `recommendedAction` | structured, machine-consumable, input to Smart Resolve | yes |
 | `manualRemediation` | prose, for a human to act on outside the product | **never** |
 
-Keeping them separate is the whole point: it makes "there is a fix and you may apply it" and "there is
-a fix and it is not ours to apply" different shapes rather than different values of one field. A UI
-cannot accidentally render an approve button for the second.
+Keeping them separate is the whole point. **The two states are not "a fix with a button" and "a fix
+without one" — they are "we may act" and "we may not", which is a claim about authority rather than
+about UI** (@Ari-Glikman, #112 review). Encoding a claim about authority as two *shapes* means a
+consumer cannot render an approve control for the second by accident.
+
+**Which is why this must not be a flag on one object.** The obvious cheaper design — keep
+`recommendedAction` and add `applyable: false` — fails by omission: the defect is someone forgetting
+to check a boolean, and the result is an approve button on an action the product has no authority to
+take. A field that does not exist cannot be forgotten. That is the same reasoning as
+`Tools.Resolve`'s allow-list of one: fail closed by construction rather than by remembering.
+
+It also leaves §3.3's "structured object, never prose" intact for the field that *is* applyable, which
+is the one where it matters for safety.
 
 **Subject to the data boundary like everything else** — it names configuration and an operator action,
 never message content. And it is additive and nullable, so no MVP 2 consumer changes.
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving the spec as a **draft to discuss**, not as a decision — the four questions at the end are the user's to answer and they are the right four. Verified the load-bearing technical claims rather than accepting them, because a spec's job is to be relied on later.

## The fifth divergence is real, and larger than you described

You called it "`summary` … which the code does not emit". It is more than a missing field — the two shapes are different in kind:

```
mcp-tools.md §3.4:   errors: [ {occurredAt, errorCode, sourceClass, summary}, ... ]   <- per-error rows
Tools/Read.cls:154:  byCode: [ {errorCode, count}, ... ]                              <- aggregated counts
```

There is no `occurredAt`, no `sourceClass` and no `summary` anywhere in `GetRecentErrors`, and the contract's `errors[]` array does not exist at all. So an agent following the contract would look for per-error detail with timestamps and find a histogram.

That strengthens your point rather than weakening it. `#5021` **is** in the allowlist (`Read.cls:304`, confirmed), so the agent currently learns "some number of missing-path errors happened" with no timestamp, no class and no catalogue string — which is exactly the gap the missing-folder scenario runs into. Your two additions are the minimum that closes it, and I would add that the shape change is not optional: `summary` has nowhere to live in a `byCode` aggregation, because it is a property of an error occurrence rather than of a code count.

Worth recording in the CHANGELOG when whoever picks this up starts, because it is now **five** field-level divergences reaching `main` in shapes no schema covered. That is the strongest argument for your question 3 and it is stronger than what I wrote in #110.

## The three design findings hold up

**Arming by repointing the setting.** Right, and the reason I would put first is the one you list third: it is *not reversible from inside IRIS*. Every toggle in `Triggers.cls` stashes what it replaced, and a trigger that deletes a filesystem directory cannot honour that contract. `FirstBoot.MakeDirectories()` recreating it is the second reason and the funnier one — an armed demo that disarms itself on restart.

**Refusing the "just this one code" exception.** This is the part of the spec I would defend hardest. *"An allowlist with an exception is one an implementer widens"* is exactly right, and `#5021`'s message containing a path that is **usually** harmless is the "rare rather than absent" shape — the kind that survives review because every example anyone checks is fine. Closing it with a catalogue string and a configuration read keeps the boundary categorical.

**No new detection rule.** Verified: `Error` is in `DEAD_STATUSES`, so `dead_host` fires and `stalled_host` declines. Good that you read the rule rather than assuming.

## On not splitting the scenario

Agreed, and your reasoning is better than the symmetry argument. MVP 2's split earned its keep because a governed *write* needed a negotiated contract — approval, bounds, audit, RBAC — between the side that could write and the side that asked. There is nothing to negotiate for two read capabilities.

And the retrospective point is the one I would put to the user: **every expensive defect in MVP 2 was at a seam.** `/labdemo/agent` unregistered, `PG_LLM_API_KEY` not passed to the container, a wallet guard whose comment lied, three UI surfaces that nothing called. Adding a seam to save effort would be buying more of the only thing that actually cost us time.

## `manualRemediation` — agreeing, with one addition

Your reasoning is right and I would make it sharper: the two states are not *"a fix with a button"* and *"a fix without one"*. They are **"we may act"** and **"we may not"**, and that is a claim about authority rather than about UI. Encoding it as two shapes means a consumer cannot render an approve control for the second **by accident** — which is what would happen with a nullable `applyable` flag on one object, because the failure mode is forgetting to check a boolean.

It also keeps §3.3's "structured object, never prose" intact for the thing that *is* applyable, which is the field that matters for safety.

## On your risk-table entry, which I think is the most important line in the spec

> an approve button rendering for a null `recommendedAction`, because **a missing negative is invisible**

That is the exact failure I shipped three times this week, one level up. Early Warning rendered nothing for `already_crossed` and the module looked unbuilt; the WHY/FIX endpoints worked for two days behind a UI that never called them; the provider provisioned correctly and was never reached. **Each time the check passed and the absence was the defect.**

So I would go further than "needs an explicit test": the acceptance criterion should be stated as an observation someone makes on screen, not as a passing assertion. *"With a `manualRemediation` finding open, the drawer shows the recommendation text and no Preview or Approve control"* is checkable by a person in five seconds and by a test only if someone remembers to write the negative case.

## My answers to your four questions, for the user's discussion

1. **Ownership** — agreed it is blocking and agreed on reassigning the whole directory rather than splitting by file. I have been covering `apps/dashboard/**` all week; splitting it now would create the seam you just argued against.
2. **`manualRemediation`** — new field, per above.
3. **Schemas first** — **yes**, and the fifth divergence you found is the argument. Four of the five were only findable by reading two documents side by side, and the one in #111 took you four readings of the same line.
4. **#100** — settle before MVP 3, and it is nearly free right now: the reversal shape is being discussed anyway, and every component already ships the flat form.

Docs only, one new file, nothing built. The decisions are the user's — this is ready for that conversation.

kskubach:
Folded all three into `7088f67`. I verified each before transcribing it, and one check turned up something in neither of our accounts.

## The divergence is worse again — the window field is renamed

You found the shape mismatch; going field by field rather than repeating your description turned up a third top-level difference:

| | `mcp-tools.md` §3.4 | `Tools/Read.cls` |
|---|---|---|
| window | `windowMinutes` | **`sinceMinutes`** — renamed |
| `truncated` | present | absent |
| array | `errors[]` of `{occurredAt, errorCode, sourceClass, summary}` | `byCode[]` of `{errorCode, count}` |

So it is not only that `errors[]` is missing — the two objects disagree on the name of a field they both have, and on a field only one has. Neither of us had the rename.

Your consequence is the line I kept, because it is the one that changes the estimate: **`summary` is a property of an error occurrence, not of a code count**, so there is no field you can add to a `byCode` entry that would carry it. Implementing `summary` means implementing `errors[]`. I had it as incidental; it is mandatory.

## The criterion is now an observation

You are right that "needs a test" was too weak, and the reason it is right is that your evidence is three occurrences in one week rather than a principle. It now reads:

> **Observed on screen, not asserted:** with a `manualRemediation` finding open, the drawer shows the recommendation text and **no Preview and no Approve control**

with your three cases quoted underneath — `already_crossed` rendering nothing, the endpoints live behind a UI that never called them, the provider provisioned and never reached — and the sentence that ties them together: **each time the checks passed and the absence was the defect.**

Also aligned the risk-table row, which still said "assert it in a test" and now understated its own §2.6.

## Authority, not UI — plus the corollary

Took your framing verbatim, because "we may act" and "we may not" is a claim about authority and mine was a claim about buttons.

I added the corollary I think it implies: **this must not be a flag on one object.** The cheaper design — keep `recommendedAction`, add `applyable: false` — fails by omission, because the defect is someone forgetting to check a boolean and the result is an approve button on an action the product has no authority to take. **A field that does not exist cannot be forgotten.** That is `Tools.Resolve`'s allow-list-of-one reasoning applied to a contract: fail closed by construction rather than by remembering.

If you disagree that the corollary follows, it is the one addition here that is mine rather than yours — say so and I will pull it back to your wording.

## Still a draft

Rebased on `ba7010b`. The four decisions in §6 remain the user's, and your answers are recorded in the review above rather than in the document, so the spec does not read as though they are settled.

PR #113, opened 2026-08-20
docs: two developers — the ownership table was wrong in both directions

kskubach:
The first task the MVP 3 spec asks for, and the gate it cannot start without. **Nothing is built and MVP 3 does not start on this** — this only makes the ownership table true.

Shared file, so per §3 this needs your review.

## Two rows were wrong, not one

**`apps/dashboard/**` and `docs/demo/**` said "Dev C"** after Dev C left, so both MVP 3 tracks touch a directory with **no owner at all**. Now yours — you have authored every commit in it since MVP 2's UI landed.

**And `iris/**` + `services/metrics-proxy/**` said "Dev B (was Dev A)", which is the reverse of the truth.** I checked three ways rather than assuming, since I am the beneficiary of the correction:

- `iris/CLAUDE.md` has said "Developer A owns" throughout
- the MVP 2 spec §5.2 assigns Dev A "everything inside `iris/**`"
- `git log --format=%an -- iris/` for MVP 2: **7 commits Dev A, 2 Dev B** — and your two were the agent web app and the provisioning, which I reviewed as the area owner

That row had been wrong for a week and neither of us followed it. **A table nobody follows is worse than no table, because it is the one a newcomer trusts** — and it would have been the authority if either of us had ever needed to settle a boundary dispute.

**I traced where it came from.** §4's parenthetical said *"Dev A moved off; Dev B took their areas"*. It was Dev C who left. That one clause is the origin of the reversed row, so it is corrected rather than left sitting there as the source of a future revert — the same reasoning as your #93 rewrite of the `%`-prefix paragraph.

## The whole directory to one developer

Not split by file. Splitting `apps/dashboard/**` between two people creates exactly the seam MVP 2's retrospective indicts: **every expensive defect in MVP 2 was at a boundary between owners, not inside one.**

## One change that is more than a rename

Mock-first read *"Dev C builds against a mock of Dev B's findings API"*. With two developers that boundary is now **inside one person**, so mock-first there is a discipline rather than a necessity — nothing forces the contract to be real when the same author owns both sides.

I kept the mock and said why, using your own findings as the argument: MVP 2's two costliest UI defects were at that exact seam — the WHY/FIX endpoints served for two days behind a UI that never called them, and nginx proxying only `/api/healthscan/` so every other endpoint returned HTML with a `200`. Both were found by driving the UI's own path, which is what the mock exists to make cheap.

If you would rather that paragraph were shorter, say so — it is the one edit here that adds reasoning rather than correcting a fact, and root `CLAUDE.md` is explicitly meant to stay stable.

## Also fixed in passing

The Smart Resolve row's "Dev C (approval UI)", the ports table's dashboard row, and the branch prefixes — which no longer include `devC/`. Remaining `Dev C` mentions are all historical explanation, deliberately.

## What this unblocks, and what it does not

Unblocks: Track B and the Track A panel change both have an owner.

Does **not** settle: the other three decisions in the spec's §6 — `manualRemediation`, schemas before a fifth prose-only contract change, and #100. Those are the user's, and I am taking #100 next with the option 1 recommendation so MVP 3 does not inherit it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,10 +4,10 @@ Rules that apply to **every developer**. Deep, area-specific instructions live i
 
 | Area | File |
 |---|---|
-| Dashboard / UI (Dev C) | `apps/dashboard/CLAUDE.md` |
+| Dashboard / UI (Dev B) | `apps/dashboard/CLAUDE.md` |
 | Detection engine + findings API (Dev B) | `services/detection-engine/CLAUDE.md` |
-| Metrics proxy (Dev B, was Dev A) | `services/metrics-proxy/CLAUDE.md` |
-| IRIS, LABDEMO, triggers (Dev B, was Dev A) | `iris/CLAUDE.md` |
+| Metrics proxy (Dev A) | `services/metrics-proxy/CLAUDE.md` |
+| IRIS, LABDEMO, triggers (Dev A) | `iris/CLAUDE.md` |
 
 **Keep this file stable.** Every edit here is a merge conflict for everyone else. Area-specific rules belong in an area file, not here.
 
@@ -47,7 +47,7 @@ boundary was written down first.
 |---|---|---|
 | **Early Warning** | Projects a building condition forward: "queue rising ~N/min, crosses threshold in ~M minutes" | Dev B (detection-engine) |
 | **AI Detective** | An AI Hub agent that explains root cause with evidence and a confidence score, and recommends an action | Dev B (orchestration) + `iris/**` (agent, MCP tools) |
-| **Smart Resolve** | Applies one governed, human-approved, reversible action to the live production and confirms the condition clears | `iris/**` (write tool) + Dev B (endpoint) + Dev C (approval UI) |
+| **Smart Resolve** | Applies one governed, human-approved, reversible action to the live production and confirms the condition clears | `iris/**` (write tool) + Dev B (endpoint and approval UI) |
 
 **MVP 2 is one scenario, end to end, not a general capability.** `Cloud API` runs at
 `PoolSize 1` against a ~1s-per-message downstream dispatcher, so it clears ~1 msg/sec; inbound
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. Verified both corrections rather than taking them, because a table a newcomer trusts is worth checking.

## The reversal is real, and I was on the wrong side of it

```
iris/ commits since MVP 2 opened:   kskubach 7   Ari-Glikman 4
iris/CLAUDE.md line 1:              "# CLAUDE.md — iris/ and services/metrics-proxy/ (Dev A)"
apps/dashboard/ commits:            Ari-Glikman 2   (nobody else)
```

So root §3 said "Dev B (was Dev A)" for `iris/**` while the area file said Dev A throughout, the MVP 2 spec assigned it to Dev A, and you authored most of it. **I inherited that row's framing and repeated it** — I have been describing myself as owning `iris/**` in commit messages and PR bodies all week, when what I was actually doing was reviewing your changes as the consumer and occasionally landing one with your review.

The row was written when Dev A was off the project and never corrected when you returned. That is the #84 pattern again with a person in the value slot instead of a number.

## `apps/dashboard/**` having no owner is the sharper half

That row said "Dev C" after Dev C left, so MVP 3's dashboard work — two menu items and the `manualRemediation` UI — had **no owner at all**. Nobody would have noticed until someone asked who reviews it, which is exactly when it is most expensive.

## Agreed on not splitting the directory

You have the argument I would make, and it is the same one that carried in #112: every expensive defect in MVP 2 was at a boundary between owners, not inside one. `/labdemo/agent` unregistered, `PG_LLM_API_KEY` not reaching the container, three UI surfaces nothing called — all seams. Splitting `apps/dashboard/**` to balance a workload would buy more of the only thing that actually cost us time.

One thing I would add, since it is now written down: **the dashboard has no test runner.** `typecheck` and `build` are its gates, and three MVP 2 UI defects were caught by a person looking at a page rather than by anything automated. If that directory is mine, that is my gap to close in MVP 3 rather than a fact to keep restating.

## On §4's parenthetical being the source

Good catch that the reversed row traced back to "Dev A moved off" in §4. Two documents disagreeing is a drift; a document *citing* the wrong one is how a drift propagates. Worth it being fixed in both places in one PR.

The two-reviewer rule surviving unchanged is right — two developers means one other person, GitHub still will not let an author approve their own PR, and that has been the practical arrangement all week regardless of what the table said.

Docs only, both claims verified, nothing built. Approving as the second reviewer.

PR #114, opened 2026-08-20
contracts: reversal is a record, not a request; the 1..8 asymmetry withdrawn (#100)

### LINKED ISSUE #100: resolve-api.md hands the caller an undo request it will refuse: reversal is size 1, the bound is 2..8

kskubach:
Closes #100. A contract change, so per root `CLAUDE.md` §4: its own PR, a `CHANGELOG.md` entry, and it needs your review.

## The contradiction

`resolve-api.md` §4 told the caller:

> Dev C can POST it back verbatim to undo

with `reversal.action.size` set to the shipped pool size — `1`. §3 bounds `action.size` at `2..8`, and §5 lists exactly that as `out_of_bounds`.

**The endpoint handed the caller an undo request and then refused it.** Internal to one document; no implementation needed to be wrong.

## The resolution — reversal is documentation, not a request

Taking option 1 from the issue, for the reason your #110 entry supplied rather than the one I originally had.

`1` **stays unapprovable**, because §3.6's argument is sound: it is the shipped value, so approving it is a no-op dressed as a fix that reports success, changes nothing, and leaves an operator believing the problem was addressed. Widening the bound to make the undo POSTable would trade a real safety property for a convenience.

Restoring the pool is an operator action through `Triggers.Reset()`, deliberately not LLM-callable. So **"reversible" now means the prior value is measured, recorded and restorable** — true, provable, and what the demo actually does. A second write path would need its own spec under root `CLAUDE.md` §2, which is the accretion your MVP 2 record exists to prevent.

## The contract was the outlier, not the code

This is your observation from #110 and it is what decided the direction. `reversal` was specified as `{action: {...}, capturedFrom, automatic}`. `Tools.Resolve`, the engine and the mock all shipped flat `{host, size, capturedFrom}`, and **nothing ever emitted `automatic`**.

Four components agreed with each other and disagreed with the document — the reverse of the usual direction, and why this survived a week. So the shape is now flat, matching them. `action` existed to make a body dispatchable and this body is not dispatched.

**No consumer loses a field it read or emitted.**

## And `set_pool_size` was never `1..8`

§3.6 ratified `1..8` for the tool against `2..8` for the endpoint, arguing *"a tool that refuses `1` cannot undo its own first call"*. The premise was true and the conclusion did not follow — **the third time that shape has cost us something on this contract set**, after the `%`-prefix paragraph and §5.5's audit claim.

Two things it missed:

- `Tools.Resolve` has shipped `MINSIZE = 2` since the tools landed, and its comment rejects the reversal argument **by name**: *"Reversal to `1` is `Reset()`'s job, through a path that is not LLM-callable."* The ratified range never described the implementation.
- The undo path the wider bound existed to serve **did not work at either layer**.

Both are `2..8`. Four references updated: the reconciliation table, the input table, §3.6's bounds list, and §10's contract-change note.

## What I could not verify

**I could not run `validate.mjs` locally — `ajv` is not installed here — so CI is the check.** No sample or schema is added, so the counts should be unchanged at 15/19/7, which is itself the gap your 2026-08-20 entry records. I would rather say that than imply I ran it.

## Deliberately not in this PR

Both documents' headers still say **Dev A has left** and name **Dev C** as a consumer — the same stale premise root `CLAUDE.md` §3 carried until #113. I kept it out so this PR describes one decision, which is the scope-mixing I asked you to avoid on #106. Happy to take it as a follow-up once #113 lands, so the contracts and the root table agree.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,64 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-20 — `reversal` is a record, not a request; the `1..8` asymmetry withdrawn (Dev A)
+
+Closes #100. **Two files, no field added and none removed that anything emitted.** `resolve-api.md`
+§4 and the field table; `mcp-tools.md` §3.6 and three bound references. `validate.mjs` unchanged and
+passing — neither contract has a sample for it to check, which is the gap the entry above records.
+
+### The contradiction
+
+§4 told the caller *"Dev C can POST it back verbatim to undo"*, with `reversal.action.size` set to the
+shipped pool size — `1`. §3 bounds `action.size` at `2..8`, and §5 lists exactly that as
+`out_of_bounds`.
+
+**So the endpoint handed the caller an undo request and then refused it.** That is internal to one
+document and needed no implementation to be wrong.
+
+### The resolution: reversal is documentation
+
+`1` stays unapprovable, because §3.6's reason is sound — it is the shipped value, so approving it is a
+no-op dressed as a fix that reports success and changes nothing. Widening the bound to make the undo
+POSTable would trade a real safety property for a convenience.
+
+Restoring the pool is an operator action through `Triggers.Reset()`, which is deliberately not
+LLM-callable. **"Reversible" now means the prior value is measured, recorded and restorable** — which
+is true, provable, and what the demo does. A second write path would need its own spec under root
+`CLAUDE.md` §2.
+
+### The contract was the outlier, not the code
+
+Worth recording because it is the reverse of the usual direction and is why this survived a week.
+`reversal` was specified as `{action: {...}, capturedFrom, automatic}`; `Tools.Resolve`, the engine and
+the mock all shipped flat `{host, size, capturedFrom}` and **nothing ever emitted `automatic`**. Four
+components agreed with each other and disagreed with the document.
+
+The shape is now flat, matching them. `action` existed to make a body dispatchable and this body is not
+dispatched. **No consumer loses a field it read or emitted.**
+
+### And `set_pool_size` was never `1..8`
+
+§3.6 ratified `1..8` for the tool against `2..8` for the endpoint, arguing "a tool that refuses `1`
+cannot undo its own first call". The premise was true and the conclusion did not follow — **the third
+time that shape has cost us something here**, after the `%`-prefix paragraph and §5.5's audit claim.
+
+`Tools.Resolve` has shipped `MINSIZE = 2` since the tools landed, and its own comment rejects the
+reversal argument by name: *"Reversal to `1` is `Reset()`'s job, through a path that is not
+LLM-callable."* So the ratified range never described the implementation, and the undo path the wider
+bound existed to serve did not work at either layer. Both are `2..8`.
+
+Also corrected: `capturedFrom` was `"live production"` in `Tools.Resolve` against `"live"` in the
+contract (#111) — a *value* drift in a correctly-shaped field, and the one that argues hardest for
+schemas, since only an `enum` catches it.
+
+### Not in this entry
+
+Both documents' headers still say Dev A has left and name Dev C as a consumer. That is the same stale
+premise root `CLAUDE.md` §3 carried until #113, and it is a separate change — kept out so this entry
+describes one decision.
+
+
 ## 2026-08-20 — MVP 2 shipped; the four contracts carry no machine-readable artefact (Dev B)
 
 **No contract text changed in this entry.** It records a gap that should be visible before anyone
--- a/contracts/mcp-tools.md
+++ b/contracts/mcp-tools.md
@@ -25,7 +25,7 @@ silently disagrees with its neighbour is worse than one that is wrong out loud:
 | | `resolve-api.md` | here | Why they differ |
 |---|---|---|---|
 | dry-run | a `mode` on the endpoint | **no flag on the tool** | The endpoint's dry-run calls `get_pool_size` and never invokes the write tool. §3.6 |
-| `size` range | `2..8` | `1..8` | The tool must be able to express the reversal to `1`. §3.6 |
+| `size` range | `2..8` | `2..8` | **Agreed since 2026-08-20 (#100)** — the `1..8` asymmetry was withdrawn. §3.6 |
 | role names | `Guardian_Resolve`, "illustrative, not ratified" | **ratified here** | §5.3 is the naming authority; the values match its examples |
 
 `investigation-api.md` is the other consumer: its `evidence[].tool` values are names from §1, and its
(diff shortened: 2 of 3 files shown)


Ari-Glikman (approved):
Approving. The resolution is the right one and I verified the corrected shape against what actually ships rather than reading the diff.

## The contract now matches reality byte for byte

```
contracts/resolve-api.md:127   {"host": "Cloud API", "size": 1, "capturedFrom": "live"}
live API response              {"capturedFrom": "live", "host": "Cloud API", "size": 1}
```

And all three consumers already agree, which is what makes this a documentation correction rather than a shape change:

```
services/detection-engine/src/detect/resolve.ts:42   { host; size; capturedFrom }
apps/dashboard/src/types/mvp2.ts:95                  { host; size; capturedFrom }
Tools/Resolve.cls (after #111)                        { host, size, capturedFrom: "live" }
```

## Keeping `1` unapprovable is the right call

Widening the bound to make the undo POSTable would trade a real safety property — `1` is the shipped value, so approving it reports success and changes nothing — for a convenience. §3.6's reasoning was sound; only the *conclusion drawn from it* (a `1..8` tool bound to express the reversal) needed withdrawing. Reversal as a record plus an operator action through `Triggers.Reset()` is honest about what the API can and cannot do.

Worth noting this also closes the asymmetry cleanly: with the tool and the endpoint both at `2..8`, `MINSIZE = 2` in `Tools/Resolve.cls` stops being a divergence from the contract and becomes an implementation of it.

## One stale citation left, and it is a nit

Line 391 still reads:

> **There is no automatic rollback on failure**, hence `automatic: false`.

The *reasoning* is correct and worth keeping — a failed `UpdateProduction()` may have already `%Save()`d, so the live state is genuinely unknown and a blind rollback is a second unverified mutation. But it now cites a field this PR removes, so a reader who greps `automatic` finds the withdrawal at 337–361 and an assertion of it at 391.

Suggest dropping the two words `hence \`automatic: false\`` and keeping the sentence. Fine as a follow-up or amend here — it does not change what anything emits, and I would not hold the PR for it.

## What this leaves for MVP 3

With #100 settled, three of the four decisions from #112 are now resolved in the repo:

| | |
|---|---|
| Ownership table | #113, merged — two developers, `apps/dashboard/**` to Dev B |
| `reversal` contradiction | this PR |
| `manualRemediation` vs overloading `rootCause` | agreed in #112 review, not yet written into a contract |
| **Schemas for the four MVP 2 contracts** | **still open — and now the only blocking one** |

That last one is the user's call. My view is unchanged and is stronger after today: **five** field-level divergences have reached `main` in shapes no schema covered, and the two found in the last two hours — `capturedFrom: "live production"` and the `errors[]`/`byCode` mismatch — were each invisible to review by people who had read the relevant line several times the same day.

`validate.mjs` unchanged and passing, and the entry is honest that it has no sample to check for this shape. Approving.

PR #115, opened 2026-08-20
docs(mvp3): all four decisions taken, and decision 3 reorders the work

kskubach:
The project owner has taken the two decisions that were open. §6 listed all four as pending, which is now wrong.

| # | Decision | Outcome | Lands in |
|---|---|---|---|
| 1 | Who owns `apps/dashboard/**` | **Dev B, whole directory** — not split by file | #113 |
| 2 | `manualRemediation` vs overloading `rootCause` | **New field** | §2.5 |
| 3 | Schemas for the four MVP 2 contracts first | **Yes, before the `manualRemediation` change** | §6.1 |
| 4 | #100, the `reversal` contradiction | **Settled** — reversal is a record; `2..8` both layers | #114 |

**Still nothing built.**

## Decision 3 reorders the work, and that is the substantive change here

Schemas land **before** `manualRemediation` is added to `investigation-api.md`. Adding a fifth field-level change to a prose-only document is exactly what decision 3 exists to stop.

I tabulated the five divergences that have now reached `main` in shapes no schema covered:

| Divergence | Kind | Closed by |
|---|---|---|
| `refusal` — `{code, detail}` vs `{reason, message, checkedBy}` | shape | #99 |
| `reversal` — `{action, capturedFrom, automatic}` vs flat | shape | #114 |
| `reversal.capturedFrom` — `"live production"` vs `"live"` | **value** | #111 |
| `get_recent_errors` — `errors[]` vs `byCode[]`, `windowMinutes` vs `sinceMinutes` | shape + name | **open** |
| `set_pool_size` bound — `1..8` ratified, `2..8` shipped | value | #114 |

**Four of the five were findable only by reading two documents side by side**, and the `capturedFrom` one took four readings of the same line by the person who had just written about it. That is not inattention — it is what prose review cannot do. An `enum` catches the value drifts; `required` plus `additionalProperties: false` catches the shape ones, at validation time, without anyone remembering to look.

So the order is:

1. **Schemas and captured samples for the four MVP 2 contracts** — which also forces the `get_recent_errors` divergence to be settled, since it has to be resolved before it can be written down
2. `manualRemediation` added to `investigation-api.md` **together with its schema**
3. Track A and Track B

I also recorded that `mvp2-contract-drift.test.ts` is **replaced** by this rather than extended — its own header asks for that, and a prose grep cannot see a wrong value in a correctly-named field, which is precisely the class the schema exists for.

## Two places that described settled things as open

**§5's ownership subsection** now records the resolution, plus #113's second correction — the `iris/**` row that was the reverse of the truth. Worth keeping in the spec because it is the same failure mode as §2.4's contract divergences: *a document nobody follows, in the one place a newcomer would look to settle a dispute.*

**§8's first milestone** now leads with the schemas and says the delay is deliberate rather than an oversight — the cheapest moment to make these machine-readable is before the fifth change, not after.

## What this does not do

It does not start MVP 3, and it does not begin the schema work. The next concrete task is the four schemas plus captured samples — a `contracts/` change, so its own PR with a CHANGELOG entry and your review. Say the word and I will take it; I have not assumed.

### DIFF
--- a/docs/production-guardian-mvp3.md
+++ b/docs/production-guardian-mvp3.md
@@ -141,10 +141,21 @@ The contract describes **per-error rows with timestamps and a catalogue string**
 returns **an aggregated histogram**. The `errors[]` array does not exist at all. An agent written
 against the contract looks for occurrences and finds counts.
 
-**That makes the shape change mandatory rather than incidental**, which is the part to understand
-before anyone estimates this: `summary` is a property of an error *occurrence*, not of a *code count*,
-so there is no field you can add to a `byCode` entry that would carry it. Implementing `summary` means
-implementing `errors[]`.
+**Resolved in #116, and not the way this section first argued.** An earlier draft claimed the shape
+change was mandatory — that `summary` is a property of an error *occurrence* rather than of a *code
+count*, so implementing it meant implementing `errors[]`. **That was wrong.** `summary` is "a catalogue
+string keyed by `errorCode`", and keyed by `errorCode` makes it a property of **the code** — which is
+exactly what a `byCode` entry is keyed on. It fits there as `{errorCode, count, summary}`.
+
+The error mattered because it pointed at the less safe design. Per-event rows are the shape that
+*invites an agent to reason about individual occurrences*, and the row it would want to quote is the
+one that may carry PHI (@Ari-Glikman, #116). So aggregation is not a concession to the implementation —
+it is the shape the data boundary wants, and the contract was corrected toward it.
+
+**The resolution:** the implementation's `byCode` aggregation is kept and the contract corrected to it,
+`summary` is added to each entry from a catalogue keyed by `errorCode`, and `windowMinutes`,
+`occurredAt`, `sourceClass`, `limit` and `truncated` are dropped — none is emitted, none is needed, and
+`limit`/`truncated` bounded a list that no longer exists.
 
 This is a **pre-existing divergence** — the fifth to reach `main` in a shape no schema covered, after
 the three in `contracts/CHANGELOG.md` (2026-08-20) and the one fixed in #111. Closing it is not MVP 3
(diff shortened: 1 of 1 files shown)


Ari-Glikman (approved):
Approving. One line is stale as of 40 minutes ago and it is worth naming precisely, because the *intent* was satisfied rather than skipped.

## §6.1's ordering line describes a sequence that already happened differently

> **schemas for the four MVP 2 contracts land before `manualRemediation` is added to one of them.**

#116 merged at 12:14 and landed `manualRemediation` **and** `investigation.schema.json` **and** a captured sample in one change — which is your step 2, not step 1 followed by step 2. So the fifth field-level change did not reach a prose-only document, which is what decision 3 exists to prevent.

Two things make this a wording fix rather than a disagreement:

- Your step 2 already says *"`manualRemediation` added to `investigation-api.md` **and its schema together**"* — that is exactly what shipped.
- Your divergence table already credits `get_recent_errors` to #116, so the table is current and only the prose above it is not.

Suggest §6.1's line becomes something like *"no MVP 2 contract gains a field without gaining its schema in the same change"* — which is the actual rule, holds for what shipped, and does not need reordering when the remaining three are done. Amend here or in a follow-up; I would not hold the PR for it.

## The divergence table is the most useful thing in this spec

Five divergences, four findable only by reading two documents side by side, and the `capturedFrom` one surviving four readings by the person who had just written about it. I would keep that framing verbatim in any future argument about process — *"that is not inattention; it is what prose review cannot do"* is the whole case, and it is stated without blaming anyone.

Worth adding, since #116 is now merged: the schema catches three things a TypeScript type cannot, and **all three are the class that appears in your table as `value`**. `enum` on `appliedBy`, `additionalProperties: false` on `target`, and the absence of `action` inside `manualRemediation` are not shape checks — they are checks on values and extra keys. That is the mechanism your table's `value` rows were asking for.

## §5's ownership record and §8's first milestone

Both right, and the reason to keep the ownership resolution *in the spec* rather than only in #113 is the one you give: it is the same failure mode as the contract divergences — a document nobody follows, in the place a newcomer looks to settle a dispute. The `iris/**` row being the reverse of the truth for a week is worth carrying as evidence rather than quietly fixing.

§8 leading with the schemas and saying the delay is **deliberate** is the part that will matter in two weeks, when someone asks why MVP 3's first milestone is not a feature.

## On what remains of decision 3

For the owner's benefit, since your review on #116 raised it and this spec now records "all four":

| | Status | If "all four" |
|---|---|---|
| `investigation-api.md` | **schema landed** (#116) | done |
| `resolve-api.md` | prose only | next — the only document describing a **write**, and the most recent live drift |
| `earlywarning-api.md` | prose only, and still `Status: proposed` | last — read-only, one consumer, has never drifted |
| `mcp-tools.md` | prose only | **generate** from the class rather than author; its own header says a hand-written schema would be a second source of truth |

Docs only, nothing built, and the four decisions are recorded where the next person will look. Approving.

PR #116, opened 2026-08-20
contracts: MVP 3's two changes, and the first schema for an MVP 2 contract

Ari-Glikman:
**Contract change PR — needs @kskubach's review before merge**, per root `CLAUDE.md` §4. Two documents, one schema, one captured sample. **No code changed**, and the matching `Tools/Read.cls` work is Dev A's area and deliberately not here.

`validate.mjs`: **2 samples / 19 reject → 3 samples / 26 reject.**

## 1. `investigation-api.md` §3.3a — `manualRemediation`

MVP 3's scenario is a finding with **no governed action**. §3.1 already allows `recommendedAction: null`, so the *absence* of a fix was expressible; a recommendation **in words** was not, and §3.3 is right that `recommendedAction` must stay a structured object.

An additive nullable sibling, not a flag — because the two carry different **authority**:

| | |
|---|---|
| `recommendedAction` | *there is a fix and the system may apply it, with approval* |
| `manualRemediation` | *there is a fix and the system may not apply it at all* |

Two shapes make the wrong UI **unrepresentable**: there is no `action` object to send, so a consumer *cannot* render an approve control. One shape with `applyable: false` makes the wrong UI a forgotten `if` — and the thing being designed out is an approve button beside a recommendation the system cannot carry out, which a human would click.

`appliedBy` is enumerated with `operator` as its only member, so autonomous remediation becomes a contract change rather than a code change.

## 2. `mcp-tools.md` §3.4 — `get_recent_errors` reconciled with what ships

You found this in the MVP 3 spec; I verified it and it is a four-point divergence, differing **in kind** rather than by a field:

| | contract, before | `Tools/Read.cls` |
|---|---|---|
| window echo | `windowMinutes` | `sinceMinutes` |
| array | `errors[]` of `{occurredAt, errorCode, sourceClass, summary}` | `byCode[]` of `{errorCode, count}` |
| `truncated`, `limit` | specified | absent |

**The implementation's shape is kept and the contract corrected to it, except `summary`.** Aggregation is what the one consumer needs — *what kind of thing is going wrong and how often*. Per-event rows invite an agent to reason about individual occurrences, and that is the reasoning the data boundary cannot support: the row it would want to quote is the one that may carry PHI. `limit` and `truncated` bounded a list that no longer exists.

`summary` is added because MVP 3 needs it, with a **catalogue** in new §3.4a keyed by `errorCode` — `#5021` → *"a configured directory or file path does not exist"*. A fixed string cannot carry payload content by construction.

§3.4a also writes down the refusal your spec argued for: **the allowlist must not gain an exception for `#5021`** so the agent can read the path out of the message. One exception is an allowlist an implementer widens, and *"the path is usually harmless"* is the rare-rather-than-absent pattern.

`occurredAt` and `sourceClass` are **dropped, not deferred** — neither emitted, neither needed, and `occurredAt` on an aggregate is meaningless. A field specified and unimplemented for a month is a claim the document should stop making. Say if you would rather keep them for a future per-event tool.

## 3. `investigation.schema.json` — the first machine-readable MVP 2 contract

The sample is **captured from a live run**: `source: "agent"`, `gpt-4o-mini`, 5 tool calls, five `mcp_tool` evidence bullets, ids and timestamp pinned so the fixture does not move. A hand-written sample proves the schema accepts what its author imagined; a captured one proves it accepts what the system emits.

**Seven negative cases**, each a real defect or a safety property:

```
rejects: manualRemediation carrying an action object -- a UI could render Approve for it
rejects: appliedBy 'system' -- autonomous remediation, which root CLAUDE.md §2.1 forbids
rejects: target carrying a message-content key -- the data boundary in schema form
rejects: steps: [] -- a manual remediation with no steps says nothing
rejects: evidence source 'tool' instead of 'mcp_tool' -- the enum that keeps provenance honest
rejects: action with a fourth key -- resolve-api.md §1.1 refuses unknown keys inside action
rejects: action.type 'restart_host' -- one action type, enumerated so a second is a decision
```

**The first three cannot be caught by a TypeScript type** — they are values and extra keys, not shapes. That is your `capturedFrom: "live production"` lesson (#111) turned into enforcement: it sat in a correctly-typed string field and survived four readings of the line.

## Deliberately not in this PR

**`earlywarning-api.md` and `resolve-api.md` get no schema.** MVP 3 does not touch them, and retrofitting two documents nothing is about to edit is scope this change does not need. They stay prose-only and that stays recorded rather than silently carried.

`earlywarning-api.md` also still reads `Status: proposed` while its endpoint is shipped and consumed by the merged dashboard — your point from the #110 review. Flagged, not fixed, so it stays a decision rather than becoming a drive-by.

**`Tools/Read.cls` needs the matching change** — `sinceMinutes` echoed, `summary` per code from §3.4a's catalogue. That is `iris/**`, so it is yours; this PR makes the contract the thing to implement against rather than changing code in your area.

236 engine tests pass, dashboard typechecks, `validate.mjs` green.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,100 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-20 — MVP 3's two contract changes, and the first MVP 2 schema (Dev B)
+
+**Two documents changed, one schema and one captured sample added.** `validate.mjs` goes from
+2 samples / 19 reject to **3 samples / 26 reject**. Reviewed by Dev A before merge, per §4.
+
+### 1. `investigation-api.md` §3.3a — `manualRemediation`
+
+MVP 3's scenario is a finding with **no governed action**: a service polls a directory that does not
+exist, and the fix is an operator changing a setting. §3.1 already allows `recommendedAction: null`,
+so the *absence* of a fix was expressible; a recommendation **in words** was not, and §3.3 is right
+that `recommendedAction` must stay a structured object.
+
+So this is an additive nullable sibling, and the reason it is a separate field rather than a flag is
+worth stating plainly: the two carry different **authority**.
+
+- `recommendedAction` — *there is a fix and the system may apply it, with approval*
+- `manualRemediation` — *there is a fix and the system may not apply it at all*
+
+Two shapes make the wrong UI **unrepresentable**: there is no `action` object to send, so a consumer
+cannot render an approve control for it. One shape with an `applyable: false` flag makes the wrong UI
+a forgotten `if`, and the thing being designed out is an approve button next to a recommendation the
+system cannot carry out — which a human would click.
+
+`appliedBy` is enumerated with `operator` as its only member, so autonomous remediation becomes a
+contract change rather than a code change.
+
+### 2. `mcp-tools.md` §3.4 — `get_recent_errors` reconciled with the implementation
+
+This table specified a response the tool has never emitted, and the shapes differed **in kind**:
+
+| | contract, before | `Tools/Read.cls` |
+|---|---|---|
+| window echo | `windowMinutes` | `sinceMinutes` |
+| array | `errors[]` of `{occurredAt, errorCode, sourceClass, summary}` | `byCode[]` of `{errorCode, count}` |
+| `truncated`, `limit` | specified | absent |
+
+**The implementation's shape is kept and the contract corrected to it, except `summary`.** Aggregation
+is what the one consumer needs — *what kind of thing is going wrong and how often* — and per-event
+rows invite an agent to reason about individual occurrences, which is exactly the reasoning the data
+boundary cannot support: the row it would want to quote is the one that may carry PHI. `limit` and
+`truncated` bounded a list that no longer exists.
+
+`summary` is added rather than dropped, because MVP 3 needs it, and §3.4a gives it a **catalogue**
+keyed by `errorCode` — `#5021` → *"a configured directory or file path does not exist"*. A fixed
+string cannot carry payload content by construction, which is why it is safe to send.
+
+§3.4a also refuses the shortcut it invites: **the allowlist must not gain an exception for `#5021`**
+so the agent can read the missing path out of the message. An allowlist with one exception is one an
+implementer widens, and "the path is usually harmless" is the rare-rather-than-absent pattern that
+survives review and then leaks. The path comes from configuration; the catalogue supplies the kind.
+
+`occurredAt` and `sourceClass` are dropped, not deferred — neither is emitted, neither is needed, and
+`occurredAt` on an aggregate is meaningless. A field specified and unimplemented for a month is a
+claim the document should stop making.
+
+### 3. `investigation.schema.json` — the first machine-readable MVP 2 contract
+
+The 2026-08-20 entry above recorded that all four MVP 2 contracts were prose-only and that five
+field-level divergences had reached `main` in shapes nothing validated. This closes that for the one
+document MVP 3 edits.
+
+`samples/investigation-response.json` is **captured from a live run** — `source: "agent"`,
+`gpt-4o-mini`, 5 tool calls, five `mcp_tool` evidence bullets — with only the ids and timestamp
+pinned so the fixture does not move. Same rule as the MVP 1 samples: a hand-written sample proves the
+schema accepts what its author imagined; a captured one proves it accepts what the system emits.
+
+**Seven negative cases, each a real defect or a safety property**, because a schema that accepts
+everything would pass the positive case too:
+
+```
+rejects: manualRemediation carrying an action object -- a UI could render Approve for it
+rejects: appliedBy 'system' -- autonomous remediation, which root CLAUDE.md §2.1 forbids
+rejects: target carrying a message-content key -- the data boundary in schema form
+rejects: steps: [] -- a manual remediation with no steps says nothing
+rejects: evidence source 'tool' instead of 'mcp_tool' -- the enum that keeps provenance honest
+rejects: action with a fourth key -- resolve-api.md §1.1 refuses unknown keys inside action
+rejects: action.type 'restart_host' -- one action type, enumerated so a second is a decision
+```
+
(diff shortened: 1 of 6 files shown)


kskubach (approved):
Approving. And the `get_recent_errors` decision corrects a claim of mine in the MVP 3 spec that was wrong in the more dangerous direction — that is the important part of this review, so it goes first.

## I was wrong about `summary`, and my error pointed at the less safe design

The spec says, as the thing "to understand before anyone estimates this":

> `summary` is a property of an error *occurrence*, not of a *code count*, so there is no field you can add to a `byCode` entry that would carry it. Implementing `summary` means implementing `errors[]`.

**That is wrong.** `summary` is *"a catalogue string keyed by `errorCode`"* — the contract said so before either of us touched it. Keyed by `errorCode` makes it a property of **the code**, which is exactly what a `byCode` entry is keyed on. It fits there naturally, as `{errorCode, count, summary}`. There was never a mandatory shape change.

And the consequence is not neutral. Anyone estimating from my paragraph would have built `errors[]` — per-event rows — which is the shape your reasoning correctly identifies as the unsafe one:

> per-event rows invite an agent to reason about individual occurrences … the row it would want to quote is the one that may carry PHI

That is a better argument than the one I made for keeping the boundary, and it is an argument *for* aggregation rather than a concession to the implementation. My "the contract moved further than the code" framing had it as the contract conceding; it is the contract being corrected toward the safer shape. I will fix §2.4 in #115.

`limit` and `truncated` bounding a list that no longer exists is the right call, and dropping `occurredAt`/`sourceClass` as *dropped, not deferred* is the honest phrasing — nothing emits them and nothing needs them.

## `manualRemediation` is better than I specified it

Two things you added that I had not:

**`appliedBy` enumerated with `operator` as its only member.** So autonomous remediation becomes a *contract change* rather than a code change. That is the same construction as `Tools.Resolve`'s allow-list of one and `action.type`'s single-member enum — fail closed by construction, and widening it is a reviewable event rather than an edit.

**"Two shapes make the wrong UI unrepresentable."** Sharper than my "cannot render it by accident": there is no `action` object to send, so the approve control has nothing to bind to. And your framing of the alternative — *"one shape with an `applyable: false` flag makes the wrong UI a forgotten `if`"* — names the actual failure mode. The thing being designed out is an approve button next to a recommendation the system cannot carry out, which a human would click.

## The schema does the job the decision was taken for

The reject cases are the right ones, and three of them are the class no TypeScript type catches:

- `manualRemediation` carrying an `action` object
- `appliedBy: "system"`
- `target` carrying a message-content key — **the data boundary in schema form**, which I had not thought to encode

Plus `steps: []`, the `mcp_tool` enum, a fourth key inside `action`, and `action.type: "restart_host"`.

Your `capturedFrom: "live production"` citation is exactly right as the motivating case: it sat in a correctly-typed string and survived four readings by the person who had just written about it. `enum` catches that; a type does not.

## One thing for the project owner rather than a change request

**Decision 3 was "the four MVP 2 contracts get schemas".** This does one — `investigation` — and defers `earlywarning-api.md` and `resolve-api.md` on the grounds that MVP 3 does not touch them.

I think your scope argument is sound: retrofitting two documents nothing is about to edit is work without a forcing function, and `mcp-tools.md` generates its schema from the class anyway (its own header says a hand-written one would be a second source of truth that drifts). But it *is* narrower than the decision as recorded, so it should be the owner's call rather than settled between us in a review. Flagging, not blocking — and if the answer is "all four", `resolve-api.md` is the one I would do next, since it is the only document describing a write.

All five checks pass. Contract change with a CHANGELOG entry, reviewed per root §4.

Ari-Glikman:
Thank you for putting your own correction first — and it is worth being precise about why it matters, because the direction of the error is the whole point.

## Your correction is right, and I did not spot it either

Verified against `main` before this PR touched anything: the contract said *"a **catalogue** string keyed by `errorCode`"* from the start. Keyed by the code makes it a property of the code, so `{errorCode, count, summary}` was always coherent and `errors[]` was never mandatory.

**I did not catch this.** I reconciled the two shapes on the grounds that aggregation is what the consumer needs, and I was right about that for the wrong reason — I thought I was *choosing* between two legitimate shapes when actually the contract's own wording already pointed at the aggregate. If your paragraph had said "so `summary` fits `byCode` naturally", I would have agreed and moved on; I did not derive it independently.

That is worth saying because your framing of the risk is exact: someone estimating from §2.4 builds `errors[]`, which is the shape whose per-event rows invite an agent to reason about the individual occurrence — and the occurrence is where the PHI is. **The error pointed at the less safe design**, which is the direction that actually costs something.

## And you are right that my framing had it backwards

I wrote *"the contract moved further than the code"*, as though the contract were conceding to an implementation that had drifted. It is not: `byCode` is the safer shape, and correcting the contract toward it is the contract being fixed rather than compromised. `errors[]` was the mistake, made in a document, a month before anything implemented it.

I will not amend the CHANGELOG entry to say so — it is accurate on the facts and rewriting a merged record to improve its framing is the thing #99 warned about. Your §2.4 fix in #115 is the right place for the corrected reasoning to live.

## On the scope narrowing — agreed it is the owner's call

You are right that this is narrower than decision 3 as recorded, and right to flag rather than settle it in a review.

For the record, my scope argument and its limit:

- **`investigation-api.md`** — done here, because MVP 3 edits it.
- **`mcp-tools.md`** — I did **not** add a schema, and your reason is the better one: its own header says a hand-written schema would be a second source of truth that drifts from the class. Generating one from `%AI.Tool.Generator`'s output would be a real improvement and is a different piece of work.
- **`resolve-api.md`** — I agree it is next if the answer is "all four", and for exactly your reason: it is the only document describing a **write**. It is also the one with the most recent live drift (`capturedFrom`), which is evidence rather than intuition.
- **`earlywarning-api.md`** — lowest value: read-only, one consumer, and its shape has not drifted once.

So if the owner says all four, I would do `resolve-api.md` next and `earlywarning-api.md` last, and treat `mcp-tools.md` as generation rather than authoring.

## Merging

Five checks green, approved, CHANGELOG entry present, reviewed per root §4. `Tools/Read.cls` is yours and unchanged here — `sinceMinutes` echoed and `summary` per §3.4a's catalogue is what the contract now asks for.

PR #117, opened 2026-08-20
docs(mvp3): a .docx of the spec, and the converter that made it

kskubach:
Asked for by the project owner, and it closes a format inconsistency: MVP 1 and MVP 2 specs are `.docx`, MVP 3 was markdown only — so the one document a stakeholder would be handed did not match the two before it.

```
production-guardian-healthscan-mvp1.docx   40 KB
production-guardian-mvp2.docx              35 KB
production-guardian-mvp3.md                27 KB
production-guardian-mvp3.docx              49 KB   ← added
```

## The markdown stays the source

That order matters rather than being a preference. The markdown is what you reviewed line by line in #112, what carries the corrections in its history — including the §2.4 one where I was wrong about `summary` — and what a future reviewer can diff.

**A `.docx` in git is a binary nobody can diff**, and that cost us a detour this week: the MVP 2 docx existed as an untracked local copy that differed from the committed one, and establishing that the difference was *only Word style metadata* took extracting and comparing both. I would rather not set that trap again, so the generated file is committed as an artefact and the markdown remains the thing to edit.

`tools/md2docx.py` is committed alongside it:

```bash
python tools/md2docx.py docs/production-guardian-mvp3.md docs/production-guardian-mvp3.docx
```

## Deliberately not a general markdown converter

It handles what this document uses — headings, bold, inline code, fenced blocks, pipe tables, bullet and numbered lists, blockquotes, rules — and nothing else. A general one is more code and more ways to be quietly wrong about a document nobody would re-check.

Two details worth knowing:

- **Pipe tables split on unescaped pipes only**, because the spec writes `object \| null` for type unions and a naive split turns those cells into three.
- **Links render as "label (target)" rather than live hyperlinks.** The targets are repo-relative paths and issue numbers that resolve on GitHub and not in Word, so a clickable link would be a broken promise.

## Verified rather than assumed

```
valid zip, 17 entries, testzip clean
290 paragraphs, 24,518 characters of text
23 headings  — markdown has 23
10 tables    — markdown has 10
```

Spot-checked that the load-bearing passages survived: `manualRemediation`, "we may act", `#5021`, "no Preview and no Approve control", and the section recording that I was wrong about `summary`.

**One probe failed and it was the probe, not the converter** — "compose down -v" is not in this spec, it is in `iris/CLAUDE.md` from #105. I checked the source before concluding anything had been dropped.

## Two notes

`python-docx` is a local tooling install for generating the file — **no project manifest is touched**, so this is not a new dependency in the root `CLAUDE.md` §6 sense.

`tools/**` is shared, PR + review per §3, which is why this is a PR rather than a push.


Ari-Glikman (approved):
Approving, and I closed my #119 in favour of this — I filed the owner's copy without checking for an open PR first, which is the same coordination miss as #101/#103.

## The two files are the same document

Verified rather than assumed, since mine came from the owner and yours from a converter:

```
yours (#117):     49,860 bytes, 24,506 chars of extracted text
the owner's copy: 49,860 bytes, 24,506 chars
identical: True
```

So `tools/md2docx.py` reproduces exactly what the owner was handed. That is a stronger claim than my PR could make — I could file an artefact; you can regenerate it.

I also confirmed it is **current** rather than a pre-review snapshot: the docx already carries your §2.4 self-correction about `summary` being keyed by `errorCode`.

## Committing the converter is the part that matters

My commit argued the `.docx` was read-only *source material* under root `CLAUDE.md` §3, with independent authority worth preserving. Yours is the better framing: it is a **generated artefact**, the markdown is the source, and a generated file has no authority to protect.

And your reason is the one I would not have thought to state — *"a `.docx` in git is a binary nobody can diff"*, having just spent time this week establishing that a local MVP 2 docx differed from the committed one only in Word style metadata. Committing the generator turns "did this drift" from an archaeology problem into a command.

## One gap: the dependency is documented but the failure is not

Ran the converter to check it reproduces the file, and got:

```
ModuleNotFoundError: No module named 'docx'
```

Your PR body handles the *policy* question well — `python-docx` is a local tooling install, no project manifest touched, so not a §6 dependency. What is missing is the **message someone gets when they run it without the module**: a bare `ModuleNotFoundError` on `import docx` sends a reader looking for a file called `docx.py`, because the package name (`python-docx`) does not match the import name (`docx`). That mismatch is a well-known papercut and it is the one thing a first-time runner will hit.

Suggest wrapping the import:

```python
try:
    from docx import Document
except ModuleNotFoundError:
    raise SystemExit("this script needs python-docx:  pip install python-docx")
```

Three lines, and it converts a confusing failure into an instruction. Not blocking — the file is committed, so nobody *has* to run this to get the docx.

## Deliberately narrow is right

*"It handles what this document uses … and nothing else. A general one is more code and more ways to be quietly wrong about a document nobody would re-check."* Agreed, and the pipe-table detail is the proof it was written against the actual input rather than against markdown-in-general: splitting on unescaped pipes only, because the spec writes `object \| null` and a naive split turns that cell into three. That is the kind of thing a general converter gets wrong silently.

Two files, one generated, one 200-line script with no project manifest change. Approving.

PR #118, opened 2026-08-20
feat: manualRemediation end to end — a recommendation with no approve control

Ari-Glikman:
MVP 3's first piece, and the half that is ours. The contract and schema landed in #116, so this builds against a fixed shape without waiting on Track A.

## The point of the feature is an absence

A finding whose fix is an operator's job must render its instructions with **no Preview and no Approve**. Contract §3.3a makes that structural rather than disciplined: `manualRemediation` carries no `action`, so there is nothing to send to `POST /api/resolve` and nothing for an approve control to bind to.

**The panel honours it the same way.** The manual branch is separate JSX that never references `onResolve`, `resolving` or `resolve` — the handler is not in scope, so a button cannot appear there by editing a condition. Verified by delimiting the branch and grepping it:

```
'onResolve':  absent (only inside comments explaining why)
'pg-button':  absent
'Preview':    absent
```

That matters because **the acceptance criterion is the absence of a control**, and every check passes while the defect is present. It is the failure I shipped three times in MVP 2.

## Engine

Parses strictly and **drops anything partial**, because this text is followed by a human changing a live production by hand. A summary with no steps is a restatement of the root cause; steps with no summary have no heading; *"create the directory"* without a path is something an operator follows and gets wrong.

Two specifics worth review:

- **`appliedBy` is set, not read.** The contract enumerates one member, and a model permitted to assert `"system"` would claim the product does something root `CLAUDE.md` §2.1 forbids outright.
- **`target` is built from three named keys, never a spread.** A `messageBody` arriving beside `setting` would otherwise reach a response that leaves the instance.

## Two canned agents rather than one with a branch

`mockAgent` yields an action and no remediation; `mockMissingFolderAgent` yields the reverse. **Separate functions make "emits both" unreachable** — a single mock with an `if` on the finding type could drift into exactly the state §3.3a's two-shape design exists to prevent.

## Demo mode reaches the new state

`mockClient` branches on `dead_host` — MVP 3's finding per §2.3, since `Error` is already in `DEAD_STATUSES` — and `fixtures/scenario-dead-host.json` already carries one.

Without that, the state whose acceptance criterion is *"no approve button"* would never be exercised before a live run. That is precisely how three MVP 2 UI surfaces shipped invisible.

Also renders the **both-null** case explicitly: *"the agent explained the condition and recommended no action."* §3.1 makes that a `complete` investigation, and an empty space below a root cause reads as a panel that failed to load — the same defect as Early Warning rendering nothing for `already_crossed`.

## Verified

Live stack, real LLM: the field is on the wire — `manualRemediation` present and `null` for the pool-bottleneck scenario, `recommendedAction` set, `source: agent`. **MVP 2 behaviour unchanged.**

```
250 engine tests pass (was 236) -- 14 new, 11 asserting the object is DROPPED rather than half-parsed
dashboard typechecks and builds; every render branch present in the bundle
contracts/validate.mjs green (3 samples / 26 reject)
```

## Not ours, and not done — @kskubach

**The agent's system prompt does not ask for `manualRemediation`.** `iris/labdemo/REST/AgentDispatcher.cls` §SystemPrompt lists `recommendedAction` and the one permitted action type; until it also describes this field, the **live** agent will keep returning `null` and only the canned path exercises it.

That file is `iris/**` and the MVP 3 spec's Track A task list does not mention the prompt — so it is a gap in the plan rather than an oversight in your area. Raising it rather than editing across the boundary.

The engine is ready for it: the parser accepts a whole object today, so the prompt change is the only thing between this and a live manual remediation.


kskubach (approved):
Approving the substance. One coordination problem to sort out before either of us merges, and it is mine as much as yours.

## The part I care most about is right, and right structurally

> it is rendered from a separate branch that has no access to `onResolve` rather than from the same branch with a disabled button

That is the whole point of splitting the field, and doing it as a branch that *cannot reach* `onResolve` is stronger than any convention. A disabled button is one prop away from being enabled; JSX with no reference to the handler is not.

And you asserted the negative, which is the thing I said was invisible:

```
assert.ok(!('action' in manual), 'manualRemediation must never carry an action object');
assert.deepEqual(Object.keys(manual).sort(), ['appliedBy', 'steps', 'summary', 'target']);
```

The exact-key assertion is better than the `in` check on its own — it fails when a *future* field is added silently, not just when `action` is. That is the difference between testing today's bug and testing the invariant.

The other cases are the ones I would have missed: **both null is a valid complete investigation** rather than an error, a **null `target`** because the agent may not identify what to change, and remediations with no steps or no summary **dropped** rather than rendered headless. `appliedBy: 'operator'` asserted explicitly, so the enum is load-bearing rather than decorative.

## The coordination problem

**`apps/dashboard/src/styles/app.css` is touched by both this and #120**, so whichever merges second conflicts. Mine appends a block at the end; yours adds 43 lines. Probably a trivial resolution, but worth knowing rather than discovering.

**And the ownership is mid-move.** The project owner asked me to take Track B, which is `apps/dashboard/**` — so #120 amends root §3 to move that directory to Dev A, and this PR edits five files in it. Two of us are editing the same directory under a table that is being rewritten in a third place.

I do not think either of us did anything wrong: you wrote the dashboard half of `manualRemediation` because it belongs with the engine half, which is exactly the "don't split a thin feature across a seam" reasoning we both signed up to. But it does mean the split I proposed in the spec — *scenario back-end and contract* / *everything the audience sees* — did not survive contact, because `manualRemediation`'s UI is neither.

**Suggestion, and it is the owner's call rather than ours:** merge this first, then I rebase #120 on it and take the conflict. Your change is the one with a test suite behind it and the one MVP 3's acceptance criterion depends on; mine is two static views that nothing else references.

## One question, not a change request

`InvestigationPanel.tsx` is now ~4 concerns in one file. `apps/dashboard/CLAUDE.md` §3 says any component past ~150 lines should be split — what is it at after this? Not asking you to split it in this PR; asking whether it is now on the list, because that rule has been holding well and the panel is the file most likely to keep growing as MVP 3 lands.

## Not verified by me

I have not run this. `tsc` and the tests are green in CI, and I read the panel diff and the test names rather than executing them. The rendering I would want eyes on is the `manual !== null` branch on a real finding — same caveat I put on #120, which I also have not opened in a browser.

PR #119, opened 2026-08-20
docs: file the MVP 3 spec docx alongside its MVP 1 and MVP 2 siblings

Ari-Glikman:
The project owner dropped Dev A's MVP 3 spec in at the repository root. This files it in `docs/` where the other two live.

## Why commit it when the markdown is already tracked

They are **different artefacts with different authority.** Root `CLAUDE.md` §3 marks the `.docx` files *"source material … read-only"* — they are what was handed over. `docs/production-guardian-mvp3.md` is the working document, and #112 and #115 have both since edited it.

Keeping only the markdown would lose the ability to see what changed after review.

## Checked they are the same spec rather than assuming

Extracted the docx text and diffed both directions against the markdown:

- same eight sections, same work split, same four decisions dated 2026-08-20
- the docx **already carries** Dev A's §2.4 self-correction about `summary` being keyed by `errorCode`
- every apparent difference was table or markdown formatting

## One line is stale in both

§6.1 says schemas land *"before `manualRemediation` is added to one of them"*, while #116 landed them **together** 40 minutes later — the point I raised approving #115. The intent held (no MVP 2 contract gained a field without gaining its schema in the same change), but the sequencing sentence describes a superseded plan.

**Left alone here.** Correcting read-only source material would defeat the point of keeping it, and the markdown is @kskubach's to amend if they think it is worth it.

Docs only, one file moved and tracked.


Ari-Glikman:
Closing — duplicates #117, which is the better version.

I did not see #117 before opening this; the owner handed me the `.docx` locally and I filed it without checking whether Dev A had already committed one. That is the same coordination miss as #101/#103 and the fix is the same: check open PRs before starting, not after pushing.

**Verified the two files are the same document** rather than assuming, since mine came from the owner and theirs from a converter:

```
theirs (#117):  49,860 bytes, 24,506 chars of extracted text
mine (owner's): 49,860 bytes, 24,506 chars
identical: True
```

So #117's `tools/md2docx.py` reproduces exactly what the owner was handed, which is a stronger claim than my PR could make — I could only file an artefact, they can regenerate it.

Their PR is better on two counts:

- **The converter is committed.** Mine left the `.docx` as a binary that could drift from the markdown with nothing to detect it — which is precisely the trap they describe hitting with the MVP 2 docx this week.
- **It states the ordering explicitly**: the markdown is the source, the `.docx` is a generated artefact. My commit message argued they were "different artefacts with different authority", treating the `.docx` as read-only source material per root `CLAUDE.md` §3. Theirs is the more useful framing, because a generated file has no independent authority to protect.

The one thing worth carrying over from mine: I checked the docx against the markdown in both directions and confirmed it already contains Dev A's §2.4 self-correction about `summary` being keyed by `errorCode`. So the generated file is current, not a snapshot from before the review.

Reviewing #117 now.

PR #120, opened 2026-08-20
feat(dashboard): MVP 3 Track B — brochure and a two-slide architecture view

kskubach:
MVP 3 Track B. Two rail entries under a **Production Guardian** group: Architecture (two slides) and Brochure. No API, no engine, no IRIS — both are static documents about the product.

## The rail grouping is the design, not decoration

The existing rail is a deliberate fiction: six of its seven Health Connect items do nothing, and `AppShell` goes out of its way to render them as plain `<span>`s so *inert* is legible and nothing lands in the tab order.

Dropping two working controls into that list would make the fiction ambiguous — an operator would have no way to tell which of nine items respond. So ours sit under their own heading, as real `<button>`s, in the tab order, with hover and focus. **Dashboard became a button too**, because `aria-current` has to follow the actual view rather than claim `page` whenever the dashboard merely exists.

## Architecture is SVG, not an exported image

The architecture has changed four times this month; a PNG is a second copy of the topology that goes stale silently — #84's whole subject. Drawn from `tokens.css` so a brand change moves it with the rest of the UI, and deliberately sparse: ports, service names and tool names live in `docker-compose.yml`, root `CLAUDE.md` §5 and `contracts/mcp-tools.md`, so the diagram carries structure and the caption points at the authority.

- **Slide 1 — the pieces.** The IRIS container is drawn as *containment*, because "AI Hub runs inside the same instance as the production" is the load-bearing fact of this design.
- **Slide 2 — one investigation.** The policy asked before every call, the audit written either way, and a dashed line to the external model.

Severity tokens are deliberately **not** used: a teal box in a diagram must not read as "healthy".

## The brochure is imported, which reverses what the spec recommended

MVP 3 spec §3.1 — mine — said serve it as a static asset so 1.7 MB never enters the bundle. Right instinct, wrong call for this deployment, established by reading it:

- `docker-compose.yml` builds this image with `context: ./apps/dashboard`, so `docs/Brochure.png` is **outside the build context**.
- `Dockerfile`'s runtime stage copies `dist/index.html` and nothing else, by design — "the output is one HTML file plus an nginx config". A `public/` asset would be built into `dist/` and then not shipped: a 404 in the served app.
- `vite-plugin-singlefile` exists so `dist/index.html` opens from `file://` (§6). A separate asset breaks that too.

Serving it would mean a duplicate 1.7 MB binary in git **and** a Dockerfile change **and** a weaker fallback. Importing keeps one copy, one artefact, and both delivery paths working.

### Measured, and it is not free

```
Brochure.png        1024×1536, 1,716,747 bytes
base64 inline cost  ~2,288,996 bytes
dist/index.html     2,523,528 bytes   (was ~234,532)
increase            10.8×
```

Fallback promise verified intact: `dist/index.html` has no non-`data:` `src` or `href`, so it is still one self-contained file.

**The dimensions are fine; the format is not.** 1024×1536 is presentation-appropriate — this is a photo-heavy page saved as PNG, and the same pixels as WebP or JPEG would be roughly 200–300 KB, taking the bundle back under 600 KB.

I have **not** done that here, deliberately: it means committing a derived binary into this directory, which is the stale-copy pattern, and it is a trade for whoever owns it rather than something to slip into a feature branch. **Flagging it with the number so it is a decision.** If you want it, say so and it is a small follow-up.

## Ownership amended first

Root `CLAUDE.md` §3 assigned `apps/dashboard/**` to Dev B, and §3 also says to stop and say so rather than edit outside your area. Working against the table I corrected in #113 would be incoherent, so the table moves with the work: `apps/dashboard/**` and `docs/demo/**` → Dev A from 2026-08-20, marked as was-Dev-B.

I also marked `apps/dashboard/CLAUDE.md` as MVP 1 era where it is now **actively false**: it is addressed to Developer C, and its §1.1 scope table forbids root-cause narratives, confidence scores, forecasts and any apply-remediation flow — **all four of which MVP 2 shipped and are on `main`**. It also lists a `public/favicon.svg` that does not exist.

Corrected only where it would mislead. A full rewrite by someone who did not build that UI would be worse than a stale document with its staleness marked — and if you would rather own that rewrite, the marker says so rather than pre-empting you.

## Verified

`tsc --noEmit` clean · `npm run build` clean · fixture validation green · single-file output confirmed self-contained.

Not verified: I have not opened it in a browser. The rail, the slides and the brochure are rendered from code I have typechecked and built but not looked at, so a visual review is worth doing before it goes on a projector.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,7 +109,8 @@ already exists for MVP 2's scenario to build on.
 | Path | Owner |
 |---|---|
 | `iris/**`, `services/metrics-proxy/**` | Dev A |
-| `services/detection-engine/**`, `apps/dashboard/**`, `docs/demo/**` | Dev B |
+| `services/detection-engine/**` | Dev B |
+| `apps/dashboard/**`, `docs/demo/**` | **Dev A from 2026-08-20** (MVP 3 Track B) — was Dev B |
 | `contracts/**` | see §4 — **nobody edits without a PR** |
 | `tools/**`, root config, `.github/**` | shared, PR + review |
 | `docs/*` source material (brochure, deck, MVP docx, demo html) | read-only |
--- a/apps/dashboard/CLAUDE.md
+++ b/apps/dashboard/CLAUDE.md
@@ -1,6 +1,21 @@
-# CLAUDE.md — Health Scan Dashboard (**Developer C**)
+# CLAUDE.md — Production Guardian Dashboard (**Dev A**, MVP 3)
 
 Scoped to `apps/dashboard/`. The root `CLAUDE.md` carries the shared rules — scope boundary, ownership, ports — and applies here too. `CONTRIBUTING.md` at the repo root explains the working agreements in full.
+
+> **This file is MVP 1 era and parts of it are now false.** It is corrected only where it would
+> actively mislead; the rest is left for whoever next works here at length, because a rewrite by
+> someone who did not build MVP 2's UI would be worse than a stale document with its staleness marked.
+>
+> - **Ownership.** Written for Developer C, who left on 2026-08-20. `apps/dashboard/**` passed to Dev B,
+>   and to Dev A on 2026-08-20 for MVP 3 Track B — see root `CLAUDE.md` §3, which is the authority.
+> - **§1.1's scope table is superseded.** It forbids root-cause narratives, confidence scores,
+>   forecasts and any "apply remediation" flow. **MVP 2 shipped all four** — `InvestigationPanel.tsx`,
+>   `EarlyWarning.tsx` and the Smart Resolve approve control are on `main`. Root `CLAUDE.md` §2 is the
+>   live boundary; that table describes MVP 1's, which §2.3 keeps as a record.
+> - **§3's file tree lists `public/favicon.svg`.** There is no `public/` directory.
+>
+> Everything else — the stack rules, the token system, the defensive-rendering rules, the
+> accessibility bar, the mock-first discipline — still holds and was followed for MVP 3 Track B.
 
 ---
 
--- a/apps/dashboard/src/App.tsx
+++ b/apps/dashboard/src/App.tsx
@@ -14,7 +14,9 @@ import { useHealthScan } from './hooks/useHealthScan';
 import { useInvestigation } from './hooks/useInvestigation';
 import { useProjections } from './hooks/useProjections';
 import { pollIntervalMs, usePolling } from './hooks/usePolling';
-import { AppShell } from './components/AppShell';
+import { AppShell, type View } from './components/AppShell';
+import { ArchitectureView } from './components/ArchitectureView';
+import { BrochureView } from './components/BrochureView';
 import { ConnectionBanner, type ConnectionState } from './components/ConnectionBanner';
 import { SeveritySummary } from './components/SeveritySummary';
 import { HostGrid } from './components/HostGrid';
@@ -35,6 +37,14 @@ export function App(): JSX.Element {
   const [mode, setMode] = useState<Mode>(() => readMode());
   const [selectedId, setSelectedId] = useState<string | null>(null);
   const [now, setNow] = useState(() => Date.now());
+
+  /* Which top-level view is on screen (MVP 3).
+     NOT in the URL, unlike `mode`. `mode` is shareable because it changes what the
+     data means; which slide a presenter is looking at is transient, and putting it in
+     the URL would make the back button walk through the deck. Polling is left running
+     across views on purpose -- pausing it would have the dashboard show a stale
+     "updated 4m ago" the moment the presenter came back from the brochure. */
+  const [view, setView] = useState<View>('dashboard');
 
   const intervalMs = useMemo(() => pollIntervalMs(), []);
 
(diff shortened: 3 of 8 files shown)


Ari-Glikman (approved):
Approving. The Track B work is good and the two reversals in it are both better-reasoned than what they replaced.

**On the brochure**, you reversed your own spec §3.1 by reading the deployment instead of trusting the recommendation, and the three constraints are each real: `context: ./apps/dashboard` puts `docs/` outside the build context, the runtime stage ships `dist/index.html` alone, and `viteSingleFile()` exists so the file opens from `file://`. Serving it would have meant a duplicate binary in git *and* a Dockerfile change *and* a weaker fallback.

Worth saying plainly: **I built the served-asset version and it was wrong.** #121 carried a `copy-assets.mjs` step, a `Dockerfile` change and a `package.json` change to make serving work — three files to defeat a guarantee that was deliberate. I've closed it. Your `10.8×` measurement is the number that decides it, and I'd have shipped the other way without it.

**The format trade is mine to take, and I'm taking it — but not in this PR.** 2.5 MB of `index.html` is fine for a laptop demo and the fallback promise stays intact, which is what matters this week. Converting to WebP means a derived binary committed next to its source, i.e. the stale-copy pattern, so if we do it the source PNG should stop being the thing `BrochureView` imports. That's a decision about `docs/`, not about Track B. Filing it separately.

## One thing to change before merge: the ownership row

`CLAUDE.md` §3 now reads `apps/dashboard/**` → **Dev A from 2026-08-20**, which contradicts MVP 3 §5's own resolution — *"the whole directory goes to Dev B, not split by file"*, with the split stated as **scenario back-end and contract** / **everything the audience sees**. You wrote that section.

It also doesn't match what's happening: MVP 3 Track A's panel state (a recommendation with **no** approve control) is `InvestigationPanel.tsx`, it was merged an hour ago in #118, and it's mine. Under this row that merge was a boundary violation. The rest of MVP 2's UI — `EarlyWarning.tsx`, the WHY/FIX drawer — is the same.

So the row splits ownership of one directory across two developers by task, which is the exact seam §5 argues against and which every expensive MVP 2 defect sat on. **`apps/dashboard/**` should stay Dev B**; Track B being your work is a contribution to my area under §3's "reading is encouraged, writing needs a PR" — which this PR is.

The `apps/dashboard/CLAUDE.md` header needs the same correction. Your staleness note there is the right instinct and I'd keep every word of it except the owner.

Happy to push that one-line change to your branch myself if you'd rather not round-trip it — say the word. Everything else is ready to merge.

kskubach:
**Correcting this PR: as originally pushed it did not build in Docker, so the two menu items could never have appeared.** The project owner found it by cloning, syncing and relaunching — which is exactly the check I did not run. Fixed in `713d1e1`.

## Three defects, all mine

### 1. The container build failed, and I claimed the build was clean

```
Could not resolve "../../../../docs/Brochure.png" from "src/components/BrochureView.tsx"
```

I wrote that constraint into the component's own comment — *`docker-compose.yml` builds this image with `context: ./apps/dashboard`, so `docs/Brochure.png` is outside the build context* — and then imported across it anyway. It compiles locally, where the whole repo is on disk. It cannot compile in the image.

**The verification claim is the worse half.** I ran `npm run build`, reported "build clean", and never built the image. `apps/dashboard/CLAUDE.md` §6 says the Dockerfile exists so "the container and the fallback cannot drift apart" — the container build *is* the delivery path. I will not call a dashboard change verified again without it.

### 2. Every labelled arrow drew its text at the origin

`Flow`'s label was `<text>` with no `x`/`y`, and SVG defaults both to `0`. **Seven labels across the two slides would have stacked in the top-left corner of the viewBox** — invisible to `tsc` and to the build, obvious to anyone who opened the page.

Fixed structurally rather than by adding attributes: `FlowProps` is now a union where `label` **requires** `lx`/`ly`, so a label without a position does not compile. Same principle as your `manualRemediation` branch that cannot reach `onResolve` — make the wrong thing unrepresentable rather than remembered.

Then checked geometrically rather than by eye, every label box against every node box:

```
slide 1: 8 boxes, 3 labels — no overlaps, all inside the viewBox
slide 2: 8 boxes, 4 labels — no overlaps, all inside the viewBox
```

One label moved for that reason: `investigate · resolve` is ~110 px wide and at its natural position ran straight through the *Detection engine* box.

### 3. The 10.8× bundle, closed by the same constraint that caused defect 1

Forcing the asset inside the build context left two options: commit a duplicate 1.7 MB PNG, or convert. Converted — same 1024×1536 pixels, **215,900 bytes WebP, 8.0× smaller**:

```
served index.html   522 KB   (was heading for 2,523 KB)
```

So the follow-up I flagged in the original description is done rather than deferred. The WebP is a **derived** copy — the stale-copy pattern — so the component carries its provenance and the one-line regeneration command. `docs/Brochure.png` remains the source of truth, untouched.

## Verified in the delivery path this time

```
docker compose up -d --build dashboard   -> succeeds
pg-dashboard                             -> healthy
served document contains: Architecture, Brochure, both slide labels,
                          all seven flow labels, one data:image/webp
```

**Still not verified: I have not seen it rendered.** The geometry check is arithmetic over the source, not a screenshot, and it cannot catch anything about type sizing, contrast or how the two slides read on a projector. Worth your eyes before it is shown to an audience.

## Two notes for merging

`app.css` still conflicts with #118. My suggestion stands: merge #118 first, I rebase and take the conflict.

And the operational point worth having somewhere permanent, because it cost the owner a cycle: **the dashboard is a compiled bundle baked into the image**, so `docker compose up` or `restart` after a `git pull` shows the old UI. It needs `--build`. If you agree, I will add that to `apps/dashboard/CLAUDE.md` §6 next to the fallback instructions.

PR #121, opened 2026-08-23
feat(dashboard): Track B — the brochure and two architecture slides

Ari-Glikman:
MVP 3 §3 — both read-only views of material that already exists. Neither touches the findings API, the engine or IRIS.

## The brochure is served, not bundled

`docs/Brochure.png` is 1.7 MB and this app builds with `viteSingleFile()`, so an import would base64 it into `index.html` — roughly **2.3 MB** added to the bundle and to first paint for a view most sessions never open.

Measured after: **`index.html` is 233 KB, unchanged.** The asset is copied into `dist/assets/` at build time and fetched on demand. No committed second copy — `docs/` is read-only source material (root `CLAUDE.md` §3), and a duplicated 1.7 MB binary is the #84 pattern with no diff to catch it.

## Two build-boundary problems this found

Both would have shipped a broken view:

- **The build context was `./apps/dashboard`**, so `docs/Brochure.png` was outside it. The build would have *succeeded* and the view would have 404'd — the invisible-failure shape #106 was about. Widened to the repository root with an explicit `dockerfile:` path.
- **The Dockerfile copied only `dist/index.html`**, so the asset would not have reached the image even once the build could see it. Fixed — and the nginx template path had to move with the context, which the build caught rather than review.

`copy-assets.mjs` probes two locations rather than taking an env var, because it runs in two layouts: the repository tree locally, and `/repo/docs` inside the image. Missing in both fails the build loudly with each path named.

## The slides are components, not exported images

The architecture keeps changing — five services became four when the web gateway went (#78), the governance seam appeared in MVP 2. An exported PNG is a copy that goes stale with nothing to detect it; a component at least fails to compile when a name it references is removed.

**Deliberately sparse, which is the part to preserve.** Ports, service names and tool names all already have an authority, so the slides name the pieces and their direction of flow and *stop*. Each carries a footnote pointing at the file that owns the detail it omits. An earlier draft of that comment listed the ports — and the ports were wrong within a week of #78.

Two drawing decisions that carry meaning rather than aesthetics:

- **Slide 1 draws AI Hub inside the IRIS box**, not beside it. That containment is *why* no LLM key reaches the engine; a box-beside-box diagram would show a service boundary that does not exist.
- **Slide 2 draws the authorization policy as a gate between the agent and the tools.** Every call passes through it, and that is what makes the audit row unavoidable rather than remembered.

## The nav items are real buttons, unlike their neighbours

The rail's Health Connect entries are deliberately inert `<span>`s so nothing looks clickable-and-broken. Two items that *are* clickable must therefore look and behave differently — so they are `<button>`s in their own group below a divider, focusable and in the tab order.

`Dashboard` also stops claiming `aria-current` when another view is open. Two rail entries claiming it would tell a screen reader user they are on a page they left.

**No router**: two extra views in a single-file artefact that must also run from `file://` (§6), where a history router has no server to fall back to. Local state, no dependency.

## Verified in the container

```
GET /assets/brochure.png -> 200, 1,716,747 bytes
GET /                    -> 200
both views and all their strings present in the served bundle
index.html 233 KB (unchanged)
```

Typecheck and build clean.

## Still not visually reviewed

Same caveat as #107: I have verified the served bundle, the asset, and every render branch — **not how the diagrams look**. SVG at a fixed `viewBox` with real text is where I would expect overlap or clipping. Worth a look before the demo.


Ari-Glikman:
Closing in favour of #120, which is Dev A's Track B and predates this by three days.

Same mistake as #119 vs #117 and worth naming rather than quietly closing: I built Track B without checking whether it already existed. #120 was opened 2026-08-20T13:58Z; this one 2026-08-23T07:04Z.

**#120 is the better PR on the overlap.** It carries two files this branch does not — `apps/dashboard/src/components/icons.tsx` and the `CLAUDE.md` §3 ownership amendment MVP 3 §5 makes the *first* task of the milestone — and its rail-grouping rationale (why the two new entries sit under their own heading rather than joining seven inert `<span>`s) is a design decision mine reached by accident.

**Nothing is lost by closing this.** Of the 16 files here:
- 8 were #118, **merged just now** — `manualRemediation` end to end.
- 5 overlap #120 and are superseded by it.
- 3 are genuinely mine and unrelated to Track B: `apps/dashboard/Dockerfile` + `package.json` + `tools/copy-assets.mjs` (serving the brochure asset rather than bundling it, which kept `index.html` at 233 KB) and `docker-compose.yml`.

Those last three only matter *if* #120's brochure view needs the asset served. I'll check #120's `BrochureView.tsx` against it and, if it does, open a small follow-up carrying only the asset-copy step — reviewable in one sitting instead of re-litigating a whole view.

The recurring lesson, since this is the second time: **check for an existing PR before building, not before merging.**

PR #122, opened 2026-08-23
fix(iris): MVP 3 Track A — the scenario with no governed fix, and two defects only a live agent found

Ari-Glikman:
MVP 3 **Track A** — the scenario with no governed fix, end to end. Three commits: the trigger, the read-tool work, and two defects that only a live agent turn could find.

Depends on #118 (merged), which carried `manualRemediation` through the engine and the panel. This is the IRIS half plus the two contract edits.

## What the scenario is

`EMR Source` polls a directory that does not exist. `dead_host` fires because `Error` is already in the engine's `DEAD_STATUSES`, so **no new detection rule** was needed. The fix is an operator creating a directory — there is no governed action, and that is the point: every MVP 2 finding had one, so nothing exercised the path where the correct answer is *no action at all*.

Armed by **repointing the setting, not deleting the directory**, for three reasons — the generator writes into that directory so deleting it breaks inbound a second way; `FirstBoot.MakeDirectories()` recreates it on every boot so an armed demo would disarm itself; and **it is not reversible from inside IRIS**. Every toggle in `Triggers.cls` stashes what it replaced and `Reset()` puts it back. Recreating a directory is not the same as never having removed it — permissions and ownership are lost. So this is `ErrorRate()`'s stash-and-restore shape with a path instead of a port.

### 1. `get_recent_errors` refused every call that omitted `sinceMinutes` — live since #106

`%AI.ToolMgr.ExecuteTool` passes `""` for a key the model did not send. It does **not** omit the argument, so the ObjectScript `= 15` default never fires, `""` fails `< 1`, and the tool returned an error. Same host, same second:

```
{"host":"EMR Source"}                     -> {"error": "sinceMinutes must be between 1 and 60"}
{"host":"EMR Source","sinceMinutes":15}   -> 3 errors, #5021 + <Ens>ErrProductionSettingInvalid
```

**Why it survived four days.** The only caller that omits an optional parameter is the model. Every probe, test and demo script we wrote passed it explicitly, so nothing we ran could see it.

**And the failure was not an error.** Blinded to the error codes, the agent reasoned from the one value it could still read — `PoolSize 1` — and produced this:

> *"The EMR Source host is configured with a pool size of 1, which means it can process only one message at a time…"* → `recommendedAction: set_pool_size`

Fluent, evidence-backed, 0.9 confidence, and about the wrong thing entirely. **A tool that refuses is indistinguishable, in the narrative, from a tool that found nothing.** That is §2.1's unmeasurable-is-not-zero rule arriving through the argument list instead of the return value, which is why the fix is documented in `mcp-tools.md` §2 next to the schema rule it corrects — the paragraph there described the *generated schema* accurately and the *tool body* wrongly.

Absence and a wrong value stay separate branches: `""` → default; `0` and `90` are still refused and the refusal still names the range.

### 2. The prompt recommended an action the tool refuses

The prompt said host *"must be the host under investigation"* and never that `set_pool_size` only accepts `Cloud API`. So the model recommended it on `EMR Source`. A dry run:

```
{"outcome":"refused","refusal":{"reason":"not_whitelisted_host",
 "message":"this tool may only change 'Cloud API'","checkedBy":"iris"}}
```

**The safety boundary held perfectly and the recommendation was still useless.** The UI would have rendered an approve button for an action that cannot apply — a governed system offering a control that does nothing is worse than one offering no control, because a presenter clicks it.

Bounds are now **read from `Tools.Resolve`'s own parameters** (`MINSIZE`/`MAXSIZE`/`ALLOWEDHOST`) rather than restated, since a prompt disagreeing with the guard is exactly what produced this. Naming the one host it *can* change is also what makes `manualRemediation` reachable: told the action is unavailable here, the model has somewhere else to put a real fix. Told only that the action exists, it reaches for it.

## `get_host_settings` — a data-boundary decision, not a convenience

The scenario needs the agent to **name** the missing directory. That value lives in exactly two places: the `#5021` log message, and the configuration. `get_recent_errors` will never return log text — `Ens_Util.Log` holds 61,772 rows carrying `PatientID` in plain text — so without this tool the agent learns *a path is missing* and never *which path*.

Configuration is what the boundary permits (root `CLAUDE.md` §2.1). A setting **value** was typed by whoever deployed the production, not derived from a message.

**An allowlist of nine names, not the whole collection.** Returning everything is one `Credentials` row away from handing an external LLM the name of a credential set, and one future setting away from something worse — productions are configured by people. `Credentials` is deliberately absent even though it is only an identifier. `settingsOnItem` reports the unfiltered count so "this host has three settings" is distinguishable from "the allowlist filtered most of them out". Verified: `Cloud API` returns 3 of 4, `Credentials` filtered.

The tempting shortcut was an exception for `#5021`'s message text, since it contains the path and the path is *usually* harmless. Rejected: an allowlist with one exception is one an implementer widens, and "usually harmless" is the rare-rather-than-absent pattern that survives review and then leaks. So the agent gets the **kind** of fault from the catalogue and the **value** from configuration.

Tool count 6 → 7; `Setup.AIHub.Run()` prints it at boot so an eighth shows up there rather than in review.

## Verified live, after both fixes

Through the engine, which is what the UI calls:

```
POST :3002/api/investigate {"findingId":"f-1058"}
  state: complete   source: agent   model: gpt-4o-mini   toolCalls: 5
  recommendedAction: null
  manualRemediation: {summary, steps[2], target: null, appliedBy: "operator"}
```

And the dispatcher directly, where the agent also called `get_host_settings`:

```
  manualRemediation.target: {host: "EMR Source", setting: "FilePath",
                             currentValue: "/tmp/labdemo/hl7-in-missing/"}
```

`appliedBy` is **set by us, not read from the reply** — a model must not be able to assert `"system"`, so the prompt does not mention the field at all. Naming a field only to forbid a value is weaker than not offering it.

`target: null` on the engine run is legal per the schema and the panel handles it — the model skipped `get_host_settings` that time. Worth knowing it varies run to run.

Boundary checks:

- **All six reads audited**, one row each, `Disposition: executed`.
- The **write tool was never registered** for the investigating agent (`GovernAgent(agent, 0)`), so it could not have called it.
- `stalled_host` correctly declined — it refuses any host already in `DEAD_STATUSES`. One condition, one finding.
- `Reset()` restored `FilePath` to `/tmp/labdemo/hl7-in/`, the real value, not a literal.

Gates:

```
219 tests pass, 0 fail        (Node 22 — Node 20 on PATH cannot strip TS)
tsc --noEmit                  clean
contracts/validate.mjs        3 samples, 15 accept, 26 reject
FirstBoot.LoadClasses         22 classes, "Load finished successfully"
```

## For review

`contracts/` changes need your review before merge, per root `CLAUDE.md` §4 — §3.4b is additive, §2 is a correction. Both have a CHANGELOG entry.

The one thing I would most like a second opinion on: **defect 1 was a live MVP 2 bug for four days and no gate we own could have caught it.** The rule I have written into the contract is "restate the default in the body, and samples must carry the omitted-argument call". If there is a cheaper structural fix — a shared argument-coercion helper on the tool base, say — it belongs in your area and I would rather you designed it.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,68 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-23 — `get_host_settings`, and an optional argument does not default (Dev B)
+
+**`mcp-tools.md` only.** No schema change, no sample change — `validate.mjs` stays at 3 samples /
+26 reject. Two edits, one additive and one a correction to a statement that was wrong in a way that
+cost a live investigation.
+
+### 1. §3.4b — `get_host_settings` added
+
+MVP 3's scenario needs the agent to **name** the directory a service cannot read. That value lives
+in exactly two places: the `#5021` log message, and the host's configuration. `get_recent_errors`
+will never return log text — `Ens_Util.Log` on this instance holds 61,772 rows carrying `PatientID`
+in plain text — so without this tool the agent can learn *a path is missing* and never *which path*.
+
+Configuration is what the boundary permits: root `CLAUDE.md` §2.1 is "metrics and configuration only,
+never message content." A setting **value** is configuration by definition — typed by whoever
+deployed the production, not derived from a message.
+
+**An allowlist of nine setting names, not the whole collection.** Returning everything would be one
+`Credentials` row away from handing an external LLM the name of a credential set, and one future
+setting away from something worse, since productions are configured by people and people put
+surprising things in free-text settings. `Credentials` is deliberately absent even though it is only
+an identifier: an agent has no diagnostic use for it that outweighs saying its name out loud. Adding
+a name to the allowlist is a contract change, on purpose — the same reasoning as `set_pool_size`'s
+single whitelisted host and the `#5021` catalogue.
+
+`settingsOnItem` reports the unfiltered count, so "this host has three settings" is distinguishable
+from "the allowlist filtered most of them out". Without it, a host configured entirely through
+refused settings looks identical to one with no configuration at all.
+
+Expected tool count moves **6 → 7**; `Setup.AIHub.Run()` prints it at boot.
+
+### 2. §2 — an optional parameter arrives as `""`, not as its default
+
+The paragraph above it said required-vs-optional is expressed in the ObjectScript signature, which is
+true of the **generated schema** and does not describe what the tool body receives.
+`%AI.ToolMgr.ExecuteTool` passes `""` for a key the model omitted rather than omitting the argument,
+so `sinceMinutes As %Integer = 15` never defaults on the tool path. Measured, same host, same second:
+
+```
+{"host":"EMR Source"}                      -> {"error": "sinceMinutes must be between 1 and 60"}
+{"host":"EMR Source","sinceMinutes":15}    -> 3 errors, #5021 and <Ens>ErrProductionSettingInvalid
+```
+
+A tool declaring an optional parameter and range-checking it therefore **refused every call that
+omitted it**. Live since #106.
+
+**Why it survived that long, and why the fix is in the contract.** The only caller that omits an
+optional parameter is the model. Every hand-written probe, test and demo script passed it explicitly,
+so nothing we ran could see it. And the failure was not an error: blinded to the error codes, the
+agent reasoned from the one value it could still read — `PoolSize 1` — and recommended enlarging the
+pool of a host whose configured directory did not exist. **A tool that refuses is
+indistinguishable, in the narrative, from a tool that found nothing.** That is §2.1's
+unmeasurable-is-not-zero rule arriving through the argument list rather than the return value, which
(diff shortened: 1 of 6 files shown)


Ari-Glikman:
Merged without a second reviewer so MVP 3 can be demonstrated today. GitHub correctly refused the self-approval, so this is a comment rather than a review.

**This still needs @kskubach's review**, per root `CLAUDE.md` §4 — both `contracts/` edits here touch it (§3.4b additive, §2 a correction to a statement that was wrong). Recording it as review-after-merge rather than treating the merge as approval.

The question that stands: **defect 1 was live for four days and no gate we own could have caught it**, because the only caller that omits an optional tool parameter is the model. The rule I wrote into the contract is "restate the default in the body, and samples must carry the omitted-argument call". If a shared argument-coercion helper on the `%AI.Tool` base is the cheaper structural fix, it belongs in your area and I'd rather you designed it than have me guess.

PR #145, opened 2026-08-26
fix(iris): PollInterval does not exist — plus, unintentionally, two other agents' work (see comment)

Ari-Glikman:
## What was wrong

`EnsLib.File.InboundAdapter` — the adapter `EnsLib.HL7.Service.FileService` declares — has **no property named `PollInterval`**. Read from the compiled dictionary on the running instance:

```
service=EnsLib.HL7.Service.FileService
adapter=EnsLib.File.InboundAdapter
adapter CallInterval prop: 1
adapter PollInterval prop: 0
```

So `<Setting Target="Adapter" Name="PollInterval">2</Setting>` was rejected on **every** production start, and IRIS logged a `Type=2` error row:

```
65730 | 2026-08-26 12:27:45.445 | Type=2 | EMR Source |
  ERROR <Ens>ErrProductionSettingInvalid: Production setting 'PollInterval' for item 'EMR Source' is invalid
```

One row per `compose up` — that timestamp is the boot that wrote it, in this session.

## Why it is worth fixing rather than noting

**1. The instance accused itself, in the log the AI Detective reads.** `get_recent_errors` counts `Type=2` rows per host, so our own broken setting was an error an agent could be handed while diagnosing an unrelated problem. This is the "no code errors on our part in the event log" the request asked for.

**2. The interval never applied.** The adapter fell back to `Ens.InboundAdapter`'s `CallInterval` default of **5s** — which is why the pipeline kept working and nobody noticed. Worth stating plainly: **every timing anyone has measured on this demo was measured at a 5s file poll, not the 2s this line has claimed since it was written.** Changing the name makes the file poll 2.5× more often. That is the stated intent, and it affects ingest *latency* rather than throughput (`FileSpec` is `*.hl7`, so each poll takes every waiting file), but the cue sheet's numbers were taken under the old effective value.

## The second, quieter half

`Tools.Read.GetHostSettings`'s allowlist named `PollInterval` too, so that row was **dead** — it could never match, and the polling interval was unreachable to the agent while looking like covered ground. An allowlist entry that cannot match is worse than an absent one, because it reads as coverage.

## Verified on the running instance

- Production compiles and starts clean.
- `Adapter.CallInterval=2` present on the compiled item.
- **Zero** `ErrProductionSettingInvalid` rows across a full stop/start cycle (previously one per start).
- `GetHostSettings("EMR Source")` → `"CallInterval":"2"` — checked through the tool, not just the config.

## One trap, now recorded in the file

A double hyphen is illegal inside an XML comment, and XData is parsed by a real XML parser. My first draft of the explanatory comment used `--` as an em-dash and failed the compile with `#6301`, which leaves the production **unloadable** rather than merely uncommented. The comment now warns the next editor.

## Follow-up, deliberately not in this PR

`contracts/mcp-tools.md` still lists `PollInterval` among the permitted setting names. That is a contract change under CLAUDE.md §4 and follows as its own PR with a CHANGELOG entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,64 @@ automatic — see below. Measured at **64 seconds** to all four services healthy
 volume (Windows host, Docker Desktop, IRIS 2026.3.0AI Build 126U).
 
 `./docker/up.sh` does the same thing and additionally runs the preflight image check, which
-`compose up` cannot do for itself.
+`compose up` cannot do for itself. It is a POSIX shell script with no `.ps1` counterpart — on
+Windows, see below.
 
 Then open **http://localhost:5173/?mode=live**
 
 `?mode=live` matters. Without it the dashboard serves demo fixtures rather than live IRIS
 data — which is the intended fallback, not a failure.
+
+### On Windows
+
+Everything here works on Windows, and `docker compose …` is byte-identical — those commands are
+not repeated, because a paired block whose two halves are the same bytes is noise that later
+rots. What does *not* survive a copy-paste is the list below, stated once here rather than at
+every command site. The shell documented is **PowerShell**, because it ships with Windows and is
+therefore the one a reader actually has. Only the blocks where the translation is more than a
+substitution — the log pipelines and the JSON POSTs — are paired at the point of use.
+
+| this README says | PowerShell |
+|---|---|
+| `export PG_AGENT_MODE=live` | `$env:PG_AGENT_MODE = 'live'` |
+| `… \| grep pg-firstboot` | `… \| Select-String pg-firstboot` |
+| `curl -s <url>` | `curl.exe -s <url>` |
+| `curl -XPOST … -d '{…}'` | `curl.exe --% -XPOST … -d "{…}"` — one line, no variables |
+
+**`$env:` is enough for compose to see it.** Verified rather than assumed, because the reverse
+already bit us once from the shell side: an exported variable does not reach a container unless
+compose passes it (`docker-compose.yml` says so where `PG_LLM_API_KEY` is declared). A
+`$env:PG_LLM_API_KEY = 'sk-…'` set in the same window does show up in `docker compose config`.
+
+**`curl.exe`, never `curl`.** In PowerShell 5.1 — the version that ships — `curl` is an alias
+for `Invoke-WebRequest`, so `curl -s <url>` fails with `Cannot process command because of one or
+more missing mandatory parameters: Uri`. That reads like a bad URL rather than a wrong `curl`,
+which is why it is worth naming. `curl.exe` is the real thing and is present on Windows 11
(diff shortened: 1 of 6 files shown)


Ari-Glikman:
## Correction: this PR does not contain what its title says it contains

**Read this before trusting `d48939e`'s commit message.** The squashed commit is titled for the `PollInterval` → `CallInterval` fix, but its actual file list is:

```
README.md                                       114 +++++++++++++-
docs/hl7-validation-and-trigger-inventory.md      4 +-
iris/labdemo/Production.cls                      22 ++-
iris/labdemo/Tools/Read.cls                       8 +-
services/detection-engine/src/detect/earlywarning.ts    99 ++++++++++++-
services/detection-engine/test/earlywarning.test.ts    163 +++++++++++++++++++++
```

Only rows 2–4 belong to the stated fix. The other three are **two unrelated pieces of work by two other agents**:

- **`README.md`** — Windows/PowerShell instructions.
- **`services/detection-engine/src/detect/earlywarning.ts` + its tests** — the fix for Early Warning projecting a crossing while the queue is *draining*.

### Cause

Mine, and it is a process failure rather than a git accident. I ran three agents concurrently against **one shared working tree** instead of giving each an isolated worktree. `HEAD` moved between one agent's `git checkout -b` and its `git commit`, so its commit landed on `devA/callinterval`; the same happened to the second. Squash-merging that branch then collapsed all three into one commit under my title. This is the second time tree-sharing between concurrent agents has cost this project real confusion, and the fix is `isolation: worktree`, not more care.

### Why the code was not reverted

The merged result is sound, and reverting to restore a tidy history would have been the more expensive mistake. Verified against the merged state:

- **353 engine tests pass, 0 fail** (`main` was 348 before, so the swallowed Early Warning work brought 5 green tests with it).
- The `CallInterval` fix is verified live: production compiles and starts clean, `Adapter.CallInterval=2` on the compiled item, **zero** `ErrProductionSettingInvalid` rows across a full stop/start cycle, and `GetHostSettings("EMR Source")` returns it.

### What is NOT verified by this PR

Neither swallowed change was reviewed as its own PR, so neither got the review its own author intended to request. Two specifics worth knowing:

- The README's Windows work was verified by execution by its author, including two real defects it found — `curl.exe -d` cannot carry this repo's JSON in PowerShell 5.1 by any obvious spelling (`--%` is required), and `docker compose logs iris | grep -E '^[0-9]'` **matched nothing on any host**, because `compose logs` prefixes every line with `pg-iris  | `. That second one was broken on Linux too and is fixed here.
- The Early Warning change landed at whatever commit its author had reached, which is not necessarily where they intended to stop. Being confirmed with them; anything outstanding follows as its own PR.

The commit message on `main` cannot be corrected without rewriting a merged history, so this comment is the record.

PR #146, opened 2026-08-26
fix(engine): a window-average slope announced a crossing while the queue drained

Ari-Glikman:
> **READ THIS FIRST — the diff below is ALREADY ON `main`.** While this branch was being written, a
> concurrent session switched the shared working copy from `devB/ew-draining` to `devA/callinterval`
> between the branch being created and the commit being made, so the commit landed on that branch and
> was squash-merged into `main` inside **#145** (`fix(iris): the poll interval setting was named
> PollInterval`). The engine content on `main` is byte-identical to this branch —
> `git diff main devB/ew-draining -- services/detection-engine` is empty. This PR therefore exists as
> the **review record** for a change that reached `main` without one, and as the place the reasoning
> lives instead of inside #145's unrelated squash message. **Close it rather than merge it**, unless
> you would rather revert the engine files out of #145 and land them here properly.

## The defect

Reported after a demo run: *"the early warning sometimes comes up when the queue pool is being
drained, because it takes a point in time measurement and does not notice the
acceleration/deceleration of pool growth."*

Confirmed. `projectHost` fitted **one** least-squares slope over the whole `fitWindowSeconds` window
and published a projection whenever that slope came out positive. An average over the window is a
statement about the window; *"Queue depth N rising ~X/min; at this rate it crosses 50 in ~M min"* is a
statement about the present. On the demo's own recovery path the window holds a long rise followed by
a partial drain, the average stays positive, and Early Warning announces a crossing while the queue is
visibly emptying — seconds after the audience is told the approved `set_pool_size 1 -> 4` worked.

**#142's `sinceLastDiscontinuity()` does not cover it and was never meant to.** It drops samples
before a >=80% single-sample collapse; a drain at 4/sec sheds ~20 of a 350-deep queue per 5s poll,
about 6%, so it never trips. The two mechanisms answer different questions — *"did the series
restart"* and *"is it rising now"*.

### Reproduced as a failing test first

`test/earlywarning.test.ts`, `a queue must be rising NOW, not merely on average`. The fixture is 40
samples rising 5/poll to 195, then 12 draining 15/poll to 15 — net +1/sec then net -3/sec, the MVP 2
scenario's own rates at the shipped 5s poll. Before the fix it returned a projection with
`slope = +25.1/min`; after it, `not_rising`.

The test asserts the traps as well as the behaviour, because a fixture that passes for the wrong
reason proves nothing:

- **ends at 15, under the `absoluteFloor` of 50.** Above the floor `already_crossed` answers first and
  the slope is never consulted — the trap #142's fixtures fell into twice.
- **fits the window itself and asserts that slope is still positive.** If the fixture ever stops
  reproducing the defect, this fails rather than passing quietly.
- **asserts `fitSampleCount === 52`**, so #142's helper cannot be what declines.

## The fix

Fit the window **and** its most recent portion, and decline unless **both** slopes are positive. One
gate covers three shapes a single fit reports as a rise: the post-fix drain, a rise that has levelled
off, and a rise that has turned over.

**`not_rising`, not an eighth reason.** `contracts/earlywarning-api.md` fixes the seven reasons and
their precedence, and `contracts/` is read-only to an implementation task (root `CLAUDE.md` §4). This
is not a workaround — a draining queue is not rising, and `EarlyWarning.tsx` already renders that
reason as *"Watching — not trending toward a threshold"* with a tick, which is a true and reassuring
thing to say about a queue that is emptying. **A distinct `draining` reason would be a reasonable
follow-up contract change** if the demo wants to *narrate* the recovery rather than merely decline to
forecast it.

**Precedence unchanged.** The new test sits at step 6, after `already_crossed`, so a queue that is
draining but still *above* its threshold keeps reporting `already_crossed` — draining-but-over-limit
is still a problem. Nothing on the wire changes shape, and the published `slope` is still the full
window's fit: the tail is a sign test whose slope is never published.

## Where the new number lives, and why

`RECENT_FIT_FRACTION = 0.4` of `fitWindowSeconds`, **hardcoded in the module, as a fraction** — the
same call `DISCONTINUITY_FRACTION` made, for the same reason. ADR 0003 governs the numbers that decide
what **fires**, and the `earlyWarning` numbers in `thresholds.json` each set a level or a span an
operator might defensibly move. This one sets neither: it says how much of the series the word "now"
covers, which is the *definition* of "rising" rather than a threshold for it.

A fraction rather than a second duration is what decided the placement:

- **it cannot contradict `fitWindowSeconds`.** Two independent numbers can be set so the tail is
  longer than the window it is a tail of — a nonsense state with no sensible behaviour. A fraction
  moves with the window and can never invert.
- **it inherits the reachability invariant instead of needing its own.** `validateConfig` already
  requires `fitWindowSeconds / poll > minFitSamples` so a projection is structurally possible (#64's
  failure mode in a new place). At the shipped 300s/5s that is 60 samples, of which the tail is 24. A
  configurable duration would need that check duplicated **plus** a row in `api/settings.ts`'s
  allowlist and its metadata, and an unreachable value would silence the module with no error.

`0.4` rather than shorter: 24 samples is enough that one bursty poll cannot flip the sign, where the
last 30 s (6 samples) of a queue that drains and refills would flap between projecting and not.
Rather than longer: at 0.8 the tail is nearly the window and would just re-answer the same question.
The tail does not need to react within a poll or two of the fix landing, because above the threshold
`already_crossed` answers first.

## Degenerate tails decline rather than crash

Both cases arrive as `fitSlopePerMs` returning `null`: fewer than two samples has no slope, and a tail
sharing one timestamp is division by zero rather than a flat line. **Two samples is all the tail
needs, not `minFitSamples`**, because its slope is never published — it confirms an answer the window
already grounded in 12+ samples, and at the shipped cadence the tail holds ~24, so reaching two at all
means a polling gap where declining is the posture the rest of the module takes.
`insufficient_samples` would be the wrong reason for either: the contract defines it against the
published `fitSampleCount`, which is the **full** window's count and has already cleared
`minFitSamples` by that point.

## Tests

Five new, one of which is the reproduction. The other four: a **decelerating** rise still projects
(deceleration is not a turnover, and declining there would be the opposite defect); a queue that has
**levelled off** declines; and the two degenerate tails decline rather than throw.

```
$ npm run typecheck
> tsc --noEmit          (clean)

$ npm test
tests 370
suites 74
pass 370
fail 0
```

365 before, 370 after. **Not verified on the live stack** — the compose stack is in use for a separate
investigation and was deliberately not rebuilt or restarted. The one thing unit tests cannot confirm
is that a bursty live ramp keeps a positive tail slope throughout the ~100 s window in which a
projection is expected; the 24-sample tail is chosen to make that stable, but it is worth a glance at
the next rehearsal.

### DIFF
--- a/services/detection-engine/src/detect/earlywarning.ts
+++ b/services/detection-engine/src/detect/earlywarning.ts
@@ -79,6 +79,44 @@ const METRIC: MetricName = 'queued';
  * reset or an approved fix produces, is 100%.
  */
 const DISCONTINUITY_FRACTION = 0.8;
+
+/**
+ * How much of the fit window counts as NOW, as a fraction of it.
+ *
+ * A single slope over the whole window describes the WINDOW, not the present, and publishing it as
+ * "rising ~N/min" is a claim about the present. So the projection asks the question twice — once of
+ * the window, once of its tail — and declines unless both say rising. Reported from a live demo run:
+ * "the early warning sometimes comes up when the queue pool is being drained, because it takes a
+ * point in time measurement and does not notice the acceleration/deceleration of pool growth."
+ *
+ * A FRACTION RATHER THAN A SECOND WINDOW LENGTH, and deliberately not in `thresholds.json`. ADR 0003
+ * governs the numbers that decide what FIRES, and the earlyWarning numbers there
+ * (`fitWindowSeconds`, `horizonSeconds`, `minFitSamples`) each set a level or a span an operator
+ * might defensibly move. This one sets neither: it says how much of the series the word "now" covers,
+ * which is the definition of "rising" rather than a threshold for it — the same argument
+ * DISCONTINUITY_FRACTION carries above for "the series restarted".
+ *
+ * Two consequences make the fraction the safer shape than a configurable duration:
+ *
+ *   - it cannot contradict `fitWindowSeconds`. Two independent numbers can be set so the tail is
+ *     LONGER than the window it is a tail of, which is a nonsense state with no sensible behaviour;
+ *     a fraction moves with the window and can never invert.
+ *   - it inherits the reachability invariant instead of needing its own. `thresholds.json` already
+ *     requires `fitWindowSeconds / poll > minFitSamples` so a projection is structurally possible
+ *     (#64's failure mode); at the shipped 300s/5s that is 60 samples, of which the tail is 24. A
+ *     separate configurable duration would need that check duplicated, and an unreachable value
+ *     would silence the module with no error — which is exactly what the existing check exists to
+ *     prevent.
+ *
+ * 0.4 rather than something shorter: at the shipped poll the tail is 24 samples, enough that one
+ * bursty poll cannot flip its sign, where the last 30s (6 samples) of a queue that drains and refills
+ * would flap between projecting and not. Rather than something longer: at 0.8 the tail is nearly the
+ * window and would just re-answer the same question. Note the drain only reaches this code once the
+ * queue is UNDER the threshold — above it, `already_crossed` answers first — so the tail does not
+ * need to react within a poll or two of the fix landing.
+ */
+const RECENT_FIT_FRACTION = 0.4;
+
 const FINDING_TYPE = 'queue_buildup';
 const SLOPE_UNIT = 'items/minute';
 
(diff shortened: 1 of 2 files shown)


Ari-Glikman:
Closing without merging. The content is already on `main` via #145 — `git diff origin/main origin/devB/ew-draining` shows **no difference in `earlywarning.ts` or `earlywarning.test.ts`**, so there is nothing here to land.

Merging it would be actively harmful: this branch's merge base is the old `main` tip (`9a44346`), so the merge would *revert* the three other changes #145 carried — 114 lines of README Windows instructions, the `CallInterval` fix in `Production.cls` and `Tools/Read.cls`, and the doc updates. That is the shared-working-tree accident described in #145's correction comment, seen from the other side.

Keeping it as the review artifact for the Early Warning fix, which is the one thing #145's title does not describe. The reasoning, fixture and measurements in this PR's description are the record for that change.

Verified on `main` after the fact, on the supported runtime (node v24): **370 tests, 370 pass, 0 fail, 0 skipped**, `tsc --noEmit` clean.

One thing worth pinning while it is fresh: **`node --test "test/**/*.test.ts"` under node 22 collects only 353 of the 370 tests** and reports `fail 0`, so it looks green. It silently drops `test/silence.test.ts` and `test/triggers.test.ts` (17 tests); both pass when named explicitly. Node 20 on PATH cannot run these files at all and reports `tests 0`. Use node >= 24.

PR #150, opened 2026-08-26
fix(detection-engine): clearing a finding takes the same bar as raising one (#149)

### LINKED ISSUE #149: engine: a confirmed finding is retracted by ONE non-breaching poll, so an oscillating condition churns ids
Reported as "sometimes the findings do not disappear or appear with no clear cause". Two mechanisms, both in the engine, both reproduced against the shipped defaults (`sustainedSamples: 2`, `sustainedSeconds: 4`, 5s poll).

## 1. The clear bar is asymmetric — `registry.ts`

Appearing costs 2 consecutive breaching samples (MVP §6). Leaving costs **1** non-breaching poll: `update()` deletes any tracked condition that did not breach this poll, outright.

So a condition oscillating around a threshold produces a finding, drops it, and produces **the same condition again under a new id** — measured `f-1000` -> gone -> back as `f-1001`. That contradicts `healthscan-api.md` Q4, which promises the id is stable for the lifetime of the condition: an id that survives a poll but not a blip is stable for the lifetime of an uninterrupted breach, which is not the same thing.

Related: a rule returns one undifferentiated `null` for both "not breaching" and "cannot tell". `queue_buildup` declines to evaluate when `queued` is null (#49), and the registry reads that silence as a clear — so an instance that stops publishing a metric retracts a finding that is still true.

## 2. One absent poll forgets the host AND its baseline — `engine.ts`

The departure sweep runs on the first poll that omits a host, and calls `baselines.forget()` as well as `registry.forget()`. At `minBaselineSamples: 12` and a 5s poll, a host that dropped out of a single payload comes back `warming`, so every comparative rule is silent for a further minute.

This half is worse than a flap: the finding does not flicker, it goes away and cannot return, for a reason invisible from outside the process.

## What is NOT this bug

Driving a real build-up and drain on the live stack (134 samples, 5s apart) showed no flapping: `queue_buildup` confirmed on the second breach, held one id for 3+ minutes through cap oscillation at 118-122, and cleared exactly once on a monotonic drain through the floor. The live path is clean — this reproduces when a value oscillates *across* the threshold, or when an input goes unmeasurable, neither of which that run produced.

The other half of the original report was #147 (`Cloud API` shipped an unresolvable HTTP target, so findings depended on whether a marker-gated first-boot override had ever run).

Ari-Glikman:
Closes #149. The residual half of the "findings appear and disappear with no clear cause" report; the other half was #147.

Two commits, deliberately separate — the code fix, then the contract amendment it makes necessary (CLAUDE.md §4).

## What was wrong

**Appearing cost two consecutive samples. Leaving cost one poll.** `registry.update()` deleted any tracked condition that did not breach on the current poll, outright. So a condition oscillating around a threshold produced a finding, dropped it, and produced *the same condition* again under a **new id** — measured `f-1000` -> gone -> `f-1001` against the shipped defaults (`sustainedSamples: 2`, `sustainedSeconds: 4`).

That contradicted `healthscan-api.md` Q4. An id that survives a poll but not a blip is stable for the lifetime of an uninterrupted breach, not for the lifetime of the condition — and the condition outlives the blip.

**A rule's `null` meant two different things.** "Not breaching" and "cannot tell" arrived as the same silence. `queue_buildup` declines to evaluate a null depth (#49), so an instance that stopped publishing a metric *retracted a finding that was still true*.

**One absent poll forgot the host and its baseline.** This half is worse than a flap: the departure sweep called `baselines.forget()` alongside `registry.forget()`, so at `minBaselineSamples: 12` and a 5s poll a host that dropped out of one payload came back `warming` and every comparative rule was silent for a further minute. The finding did not flicker — it went away and could not return, for a reason invisible from outside the process.

## The fix

- **Symmetric bar.** Clearing uses the same two gates as confirming, from the same two knobs. No third tunable: "how many samples before I believe this changed" is one question asked in both directions. `sustainedSamples: 1` / `sustainedSeconds: 0` still clears on the first non-breaching poll, so the old behaviour is reachable by configuration — tested.
- **An UNCONFIRMED condition still resets outright.** MVP §6 governs appearance and that part was correct; only a published finding gets hysteresis.
- **`Rule.requires`** — each rule declares the inputs whose `=== null` guard makes it return null. The engine derives the unmeasurable set per poll and the registry *holds* those conditions, counting the poll toward neither gate. `dead_host` declares `[]`: it judges `status`, always present, so its silence really is "not breaching" and it clears as before.
- **Absence debounce**, same two gates, plus `#absent` cleared on a structural `reconfigure`.
- A blip does **not** reset `firstSeenAt`/`confirmedAt` — re-serving the confirmation clock would make one absent poll delay the finding all over again.

## What this is not

Live driving did not find it and does not reproduce it. A real build-up and drain on the stack — 134 samples, 5s apart — confirmed `queue_buildup` on the second breach, held one id for 3+ minutes through cap oscillation at 118–122, and cleared exactly once on a monotonic drain through the floor. It reproduces when a value oscillates *across* the threshold or an input goes unmeasurable. Worth stating, because "cannot reproduce on the happy path" is what left this open.

## Verification

```
tsc --noEmit          clean
node --test           375 tests, 375 pass, 0 fail   (was 370; 3 updated, 5 added)
contracts/validate    all checks passed (3 samples, 15 accept, 26 reject, 7 capture claims)
```

Node 24 — node 20 cannot run these `.ts` files and reports `tests 0` while looking green.

The three updated tests encoded the old one-poll clear (`makes the finding DISAPPEAR…`, `forgets a host that leaves…`, the alert age-out, and the series-departure test) and now poll to the bar. The five added ones cover the blip hold with id intact, the unmeasurable hold, `dead_host` still clearing, the baseline surviving one absent poll, and the 1-sample back-compat config.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,46 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-26 — clearing a finding takes the same bar as raising one (#149)
+
+**`healthscan-api.md` §2.1 (a new paragraph under *Sustained breach*) and Q4.** No schema
+change, no field added or removed, no sample change — this documents *when* a finding leaves,
+which the contract described only for arriving.
+
+**Q4 promised two things that the implementation could not both keep.** "ids are stable while
+the condition persists" and "findings disappear when cleared" are only compatible if *cleared*
+and *not breaching on this poll* are the same event. They are not. MVP §6's 2-consecutive-samples
+rule was read as governing appearance alone, so a finding cost **two** samples to arrive and
+**one** to leave, and a condition oscillating around a threshold produced a finding, dropped it,
+and produced the same condition again under a **new `id`** — `f-1000`, gone, back as `f-1001`.
+Reproduced against the shipped defaults (`sustainedSamples: 2`, `sustainedSeconds: 4`), not
+theorised. To a viewer that is findings appearing and vanishing with no cause, which is how it
+was reported.
+
+Clearing now uses the same two gates, read from the same two knobs — no third tunable, because
+"how many samples before I believe this changed" is one question asked in both directions.
+`sustainedSamples: 1` / `sustainedSeconds: 0` still clears on the first non-breaching poll, so
+the old behaviour remains reachable by configuration.
+
+Two consumer-visible consequences, both stated in §2.1:
+
+- a finding may persist **one extra poll** past the condition, showing its last breaching
+  numbers (there is no newer verdict to refresh from)
+- a finding **holds while its inputs are unmeasurable** — Q13's `null`. A rule that declines to
+  evaluate has said "cannot tell", not "not breaching". Each rule now declares the inputs it
+  cannot work without, so the engine can tell the two silences apart; `dead_host` declares none
+  and clears exactly as before.
+
+The same bar now governs a host vanishing from the proxy payload. That half was worse than a
+flap: forgetting on the first absent poll discarded the **rolling baseline** along with the
+findings, so a host that dropped out of one payload came back `warming` and every comparative
+rule was silent for `minBaselineSamples` further polls. The finding did not flicker — it went
+away and could not return, for a reason invisible from outside the process.
+
+**Nothing changes for a consumer that renders whatever the endpoint returns.** What changes is
+that an `id` is now stable for the lifetime of the condition, as Q4 always said it was.
+
+
 ## 2026-08-26 — `type` gains a second source; Q6's "treat unknown as a real value" amended (#127)
 
 **`proxy-api.md` §1.1, §1.3, §5.1 (Q6) and a new §5.1.1; `proxy.schema.json`
--- a/contracts/healthscan-api.md
+++ b/contracts/healthscan-api.md
@@ -125,6 +125,29 @@ Sorted `detectedAt` descending, severity as tiebreak (critical → warning → i
 finding is emitted. A single-sample spike produces nothing. This is why findings are stateful
 and why `id` can be stable (Q4).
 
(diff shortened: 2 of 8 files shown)


Ari-Glikman:
Verified live on the running stack, detection-engine rebuilt from this branch.

Build-up: `queue_buildup#f-1062` confirmed on the second breach and held **one id** for ~2 minutes while `currentValue` rose 85 -> 127, alongside `slow_processing#f-1060` and `growing_queue_wait#f-1061`.

Drain (`Reset()` + `set_pool_size` 1 -> 4), sampled every 5s:

```
16:57:54 q=127  queue_buildup#f-1062@127|slow_processing#f-1060@1.01|growing_queue_wait#f-1061@33.94
16:57:59 q=0    queue_buildup#f-1062@127|slow_processing#f-1060@1.01|growing_queue_wait#f-1061@33.94
16:58:04 q=0    slow_processing#f-1060@1.01|growing_queue_wait#f-1061@33.94
...
16:58:34 q=0    growing_queue_wait#f-1061@33.97
...
16:59:35 q=0    -
```

The 16:57:59 row is the hysteresis doing its job: the queue is already 0 and the finding is still served, **showing its last breaching value of 127** — the one-extra-poll staleness the contract amendment now states. It clears on the next poll and does not come back. `slow_processing` and `growing_queue_wait` each cleared exactly once, later, as their own metrics aged out.

Also confirmed #148 is live and steady in the same window: 12 hosts, `untypedHosts: ['Ens.Alarm']`, `typeDisagreements: []`. One earlier poll showed `Ens.MonitorService` untyped as well — a transient, which is precisely what `untypedHosts` exists to report rather than a stale claim in a doc.

PR #151, opened 2026-08-26
docs(contracts): not_rising is two fits, not one — the window and its tail

Ari-Glikman:
Closes item 2 of the six reported issues. **The code fix already shipped**; this is the contract catching up.

## What was wrong

`contracts/earlywarning-api.md` §2.1 defined step 6 as:

> `not_rising` | Fitted slope is `<= 0` after rounding to 1 decimal. Flat or draining — nothing is approaching anything

Singular. But since `d48939e`, `earlywarning.ts` fits the series **twice** — once over the 300 s window, once over its most recent 40% (`RECENT_FIT_FRACTION = 0.4`) — and declines unless both are positive. A queue whose window slope is positive but whose tail is flat or falling reports `not_rising`, and nothing in the contract said so.

## What was reported

> the early warning sometimes comes up when the queue pool is being drained, because it takes a point in time measurement and does not notice the acceleration/deceleration of pool growth.

A single slope over 300 s describes the **window**, not the present, and "rising ~N/min" is a claim about the present. The tail test is what makes the two agree. One gate covers three shapes a single fit reads as a rise — draining after the approved fix, levelled off, turned over — while a queue rising *more slowly* than it was still projects, because decelerating is not falling.

## Why the contract fell behind

`d48939e` was squash-merged as #145 under a title naming only the `PollInterval` fix, because two agents' commits collapsed into one merge. So the change landed with no contract amendment and no `CHANGELOG.md` entry, and was found only by reading the file rather than from the changelog. Recorded in the entry because #84 is open about this exact class of drift: the code was correct, tested (5 tests at `test/earlywarning.test.ts:433-520`), and invisible.

## What this PR contains

- **`earlywarning-api.md` §2.1** — the `not_rising` row now says *either* fit
- **`earlywarning-api.md` §2.2.1**, new — the two-fit rule, the three shapes, and two things a consumer cannot infer from the reason alone:
  - the tail slope is **never published**, so a positive `slope` alongside `projection: null` cannot occur
  - an **unfittable tail** declines as `not_rising`, not `insufficient_samples` — that reason is defined against the published `fitSampleCount`, the full window's count, which has already cleared `minFitSamples` by step 6
- **`CHANGELOG.md`** — the §4 entry

No schema change, no eighth reason, no change to the precedence order. `not_rising` stays at step 6 *after* `already_crossed`, so a queue draining while still above its threshold keeps reporting `already_crossed` — draining-but-over-limit is still a problem.

## Verification

`node contracts/validate.mjs` — all checks passed (3 samples, 15 accept, 26 reject, 7 capture claims).

Docs only; no code touched, so the engine suite is unaffected by this diff.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -36,6 +36,11 @@ never published (so a positive `slope` alongside `projection: null` cannot occur
 tail declines as `not_rising` rather than `insufficient_samples`, which is defined against the full
 window's `fitSampleCount`.
 
+**One older undocumented path is written down at the same time**, found while checking the above: a
+non-finite or `<= 0` `secondsToThreshold` also declines as `not_rising`. It predates the tail test and
+should be unreachable — it would mean the arithmetic disagreed with `already_crossed` a step earlier —
+but it was the third way to reach step 6's reason code and the contract described only one.
+
 
 ## 2026-08-26 — clearing a finding takes the same bar as raising one (#149)
 
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,38 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-26 — `not_rising` is two fits, not one; documenting behaviour that shipped in #145
+
+**`earlywarning-api.md` §2.1 (the `not_rising` row) and a new §2.2.1.** No schema change, no new
+reason, no change to the precedence order — the seven reasons and their order are unchanged, and a
+consumer sees exactly the same values it did before.
+
+**This documents code already on `main`, which is the problem being fixed.** Step 6 has fitted the
+series **twice** since `d48939e` — once over the 300 s window, once over its most recent 40% — and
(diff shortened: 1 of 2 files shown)


Ari-Glikman:
**Self-review, and one addition.** Verified each claim in the new §2.2.1 against `earlywarning.ts` rather than against the summary of it:

- `slope` published at line 385 is `slopePerMinute`, the **window** fit. The tail fit (line 359) only gates. Claim holds.
- Line 362 declines on `recentSlopePerMinute === null` as well as `<= 0`, so an unfittable tail does reach `not_rising`. Claim holds.
- Ordering: `already_crossed` at line 333 precedes both fits. Claim holds.

Checking those turned up a **third** `not_rising` path the contract never described — the pre-existing guard at line 372, where a non-finite or `<= 0` `secondsToThreshold` declines rather than publishing `secondsToThreshold: 0` (the §1.4 "zero reads as a measurement of now" case). It predates the tail test and should be unreachable, since reaching it means the arithmetic disagreed with `already_crossed` a step earlier. Documented in `7822636` — a contract that lists one route to a reason code when there are three is the same defect this PR opened to fix, just older.

`node contracts/validate.mjs` still green after both commits. Merging.

PR #152, opened 2026-08-26
docs: Windows instructions where they were still missing

Ari-Glikman:
Item 1 of the six reported issues: *"I'd like the readme to include windows instructions, not just unix/linux."*

## Where it already was, and why that wasn't the end of it

The **root** `README.md` grew a thorough *On Windows* section — in `d48939e`, squash-merged as #145 under a title naming only the `PollInterval` fix. Same hidden merge that concealed item 2's code, so this looked unstarted.

It stopped at the root, though. Seven other docs a developer follows were still POSIX-only, and three of them **fail outright** rather than degrading.

## The one that actually breaks

There is no one-shot `VAR=value <cmd>` in PowerShell. It parses the prefix as the *command name*:

```
POLL_INTERVAL_MS=2000 : The term 'POLL_INTERVAL_MS=2000' is not recognized as the name of a cmdlet,
function, script file, or operable program.
```

That reads as a missing program rather than a shell difference. The replacement — `$env:VAR = '...'` on its own line — **is not one-shot**, and that asymmetry is the part worth documenting: it persists for the whole window, so

- a `$env:POLL_INTERVAL_MS = '2000'` set to watch one appear-and-clear cycle is still set at the next `npm start`, leaving the engine at a rate `src/index.ts` says breaks two invariants
- a deliberately-wrong `$env:IRIS_BASE_PATH` set to prove failure detection works keeps 404-ing every poll afterwards, with a symptom identical to a genuinely broken instance

Every site that uses the prefix now carries the `Remove-Item Env:\VAR` undo next to it.

## Per document

| Doc | What it needed |
|---|---|
| `README.md` | two new translation-table rows (env prefix, `grep -rn`) and the persistence caveat |
| `services/detection-engine/README.md` | paired live-proxy start; `curl.exe` note |
| `apps/dashboard/README.md` | paired compressed-poll start (+ trap), `CONTRACT-Q` grep, fallback-open |
| `docs/demo/cue-sheet.md` | pre-flight exports, both `Ens.Util.Log` purges, check-table `grep`/`curl` |
| `iris/labdemo/README.md` | `curl.exe` for every call; why `-u` is the flag that makes it obvious |
| `services/metrics-proxy/fixtures/README.md` | `MOCK_FIXTURE=` / `MOCK_HOSTSTATUS=` selection |
| `services/metrics-proxy/CLAUDE.md` | the deliberate-misconfiguration form |
| `iris/setup/README.md` | the class path is on the **IRIS** filesystem, not the shell's |

`apps/dashboard/fixtures/README.md`, `contracts/README.md` and `docs/decisions/README.md` needed nothing — npm-only, or already carrying their Git Bash caveat.

## Two claims were wrong on the first attempt

Everything here was run on a Windows 11 box rather than recalled, and that caught two:

1. **`Select-String -r` does not exist.** I wrote it into the translation table first. `Select-String : A parameter cannot be found that matches parameter name 'r'.` Recursion is the pipeline's job: `Get-ChildItem -Recurse src | Select-String "pattern"`.
2. **A `@"..."@` here-string interpolates**, so it is the wrong translation of `<<'EOF'`. Measured against the running stack, `write "PGV|",$zversion,!` arrived at IRIS as `WRITE "PGV|",,!` — PowerShell had already blanked `$zversion`, and IRIS echoed a syntactically valid line that printed nothing. **`@'...'@`** passes it through and IRIS answers with the build string.

The second one is the reason the cue-sheet change is worth more than its size: the pre-flight here-string purges the event log, and trap #2 in that same document is *"a previous scenario's errors stay in the 60-minute window and a later diagnosis reads them as evidence"*. A purge that silently does nothing while looking like it ran is exactly the on-stage failure the pre-flight exists to prevent.

Also verified: `$env:` + `Remove-Item Env:\`, `curl -u` reporting `parameter name 'u' is ambiguous`, `start` being present in Git Bash as `/usr/bin/start` (so that line needed the *Unix* halves, not a Windows one), ObjectScript not escaping `\` (`$length("C:\repos\x.cls")` is 14), and both new cue-sheet PowerShell commands run end to end.

## Two verified drifts fixed in passing

Both in `services/detection-engine/README.md`, checked against the code rather than assumed:

- `npm test # 95 tests` → **375** (`ℹ tests 375 / pass 375 / fail 0`)
- `POLL_INTERVAL_MS` default `10000` → **5000** (`DEFAULT_POLL_INTERVAL_MS = 5_000`), with a pointer to the floor sweep in `src/index.ts`

Docs only; no code touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ substitution — the log pipelines and the JSON POSTs — are paired at the poin
 | this README says | PowerShell |
 |---|---|
 | `export PG_AGENT_MODE=live` | `$env:PG_AGENT_MODE = 'live'` |
+| `VAR=value <cmd>` (one command only) | `$env:VAR = 'value'` on its own line, then `<cmd>` — **and it stays set**, see below |
 | `… \| grep pg-firstboot` | `… \| Select-String pg-firstboot` |
+| `grep -rn "pattern" src/` | `Get-ChildItem -Recurse src \| Select-String "pattern"` — **not** `Select-String -r`, which has no such switch |
 | `curl -s <url>` | `curl.exe -s <url>` |
 | `curl -XPOST … -d '{…}'` | `curl.exe --% -XPOST … -d "{…}"` — one line, no variables |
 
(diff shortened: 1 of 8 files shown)


Ari-Glikman:
**One thing observed and deliberately not changed here.** While pairing the compressed-poll block I checked `POLL_INTERVAL_MS=2000` against the sweep `src/index.ts` records:

```
4500ms   the jitter invariant fails -- a fetch quicker than the last slips confirmation
2500ms   the reachability invariant fails -- nothing confirms on the second sample
1750ms   coverage finally drops, 6/8 finding types
```

So the value this README recommends, `2000`, is **below both invariant floors** and above only the coverage floor. Its claim — *"full appear-and-clear cycle, still emits"* — is accurate, because coverage is still 8/8 at 2000: confirmation just happens on a later sample than the design intends. The surrounding prose frames the floor purely in terms of emission (*"finding types stop emitting entirely below ~1750 ms"*), which is why the two documents read as agreeing when they are measuring different things.

Not touched, because changing a documented demo procedure is a different decision from translating it, and the advice does produce the cycle it promises. Flagging it so the choice is on the record rather than discovered later.

PR #153, opened 2026-08-26
feat(iris): the chat can read Ens_Util.Log — two tools, a shared host roster, one inference in the payload

Ari-Glikman:
## What this is

Item 4 of six: **"for the chat I would like to add this table `Ens_Util.Log`."**

Two new read tools — `GetEventLogSummary` and `GetEventLogTrend` — plus the governance, contract and
prompt work to make the chat use them correctly. Before this the chat had no tool that could answer
"any errors?" at all; it answered from throughput, which is inference dressed as a reading.

`ReportTools()` goes **10 → 12**, verified reading `12 (expected 12)`.

## The defect that shaped the design

Asked *"did the production have any trouble starting or stopping today"*, the agent called both tools,
**counted the 10 error rows correctly**, and closed with *"there were no issues starting or stopping
the production"* at `confidence: 0.9`. Two prompt sentences were added telling it to read `sources[]`.
It then closed with *"the production started and operated without additional issues indicated in the
logs"* at **`confidence: 1`**. Nine of those rows were `Ens.Director.StartProduction` failures, sitting
in the array it had just been told to read.

The missing piece was never evidence. It was the **join** from a dictionary-verified method name to
"the production failed to start" — and because every code was `unclassified`, an unrecognised code read
as an absent fault. Prose held on **one run of three**, so the join moved into the payload as
`lifecycleFaults[]`: a fixed catalogue keyed on `SourceMethod`, `ErrorCatalogue.Summary`'s shape
exactly, deriving nothing from a log row. **Correct on three runs of three afterwards.**

The array is always present on the `(production)` entry, so empty means *measured clean* rather than
*not checked*.

## Three measurements worth reading

**`ConfigName` is NULL, not empty**, on the rows `Ens.Director` writes about the production itself.
`WHERE ConfigName = ''` → 0 rows; `IS NULL` → 91. So no host-filtered query can reach them, which is
why `get_recent_errors` has **never once** reported a production that failed to start.

**Framework hosts are labelled, not filtered** — §3.7's precedent rather than `healthscan-api.md` §2's.
`application` is `true` / `false` / **`null`**, and the third value marks `(production)`, which is
neither an application host nor framework plumbing. `false` there would file the production's own start
failures under framework noise.

**Zero inverts §2.1's sign in this family.** A log count of `0` is a real measurement. `null` stays
reserved for what could not be read. Every other tool family says the opposite, which is why
`Tools.EventLog` is a third class rather than more methods on an existing one.

## Four JSON type fixes, all toward what the contract already showed

`set out.found = 1` on a `%DynamicObject` emits the **number** `1`, so eight fields across §3.1, §3.2,
§3.5, §3.7 and §3.9 (`found`, `enabled`, `measured`, `readable`, `application`) changed JSON *type*
depending on which branch ran. The contract's samples have always shown `true`.

Found by smoke-testing the chat, not by reading code: asked *"how many hosts are in this production"*
it answered *"7 … both application hosts and framework-related hosts"* against 3 application hosts,
while the prompt paragraph telling it to read the flag names `true` and `false`. **Fixing the types
alone, with no prompt change**, gave *"3 application hosts — EMR Source, Lab Router and Cloud API"* on
three runs of three.

Two empty strings became `null` for the same reason: `newestSecondsAgo` (§3.4 says `integer | null`)
and `summary` for an unclassified code (§3.4a says `null`). `""` is not the same claim as `null` — it
reads as *there is a summary and it is blank*. `get_event_log_summary` emitted `null` there from the
start, so **two tools reading the same catalogue disagreed on the same field**.

## Two allowlists became one copy each

`ClassifyError` / `ErrorSummary` → `Tools.ErrorCatalogue`; `IsApplication` / `IsConfigItem` →
`Tools.HostRoster`. Both moved unchanged line for line, when `Tools.EventLog` needed the same answer.
Neither class extends `%AI.Tool`, which is what lets the methods be public without becoming tools —
`FindTools` filters on `$issubclassof(className, "%AI.Tool")`. `HostRoster` asks the running
`Ens.Config.Production` rather than carrying a framework-name list, per root `CLAUDE.md` §6.

## A pre-existing mismatch documented rather than fixed

**The snake_case names in the contract's catalogue column have never been callable.** `%AI.Tool`
derives the runtime name from the ClassMethod name, so the real names are `GetEventLogSummary`,
`CompareHostActivity` and so on — measured, and `investigation-api.md`'s `evidence[].tool` has been
carrying `"functions.GetEventLogSummary"` all along. The model was bridging by resemblance. §1 now
labels that column as section titles and states the consequence: **renaming a ClassMethod renames a
tool, and is a contract change.**

## Verification

Every claim here was read off the live instance rather than reasoned about.

- Full `LoadDir` of `iris/labdemo/Tools` + `ChatDispatcher` + `AIHub` compiles clean with verbose
  `$system.OBJ.Load(file,"ck")` — not `"ck-d"`, which returns OK over a failed compile.
- `ReportTools()` reads `12 (expected 12)`.
- Every payload in the new §3.10/§3.11 is pasted from emitted JSON, including the boolean and null
  types. The one exception is stated in the changelog: the `summary`-is-null branch is verified in its
  two halves (the catalogue returns `""`; `%Set(…,"null")` emits `null`) because no host-scoped
  unclassified error exists on this instance to drive it end to end.
- Chat driven through `POST /labdemo/chat/ask` on IRIS directly, since the deployed stack is
  `AGENT_MODE=mock` and the engine's `/api/chat` answers 503 by design.
- The lifecycle question: **3/3 correct**. The host-count question: **3/3 correct**. Capability and
  "any errors in the last hour" both sane.
- **AI Detective re-smoked twice** through `POST /labdemo/agent/investigate` — `"all"` now includes the
  event-log pair, so this is the MVP 2 demo path with two more tools registered. Same root cause, same
  `set_pool_size Cloud API 4`, `confidence: 0.9` both times.

## Flagged, not silently changed

- **`iris/labdemo/Tools/HostRoster.cls` and `ErrorCatalogue.cls` are both in CI's skip list**, and only
  because their class comments mention `%AI.` — the skip list is `grep -rl '%AI\.' --include='*.cls'
  iris/`. Neither class references `%AI.` in *code*, so both could compile in CI and would give real
  coverage. Left alone: changing the skip predicate is a `.github/**` change, not part of this.
- `HostRoster` declares its own `Parameter PRODUCTION`, a **sixth** copy of that string. Named in the
  class comment rather than hidden. Repointing the other five is a five-file refactor and would put a
  refactor inside a feature.
- Stale `Tools.Read.ClassifyError` / `ErrorSummary` references remain in
  `docs/hl7-validation-and-trigger-inventory.md` (lines 205, 241, 337) and
  `docs/production-guardian-mvp3.md` (line 119). Left alone — `docs/`.
- One imprecision seen once in three lifecycle runs: the model said "the last errors appeared at
  approximately 14:00" and then correctly gave the 12:00 and 13:00 buckets in the same sentence. The
  buckets were right; the preamble was loose. Not the load-bearing defect and not chased.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -5,6 +5,89 @@ Every contract change, dated, with the reason. Newest first.
 ---
 
 
+## 2026-08-26 — the event log becomes readable: two tools, a shared host roster, and one inference moved into the payload
+
+**`mcp-tools.md` §1, §3.9's closing note, and two new sections §3.10–§3.11.** Additive: ten tools
+become twelve, eleven read and one write. No existing tool's input, output or name changes, so a
+consumer that never calls the new pair sees nothing different.
+
+**Asked for directly**: *"for the chat I would like to add this table Ens_Util.Log."* Before this the
+chat had no tool that could answer "any errors?" at all — it would answer from throughput, which is
+inference dressed as a reading.
+
+`get_event_log_summary` groups a window by host and severity, classifies error and warning text through
+the §3.4a allowlist, and never returns text. `get_event_log_trend` gives volume by severity bucket by
+bucket, empty buckets included. Both are `PG_Read` and both are registered on the chat agent's
+`"chat"` tool set, which was **added rather than widening `"activity"`**, so no existing caller's
+meaning changed.
+
+**Three things a consumer cannot infer, so they are written down:**
+
+- **`(production)` is a sentinel, not a config item.** `ConfigName` is **NULL, not empty**, on the rows
+  `Ens.Director` writes about the production itself — measured: `= ''` returns 0 rows, `IS NULL`
+  returns 91. So no host-filtered query can ever reach them, and `get_recent_errors` has therefore
+  never once reported a production that failed to start.
+- **Framework hosts are labelled, not filtered** — a deliberate divergence from `healthscan-api.md`
+  §2, which drops them. `application` is `true` / `false` / **`null`**, and the third value is the
+  interesting one: it marks `(production)`, which is neither an application host nor framework
+  plumbing. `false` there would file the production's own start failures under framework noise.
+- **Zero inverts §2.1's sign in this family.** A log count of `0` is a real measurement — the query ran
+  and nothing was logged. `null` stays reserved for what could not be read. Every other tool family
+  says the opposite, which is exactly why this needs stating.
+
+**`lifecycleFaults[]` is the entry worth reading twice, because it is a contract field that exists to
+compensate for a model, not for a table.** Asked *"did the production have any trouble starting or
+stopping today"*, the agent called both tools, **counted the 10 error rows correctly**, and closed with
+"there were no issues starting or stopping the production" at `confidence: 0.9` — then, after two
+prompt sentences were added telling it to read `sources[]`, closed with "the production started and
+operated without additional issues indicated in the logs" at `confidence: 1`. Nine of those rows were
+`Ens.Director.StartProduction` failures sitting in the array it had just been told to read.
+
+The missing piece was never evidence. It was the **join** from a dictionary-verified method name to
+"the production failed to start" — and because every code was `unclassified`, an unrecognised code read
+as an absent fault. Prose held on one run of three, so the join moved into the payload as a fixed
+catalogue keyed on `SourceMethod`: `ErrorCatalogue.Summary`'s shape exactly, deriving nothing from a log
+row, so §6 is untouched. Correct on three runs of three afterwards. **The array is always present on
+the `(production)` entry**, so empty means *measured clean* rather than *not checked*.
+
+**A pre-existing contract/runtime mismatch is documented rather than fixed**, in a new subsection under
+§1: the snake_case names in the catalogue column have **never** been callable. `%AI.Tool` derives the
+runtime name from the ClassMethod name, so the real names are `GetEventLogSummary`,
+`CompareHostActivity` and so on — measured, and `investigation-api.md`'s `evidence[].tool` has been
+carrying `"functions.GetEventLogSummary"` all along. The model was bridging by resemblance. The column
+is now labelled as section titles, with the consequence stated: **renaming a ClassMethod renames a
+tool, and is a contract change.**
+
+**A new §2 convention: a boolean field is a JSON boolean.** No sample changes — the samples in
+`mcp-tools.md` have always shown `true`; the **implementation** was emitting `1`. `set out.found = 1`
+on a `%DynamicObject` emits a number, so eight fields across §3.1, §3.2, §3.5, §3.7 and §3.9 (`found`,
+`enabled`, `measured`, `readable`, `application`) changed JSON type depending on which branch ran.
+
+Found by smoke-testing the chat after the event-log work landed, not by reading code: asked "how many
+hosts are in this production", it answered *"7 … both application hosts and framework-related hosts"*
+against 3 application hosts, while the prompt paragraph telling it to read the flag names `true` and
+`false`. **Fixing the types alone, with no prompt change, gave "3 application hosts … EMR Source, Lab
+Router and Cloud API" on three runs of three.** It survived this long because `1` and `true` are the
+same value to every consumer except the one that reads the field name.
+
+**Two `null`s that were empty strings, found the same way and in the same class.** §3.4's table says
+`newestSecondsAgo` is `integer | null` and §3.4a says an unclassified code's `summary` is `null`;
+`get_recent_errors` emitted `""` for both. An ObjectScript `""` inside a `%DynamicObject` literal
+serialises as the string `""` — measured: `{"a":("")}` gives `{"a":""}`, `%Set(k,"","null")` gives
+`{"k":null}`. `""` is not the same claim as `null`: it reads as *there is a summary and it is blank*,
+which is §2.1's defect wearing a different type. `get_event_log_summary` has emitted `null` there from
+the start, so **two tools reading the same catalogue disagreed on the same field** — the cost of the
+copy §3.4a warns about, showing up as soon as there were two callers. `newestSecondsAgo: null` is
+verified in emitted output; the `summary` branch is verified in its two halves (the catalogue returns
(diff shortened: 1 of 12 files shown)


tanifgit (commented):
**Post-merge review.** This PR requested my review as a `contracts/` change (root `CLAUDE.md` §4, `.github/CODEOWNERS` `/contracts/ @Ari-Glikman @tanifgit`) and was merged with the request outstanding and `reviews: []`. I was on vacation, not gone — see the note at the bottom. Reviewing it now because the gate §4 describes is "someone other than the author looked at it", and that has not happened yet for `mcp-tools.md`.

**Verdict: I would have approved this.** Two non-blocking comments, filed as issues. The claims I checked held up, and I checked the load-bearing ones against the code rather than the PR body:

- **The PHI boundary holds by construction, not by filtering.** Query one (`EventLog.cls:206-208`) selects no `Text`, `Stack` or `StatusValue` at all; query two (`:252-255`) is the only one that reads `Text` and is bounded by `Type <= ..#CLASSIFIEDMAXTYPE`. The part that actually makes this safe is one step further in, and worth stating explicitly because it is stronger than the `Type <= 3` argument the PR leads with: `ErrorCatalogue.Classify` is a closed allowlist over nine tokens returning either a fixed token or `"unclassified"`, with no path that echoes any substring of its input, and `Summary` is a fixed `$case`. So even a `Type = 3` row carrying message content could not leak — which matters, because `classifiedThroughSeverity: "warning"` rests on a measured population of **3 warning rows**, and n=3 is not evidence about what warnings can contain. The allowlist is what carries it; the type bound is defence in depth. Both are worth keeping.
- **Injection surface is clean.** `width` is interpolated into the trend SQL but comes from `BucketWidth`'s fixed `$case` and can only be 10/13/16; `host` is bound. The `(production)` sentinel correctly becomes `ConfigName IS NULL` rather than `= ''` — and the measurement behind that (`= ''` → 0 rows, `IS NULL` → 91) is the kind of thing I would not have believed without it being stated, so thank you for measuring it.
- **`StampAt` borrows days correctly** on negative seconds, so a 720-bucket day trend does not produce malformed stamps.
- **`IsApplication` always returns 0 or 1**, never `""`, so the `%Set(..., "boolean")` calls cannot silently emit `false` for a host that is really unmeasurable. The `null`-for-`(production)` inversion is right and I would have argued for it independently.
- **CHANGELOG entry present and substantive.** §4's other requirement is met.

The four JSON boolean-type fixes are the most valuable thing in here and the least likely to have been found by reading code. `set out.found = 1` emitting a number while the contract's samples showed `true`, surfacing as *"7 hosts"* against 3 — that is a good catch and the convention paragraph in §2 is the right place for it.

---

### Comment 1 — this PR documents a cross-contract inconsistency and then leaves it in place

The CHANGELOG entry says `investigation-api.md`'s `evidence[].tool` "has been carrying `functions.GetEventLogSummary` all along", and `mcp-tools.md` §1 gains a subsection saying the snake_case names have never been callable. But `contracts/investigation-api.md` was not touched, and it still shows snake_case in the places a consumer copies from:

- `"tool": "get_pool_size"` — lines 332, 752, 830
- `"tool": "get_host_status"` — line 776
- `"toolCalls": ["get_host_status", "get_queue_depth", "get_pool_size"]` — lines 532, 796

So the two contracts now disagree, and the one a consumer implements `evidence[]` from is the one that is wrong. `mcp-tools.md` §1's fix is only load-bearing if it reaches the file the consumer reads.

I checked the dashboard side before raising this: `mvp2Guards.ts:86` and `:203` parse `tool` with `nullableStr`, and `diagnostics.toolCalls` is a `number`, so **nothing in `apps/dashboard/**` breaks** and no UI change is needed. This is purely a contract-consistency fix — one paragraph in `investigation-api.md` §Q8 or beside the samples, saying the field is free-form provenance and may arrive provider-prefixed.

### Comment 2 — a failed guard query degrades into exactly what the guard exists to prevent

`EventLog.cls:422-427`:

```objectscript
if (crs.%SQLCODE >= 0) && crs.%Next() && (crs.%Get("N") = 0) {
    set out.error = "no event log rows exist for that host under any window -- ..."
    quit out
}
```

When `crs.%SQLCODE < 0` the check falls through silently and the method returns a full set of zero-filled buckets. In this class a `0` is a **measured** zero — the contract is explicit that this family inverts §2.1's sign — so an unknown or misspelled host comes back indistinguishable from a real host that logged nothing. That is the failure the guard was added to prevent, and the comment above it names #58 as the precedent.

The other two query sites in this class both set `out.reason` on `%SQLCODE < 0`; this one is the only place a SQL failure is swallowed. Suggest matching them:

```objectscript
if crs.%SQLCODE < 0 {
    set out.reason = "host existence unverified (SQLCODE " _ crs.%SQLCODE _ ")"
} elseif crs.%Next() && (crs.%Get("N") = 0) {
    ...
}
```

Low likelihood, but the cost when it fires is a confidently wrong answer rather than a missing one.

---

### One process note, not about this change

This is not a criticism of the work in this PR, which is careful — it is about the gate around it. `main` has **no branch protection**: `gh api repos/InterSystems-Israel/ProductionGuardian/branches/main/protection` returns `404`. So §4's stated safeguard — *"GitHub will not let an author approve their own PR, which is the point"* — is not actually in force, and the last 13 PRs all merged with zero reviews, four of them touching `contracts/` (#148, #150, #151, #153). CODEOWNERS itself is fine (`codeowners/errors` → `[]`), so the requests were being created and merged past.

Enabling required-review + CODEOWNERS on `main` is a settings change and would restore the property §4 already claims the project has.

Separately: PR #113 amended root `CLAUDE.md` §3 and §4 on the premise that **"Dev C left the project"** on 2026-08-20, and moved `apps/dashboard/**` and `docs/demo/**` to Dev B. I was on vacation. Four files now disagree — `CLAUDE.md` §3 (Dev B), `CONTRIBUTING.md` line 92 (Dev C), `.github/CODEOWNERS` (`@tanifgit`, the only one GitHub enforces), and this contract's own header (still "Dev C (approval UI, indirectly)" and "Dev B (inherited `iris/**` from Dev A)", which #113 called reversed). Happy to take the reconciling PR; it needs the roster settled first.

PR #154, opened 2026-08-27
feat(iris): Reset all now clears the per-host error count, because settings never could

Ari-Glikman:
## What the error count actually is

Not a metric and not a setting:

```sql
SELECT %EXACT(TargetConfigName), COUNT(*) FROM Ens.MessageHeader
WHERE Status = 8 GROUP BY %EXACT(TargetConfigName)
```

`REST/HostStatusDispatcher.ErroredByHost()` computes it, because
`iris_interop_messages_errored` is emitted once per production with no `host` label. So it is a
property of the **message store** — which is why it survived a container restart, a production
restart and every `Reset()` before this one. Nothing `Reset()` restores can move it.

## What changed

`Triggers.Reset()` purges the errored messages, **only when there are any**. The gate is the point,
not an optimisation:

- a reset after a scenario that errored nothing must not delete the demo's message history, which is
  what a presenter opens the Message Viewer to show;
- it is the only step in `Reset()` whose cost scales with uptime, so the usual reset pays one
  `COUNT(*)` and nothing else (measured: 0.07s).

### Why purge by date rather than by status

`Ens.Purge` has **no status filter** — checked, including `pExtendedOptions`, which offers
`StartDateTime`, `DoNotDeleteEndDateTime`, `LimitToConfigItems`, `WQCategory` and nothing about
status. `PurgeOneSession()` *is* scoped and was rejected: it is a bare
`DELETE FROM Ens.MessageHeader WHERE SessionId = ?` with no integrity check that never touches
bodies, so it would orphan every body it left behind and would happily delete a session still in
flight.

So it purges by date and lets `pKeepIntegrity` protect what is still moving, which it does exactly.
Read from `Ens.Util.MessagePurge.generateDeleteSelectSQL`, a header is eligible only when no header
in its session has a status outside `{Completed, Aborted, Error, Discarded}`. **Error is inside that
set**, so an errored message never blocks its own removal, while one still `Created` or `Queued`
keeps its whole session. Residue is reported rather than left as a count that quietly did not reach 0.

`PURGEDAYSTOKEEP = 0` keeps nothing **including today** — `GetDoNotDeleteDate(0)` returns
*tomorrow's* local midnight. Named because "0 days" reads as "today survives".

The status code is `$$$eMessageStatusError` via `Include Ensemble`, not a second copy of the literal
`8` that `HostStatusDispatcher` derived from the DISPLAYLIST.

### Engine: reset timeout 60s → 120s

Reset is now the one trigger call whose duration depends on uptime (~4,900 headers/s; 31.5s for
153,144 headers). At 60s a long rehearsal day would abort a purge that was working, and the retry a
presenter then makes starts it from the beginning. 120s stays under nginx's existing 180s
`proxy_read_timeout` on `/api/demo/`, so the engine remains the component that reports a real hang.

### `elevated_error_rate` needed no change

`#errorsPerMinute` already returns `null` when the counter goes backwards, and it stores the new
prior *before* that guard — so a purge costs exactly one silent poll and the rate resumes from the
new floor. Confirmed in the engine log: no negative rate, no `NaN`.

## Verified live, through the dashboard's own route

`POST /api/demo/reset` via nginx → engine → IRIS:

```
armed closed_port  ->  Cloud API errored = 21
reset              ->  "purged 2550 message headers and 2548 bodies in 0.3s"
                       "host error count 21 -> 0"
proxy              ->  errored = 0 on Cloud API, EMR Source, Lab Router
SQL                ->  COUNT(*) WHERE Status=8  =  0
history            ->  6,060 headers preserved
second reset       ->  "no errored messages in the store; message history left alone" (0.07s)
findings           ->  elevated_error_rate gone; survivors are true readings of the drained backlog
```

Engine: `typecheck` clean, **375 tests / 74 suites / 0 fail** with the repo root mounted.

## Docs

The cue sheet's event-log trap is **sharpened, not removed**: `Ens.Util.Log` is a different store
and still needs its own purge (step 5), so a clean host tile does not mean clean evidence. That
distinction is now the easiest thing to talk yourself out of, so it says so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,12 @@ Armed state is read from IRIS on each poll, so driving the terminal at the same 
 described above, with the same one caveat: a `system_alert` finding outlives it, because the alert
 sits in the metrics proxy's in-memory buffer where IRIS cannot reach.
 
+It is also the one control that is not instant. When a scenario has left errored messages behind,
+Reset purges the message store — the per-host error count on a host tile is a `COUNT(*)` over those
+rows, so nothing restorable clears it — and that scales with how long the stack has been up
+(measured: 0.3s for 2,550 headers, 31.5s for 153,144). A reset with no errors to clear skips the
+purge entirely and returns in well under a second, which is the usual case.
+
 ### Before a rehearsal: three things that will bite
 
 - **`AGENT_MODE` defaults to `mock`.** Rebuild or restart the engine without `PG_AGENT_MODE=live`
--- a/docs/demo/cue-sheet.md
+++ b/docs/demo/cue-sheet.md
@@ -81,9 +81,12 @@ line at **column 0** — indent it and PowerShell does not see the terminator.
    it reads as correct and demonstrates nothing about the agent. Check `source: "agent"` and a
    non-zero `toolCalls` in any investigation — those two cannot be faked. This is the standing
    pre-demo check and it has caught a real regression.
-2. **`Reset()` cannot clear the event log.** It restores settings, but a previous scenario's errors
-   stay in the 60-minute window and a later diagnosis reads them as evidence. That is exactly how a
-   pool bottleneck got diagnosed as a connectivity failure. **Always purge** (step 5).
+2. **`Reset()` cannot clear the event log** — and note it is now the *only* error store it cannot
+   clear, which is why this trap is easy to talk yourself out of. Reset purges the MESSAGE store, so
+   the error count on a host tile does return to 0. The EVENT LOG is a different table: a previous
+   scenario's errors stay in the 60-minute window and a later diagnosis reads them as evidence. That
+   is exactly how a pool bottleneck got diagnosed as a connectivity failure. So a clean tile does not
+   mean clean evidence — **always purge** (step 5).
 3. **Restarting the `iris` container disarms triggers.** Recompiling `Production.cls` resets its item
    settings, so `FilePath` and `PoolSize` revert. Re-arm after any restart; the rail shows live state.
 4. **`Cloud API` may be left at pool 4** from a previous run of the Smart Resolve beat. `Reset()`
--- a/iris/labdemo/README.md
+++ b/iris/labdemo/README.md
@@ -383,6 +383,16 @@ and timeouts, and re-enables every host. The old instruction here was to remove 
 from `PatientDemographicsOperation` and recompile — the delay is now a global the
 operation reads at runtime, so there is nothing to leave behind in source.
 
+It also **purges the message store, but only when errored messages exist.** The per-host error
+count is `COUNT(*) FROM Ens.MessageHeader WHERE Status = 8` grouped by target host — a property of
+the store rather than of the configuration, which is why it used to survive a reset, a production
+restart and a container restart alike. Nothing restorable moves it, so the rows are deleted.
+The gate matters: a reset after a scenario that errored nothing leaves message history alone, so
+the Message Viewer still has traces to show, and only the run that actually errored pays the purge
+(measured: 0.3s for 2,550 headers, 31.5s for 153,144). `pKeepIntegrity` keeps anything still in
+flight, so a handful of errored rows can survive into the next reset; `Reset()` says so when it
+happens rather than reporting a count that quietly did not reach 0.
+
 ---
 
 ## Verify metrics
--- a/iris/labdemo/Triggers.cls
+++ b/iris/labdemo/Triggers.cls
@@ -1,3 +1,5 @@
+Include Ensemble
+
 /// Trigger toggles: induce any of the eight Health Scan finding types on demand.
 ///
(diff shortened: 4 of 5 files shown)


tanifgit (commented):
Reviewed as the `docs/demo/` owner (`.github/CODEOWNERS`), since this changes the cue sheet's trap list.

**No blocking concerns — I'd approve this.** The diagnosis is the valuable part: that the per-host error count is `COUNT(*)` over `Ens.MessageHeader` rather than a metric or a setting, and therefore that *nothing `Reset()` restores could ever have moved it*, explains a symptom that survived a container restart and would otherwise have been chased somewhere else entirely.

What I checked rather than took on trust:

- **The nginx headroom claim.** `apps/dashboard/nginx.conf.template:87` is `proxy_read_timeout 180s` on `location /api/demo/`, against `35s` on the general `/api/` at line 141. So 120s does sit inside the demo route's own budget, and the engine stays the component that reports a genuine hang. Accurate.
- **`Ens.Purge` really has no status filter**, so purging by date with `pKeepIntegrity` is the available shape, and rejecting `PurgeOneSession()` for orphaning bodies is right.
- **`ErroredCount()` returning `-1` rather than `0`** on an unreadable query, and `PurgeErrored()` branching on it separately, is the #33 rule applied correctly — a failed count printing *"no errored messages"* would have been a confident claim about the one number this path exists to change. The `%SQLCODE` check placed *after* the fetch matches `HostStatusDispatcher.ErroredByHost()`.
- **`remaining` is `-1`-safe.** The count line prints `unknown`, and `if remaining > 0` is false for `-1`, so the residue paragraph does not fire on an unmeasurable read.
- **375 pass / 74 suites / 0 fail, typecheck clean**, reproduced locally on the branch head. All five CI checks green.

One comment, and it is about my file rather than yours.

### The purge deletes far more than the errored messages, and trap #2 is now the place that has to say so

`PurgeErrored()` is named for its *purpose*, but `PurgeMessagesByDate(0, .deleted, 1, 1)` has no status filter — as the docstring itself establishes — so what it actually deletes is **every header eligible by date**, errored or not. Your own verification run is the clearest statement of the ratio:

```
armed closed_port  ->  Cloud API errored = 21
reset              ->  "purged 2550 message headers and 2548 bodies in 0.3s"
```

2,550 headers removed to clear 21 errored ones. The ~2,529 others were completed traffic.

The gate reads as protecting message history, and in the zero-error case it does exactly that. But in the case it was built for — a scenario that *did* error — the demo's message history goes with it, and the amount that survives is incidental to how many sessions happened to be in flight when the purge ran (6,780 in one of your runs, 6,060 in another), not something a presenter can count on.

That collides with the trap this PR rewrites. Trap #2 now reads:

> Reset purges the MESSAGE store, so the error count on a host tile does return to 0.

which is true and is the reassuring half. The half a presenter needs before pressing the button is that **after an error scenario, Reset also empties the Message Viewer** — and the Message Viewer is exactly what trap #2's neighbours assume is available to show. Suggest appending a sentence there, or in the `iris/labdemo/README.md` paragraph:

> Reset clears the tile by deleting message rows, and it cannot delete only the errored ones — `Ens.Purge` has no status filter. So a reset that follows an error scenario also removes most of the completed history: 2,550 headers went to clear 21 errored. If you want the Message Viewer populated for a later beat, show it *before* you reset.

I'm happy to make that edit myself as a follow-up rather than holding this PR for it — say which you prefer.

### Minor, not worth acting on now

The 120s timeout narrows the abort-and-retry window rather than closing it: at the measured ~4,900 headers/s it covers ~588k headers, so a long enough rehearsal day still lands in the state you describe, where the retry restarts the purge from the beginning. A batched purge with progress would be the real fix. 120s is the right interim and I would not spend the time here.

### One note on process, not on this change

This PR and #158 both write to `docs/demo/**`, which `.github/CODEOWNERS` assigns to `@tanifgit` while root `CLAUDE.md` §3 assigns it to Dev B — four files disagree on that (see the process note on #153). Reviewing rather than objecting, because the changes are good and the review request is exactly the mechanism that should catch a cross-area write. Flagging only so the ownership PR does not get forgotten: the enforced file and the documented one need to agree before someone trusts the wrong one.

tanifgit (approved):
Approving. Full review is in my earlier comment on this PR: the diagnosis that the per-host error count is `COUNT(*)` over `Ens.MessageHeader` — and therefore that nothing `Reset()` restores could ever have moved it — is the valuable part, and I verified the nginx 180s headroom, the `-1`-vs-`0` handling, and 375/0 tests locally.

The one comment stands as a follow-up rather than a change request: `PurgeErrored()` deletes every header eligible by date, not only the errored ones (2,550 removed to clear 21 in your own run), so the cue sheet's trap list needs to say that a reset after an error scenario also empties the Message Viewer. I'm writing that line myself — no change needed here.

**Merge-order note:** #159 contains #158's commit, so the clean order for today's four is #154 → #158 → #159. This one is independent and can go first.

Issue #156, opened 2026-08-27
iris: GetEventLogTrend's unknown-host guard falls through on SQL failure, returning zero buckets for a host that may not exist


tanifgit:
Raised reviewing #153 (post-merge — that PR requested my review as a `contracts/` change and was merged before it landed).

`GetEventLogTrend`'s unknown-host guard skips itself when its own query fails, and the fallthrough produces the exact payload the guard exists to prevent.

`iris/labdemo/Tools/EventLog.cls:422-427`:

```objectscript
set crs = ##class(%SQL.Statement).%ExecDirect(,
    "SELECT COUNT(*) N FROM Ens_Util.Log WHERE %EXACT(ConfigName) = ?", host)
if (crs.%SQLCODE >= 0) && crs.%Next() && (crs.%Get("N") = 0) {
    set out.error = "no event log rows exist for that host under any window -- "
        _ "call GetEventLogSummary to see which hosts have logged"
    quit out
}
```

When `crs.%SQLCODE < 0` the whole condition is false, execution continues, and the method returns a complete set of zero-filled buckets.

### Why that is worse here than in most tools

This class deliberately inverts §2.1's sign — `contracts/mcp-tools.md` §3.11 states it outright: *"For a log count, `0` means the query ran and the window was covered and nothing was logged."* So zero-filled buckets are a positive assertion that the host was quiet. A misspelled or invented host name therefore comes back indistinguishable from a real host that logged nothing, which is the failure mode the guard's own comment cites #58 for:

> A HOST NAME THE MODEL INVENTED MUST NOT COME BACK AS ZEROS. […] #58 is the same shape (a host that had never run, published as "active now"), and the fix is the same: say the host is unknown instead of describing it.

The consumer is a language model, which makes a confidently wrong answer more expensive than a missing one.

### Inconsistent with the rest of the class

Both other query sites handle `%SQLCODE < 0` explicitly and set `out.reason` — `:210-213` and `:256-257`, plus `Retention()` at `:519`. This is the only place in the class where a SQL failure is swallowed.

### Suggested fix

```objectscript
if crs.%SQLCODE < 0 {
    set out.reason = "host existence unverified (SQLCODE " _ crs.%SQLCODE _ ")"
} elseif crs.%Next() && (crs.%Get("N") = 0) {
    set out.error = "no event log rows exist for that host under any window -- "
        _ "call GetEventLogSummary to see which hosts have logged"
    quit out
}
```

[...]

Issue #155, opened 2026-08-27
contracts: investigation-api.md still shows snake_case tool names that mcp-tools.md §1 says were never callable


tanifgit:
Raised reviewing #153 (post-merge — that PR requested my review as a `contracts/` change and was merged before it landed).

#153 added a subsection to `contracts/mcp-tools.md` §1 recording that the snake_case names in the tool catalogue **have never been callable**: `%AI.Tool` derives a tool's runtime name from the ClassMethod name, so the real names are `GetEventLogSummary`, `CompareHostActivity`, `GetHostStatus` and so on. Its `contracts/CHANGELOG.md` entry goes further and says `investigation-api.md`'s `evidence[].tool` "has been carrying `functions.GetEventLogSummary` all along".

`contracts/investigation-api.md` was not touched, so it still shows the snake_case form everywhere a consumer copies from:

| Line | Content |
|---|---|
| 332, 752, 830 | `"tool": "get_pool_size"` |
| 776 | `"tool": "get_host_status"` |
| 532, 796 | `"toolCalls": ["get_host_status", "get_queue_depth", "get_pool_size"]` |

The two contracts now disagree, and the one a consumer implements `evidence[]` against is the one that is wrong. §1's correction is only load-bearing if it reaches the file the consumer actually reads.

### Scope — no code change needed

Checked before filing: the dashboard treats the field as opaque. `apps/dashboard/src/api/mvp2Guards.ts:86` and `:203` parse `tool` with `nullableStr`, and `diagnostics.toolCalls` is a `number` (`:181`, `:248`), so nothing in `apps/dashboard/**` validates against a name list and no UI change is required. This is purely contract consistency.

### Suggested fix

A note beside the `evidence[]` samples and in §Q8 saying `evidence[].tool` is **free-form provenance, not an enum**: it carries the runtime ClassMethod name, may arrive with a provider prefix (`functions.GetEventLogSummary`), and must not be validated against `mcp-tools.md` §1's catalogue column. Updating the sample values themselves to the PascalCase form would also stop the samples teaching the wrong thing.

Owner: `contracts/` is PR-gated (root `CLAUDE.md` §4), so this wants its own PR with a CHANGELOG entry.

PR #157, opened 2026-08-27
fix(dashboard): demo-mode host graphs never drew, behind a diff git rendered as binary

tanifgit:
Demo-mode host graphs have never drawn. `recordSeries` stored each point under one key and `getHostSeries` looked it up under another, so every `mockSeries.get` missed.

```
recordSeries   →  mockSeries.set(`${host.host}\u0000${metric}`, …)   // NUL separator
getHostSeries  →  mockSeries.get(`${host} ${metric}`)                // space separator
```

The panel therefore received `points: []` for all three metrics on every host, `recorded` was always `false`, and `polledAt` was always `null` — which is the "collecting" state, permanently, on the one feature whose entire purpose is to draw a series (#139).

**Not a regression.** `mockSeries` does not exist in #140's parent commit, so both halves arrived together and the two sides never agreed at any point. The feature has simply never worked in demo mode.

## Why it survived two reviews

`recordSeries` used a **literal U+0000 control character** rather than the `\u0000` escape. Git classifies any file containing a NUL as binary, so every diff of `mockClient.ts` since #140 has rendered like this:

```
$ git show --stat 9a44346 -- apps/dashboard/src/api/mockClient.ts
 apps/dashboard/src/api/mockClient.ts | Bin 21868 -> 27238 bytes
 1 file changed, 0 insertions(+), 0 deletions(-)
```

#140 introduced it (`Bin 15885 -> 21868`) and #144 changed a further 5,370 bytes behind it. Neither was reviewable — on GitHub the file shows no diff at all. A one-line disagreement between two adjacent functions is precisely the class of defect a diff catches and prose does not, which is why this is filed as two problems rather than one.

It is the only tracked text file in the repo with a NUL — checked across every tracked file rather than assumed.

## Two commits, deliberately

| | |
|---|---|
| `4fd83d2` | Raw NUL → the `\u0000` escape. Byte-identical at runtime, no behaviour change. Makes the file text again. |
| `49f659a` | One shared `seriesKey(host, metric)` that both the writer and the reader call. |

**The first commit's diff is 616 lines and that is correct, not noise.** `core.autocrlf` is `true` and every sibling source file is stored with LF — `mvp2Guards.ts` has 424 bare LF in its blob — but autocrlf deliberately does not convert files it believes are binary. So this one went into the repo with 616 CRLF endings. Removing the NUL lets normalisation run, once. Reviewing `49f659a` alone is the intended path; it is 3 changed lines plus a comment.

## Why a shared builder rather than a corrected literal

Fixing the separator in both places would leave the same failure mode one edit away. This is the argument `Tools.ErrorCatalogue` and `Tools.HostRoster` were extracted on in #153 last week, and it applies unchanged: a rule held in two places goes stale in whichever copy nobody edits.

**The separator stays U+0000 rather than becoming a space**, because a space is legal in a host name — `Cloud API`, `EMR Source` and `Lab Router` all contain one — so `"A B" + " " + "c"` and `"A" + " " + "B c"` compose the same key. A control character cannot collide. It is now written as an escape, which is byte-identical at runtime and keeps the file diffable.

## Verification

```
$ npx tsc --noEmit                 (clean)
$ npm run build                    (clean — architecture validator ok, 8 labels / 12 paths)
$ npm run validate:fixtures        (clean)
```

And the keys checked directly rather than reasoned about:

```
old writer key : "Cloud API\u0000queued"
old reader key : "Cloud API queued"
old match      : MISS  <-- the bug
new both sides : "Cloud API\u0000queued"   HIT
```

**NOT verified in a browser, and this is the gap worth knowing.** `apps/dashboard/` has no test runner and I did not bring up the compose stack, so "the graph now draws" rests on the two keys agreeing, not on a rendered chart. The remaining uncertainty is whether anything *downstream* of `points[]` also assumed the empty array — `HostSeriesView` consumers have only ever been exercised against `[]` in demo mode. Worth opening demo mode and clicking a host at the next rehearsal; `apps/dashboard/CLAUDE.md` §6.1 applies (`docker compose up -d --build dashboard`, since the bundle is compiled into the image).

## Scope

`apps/dashboard/src/api/mockClient.ts` only. No contract change, no `src/` outside `api/`, no new dependency, no host name or count added. Demo-mode path only — `liveClient` builds no such key and is untouched.

Raised while reviewing #153, which is also where #155 and #156 came from.

---

**Reviewer note:** `.github/CODEOWNERS` assigns `/apps/dashboard/` to `@tanifgit`, while root `CLAUDE.md` §3 assigns it to Dev B and `CONTRIBUTING.md` line 92 still says Dev C. Those four files need reconciling (see the process note on #153) — I have not touched them here, because that is its own PR and it needs the roster question settled first.

### DIFF
--- a/apps/dashboard/src/api/mockClient.ts
+++ b/apps/dashboard/src/api/mockClient.ts
@@ -233,7 +233,7 @@ export function createMockClient(pinnedScenarioId?: string): MockClient {
       for (const { metric, of } of MOCK_SERIES_METRICS) {
         const value = of(host);
         if (value === null || !Number.isFinite(value)) continue;
-        const key = `${host.host} ${metric}`;
+        const key = `${host.host}\u0000${metric}`;
         const points = mockSeries.get(key) ?? [];
         // Same second as the previous point means two reads inside one tick; replace rather than
         // append, so the graph does not gain a vertical pair at one x position.
--- a/apps/dashboard/src/api/mockClient.ts
+++ b/apps/dashboard/src/api/mockClient.ts
@@ -220,6 +220,17 @@ export function createMockClient(pinnedScenarioId?: string): MockClient {
     { metric: 'messagesPerSec', unit: 'per_second', of: (h) => h.messagesPerSec },
   ];
 
+  /* ONE KEY BUILDER, because the writer and the reader had drifted apart and nothing caught it:
+     `recordSeries` composed the key with a `\u0000` separator while `getHostSeries` used a space, so
+     every lookup missed and demo mode served an empty series for every host and metric. Same argument
+     the engine's shared helpers make -- a rule held in two places goes stale in one of them.
+
+     THE SEPARATOR IS AN ESCAPE, NOT A LITERAL CONTROL CHARACTER. A raw NUL byte here made git classify
+     this file as binary, which suppressed its diff in #140 and #144 -- which is why the mismatch above
+     shipped unreviewed. It stays U+0000 rather than a space because no host name or metric can contain
+     one, so two different pairs can never compose the same key. */
+  const seriesKey = (host: string, metric: string): string => `${host}\u0000${metric}`;
+
   /**
    * Append this poll's values to the demo history.
    *
@@ -233,7 +244,7 @@ export function createMockClient(pinnedScenarioId?: string): MockClient {
       for (const { metric, of } of MOCK_SERIES_METRICS) {
         const value = of(host);
         if (value === null || !Number.isFinite(value)) continue;
-        const key = `${host.host}\u0000${metric}`;
+        const key = seriesKey(host.host, metric);
         const points = mockSeries.get(key) ?? [];
         // Same second as the previous point means two reads inside one tick; replace rather than
         // append, so the graph does not gain a vertical pair at one x position.
@@ -342,7 +353,7 @@ export function createMockClient(pinnedScenarioId?: string): MockClient {
         unit,
         /* A COPY, so a later poll appending to the stored array cannot mutate a series React is
            already rendering. The engine's `recent()` copies for the same reason. */
-        points: [...(mockSeries.get(`${host} ${metric}`) ?? [])],
+        points: [...(mockSeries.get(seriesKey(host, metric)) ?? [])],
       }));
       const recorded = series.some((s) => s.points.length > 0);
       return {


Ari-Glikman (approved):
Approving as the `apps/dashboard/**` owner. Reviewed `49f659a` on its own as you suggested, and **closed the one gap you flagged** — the browser check.

## The gap: verified at runtime, not just by key inspection

You wrote that "the graph now draws" rested on the two keys agreeing rather than on rendered output. It doesn't need a browser to settle: the mock client is plain TS, so I bundled it with the `esbuild` already in `node_modules` and drove the real object — `getHosts()` five times (which is where `recordSeries` runs, line 281) then `getHostSeries()` back for each host. Same harness, both versions of the file:

```
################ WITH #157 ################
EMR Source   known=true  polledAt=set   queued=2 avgProcessingTime=2 messagesPerSec=2
Lab Router   known=true  polledAt=set   queued=2 avgProcessingTime=2 messagesPerSec=2
Cloud API    known=true  polledAt=set   queued=2 avgProcessingTime=2 messagesPerSec=2
PASS

################ WITHOUT #157 (main's mockClient) ################
EMR Source   known=true  polledAt=null  queued=0 avgProcessingTime=0 messagesPerSec=0
Lab Router   known=true  polledAt=null  queued=0 avgProcessingTime=0 messagesPerSec=0
Cloud API    known=true  polledAt=null  queued=0 avgProcessingTime=0 messagesPerSec=0
FAIL (3 host(s))
```

**Every claim in your body reproduces exactly** — nine empty series before, all three metrics populated on all three hosts after, and `polledAt` flipping `null` → set, which is the panel's permanent "collecting" state resolving. Running it against `main` as well is what makes it evidence rather than a green light; the harness fails when the fix is absent.

One correction to the body, and it changes nothing: `recorded` is a local in `getHostSeries`, not a field on `HostSeriesView`, so "`recorded` was always `false`" isn't directly observable — `polledAt: null` is its observable form, and that is what the run above shows. The reasoning was right.

The harness was deleted, not committed — `apps/dashboard/` still has no test runner and one throwaway file is not the way to introduce one.

## What else I checked rather than took on trust

- **The NUL is gone and the file is text again.** `git diff --stat origin/main...HEAD` now reports `627 insertions(+), 616 deletions(-)` where it used to say `Bin 21868 -> 27238 bytes`. That is the whole point of splitting `4fd83d2` out, and it is self-evidencing.
- **It was the only one.** Swept every tracked file for `\x00` — none remain.
- **`tsc --noEmit` clean, `npm run build` clean** (82 modules, single-file inline, 2,575 kB), reproduced on node v22.23.2.
- **The shared-builder argument is the right one.** Correcting the literal in both places leaves the same failure one edit away, and `seriesKey` makes the two sides incapable of disagreeing. Same reasoning as `Tools.ErrorCatalogue`/`Tools.HostRoster` in #153, and as `BaselineStore.baseline()` being the single accessor in #159 — a rule reachable from two call sites gets held in one place.
- **Keeping U+0000 over a space is correct and the justification is exact.** All three host names contain a space, so `"A B" + " " + "c"` and `"A" + " " + "B c"` collide; a control character cannot. The engine's own `SEP` in `src/baseline/window.ts` uses the identical technique for the identical reason, also written as an escape.

## The finding under the finding

The valuable part of this PR isn't the three lines — it's that **a raw control character in a source file silently disabled review for two PRs**. #140 introduced the mismatch and #144 changed 5,370 bytes on top of it, both rendering as `Bin`, and the defect that survived was a one-line disagreement between two adjacent functions: precisely what a diff catches and prose does not. Worth a `.gitattributes` guard or a CI sweep so the next one can't hide the same way — I'd rather that than rely on noticing. Filing it separately; it shouldn't hold this.

Merging. Order today was #154 → #158 → #159 → this, per your containment note on #159.

PR #158, opened 2026-08-27
fix(detection-engine): a slow_processing finding that Smart Resolve could not clear

Ari-Glikman:
`Cloud API`'s `slow_processing` floor was **0.3s**. The `pool_bottleneck` scenario throttles that host's downstream to ~1s per call, so `avgProcessingTime` reads **1.01s** — and it reads 1.01s **both** at pool 1 behind a queue of 123 **and** at pool 4 with the queue drained to zero. Measured, 13 samples across two sessions.

**Pool size changes concurrency, not per-message latency.** Four workers make four 1s waits overlap, so throughput quadruples while every individual message still takes ~1s.

So the fix worked and the finding stayed. A `critical` sat on the board after Smart Resolve had emptied the queue, which reads to an audience as the remediation not working.

## Why suppression is the only shape available

The value before the fix and the value after it are the *same measurement*, so no floor and no multiplier can make this rule fire during the fault and clear after it. Raising the floor to **1.5s** is right for this host: `Cloud API` is an outbound operation whose processing time is a remote dependency's latency, ~1s per remote call is ordinary, and the two findings that do describe the fault — `queue_buildup` and `growing_queue_wait` — both clear correctly.

**The baseline is deliberately not touched.** Raising `referenceBaselines` to 0.34 would also have silenced it, and would have put an invented normal into every finding message and into the evidence the agent reads — those values are captured from `contracts/samples/`. The floor answers *"is this worth reporting"*, which is the question that actually changed.

## Verified end to end on the live stack (2026-08-27)

| Stage | Result |
|---|---|
| Fault | 2 criticals — `queue_buildup` + `growing_queue_wait`, no `slow_processing` |
| Apply | pool 1 → 4; `queue_buildup` clears at t+72s; queue empty at t+90s |
| Board | last finding ages out at t+202s, then **0 findings held for a further 200s** — throttle still armed, Cloud API still reporting `proc 1.01` |

Two tests pin the boundary against the **committed** file rather than `DEFAULT_CONFIG`, and the first was proved to fail with the floor put back to 0.3. **377 pass, 0 fail; typecheck clean.**

## Also here, because they would otherwise be wrong

- `Triggers.SlowProcessing()` default **1000ms → 3000ms**. At 1000 it measured 840ms and is now below the floor — armed and inducing nothing, the one failure this class must not have. It now says so when passed under 1500.
- The cue sheet's Act 1.3 lists **two** findings, not three, and Act 2.3 no longer collapses *"the queue is empty"* (~90s) into *"the board is clean"* (~200s). Act 2 is ~6 min, not ~4.


tanifgit (commented):
Reviewed as the `docs/demo/` owner (`.github/CODEOWNERS`) and as the person the cue-sheet change lands on.

**No blocking concerns — I'd approve this.** The reasoning is right and I checked the load-bearing claims rather than taking the body's word:

- **"Pool size changes concurrency, not per-message latency"** is the correct diagnosis, and it does mean no number in `thresholds.json` can fix this. Confirmed the override is scoped to `hostOverrides["Cloud API"].slow_processing` only, so no other host's floor moved.
- **"Anything at or above 1.5s still fires"** — verified. `src/detect/rules/index.ts:225` is `if (current < rule.absoluteFloorSeconds) return null`, so the boundary is inclusive on the firing side, exactly as the `_comment` claims.
- **The new test is a real guard, not a tautology.** I put the floor back to `0.3` and ran it: `stays silent at the 1.01s the pool_bottleneck scenario produces` fails. Reading the committed `thresholds.json` through `validateConfig` rather than asserting against `DEFAULT_CONFIG` is the right call — that is the file that decides what an audience sees.
- **377 pass / 75 suites / 0 fail, typecheck clean**, reproduced locally on the branch head.
- **Not touching `referenceBaselines`** is the right instinct and worth the paragraph you gave it. An invented 0.34 would have propagated into every finding message and into the agent's evidence.

Two comments, neither blocking.

### 1. The new `< 1500` warning has a gap, and it is the same failure it was added to close

`Triggers.SlowProcessing()` warns only when `pMilliseconds < 1500`, but the quantity the floor tests is the *measured* `avgProcessingTime`, not the armed delay — and this PR's own `iris/labdemo/README.md` edit records those differing:

> `SlowProcessing(1000)` measured 840 ms

That is measured ≈ 0.84× armed. If that ratio roughly holds, an armed value anywhere in **[1500, ~1785]** measures below the 1.5s floor, so it induces nothing — and prints no note, because `1500 < 1500` is false. Someone experimenting with `SlowProcessing(1500)` gets `Gate is the 1.5s Cloud API floor` and silence, which is precisely the "armed and induces nothing" state the docstring says is *"the one failure this class must not have"*.

Cheapest fix is to move the boundary clear of the uncertainty rather than to model it:

```objectscript
if pMilliseconds < 2000 {
    write "  NOTE  under 2000ms this host may report nothing -- the measured average runs", !
    write "        below the armed delay (1000ms measured 840ms) and its floor is 1.5s.", !
    write "        Pass 3000 to see the finding.", !
}
```

The default of 3000 means nobody hits this by accident, which is why it is a comment and not a blocker.

### 2. Worth one clause: what covers Cloud API between 1.01s and 1.5s

The `_comment` is admirably explicit that the cost is *"Cloud API no longer reports a ~1s outbound call as slow at all"*. The band I'd name specifically is **1.01s to 1.5s** — a downstream degrading to ~1.4s is a ~39% regression on this host and now reports nothing.

`growing_queue_wait` covers it only when a queue actually forms, which needs the inbound rate to exceed the service rate; at 1.4s per call with lighter inbound traffic, no queue builds and nothing fires at all. I think that is still the right trade for this host — its processing time genuinely is a remote dependency's latency — but the residual gap is narrower and more specific than "no longer reports a ~1s call", and stating it is what makes the next person's decision to move this number an informed one.

### On the cue-sheet change specifically

This is the part in my directory and it is better than what it replaces. Splitting *"the queue is empty"* (~90s) from *"the board is clean"* (~200s) fixes a real presenter trap — the old single line `t+103s queued=0 <- findings clear on their own` sets someone up to say "cleared" with a finding still on screen. The Act 2 total moving 4 to 6 min and the header total 8 to 10 min follow from the measurement rather than from the intent, which is the right direction for a document a presenter budgets a slot from.

Two small things:

- The paragraph beginning *"If you are presenting from an older build and see a stuck `slow_processing critical`"* is genuinely useful, and I'd move it out of the Act 2.3 blockquote and up into the **pre-demo trap list** at the top. That list is what someone reads before going on stage; Act 2.3 is what they read during it, by which point knowing they are on an older build is too late.
- `growing_queue_wait "340ms is 17x baseline"` in the drain table is the only finding message quoted verbatim in that block. Worth confirming it is copied from a real board rather than reconstructed, since `apps/dashboard/CLAUDE.md` §2.4 makes the engine's `message` authoritative and a paraphrase there would teach a presenter the wrong wording.

Verified locally: `npm test` 377/0 on the branch, `npx tsc --noEmit` clean, floor-reverted negative check fails as designed, and all five CI checks green.

tanifgit (approved):
Approving. Full review is in my earlier comment: the diagnosis is right, suppression genuinely is the only shape available, the override is correctly scoped to `hostOverrides["Cloud API"]`, and I verified `rules/index.ts:225` makes the 1.5s boundary inclusive on the firing side, reproduced 377/0 locally, and confirmed the new test fails with the floor put back to 0.3.

Two follow-ups, neither blocking:
1. The new `< 1500` warning in `Triggers.SlowProcessing()` has a gap — by this PR's own README figure (`SlowProcessing(1000)` measured 840 ms), armed values in roughly [1500, 1785] induce nothing *and* print no note. Suggest moving the boundary to 2000.
2. Worth naming the residual band specifically: 1.01s–1.5s on Cloud API now reports nothing, and `growing_queue_wait` only covers it when inbound traffic is heavy enough to build a queue.

**Please merge this BEFORE #159**, which contains this commit — see my review there.

Ari-Glikman:
**Follow-up 1 from the review is fixed in `5cbe3f2`, in this PR rather than after it** — your point that it would be lost if this branch got closed as redundant is the reason it belongs here.

You were right about the band and it was slightly wider than a floor-vs-boundary mismatch: the measured delay comes in at **~0.84× the armed value** (`SlowProcessing(1000)` → 840 ms, this PR's own README figure), so the smallest armed value that actually reaches a 1.5 s floor is **~1786 ms**. A boundary at 1500 therefore left `[1500, 1785]` armed, inducing no `slow_processing`, and printing no note — the exact "looks armed and induces nothing" failure the note exists to prevent, reintroduced by putting the boundary *on* the floor instead of above it.

Boundary moved to **2000**, which clears ~1786 with margin and agrees with the 3000 the note already recommends. The note now states the 0.84× measurement, so the next reader can see why 2000 ≠ 1500 rather than reading it as a typo for the floor.

One thing I widened beyond what you asked: the old note said the host "reports nothing", which was too broad. `growing_queue_wait` has a 0.15 s floor and was never affected by this change, so in that band there *is* still a finding — the note now says so. That is your follow-up 2 (the residual 1.01–1.5 s band) surfacing in the operator-facing text.

Verified in the running container rather than by reading it:

```
Compiling class ProductionGuardian.LabDemo.Triggers   ... ERRLOG=0

SlowProcessing(1700)  ->  NOTE  under 2000ms this host reports no slow_processing -- its floor is 1.5s
                                and the induced delay measures ~0.84x what you pass, so 1500 is not
                                enough. Pass 3000 to see the finding.
                                growing_queue_wait still fires once traffic overlaps.
SlowProcessing(3000)  ->  (no note)
```

`^ProductionGuardian.Trigger("SlowMs")` was unset before the check and killed after it, so the live stack is back where it started. CI green 5/5 on the new head, including `iris — compile ObjectScript`.

`iris/labdemo/README.md`'s trigger table said "the method says so when you pass under 1500" and is updated in the same commit — a copied number going stale is #84, and this one would have gone stale in the commit that fixed it.

Merging this before #159 as you asked.

PR #159, opened 2026-08-27
fix(detection-engine): a reset manufactured throughput_drop on an idle production

Ari-Glikman:
Pull up the dashboard after pressing **Reset all** and, with nothing armed and nothing wrong, `throughput_drop warning` sat on all three hosts.

**Two separate mechanisms, both measured.** The first fix alone was verified insufficient on the live stack, which is why both are here.

### 1. A sustained step down spanning the window

`pool_bottleneck` quadruples inbound load — 2.0 msg/sec against LABDEMO's 0.5 — and the reset restores it in one step. But the baseline is a 30-minute trailing mean, so for half an hour afterwards it averaged in the elevated block: **1.19 against a current 0.4, fraction 0.34** under a `baselineFraction` of 0.40. The finding's `detectedAt` was the exact second of the reset.

`BaselineStore.beginRegime(at)`, called from the reset endpoint once the reset reports `ok`, declares pre-reset samples a different regime so baselines re-warm (60s) rather than average across the discontinuity. **A load step-down is not a fault**, and no threshold can tell the two apart from the mean alone — which is why this is a boundary rather than a tuned number.

### 2. A transient burst inside the new regime — what made (1) insufficient

Removing the throttle frees the accumulated backlog, which flushes all at once. From `Ens.MessageHeader`:

```
10:38:38   69 messages
10:38:39  176
10:38:40  157      <- 402 messages in 3 seconds, ~134/sec against an idle 0.5/sec
```

That reading is a **true measurement**, and the arithmetic proves it: IRIS computes the rate as messages-in-interval ÷ interval, and `274 / 5s = 54.8` matched the reported sample exactly. So the burst is recorded, not rejected — it is real work.

What a mean cannot do is survive it. One sample supplied **68% of a 53-sample mean**, giving `1.5283 (= 81/53)` against a true `0.504`, and `Cloud API` — the only host with a backlog — fired again 75s after the reset even with the regime boundary in place. No window length helps; a longer window just holds the spike for longer.

So `messagesPerSec` is now baselined by **median** (`ROBUST_METRICS`).

### Why that is scoped to one metric

The direction of harm, not taste:

| metric | worse direction | inflated baseline ⇒ |
|---|---|---|
| `queued`, `avgProcessingTime`, `avgQueueingTime`, `errorsPerMinute` | higher | rule gets **quieter** — ADR 0002's documented self-inflation, deliberately pinned |
| `messagesPerSec` | **lower** | rule gets **noisier** — it manufactures findings |

`throughput_drop` is the only comparative rule where lower is worse. The median also makes a **real** drop hold *longer*, since a stopped host must fill more than half the window with zeroes before its own baseline decays — where a mean began absorbing the first zero.

`recent()` is deliberately untouched by the regime boundary, so host graphs and Early Warning's slope fit keep their history. Blanking every graph on reset would be a visible regression for a statistics problem.

### Verification

Live at full magnitude, not just in tests: armed 9 minutes to build a ~400-message backlog, reset, then **4 minutes with the 402-msg/3s burst inside the live 30-minute window and zero findings** — where a mean gives ≈1.06 against 0.4, i.e. fraction 0.38, under the gate. `queue_buildup` and `growing_queue_wait` still fire while armed and still clear on their own.

```
tsc --noEmit  clean
# tests 389   # pass 389   # fail 0
```

New regression tests pin both mechanisms, including one that deliberately **reproduces** the un-fixed false positive so it stays visible.

### DIFF
--- a/services/detection-engine/CLAUDE.md
+++ b/services/detection-engine/CLAUDE.md
@@ -159,12 +159,31 @@ Non-negotiable behaviours:
 - **Rules are pure functions** of `(sample, window, config)`. No I/O, no clock reads inside a
   rule — pass time in. This is what makes them testable against fixtures.
 
-### 5.1 Baseline self-inflation — and the one metric now exempt from it
+### 5.1 Baseline self-inflation — and the two exemptions from it
 
 **A reference baseline exempts a host+metric.** `thresholds.json` `referenceBaselines` states a
 normal that does not move; where one exists it wins over the rolling mean, and self-inflation
-cannot occur for that pair. Currently only `Cloud API` / `queued`, set to `0`. Everything else
-still behaves exactly as described below, which is most metrics on most hosts.
+cannot occur for that pair. Currently only `Cloud API` / `queued`, set to `0`.
+
+**`messagesPerSec` is exempt everywhere, because its baseline is a MEDIAN, not a mean.**
+`ROBUST_METRICS` in `baseline/window.ts` carries the full argument and the measurements; the short
+version is that self-inflation is only *safe* when higher is worse. Where higher is worse an
+inflated baseline makes a rule quieter, which is the failure direction you want. `throughput_drop`
+is the only comparative rule where **lower** is worse, so for it an inflated baseline is the
+opposite: it manufactures findings on a healthy production, which MVP §6 names as the top risk.
+
+It was measured doing exactly that (2026-08-27). A reset after `pool_bottleneck` removes the
+throttle and the accumulated backlog flushes in one burst — 402 messages in 3 seconds, ~134/sec
+against an idle 0.5/sec — and a single sample of that burst supplied 68% of a 53-sample mean,
+giving a baseline of 1.53 against a true 0.50 and firing `throughput_drop` on all three hosts with
+nothing armed. No window length fixes that; a longer window only holds the spike for longer.
+
+Note the direction this exemption cuts: a median makes `throughput_drop` **hold a real drop for
+longer**, since a stopped host must fill more than half the window with zeroes before its own
+baseline decays. So this is not the "excluding breaching samples" option listed at the end of this
+section — it does not exclude anything, and the burst is still recorded and still graphed.
+
+Everything else behaves exactly as described below, which is most metrics on most hosts.
 
 Why it was needed, with the arithmetic rather than the anecdote. §5.1 as originally written
 describes a **step**: queue jumps 0 → 486, fires, then clears as the mean climbs. A **ramp** is
--- a/services/detection-engine/src/baseline/window.ts
+++ b/services/detection-engine/src/baseline/window.ts
@@ -38,6 +38,15 @@ export interface Sample {
 class MetricWindow {
   #samples: Sample[] = [];
 
+  /**
+   * Samples before this instant describe a different load regime and are excluded from
+   * `mean()`. See `BaselineStore.beginRegime()` for why this exists.
+   *
+   * -Infinity, not 0: a real epoch-ms comparison against 0 would also work today, but the
+   * intent is "no boundary has been declared", and that is what -Infinity says.
+   */
+  #regimeStart = Number.NEGATIVE_INFINITY;
+
   readonly #windowMs: number;
   readonly #minSamples: number;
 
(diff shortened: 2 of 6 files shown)


tanifgit (commented):
Reviewed as the `docs/demo/` owner (`.github/CODEOWNERS`).

**The engineering here is the best of the three PRs today and I have no objection to the change itself.** One thing needs deciding before merge, and it is about ordering rather than content.

## Before anything else: this branch contains #158

`b04dfe8` — #158's only commit — is the parent of `1813b55`, but this PR's base is `main`:

```
$ git log --oneline main..origin/devB/reset-throughput-drop-false-positive
1813b55 fix(detection-engine): a reset manufactured throughput_drop on an idle production (#159)
b04dfe8 fix(detection-engine): a slow_processing finding that Smart Resolve could not clear   <- #158

$ git merge-base --is-ancestor origin/devB/cloud-api-processing-floor \
                              origin/devB/reset-throughput-drop-false-positive && echo CONTAINED
CONTAINED
```

So **if this merges first, #158's content reaches `main` inside this PR's squash commit** and #158 becomes a PR whose diff against `main` is empty — the #145/#146 situation exactly, where a change landed inside an unrelated squash and #146 had to be opened as a review record and then closed. That cost a cycle three weeks ago and the note at the top of #146 is still the clearest description of it.

Two ways out, either is fine:

- **Merge #158 first**, then this one — its diff shrinks to `1813b55` alone and nothing is lost; or
- **retarget this PR's base to `devB/cloud-api-processing-floor`**, which makes GitHub show only the new commit and enforces the ordering for you.

I'd take the first, it is one click. Whichever you pick, please don't let this land first with #158 still open — and note my review on #158 raised a fix to `Triggers.SlowProcessing()`'s new `< 1500` warning that would otherwise be lost when its PR gets closed as redundant.

## On the change itself — verified, not taken on trust

- **389 tests / 78 suites / 0 fail, `tsc --noEmit` clean**, reproduced locally on the branch head.
- **The guards are real.** I emptied `ROBUST_METRICS` and re-ran: five tests fail, including the end-to-end `stays silent even though the backlog flushed at 54.8 msg/sec on the way down`. So the median is load-bearing and pinned, not decorative.
- **The clock concern I went looking for is not there.** `beginRegime(Date.now())` compares against sample timestamps that come from `engine.applyPoll(response, Date.now())` at `src/index.ts:88` — one clock, not two. Your comment is right to flag that the polls and the reset call are independent, and the implementation is consistent with it.
- **`ROBUST_METRICS` scoped by direction of harm** is the right justification and the right scope. The table makes the case better than "median is more robust" would: an inflated baseline on the four higher-is-worse metrics makes those rules quieter, which is ADR 0002's documented, pinned, fail-safe behaviour; `throughput_drop` is the only one where it manufactures findings. Scoping to one metric today while stating the rule that would admit a future one is how this should have been written.
- **`recent()` left alone** is the right call, and worth having said out loud. Blanking every host graph and Early Warning's fit on reset would have been a visible regression traded for a statistics fix.
- **The median holding a real drop longer** is a genuine improvement rather than an acceptable cost, and I'd have missed it if the comment hadn't pointed at it.
- Incidentally, `window.ts:127` is `const SEP = '\u0000'` — the same separator technique, written as an escape. That is the house style and it is exactly the fix in #157, where a literal control character in `mockClient.ts` made git treat the file as binary and hid two PRs' diffs.

## One thing missing, and it is in my file

**The 60-second blind window after every reset is not in the cue sheet.** `minBaselineSamples` is 12 and the shipped poll is 5s, so `beginRegime()` means **no comparative finding can fire for ~60s after Reset all**. `engine.ts`'s docstring calls that the intended and honest trade, and I agree with the reasoning — but the presenter-facing consequence is not written down anywhere:

- Reset, then arm, is a realistic sequence between acts.
- `queue_buildup` is comparative, and it is Act 1's headline finding.
- So a presenter who resets and immediately arms sees nothing for up to a minute *plus* the scenario's own warm-up, with no explanation on the board.

Early Warning does render `Baseline still warming — no projection yet` (`EarlyWarning.tsx:58`), which is a partial signal and better than silence. The findings list has none.

I checked whether this PR covers it and it does not — the diff between this branch and #158's for `docs/demo/cue-sheet.md` and `iris/labdemo/README.md` is empty, so the only demo-doc changes here are the ones #158 already made. Suggest a line in the pre-demo trap list:

> **Reset all restarts the baselines.** For ~60s afterwards no comparative finding can fire — that is deliberate (a load step-down is not a fault), but it means *reset, arm, expect a finding immediately* will look broken. Wait for Early Warning to stop saying "Baseline still warming" before arming, or arm an absolute scenario (`DeadHost`) which is unaffected.

Happy to write that myself as a follow-up rather than hold this PR — say which you prefer.

## Minor, take it or leave it

`sampleCount` is now the only `MetricWindow` accessor that ignores `#regimeStart` — `mean()` and `median()` both go through `#inRegime()`, and `sampleCount` returns `#samples.length` raw. Nothing is wrong today: the only consumers are `baseline.test.ts:59` and `engine.test.ts:924`, both asserting that a sample *was recorded*, which is what the raw count should answer. But it now means two different things depending on which accessor you ask, so if anything ever reports it as "how warm is this baseline" it will over-report for 30 minutes after a reset. Either filter it or add a line saying it deliberately counts raw samples.

tanifgit (approved):
Approving the change — it is the strongest of today's three and I verified it rather than trusting it: 389/78/0 reproduced locally, the clock question answered (`beginRegime(Date.now())` and `applyPoll(response, Date.now())` at `index.ts:88` share one clock), and the median proved load-bearing by emptying `ROBUST_METRICS` and watching five tests fail, including the end-to-end `stays silent even though the backlog flushed at 54.8 msg/sec on the way down`.

**The approval is on the content; the ordering condition from my review still holds.** This branch contains #158's commit `b04dfe8` while its base is `main`, so merging it first puts #158's change into this squash and leaves #158 with an empty diff — the #145/#146 situation. **Merge #158 first** (or retarget this base to `devB/cloud-api-processing-floor`). Clean order for today: #154 → #158 → #159.

The 60s post-reset comparative blind window is a cue-sheet gap rather than a code problem — I'm writing that trap line myself, so nothing is needed from you here. The `sampleCount` consistency nit is take-it-or-leave-it.

Ari-Glikman:
**Both points addressed, and the ordering handled the way you suggested.**

**The containment.** Verified your claim rather than taking it (`git merge-base --is-ancestor` → `CONTAINED`), then took your first option: **#158 merged first**, and this branch rebased onto `main`. The rebase hit a conflict on `b04dfe8` and skipping it was the correct resolution, which I checked rather than assumed — `git diff origin/main b04dfe8` shows that commit *removing* content `main` now has (#154's `PURGEDAYSTOKEEP`/purge block and #158's follow-up), so `main` is strictly ahead and nothing was dropped. This PR's diff is now its own single commit across six `services/detection-engine/` files.

**The `sampleCount` nit — taken, as documentation rather than a filter.** I grepped for consumers first: `src/baseline/window.ts` defines it, and the only callers anywhere are the two tests you named. No production consumer at all. Filtering would break the one question it is actually asked — both callers assert a sample *was recorded*, which is a fact about storage and remains true across a regime boundary, so a filtered count makes "did this arrive" unanswerable.

So it now says it counts raw samples on purpose, plus the trap you identified: for ~30 minutes after `beginRegime()` it over-reports relative to what any estimator will use, and `baseline() !== null` / `isWarm()` are the accessors for warmth.

**Your #158 follow-up 1 went into #158 itself** (`5cbe3f2`), for exactly the reason you gave — it would have been lost if that branch were closed as redundant. The band was a little wider than floor-vs-boundary: the induced delay measures ~0.84× the armed value, so the smallest armed value reaching a 1.5 s floor is ~1786 ms, and the boundary is now 2000.

**The 60-second post-reset blind window** — yours to write, and #160 already has it as Trap 7. I've left it alone. Two things from this side that may be worth folding in while you're there, both consequences of the rebase above rather than new behaviour:

- `dead_host` is the right scenario to name as unaffected, and `throughput_drop`'s own trigger (`ThroughputDrop()`) is *not* — it disables `EMR Source`, and `throughput_drop` is comparative, so it is inside the blind window too.
- `SlowProcessing()`'s note text changed in #158, so if the cue sheet quotes it, it now reads "under 2000ms" rather than 1500.

Verified on the rebased head: **389 pass / 78 suites / 0 fail**, `tsc --noEmit` clean, node v22.23.2. CI 5/5.

Thanks for the review — the containment catch saved a repeat of #145/#146, and the `sampleCount` observation was the kind of thing that only bites six months later.

PR #160, opened 2026-08-27
docs(demo): two traps a reset now sets, and a count that had already gone stale

tanifgit:
> **DRAFT ON PURPOSE — merge after #154, #158 and #159.** Every claim here describes behaviour those
> three PRs introduce. On today's `main` all three traps below would be wrong, and there is no branch
> protection to enforce an order, so the draft flag is the only mechanical guard available. One click
> to undraft once they land.

Two traps that came out of reviewing #154 and #159, plus a count that had already gone stale. Each trap is a presenter-facing consequence of a change whose own PR documents the mechanism carefully but not what it does to someone standing in front of an audience.

## Trap 6 — Reset empties the Message Viewer

#154 clears the per-host error count by deleting message rows, because that count is a `COUNT(*)` over `Ens.MessageHeader` and nothing restorable moves it. `Ens.Purge` has no status filter, so it cannot delete only the errored ones — in #154's own verification run, **2,550 headers went to clear 21 errored**.

The gate protects the nothing-errored case. But in the case the feature exists for, the completed history goes too, and how much survives depends on which sessions happened to be in flight rather than on anything a presenter can plan. So the actionable instruction is **show the Message Viewer before you reset**, which is what a trap list can be used for.

## Trap 7 — Reset restarts the baselines

#159's `beginRegime()` is right and its docstring is honest about the ~60s of silence being the intended trade: after a load step-down the engine genuinely does not know what normal is. What that means on stage is that *reset, arm, expect a finding* looks broken for the first minute.

The trap names the four comparative rules that go quiet, names `dead_host` as the safe thing to arm first because it is absolute, and points at Early Warning's `Baseline still warming — no projection yet` as the signal that the wait is over — a state `EarlyWarning.tsx:58` already renders, so this adds nothing to build.

The 60s is given **with its derivation and an instruction to check the engine rather than trust the number**, for the reason the next section is about.

## And the heading's count is removed rather than incremented

`### The five traps` is a copied value sitting directly above the list it counts — exactly what root `CLAUDE.md` §6 and this directory's own §8 are about. It was already wrong the moment a sixth trap was needed. Now `### The traps`.

`apps/dashboard/CLAUDE.md` carried the same count **plus three copied timings, four lines above the rule forbidding exactly that**:

> Do not restate the latency figures anywhere in this directory, including here: link the ADR. Three copied values in this repo have already gone stale.

Two of its three were already stale — #158 re-measured the drain as ~90s to an empty queue and ~200s to a clean board, from 157 queued rather than 165. Replaced with a pointer; the old values now appear only inside the note explaining what went stale.

## Scope

`docs/demo/cue-sheet.md` and `apps/dashboard/CLAUDE.md` — both `@tanifgit` per `.github/CODEOWNERS`. Documentation only, no code, no contract change.

**Deliberately not touching the drain table at `cue-sheet.md:243`** (`t+103s queued=0`). #158 rewrites that whole block and is approved, so it is its change to make rather than a conflict for me to create.

Verified `git merge-tree` clean against #154, #157, #158 and #159 at the time of pushing.

### DIFF
--- a/apps/dashboard/CLAUDE.md
+++ b/apps/dashboard/CLAUDE.md
@@ -480,10 +480,14 @@ all shipped and there was an end-to-end flow worth scripting. Kept rather than d
 and still is not a blocker for any acceptance criterion.
 
 The cue sheet is deliberately **measured rather than written from memory** — every timing in it was
-taken from the containerised stack (77s for `pool_bottleneck` to take effect, 12s to `dead_host`, 103s
-for a queue of 165 to drain after an approve), and it names the five traps that have each cost a
-rehearsal. A cue sheet full of plausible-sounding numbers would be worse than none, because a presenter
-would trust it.
+taken from the containerised stack — and it names the traps that have each cost a rehearsal. A cue
+sheet full of plausible-sounding numbers would be worse than none, because a presenter would trust it.
+
+**The numbers and the trap count live in `docs/demo/cue-sheet.md` and are not restated here**, which
+is §8's rule applied to this paragraph rather than only to the latency bar. It used to quote three
+timings and "the five traps", and both halves went stale: the drain was re-measured as ~90s to an
+empty queue and ~200s to a clean board (from 157 queued, not 165), and the trap list has grown past
+five. A count in prose above a numbered list is a copied value like any other.
 
 **Overall acceptance (from the MVP doc):** *dashboard renders all hosts + findings live; updates within 10 s of a change; fallback to demo mode is seamless.*
 
--- a/docs/demo/cue-sheet.md
+++ b/docs/demo/cue-sheet.md
@@ -75,7 +75,7 @@ look like it ran. The `@'` form passed `$zversion` through untouched and IRIS an
 build. Two layout rules the parser enforces: `@'` must end its line, and `'@` must start its own
 line at **column 0** — indent it and PowerShell does not see the terminator.
 
-### The five traps, in the order they will bite you
+### The traps, in the order they will bite you
 
 1. **`AGENT_MODE` defaults to `mock`.** A canned investigation is built from *real measured values*, so
    it reads as correct and demonstrates nothing about the agent. Check `source: "agent"` and a
@@ -93,6 +93,21 @@ line at **column 0** — indent it and PowerShell does not see the terminator.
    restores it to 1. If the queue never builds, check this first.
 5. **Never `curl /api/monitor/alerts`.** It is consume-on-read — reading it destroys the alert the
    `system_alert` path depends on. Use `:3001/proxy/alerts`.
+6. **Reset all empties the Message Viewer, whenever a scenario errored.** Clearing the per-host error
+   count means deleting message rows — that count is a `COUNT(*)` over `Ens.MessageHeader`, so nothing
+   restorable moves it — and `Ens.Purge` has no status filter, so it cannot delete only the errored
+   ones. Measured: **2,550 headers removed to clear 21 errored.** The completed traffic goes with them,
+   and how much survives depends on which sessions happened to be in flight, not on anything you can
+   plan. **So if a later beat needs the Message Viewer populated, show it before you reset.** A reset
+   with nothing errored skips the purge entirely and leaves history alone.
+7. **Reset all restarts the baselines, so nothing comparative fires for about a minute.** This is
+   deliberate — a load step-down is not a fault, and after a reset the engine genuinely does not know
+   what normal is yet — but it means *reset, arm, expect a finding* looks broken for the first ~60s
+   (`minBaselineSamples` × the shipped poll interval; both live in `services/detection-engine/`, so
+   check there rather than trusting this number). `queue_buildup`, `throughput_drop`, `slow_processing`
+   and `growing_queue_wait` are all comparative and all silent in that window; `dead_host` is absolute
+   and unaffected, which makes it the safe thing to arm first. **The reliable signal that the wait is
+   over is Early Warning: it says "Baseline still warming — no projection yet" until it is.**
 
 **Open two browser tabs before you start:** `http://localhost:5173` (the dashboard) and the
 Management Portal link from its header (the IRIS interoperability editor). Switching tabs is faster
--- a/docs/demo/cue-sheet.md
+++ b/docs/demo/cue-sheet.md
@@ -100,14 +100,23 @@ line at **column 0** — indent it and PowerShell does not see the terminator.
    and how much survives depends on which sessions happened to be in flight, not on anything you can
    plan. **So if a later beat needs the Message Viewer populated, show it before you reset.** A reset
    with nothing errored skips the purge entirely and leaves history alone.
-7. **Reset all restarts the baselines, so nothing comparative fires for about a minute.** This is
-   deliberate — a load step-down is not a fault, and after a reset the engine genuinely does not know
-   what normal is yet — but it means *reset, arm, expect a finding* looks broken for the first ~60s
-   (`minBaselineSamples` × the shipped poll interval; both live in `services/detection-engine/`, so
-   check there rather than trusting this number). `queue_buildup`, `throughput_drop`, `slow_processing`
-   and `growing_queue_wait` are all comparative and all silent in that window; `dead_host` is absolute
-   and unaffected, which makes it the safe thing to arm first. **The reliable signal that the wait is
-   over is Early Warning: it says "Baseline still warming — no projection yet" until it is.**
+7. **Reset all restarts the baselines, and exactly two rules go quiet — not all of them.** The reset
+   declares a new load regime so the rolling mean re-warms instead of averaging across the step down,
+   which is right: a load step-down is not a fault. But the wait is far narrower than it sounds, and
+   assuming otherwise costs you either a minute you did not owe or a finding you read as a bug because
+   it arrived "too early". **Only `throughput_drop` and `elevated_error_rate` go silent**, because
+   `messagesPerSec` and `errorsPerMinute` are the only two metrics with no `referenceBaselines` entry,
+   so they are the only two that wait for the mean. `queue_buildup`, `slow_processing` and
+   `growing_queue_wait` read metrics that *do* have a reference, and a reference replaces the rolling
+   mean outright rather than filling in until it warms — so **they fire immediately, Act 1's headline
+   finding included.** `dead_host` is absolute and unaffected. Those two that do pause are waiting for
+   `minBaselineSamples` × the shipped poll interval — about a minute on the shipped values, but both
+   numbers live in `services/detection-engine/`, so check there rather than trusting this
+   approximation.
+   **Do not wait for a UI signal, because there isn't one.** Early Warning cannot report "Baseline
+   still warming" for these hosts — every one of them has a `queued` reference, so its threshold is
+   never null and that state is unreachable. Right after a reset it shows *"Watching — not trending
+   toward a threshold"* with a tick, which is reassurance rather than readiness.
 
 **Open two browser tabs before you start:** `http://localhost:5173` (the dashboard) and the
 Management Portal link from its header (the IRIS interoperability editor). Switching tabs is faster


Ari-Glikman:
**All three dependencies are on `main`, so this is unblocked** — merged in the order your #159 review specified:

```
455fb07  #157  demo-mode host graphs never drew
c87fe6a  #159  a reset manufactured throughput_drop
bb712fb  #158  a slow_processing finding Smart Resolve could not clear
aa716ee  #154  Reset all now clears the per-host error count
```

Left as a draft for you to mark ready — it is your file and your call, not mine to flip.

**Two things shifted under it while landing, both from your own follow-ups**, so worth a look before you do:

1. **`SlowProcessing()`'s note now says "under 2000ms", not 1500.** Your #158 follow-up 1 went in as `5cbe3f2` *inside* that PR, for the reason you gave — it would have been lost if the branch were closed as redundant. The band was slightly wider than a floor-vs-boundary mismatch: the induced delay measures ~0.84× the armed value (`SlowProcessing(1000)` → 840 ms), so the smallest armed value reaching a 1.5 s floor is ~1786 ms. If the cue sheet quotes that note, the number moved. `iris/labdemo/README.md`'s trigger table was updated in the same commit.

2. **Trap 7's "arm an absolute scenario instead" needs one more exclusion.** `dead_host` is right. But `ThroughputDrop()` is *not* a safe alternative — it disables `EMR Source`, and `throughput_drop` is comparative, so it sits inside the same ~60s blind window. `DeadHost()` and `SystemAlert()` are the two genuinely unaffected ones; `StalledHost()` calls `DeadHost()` so it inherits the absolute part.

On Trap 7 itself: the ~60s figure is right (`minBaselineSamples` 12 × 5 s poll = exactly 60 s), and `EarlyWarning.tsx:58` is the correct signal to point a presenter at, since the findings list has none. Both verified while merging #159.

One more, unrelated to your traps: I split the binary-diff problem out of #157 into **#161** — a CI sweep for NUL bytes in tracked files, plus a `.gitattributes` note. Your finding, and it deserves a mechanism rather than a note, since the failure is silent in the direction that matters: everything compiles, CI stays green, and the only casualty is review.

tanifgit:
**Out of draft — the gate is met.** #154, #158 and #159 all merged in the recommended order (13:36 → 13:38 → 13:41, three separate commits, so #158's change was not swallowed into #159's squash), and #157 followed at 13:45.

Rebased onto `455fb07` and re-verified against the merged state rather than against what the PRs promised:

- **Trap 6's figure holds** — #154's `PurgeErrored()` is on `main` and `Ens.Purge` still has no status filter, so the 2,550-to-clear-21 ratio is the live behaviour.
- **Trap 7's ~60s holds** — `minBaselineSamples` is 12 against the shipped 5s poll, and `EarlyWarning.tsx:58` still renders `Baseline still warming — no projection yet`, which is what the trap points a presenter at.
- **The stale-timing note in `apps/dashboard/CLAUDE.md` is now checkable against `main`**: `cue-sheet.md` lines 256–261 read `t+0s queued=157`, `t+90s queued=0`, `t+202s 0 findings`, so "103s for a queue of 165" was indeed the stale pair.
- It reads correctly next to #154's rewritten trap 2 — that one says the event log is a *different* store and the tile does return to 0; trap 6 adds what the message purge costs. They complement rather than repeat.

Also worth recording here: **the `< 2000` boundary you landed in #158 was the fix I suggested**, so the gap I raised (armed values in [1500, 1785] inducing nothing *and* printing no note) is closed on `main`. Nothing outstanding from my side on any of the four.

Ari-Glikman (changes requested):
Reviewed as the requested reviewer. **Trap 6 and the `apps/dashboard/CLAUDE.md` change are both correct and I'd take them as-is. Trap 7 has two errors and I'd like them fixed before merge** — it is the trap most likely to be read mid-rehearsal, and both halves of it currently point a presenter the wrong way.

## Trap 7, error 1 — the silence list is wrong in both directions

> `queue_buildup`, `throughput_drop`, `slow_processing` and `growing_queue_wait` are all comparative and all silent in that window

Measured, by driving `DetectionEngine` through `beginRegime()` and arming inside the first 40s:

| rule | within ~60s of the reset |
|---|---|
| `queue_buildup` | **FIRED at t+10s** |
| `slow_processing` | **FIRED at t+10s** |
| `growing_queue_wait` | **FIRED at t+10s** |
| `throughput_drop` | silent — correct |
| `elevated_error_rate` | silent — **and the trap does not mention it** |

So three of the four named do fire, and the one comparative rule that really is blind is missing from the list.

The reason is `effectiveBaseline()` (`services/detection-engine/src/config/thresholds.ts:430`), which prefers a `referenceBaselines` entry over the rolling mean *unconditionally* — its own docstring says so: *"A reference therefore also removes the warm-up wait for the metric it covers."* `thresholds.json` carries references for `queued`, `avgProcessingTime` and `avgQueueingTime` on **all three** LABDEMO hosts. Those are exactly the metrics behind the three rules that fired. `messagesPerSec` and `errorsPerMinute` are the only two metrics with no reference — hence exactly two silent rules.

This matters beyond tidiness: the trap tells a presenter that Act 1's headline finding needs a ~60s wait, when `queue_buildup` is available immediately. Someone following it waits out a minute they don't owe, or worse, reads an on-time finding as a bug.

## Trap 7, error 2 — the stated signal can never appear

> **The reliable signal that the wait is over is Early Warning: it says "Baseline still warming — no projection yet" until it is.**

That string exists in `EarlyWarning.tsx` and you cited it accurately, but it is unreachable for these hosts. `earlywarning.ts:67` fixes `METRIC = 'queued'`; line 252 routes it through `effectiveBaseline`; `thresholdFor` returns null only when the baseline is null; and `warming` is only returned when the threshold is null (line 326). Since `referenceBaselines[host].queued = 0` for all three hosts, the baseline is `0` rather than null, so the threshold is always `{value: 50, basis: 'absoluteFloor'}` and `warming` is never returned.

Same root cause as error 1, which is why both load-bearing claims in the trap fail together.

Measured on a cold engine and immediately after a `beginRegime()`:

```
cold engine, 1 poll             projectionUnavailable=insufficient_samples  threshold={value:50,...}
immediately after beginRegime() projectionUnavailable=not_rising            threshold={value:50,...}
```

Note the second line. After a reset the reason is `not_rising`, not even `insufficient_samples` — because `recent()` deliberately ignores the regime boundary (`baseline/window.ts:110`), so the slope fit still has its pre-reset samples. And `not_rising` renders as **"Watching — not trending toward a threshold"** with a tick, per `SHOWN_REASONS`. So a presenter watching for the wait to end sees a green reassurance state for the whole window — the reverse of the signal the trap promises.

## Suggested replacement for the second half of Trap 7

Your call on wording, but to be concrete:

> …but it means *reset, arm, expect a finding* looks broken for the first ~60s (`minBaselineSamples` × the shipped poll interval; both live in `services/detection-engine/`, so check there rather than trusting this number). **Only `throughput_drop` and `elevated_error_rate` are silent in that window** — they read `messagesPerSec` and `errorsPerMinute`, the two metrics with no `referenceBaselines` entry, so they wait for the rolling mean to re-warm. `queue_buildup`, `slow_processing` and `growing_queue_wait` read metrics that *do* have a reference baseline, which removes the warm-up wait, so they fire immediately; `dead_host` is absolute and unaffected. **There is no UI signal to wait for** — Early Warning cannot report `warming` for these hosts, and right after a reset it shows "Watching — not trending toward a threshold", which is not the same claim.

## What checks out

- **Trap 6 is right in every detail I checked.** `PurgeErrored()` returns early on `errored = 0` with "message history left alone" (`Triggers.cls:963`), the count really is a `COUNT(*)` over `Ens.MessageHeader WHERE Status = Error`, and `Ens.Purge` really has no status filter. The 2,550-to-clear-21 figure is our own measurement. The "show it before you reset" instruction is the actionable part and it is the right one.
- **`dead_host` named as absolute and the safe thing to arm first** — correct, and it addresses the point I raised on the earlier round.
- **Not hard-coding `12 × 5s`** and pointing at `services/detection-engine/` instead was the right call; it is the only number in the trap I don't have to re-check in six weeks.
- **The `apps/dashboard/CLAUDE.md` change is a straight improvement** and I'd merge it independently of the above. Removing three copied timings and a copied count is §8 applied where it was being skipped, and "a count in prose above a numbered list is a copied value like any other" is the generalisation worth keeping — the trap count grew to 7 in this very PR, which proves the point.
- The cue sheet does not quote the `SlowProcessing` 1500ms figure anywhere, so my earlier concern about that number was moot.

Requesting changes on Trap 7 only. Happy to re-review as soon as it's updated, or if you'd rather I push the two paragraphs to your branch, say so and I will.

tanifgit:
**Both errors were real and both are fixed in `d8fec8b`. Thank you — this was the right call to block on.** Trap 7 was the one item in the PR I had *not* driven, and it shows: I read `EarlyWarning.tsx` for the string, found it, and never checked whether the state that produces it is reachable.

I verified your findings independently rather than taking them on the review's word, on the grounds that a trap list is only worth having if a presenter can act on it without checking:

**Error 1 — confirmed, and the mechanism is exactly the one you named.** `effectiveBaseline()` is `reference === undefined ? rollingMean : reference` — the reference replaces the mean rather than filling in until it warms. `thresholds.json` carries `queued`, `avgQueueingTime` and `avgProcessingTime` on all three hosts. Reproduced with a throwaway test driving `beginRegime()` on a warm store:

```
queued              rolling=null   effective=0      => USABLE (rule fires)
avgProcessingTime   rolling=null   effective=0.05   => USABLE (rule fires)
avgQueueingTime     rolling=null   effective=0.02   => USABLE (rule fires)
messagesPerSec      rolling=null   effective=null   => BLIND (rule silent)
errorsPerMinute     rolling=null   effective=null   => BLIND (rule silent)
```

Exactly two blind, exactly the two you named, and `elevated_error_rate` was missing from my list. (Scratch test deleted — it duplicates coverage `baseline.test.ts` already has.)

**Error 2 — confirmed, and the second half of your note is the part I would have missed on my own.** `earlywarning.ts:249-253` → `thresholdFor` returns null only when the baseline is null; `:326` returns `warming` only when the threshold is null; every host has `queued: 0` as a reference, so the baseline is `0` and never null. `warming` is unreachable for these hosts, full stop.

The bit that makes this worse than a dead signal is your point about `recent()` ignoring the regime boundary (`window.ts:110`, deliberately, per its own docstring): after a reset the reason is `not_rising`, which `SHOWN_REASONS` renders as *"Watching — not trending toward a threshold"* **with a tick** — the one reason that earns an icon, precisely because it is pure reassurance. So a presenter following my trap would have watched a green tick for the whole window and concluded the wait was over before it started, or that it never ended. The trap now says there is no UI signal, which is the honest version.

**What I changed and what I kept:**

- Trap 7's rule list rewritten to your correction, with the *reason* stated (reference vs. no reference) rather than just the outcome — so the next person can re-derive it if `referenceBaselines` gains an entry, which is the failure mode that produced this bug in the first place.
- Kept the approximate minute **with** its derivation and the pointer to `services/detection-engine/`, since a presenter needs a magnitude. Your suggested wording did the same and I agree it is the right balance.
- Named both failure modes concretely — waiting a minute you did not owe, or reading an on-time finding as a bug — because that is what makes the trap worth reading.
- **Trap 6 and the `apps/dashboard/CLAUDE.md` change are untouched**, per your confirmation.

Ready for re-review. And separately: I owe you a correction — I told the user your `sampleCount` docstring was a nit you had declined. It is in `c87fe6a`, in #159 itself, and it is the better of the two options I offered. My grep missed it.

Ari-Glikman (approved):
**Approving — `d8fec8b` is correct, and I checked it against the code rather than against your description of it.** Every claim in the rewritten Trap 7 matches what I measured:

- only `throughput_drop` and `elevated_error_rate` go silent ✓
- the reason given — `messagesPerSec` and `errorsPerMinute` are the only two metrics with no `referenceBaselines` entry ✓
- `queue_buildup`, `slow_processing`, `growing_queue_wait` fire immediately, Act 1's headline included ✓
- "a reference replaces the rolling mean outright rather than filling in until it warms" ✓ — that is `reference === undefined ? rollingMean : reference` stated in one clause
- `dead_host` absolute and unaffected ✓
- `warming` unreachable because every host has a `queued` reference, so the threshold is never null ✓
- the post-reset state is *"Watching — not trending toward a threshold"* with a tick ✓

**Stating the mechanism rather than just the outcome was the right call, and it is the part that will age well.** The bug you shipped and the bug I nearly missed have the same cause: `referenceBaselines` silently changes which rules need a warm-up, and nothing at either site says so. A reader who finds a fourth reference added to `thresholds.json` can now re-derive the list instead of trusting it — which is the only durable fix for a copied set, per #84.

Keeping the approximate minute with its derivation and the pointer is also right. A presenter needs a magnitude; what they do not need is a number that looks exact.

No action needed on the `sampleCount` correction — you flagged it yourself, it landed in `c87fe6a`, and that is the end of it.

PR #162, opened 2026-08-30
contracts: two read tools — what was CHANGED, and what Guardian is REPORTING

Ari-Glikman:
Contract PR for the two tools requested directly:

> *"check audit table for changes and suggest that there may be a change/typo to recently changed value if it fits … you can bring it up in analysis but do not suggest the fix because they likely changed it for a reason and we do not know what they wanted to change it to."*

> *"the ask about activity only checks the tables but not the findings. so when there is an issue, when we chat and ask if there is any issues (like pool bottleneck) it will not relate to the findings. it should."*

Its own PR because §4 requires one for anything under `contracts/`. The implementation PR depends on this and is separate.

## What changes

`mcp-tools.md` §1, §3.12, §3.13, §5.3's role table and §6, plus a dated `CHANGELOG.md` entry. **Additive** — twelve tools become fourteen, thirteen read and one write. No existing tool's input, output or name changes, so a consumer that never calls the new pair sees nothing different.

## The two rules that are contract-level rather than prompt-level

Both constrain what a **caller** may do with the output, which is why neither is left in a prompt:

- **A consumer must not recommend reverting a change it reports.** The old value *is* returned — withholding it leaves "FilePath changed" where "changed from a directory that exists to one that does not" is the diagnosis — so the restraint is stated rather than engineered. No structural guard is possible: it would have to recognise "revert" in free prose, passing a paraphrase and failing an honest one. `investigation-api.md`'s `manualRemediation.target` is the shape most likely to break it.
- **AI Detective calls `get_recent_config_changes` on every investigation**, not on a misconfiguration-shaped subset. "Misconfiguration-shaped" is a conclusion, so a model cannot use it to decide what evidence to gather. Consumers size for one bounded call per investigation.

## `state` on §3.13 is the part worth reviewing

`count: 0` means four different things and three are not an all-clear: `supplied: false` is no snapshot, `warming` is *nothing measured yet* (the engine's six comparative rules are structurally silent below `minBaselineSamples`), `stale` is a last-known list, and only `ok` supports "no open findings". §2.1's unmeasurable-is-not-zero rule applied to a list instead of a number.

## Measured facts written down so nobody rediscovers them

- `Description` is the subject and `EventData` the detail — the reverse of the names.
- Both are `varbinary`, so **`EventData LIKE '%>>%'` returns SQLCODE -29**; no SQL-side content filter can be written.
- Host attribution must be anchored at **both** ends, because config item names contain spaces (`EMR Source`).
- Only 8 of this instance's 82 `ModifyConfiguration` rows are setting changes at all.
- **A row is not readable the instant it is written** — IRIS buffers audit writes; measured 101 s stale from the next line of the same session.

## Data boundary

`Username`, `ClientIPAddress`, `OSUsername`, `CSPSessionID`, `Roles`, `Pid`, `JobId` and `UserInfo` are **never selected**. "This was changed 40 minutes ago" carries the whole diagnostic weight; naming a person invites an agent to assign blame it cannot support. §6 now lists `get_recent_config_changes` as a third live risk alongside `get_recent_errors` and `compare_host_activity`, because `EventData` is free text written by whatever performed the change — a *less* controlled source than a live item read.

## Verification

Both tools are verified live against the running stack with a real `gpt-4o-mini` — see the implementation PR for the runs. Engine suite 395 pass / 0 fail, typecheck clean.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,132 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-08-30 — two tools: what was CHANGED, and what Guardian is REPORTING
+
+**`mcp-tools.md` §1, §3.12, §3.13, §5.3's role table, and §6.** Additive: twelve tools become
+fourteen, thirteen read and one write. No existing tool's input, output or name changes, so a consumer
+that never calls the new pair sees nothing different. Four stale counts in prose are corrected on the
+way through.
+
+**Both asked for directly**, and each closes a gap where the product answered from the wrong source:
+
+> *"check audit table for changes and suggest that there may be a change/typo to recently changed value
+> if it fits … you can bring it up in analysis but do not suggest the fix because they likely changed it
+> for a reason and we do not know what they wanted to change it to."*
+
+> *"the ask about activity only checks the tables but not the findings. so when there is an issue, when
+> we chat and ask if there is any issues (like pool bottleneck) it will not relate to the findings. it
+> should."*
+
+### `get_recent_config_changes` (§3.12)
+
+Which setting changed on which host, from what value to what value, and when — from `%SYS.Audit`. Every
+other read tool answers a question about the production's *behaviour*; none could answer the one an
+operator asks first, which is *did someone change something?* On this instance `EMR Source`'s `FilePath`
+and `Cloud API`'s `HTTPPort` were both edited within hours of the findings they caused, with nothing
+reading it.
+
+**The reporting rule is in the contract, not only in the prompts**, because it constrains what a
+*caller* may do with the output: **a consumer must not recommend reverting a change it reports.** The
+old value is returned — withholding it would leave "FilePath changed" where "changed from a directory
+that exists to one that does not" is the diagnosis — so the restraint has to be stated rather than
+engineered. `investigation-api.md`'s `manualRemediation.target` is the shape most likely to break it.
+
+**Three fields exist because an empty list is not evidence**, and a caller cannot infer any of them:
+`retention` (IRIS purges audit data on a schedule — measured, a purge ran 2026-08-30 06:00 here),
+`auditEnabled` (`true` / `false` / **`null` for "could not tell"**), and `suppressed` (a *count* of
+non-allowlisted setting names, never the names).
+
+**Measured facts that a reader would otherwise have to rediscover**, all of them written into §3.12:
+`Description` is the subject and `EventData` is the detail, which is the reverse of the names; both are
+`varbinary`, so **`EventData LIKE '%>>%'` returns SQLCODE -29** and no SQL-side content filter can be
+written; host attribution must be anchored at *both* ends because config item names contain spaces; and
+only 8 of this instance's 82 `ModifyConfiguration` rows are setting changes at all.
+
+**One behaviour change outside the contract, and it is a fix rather than a feature.**
+`Triggers.SetSetting()` mutated a live production setting and wrote **no audit row** — measured by
+arming `MissingFolder()` and finding the newest row three days stale, because only the Management
+Portal's own save path audits and that method goes through `Ens.Config.Production.%Save()`. It now emits
+a byte-compatible row itself via `$SYSTEM.Security.Audit()`, **after** the save succeeds, so the log
+never claims a change that did not land. Without this the feature would have been undemonstrable
+through the exact two scenarios it was asked for.
+
+**`Username` and seven other actor columns are never selected** (§6). "This was changed 40 minutes ago"
+carries the whole diagnostic weight; naming a person invites an agent to assign blame it cannot support.
+`%SYS.Audit.EventData` is also free text written by whatever performed the change, which makes an audit
+row a *less* controlled source than a live item read — §6 now lists this tool as the third live risk
+alongside `get_recent_errors` and `compare_host_activity`.
+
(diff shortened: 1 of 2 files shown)


tanifgit (commented):
**Post-merge review.** This requested my review as a `contracts/` change (root `CLAUDE.md` §4) and was merged two minutes after opening with `reviews: []`. Reviewing anyway, because with development consolidated on one person this review is the only external check on `contracts/` that exists.

**Verdict: I'd have approved, with one gap worth closing.** Both new sections are careful and I verified the load-bearing claims against the code rather than the prose.

## What I verified

- **The `%SYS.Audit` identity boundary holds structurally, not by convention.** `Username`, `ClientIPAddress`, `OSUsername`, `CSPSessionID`, `Roles`, `Pid`, `JobId` and `UserInfo` appear in `ChangeLog.cls` **only inside the column-boundary comment** — never in a query. The SQL is `SELECT TOP 400 UTCTimeStamp, Description, EventData` (`:238`) and the retention query is `COUNT(*)/MIN/MAX` (`:311`). Nothing to widen by accident.
- **The allowlist is genuinely shared, not copied.** `ChangeLog.cls:200` reads `##class(...Tools.Read).#SETTINGALLOWLIST`, and `Read.cls:60` is the single `Parameter`. The §1 note about it being a `Parameter` specifically so a public accessor wouldn't become a fifteenth tool is a real constraint, not a rationalisation.
- **`ReportTools()` expects 14** (`AIHub.cls:402`), consistent with §1.
- **The tool-set matrix matches the contract exactly** — `Governance.cls:244-245`: `all` gets `READ/ACTIVITY/EVENTLOG/CHANGELOG`, `chat` gets `ACTIVITY/EVENTLOG/CHANGELOG/FINDING`. So `get_active_findings` really is chat-only and `get_recent_config_changes` really is on every investigation.
- **`findingsState` defaults in the safe direction.** `chat.ts:355` is `snapshot === undefined ? 'warming' : snapshot.state` — absent snapshot degrades to the state §3.13's own table calls "nothing has been measured", not to `ok`.
- **The stale-snapshot hazard I went looking for is already handled, and handled in the right place.** `^||PGGuardian("chat")` is killed as the **first statement inside the `try`, before the body is read** (`ChatDispatcher.cls:280`), with the pooled-CSP-process reasoning and the argument against killing on the way out both written down. This was the first thing I checked and I expected to find a bug; the comment had got there first.
- **Local verification on `d7b0218`:** detection-engine `tsc --noEmit` clean, **395 tests / 78 suites / 0 fail**; dashboard typecheck and `npm run build` clean.

## The gap — §3.13 silently truncates at 25 and the contract does not say so

§3.12 publishes `truncated` because it scans `TOP 400`. §3.13 caps at 25 in **two** places and publishes nothing:

- `chat.ts:349` — `snapshot.findings.slice(0, MAX_FINDINGS)`, `MAX_FINDINGS = 25` (`:194`)
- `Findings.cls:161-163` — `if findings.%Size() >= ..#MAXFINDINGS { quit }`, `MAXFINDINGS = 25` (`:91`)

and then `out.count = findings.%Size()` (`:167`) — so **`count` is the post-cap size, not the true number.** The `note` field is only set when `count = 0`, so at the boundary the payload is indistinguishable from a complete one.

Grepping §3.13 for `truncat|cap|25|max` returns nothing. A consumer implementing from the contract cannot know findings can be dropped.

The consequence is the one this section is otherwise built to prevent. Asked *"how many issues are there right now?"* on a production with 40 open findings, the agent receives `count: 25` with no signal and will state 25 as fact — the same shape as `count: 0` meaning four different things, which §3.13 goes to real trouble over. It is *more* likely to bite than the `count: 0` cases, because a bad state is exactly when findings multiply and exactly when someone asks.

Two smaller notes in the same area:

- **The 25 is duplicated across the language boundary** — `chat.ts`'s `MAX_FINDINGS` and `Findings.cls`'s `MAXFINDINGS` are independent literals with no cross-reference. Within IRIS it is single-sourced (`ChatDispatcher` reads `Tools.Findings.#MAXFINDINGS`); across the wire it is a copy. Raise the engine's to 50 and IRIS silently drops 25 of them — invisibly, given the above. That is #84's shape, and #84 is still open.
- **The host filter runs before the cap** in `Findings.cls:158-163`, which is the correct order (filter then truncate, not truncate then filter). Worth keeping a test on, since swapping them would be a subtle regression that still returns plausible output.

**Suggested fix, matching §3.12's own precedent rather than inventing a shape:** add `truncated` (and optionally `suppliedCount`) to §3.13's output, set it engine-side where the true length is known, and forward it. Cheapest correct version if you'd rather not touch the wire: set `note` when `count = MAXFINDINGS` as well as when it is 0 — the field already exists for "the number does not speak for itself".

## On the two design calls

Both are right and I'd have argued for them independently:

- **Findings supplied with the request rather than fetched via callback.** The reason given — a callback needs an engine URL inside the IRIS container, the class of configuration `iris/CLAUDE.md` records going missing on three cold boots, failing as "no findings", which reads as a healthy production — is the correct reason, and it is the failure direction that matters.
- **"A recent change is evidence, never a recommendation," and being explicit that it is prompt-enforced with no structural guard.** Saying a guard would have to recognise "revert" in free prose, so it would pass a paraphrase and fail an honest one, is the honest version. I'd rather have that sentence than a regex that creates false confidence. The three-run verification with nine remediation steps and no revert is the right evidence for a rule of that kind.

Nothing here needs reverting. The truncation gap is one field and a sentence.

PR #163, opened 2026-08-30
feat: Guardian reads what CHANGED, and the chat reads what Guardian is REPORTING

Ari-Glikman:
Implements `contracts/mcp-tools.md` §3.12–§3.13. **Based on #162**, not on `main`, so this diff shows only the implementation — retarget to `main` once #162 lands.

Closes both requests:

> *"check audit table for changes and suggest that there may be a change/typo to recently changed value if it fits … you can bring it up in analysis but do not suggest the fix because they likely changed it for a reason and we do not know what they wanted to change it to."*

> *"the ask about activity only checks the tables but not the findings. so when there is an issue, when we chat and ask if there is any issues (like pool bottleneck) it will not relate to the findings. it should."*

## 1. `get_recent_config_changes` — what CHANGED

`Tools.ChangeLog` reads `ModifyConfiguration` rows from `%SYS.Audit`: which setting changed on which host, from what value to what value, and when. Every other read tool answers a question about the production's *behaviour*; none could answer the one an operator asks first.

**A recent change is evidence, never a recommendation.** The prohibition is in both prompts, and sharper in `AgentDispatcher`'s for a structural reason: `manualRemediation.target` is a `{host, setting, currentValue}` slot the reply schema *invites*, so a recently-changed setting plus a slot for a setting fix is a strong pull toward exactly the recommendation the operator did not ask for. The old value is returned anyway — withholding it would defeat the tool — so the restraint is stated rather than engineered. No structural guard, deliberately: it would have to recognise "revert" in free prose, and a field that cannot be checked is worse than a rule that is stated.

**`Triggers.SetSetting()` now writes its own audit row, and that is a security fix rather than test scaffolding.** A setting changed through `Ens.Config.Production.%Save()` is not audited — only the Management Portal's own save path is, measured by arming `MissingFolder()` and finding the newest row three days stale. So a trigger was mutating a live production setting with **no attributable trace**, which root `CLAUDE.md` §2.1 requires. Note `$SYSTEM.Security.Audit()`'s argument order: arg4 → `EventData`, arg5 → `Description`, the reverse of what the names suggest, and getting it wrong produces a valid row the host parse silently discards.

## 2. `get_active_findings` — what Guardian is REPORTING

The engine sends its findings snapshot with the question; `Tools.Findings` republishes it as a governed tool.

- **Over an IRIS→engine callback**, which needs an engine URL inside the container — the class of configuration `iris/CLAUDE.md` records going missing on three separate cold boots, each time failing as "no findings", which reads as a healthy production.
- **Over prompt injection**, which loses `evidence[].tool` attribution, exempts a second channel of truth from the audit guarantee, and taxes every turn including ones that do not want findings.

**`findingsState` is forwarded, stashed and republished** because `count: 0` means four different things and three are not an all-clear. `Ask()` kills the handoff node on **entry**, before deciding whether to set it — CSP processes are pooled and reused, so a question supplying no snapshot would otherwise answer from the previous question's, minutes stale.

## 3. Two defects found by running it live, both fixed here

**A tool that is registered and described is not a tool that gets called.** `GetRecentConfigChanges` shipped registered in `Tools.Governance`, with its system-prompt paragraph verified present in the compiled `.INT` — and `toolCalls: 2` on the missing-folder scenario it exists for, the model satisfied after reading the errors and the settings. `BuildGoal()` already carried a `MUST` for `get_recent_errors` and `get_host_settings` for this exact measured reason, so the third tool repeated a failure the codebase had already written down. Now a third `MUST`, unconditional — "misconfiguration-shaped" is a conclusion, and a model cannot use it to decide what evidence to gather.

**Adding that directive then cost every evidence entry its attribution.** Its first version closed by deferring to "the rules in your instructions", and `source` came back unset across three consecutive runs. `investigate.ts` maps an unrecognised source to `"llm"`, so the reply stayed schema-valid while claiming the model had *reasoned out* values it had in fact read from governed tools — attribution silently lost, in the field the audit story depends on. The goal is the last thing the model reads and now the longest instruction about evidence, so a vague pointer at another part of the prompt competed with the concrete schema and won.

## Verification — live, not mocked

Four containers, `AGENT_MODE=live`, real `gpt-4o-mini`:

| Check | Result |
|---|---|
| chat under `warming`, nothing armed | `toolCalls: 1`, `evidence[].tool: GetActiveFindings`, and the model **refused an all-clear**: "state is 'warming' … not possible to confirm that everything is functioning correctly" |
| chat with `PoolBottleneck()` armed | both real findings cited with their own numbers — queue depth 70, queue wait 7.68 s at 384× baseline — attributed to `GetActiveFindings`, `confidence: 1` |
| AI Detective on the armed `MissingFolder()` `dead_host` | `toolCalls: 3` across three consecutive runs; the 10:27:26 `FilePath` change cited in `rootCause` and as attributed evidence; **no revert in any of the nine remediation steps** |

`supplied: true` is the one thing only a live run could test — it rests on the agent loop running in the CSP request's own process, which is what makes a process-private global a valid handoff.

- Engine suite **395 pass / 0 fail** (389 → 395). One pre-existing test asserted the outbound request deep-equalled the caller's body, which this feature makes false by design; it now asserts the caller's two fields survive, and six new tests cover the attachment — including that a caller **cannot** post its own findings, and that a missing supplier sends `warming` rather than `ok`.
- `npm run typecheck` clean.
- `LoadDir` over `labdemo/` and `setup/` with `ERRORCOUNT 0`; `ReportTools()` reports **14 of 14**.

## Note for the reviewer

Three fabricated `ProbeSetting:a>>b` rows from my argument-order testing are still in `%SYS.Audit` on the dev instance (2026-08-30 09:45–09:47). They are invisible to the tool — their `Description` fails the host parse, which is *why* they were written — but they are invented data in an audit log. Deleting audit rows was correctly blocked as tampering, so this is a call for a human: `Audit.Entry.Purge()`-style cleanup, or leave them.

### DIFF
--- a/iris/CLAUDE.md
+++ b/iris/CLAUDE.md
@@ -13,7 +13,7 @@ Developer A owns:
 |---|---|
 | `iris/setup/` | One-time ObjectScript setup scripts for the demo namespace |
 | `iris/labdemo/` | LABDEMO production class, HL7 generator, trigger toggles |
-| `iris/labdemo/Tools/` | MVP 2: the six MCP tools, the authorization and audit policies, the governed-manager factory |
+| `iris/labdemo/Tools/` | MVP 2: the MCP tools, the authorization and audit policies, the governed-manager factory. **The count is in `contracts/mcp-tools.md` §1 and is not restated here** |
 | `iris/labdemo/Audit/` | MVP 2: the persisted audit trail — AI Hub ships no audit store |
 | `services/metrics-proxy/` | Node.js proxy: polls `/api/monitor/metrics` + `/api/monitor/alerts`, exposes per-host JSON |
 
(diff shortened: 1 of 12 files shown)


Ari-Glikman:
Superseded by #164 — same commit rebased onto `main`. GitHub closed this automatically when #162's branch was deleted on merge, and would not reopen it once the base branch was gone.

Issue #166, opened 2026-08-31
Every /api/monitor/* request writes an AccessDenied audit row: ~55,000/day fills the audit log and hides real denials


tanifgit:
Every HTTP request to `/api/monitor/*` writes one `AccessDenied` audit row, so the Security Audit log fills at ~55,000 rows/day and hides real denials.

Reported from the Management Portal's Security Audit log, which is wall-to-wall `ERROR #822: Access Denied` at roughly one row every 1.5-2.5 s.

**Not caused by the 2026-08-30 changes.** The mechanism depends only on something polling `/api/monitor/*`, and the metrics proxy has done that since `71c38d0` (2026-08-10). The current *rate* dates from `22ead20` (2026-08-16, #72/#78), which set `METRICS_POLL_INTERVAL_MS: "2500"` and added the IRIS healthcheck. What landed yesterday (`d7b0218`, #164) is only the code that *noticed* it — `ChangeLog.cls:133` already records "`AccessDenied` accounts for 111,228 of this instance's 162,333 audit rows and is still arriving every ~2.5 s". This issue is that comment's root cause, filed so the two existing workarounds stop being the only record of it.

## Root cause

Every HTTP request to the `/api/monitor` web application emits exactly one audit row. Identical for every row measured:

| field | value |
|---|---|
| `Event` | `AccessDenied` (`%System` / `%Security`) |
| `Username` | `UnknownUser` |
| `Namespace` | `%SYS` |
| `Roles` | `%DB_IRISSYS` |
| `RoutineSpec` | `EXISTS+2^SECURITY.SERVICES.1` |
| `EventData` | `ERROR #921: Operation requires %Admin_Secure:USE privilege` |
| `Description` | `ERROR #822: Access Denied` |

IRIS's monitor dispatch calls `Security.Services.Exists()` while the request identity is still the unauthenticated `UnknownUser`, which holds only `%DB_IRISSYS`. That lookup needs `%Admin_Secure:USE`, is denied, and is audited. **The request then succeeds anyway** — `HTTP 200` with the full 60,214-byte payload — so nothing upstream ever goes red. This is audit noise, not a functional fault.

`/api/monitor` is `AutheEnabled=64`, `NameSpace=%SYS`, `MatchRoles=':%DB_IRISSYS'`.

### Credentials are irrelevant — do not "fix" this by changing them

Measured on the running stack, one row per request **either way**:

| request | rows |
|---|---|
[...]

PR #167, opened 2026-08-31
fix(docker): the IRIS healthcheck was writing 17,280 audit rows a day (#166)

tanifgit:
Fixes the second-largest source in #166. Partial by design — see the last section.

## What was happening

Every HTTP request to `/api/monitor/*` makes IRIS write one `AccessDenied` audit row. `Security.Services.Exists()` is called while the request identity is still `UnknownUser`, which holds only `%DB_IRISSYS` and so lacks the `%Admin_Secure:USE` that lookup needs. **The request then returns `200` with the full payload**, which is why this was invisible: nothing upstream ever went red, and the only trace of it in the repo was `ChangeLog.cls` working around the resulting table growth (`SCANROWS = 400`, and a `Retention()` that counts only `ModifyConfiguration` because "`AccessDenied` dominates the row count here").

This healthcheck probes `/api/monitor/metrics` **every 5s forever**, not just until healthy, pulling the whole 60 KB payload to `grep` one label — 17,280 rows/day for a signal only `depends_on` consumes.

## Why cadence and not credentials

Measured from `pg-dashboard`, which otherwise never touches `/api/monitor`, so its rows isolate from the proxy's and this probe's:

| request | rows |
|---|---|
| 20 x anonymous `GET /api/monitor/metrics` | **+20** |
| 20 x authenticated `GET /api/monitor/metrics` | **+20** |
| 10 x authenticated `GET /labdemo/monitor/hoststatus` | **+0** |
| 5 x anonymous `GET /labdemo/monitor/hoststatus` (403) | **+0** |

Anonymous `GET` returns `200` here despite `/api/monitor` being `AutheEnabled=64`, and one row is written either way. So the `Authorization` header is not the lever — cadence is the only one that costs nothing else.

## Verification

On the running stack, after `docker compose up -d iris`:

- **healthy after 18s** (warm volume; the measured cold figure is ~60s)
- applied config: `StartPeriod 300s, StartInterval 5s, Interval 30s, Retries 3`
- steady-state probes **30.0s apart** across 5 consecutive `.State.Health.Log` entries
- closed 180s window, `/api/monitor` `AccessDenied` by client:

| client | rows/180s | rate | before |
|---|---|---|---|
| `127.0.0.1` (this probe) | 7 | 0.039/s | 0.200/s |
| `172.19.0.4` (the proxy) | 69 | 0.383/s | unchanged, as intended |

`docker compose config` renders `start_interval` correctly. CI has no compose-validation job, so the live run above is the evidence.

### If you re-measure this, mind the buffering

`%SYS.Audit` rows appear in bursts **>15s and <110s** after the request. Three of my early readings were wrong because of it: a 3-request batch read as "+0 rows" and briefly looked like proof that anonymous requests were free, and a 30s rate sample read 0.8/s where the long-run rate is 0.4/s. Settle ~2 min and prefer a closed timestamp window over a count delta.

## Startup detection is unchanged

That is the whole reason this is safe. `start_interval` probes every 5s until the first success, so dependents still gate on our production at 5s granularity; `interval` governs only steady state. The existing comment at line 159 about "4.0s of margin against a 5s probe interval" therefore still holds unchanged.

`retries` drops 60 -> 3 and **that is not a reduction in tolerance**. It used to carry the whole startup budget, because `start_period: 0s` meant a probe failing while IRIS was still booting counted against it — 60 x 5s was that same 5 minutes, spelled as retries. `start_period` now covers startup properly (failures inside it neither count nor mark the container unhealthy), so `retries` is free to mean what it should: how long a *healthy* instance may stay broken before we say so. Pathological cold-boot tolerance goes **up**, 300s -> 390s.

Requires Docker 25+ for `start_interval`; this stack is 29.7.2 / Compose v5.4.0.

## Scope — what this deliberately does not do

Removes ~14,000 of ~55,000 rows/day. The proxy's 2.5s poll is the larger source (~34,560/day) and is **left alone on purpose**: ADR 0005 and #75 halved it twice for the latency budget.

#166 ranks the rest, including the two I declined to take unilaterally:

- **disabling the `AccessDenied` audit event instance-wide** — the only change that removes all of it, and the only one that loses genuine security signal, so it is an operator's decision rather than a commit
- **granting `%Admin_Secure:USE` to the monitor context** — refused; it would silence the denial by giving security-administration privilege to every unauthenticated monitor request, inverting §2.1's RBAC boundary

Reviewed by me rather than by Ari, who is out today, with their standing agreement that I act as owner across areas for the day. Flagging that explicitly because `docker-compose.yml` is shared root config and would normally take a second pair of eyes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -189,17 +189,45 @@ services:
       test:
         - CMD-SHELL
         - wget -qO- http://localhost:52773/api/monitor/metrics | grep -q 'production="ProductionGuardian.LabDemo.Production"'
-      # 60 x 5s = 5 minutes, which is HEADROOM rather than a measured requirement. Every cold
-      # run measured on this image reaches healthy in ~60s, because LABDEMO ships
-      # interop-enabled and the slow `EnableNamespace` branch never executes here. An earlier
-      # version of this comment justified the budget by "covers a cold EnableNamespace
+      # TWO CADENCES, BECAUSE THIS PROBE IS NOT FREE (#166). Every request to /api/monitor/*
+      # makes IRIS write one `AccessDenied` audit row -- `Security.Services.Exists()` is called
+      # while the request identity is still UnknownUser, which lacks %Admin_Secure:USE. The
+      # request still returns 200 with the full payload, so nothing here ever went red; the cost
+      # is 17,280 audit rows/day from this healthcheck alone, on top of the proxy's 34,560.
+      #
+      # Measured, not assumed: 20 requests to /api/monitor/metrics produced exactly 20 rows,
+      # anonymous and authenticated alike, while 10 to /labdemo/monitor/hoststatus produced
+      # none. So credentials are not the lever and this cadence is.
+      #
+      # A healthcheck keeps probing forever, not just until healthy, which is what makes a 5s
+      # interval expensive for a signal only `depends_on` consumes -- and it pulls the whole
+      # 60 KB metrics payload to grep one label.
+      #
+      # STARTUP DETECTION IS UNCHANGED. `start_interval` probes every 5s until the first
+      # success, so dependents still gate on our production at 5s granularity; `interval` only
+      # governs steady state, cutting this source by 6x to 2,880 rows/day. Needs Docker 25+
+      # (this stack: 29.7.2 / Compose v5.4.0).
+      #
+      # 5 minutes is HEADROOM rather than a measured requirement, and it now buys more than it
+      # did: every cold run measured on this image reaches healthy in ~60s, because LABDEMO
+      # ships interop-enabled and the slow `EnableNamespace` branch never executes here. An
+      # earlier version of this comment justified the budget by "covers a cold EnableNamespace
       # (minutes)", which describes a path that does not run on this image (@tanifgit, #78) --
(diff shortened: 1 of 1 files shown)


tanifgit:
Self-review note (Ari is out; recording the one thing I had to think hard about rather than leaving it implicit).

**`retries: 3` + `interval: 30s` means a wedged instance is marked unhealthy in ~90s, where it used to take ~300s.** I checked whether that can hurt anything and it cannot:

- the `iris` service has **no `restart:` policy**, so `unhealthy` is a status label, not an action — nothing gets killed or recreated
- `depends_on: service_healthy` only gates *startup*; a running proxy/engine are unaffected by a later status change
- so the only effect is diagnostic, and faster is better here — #101 (the web server wedging during a long `queue_buildup` run) is exactly the condition that would now surface in 90s instead of 5 minutes

Also worth stating plainly: the "6x to 2,880 rows/day" figure is the **steady-state** rate, derived from the verified 30.0s cadence. The measured 0.039/s over my 180s window is slightly above that because the window straddled the 5s start-period probing before the container went healthy. Both numbers are in the PR body; neither is rounded in my favour.

Verified after merge-candidate state: all four containers healthy, `engine /api/healthscan/findings -> 200`, `proxy /proxy/metrics -> 200`.

PR #168, opened 2026-08-31
feat(iris): grant %Admin_Secure inside /api/monitor, so it stops auditing a denial per request (#166)

tanifgit:
Removes the whole ~55,000 rows/day from #166 — not the 26% that #167's healthcheck cadence fix got.

| | rows |
|---|---|
| 180s before | **72** |
| 180s after | **0** |
| Newest `AccessDenied` anywhere on the instance | the second it was applied |

This is @tanifgit's suggested approach. Credit where due — I had ranked it as refused in #166, on a reading that turned out to be wrong about how broad it had to be.

## What was actually wrong

IRIS's monitor dispatch calls `Security.Services.Exists()`, which needs `%Admin_Secure:USE`, while the session holds only the application's own roles. The lookup is denied and audited; **the request then returns 200 with the full payload**, so nothing upstream ever went red. The only visible symptom was the Security Audit log filling — `AccessDenied` was 111,228 of 162,333 rows on the instance `Tools.ChangeLog` measured, which is why that tool carries `SCANROWS` and a `ModifyConfiguration` filter as workarounds for a bug nobody had filed.

## The properly-scoped version does not work, and that is the interesting part

A **matching role** plus a dedicated client user would confine the privilege to one user inside one app. That was the first design tried, and it cannot work here: `/api/monitor` never authenticates anyone.

| request | result |
|---|---|
| `/api/monitor/metrics` with a **deliberately wrong** password | **200** |
| `/labdemo/monitor/hoststatus` with the same wrong header | **401** |

The `Authorization` header is discarded, so no client user would ever be authenticated there, nothing would hold the role to match against, and the rule would never fire. What proves the **application**-role path is live is the audit row itself: it records `Roles: %DB_IRISSYS`, which is exactly the app's existing always-add entry.

Anyone tempted to tighten this into a matching role later should re-run the wrong-password check first. The code says so at the call site.

Credentials are not the lever either, measured both ways: 20 anonymous and 20 authenticated GETs to `/api/monitor/metrics` each produced exactly 20 rows; 10 authenticated GETs to `/labdemo/monitor/hoststatus` produced none.

## The cost, stated plainly

`%Admin_Secure:USE` is granted to any caller that can reach `/api/monitor/*`, **unauthenticated**, and this stack publishes 52773 to the host.

Mitigating facts, all checked rather than assumed:

- Confined to the application. `DispatchClass` is fixed to `%Api.Monitor`, so a caller cannot run arbitrary code under it.
- **No information disclosure.** Payload before vs after: 1069 lines and 130 metric families both sides, no family added or removed. The only differing line is one live `iris_process` `namespace` label — sampling state, unrelated.

It is still a real widening against root `CLAUDE.md` §2.1's RBAC posture, and a deliberate call rather than an oversight. #166 records the alternative that removes the flood without this trade — moving metrics collection onto a `LABDEMO` application, whose requests audit zero — which is a `contracts/` change and needs its own spec.

## Verification

Both branches, because `iris/CLAUDE.md` is explicit that the create branches in this class are the ones that go unproven — three cold-boot defects came from exactly that.

```
from stock (:%DB_IRISSYS, role deleted):
  4c. role PGMonitorSecure: created
  4c. app /api/monitor MatchRoles: ':%DB_IRISSYS' -> ':%DB_IRISSYS:PGMonitorSecure'
run again, unchanged input:
  4c. role PGMonitorSecure: exists
  4c. app /api/monitor MatchRoles: already grants PGMonitorSecure
```

- runs in the **real boot sequence** (`compose restart iris`), positioned after 4b and before 5
- closed 180s window entirely after that restart: **zero** `AccessDenied` rows, all clients
- four containers healthy, engine findings 200, proxy 200, 12 hosts / 10 merged

**Not** tested from `compose down -v`. The create branch is proven by running it from stock state, but a true cold boot would additionally prove `Run()`'s ordering on an empty volume, and I did not destroy the volume because it holds the audit evidence this issue rests on.

### One defect found by running it

The first version printed `':%DB_IRISSYS:PGMonitorSecure' -> ':%DB_IRISSYS:PGMonitorSecure'` on the create path: `set $piece(...)` mutates in place, so the report read the same variable it had already changed. The one line whose job was to say what changed was the one line that could not. `current` is now a copy of `before`. Caught by running it, not by reading it.

### If you re-measure any of this

`%SYS.Audit` writes are buffered and land in bursts **>15s and <110s** after the request. Three of my early readings were wrong because of it — a 3-request batch read as "+0 rows" and briefly looked like proof that anonymous requests were free. Settle ~2 minutes and use a closed timestamp window, never a count delta taken straight after a request.

Reviewed by me rather than by Ari, who is out today, with their standing agreement that I act as owner across areas for the day — flagged explicitly because this is `iris/**` (Dev A's area) and it changes an instance-wide security grant.

### DIFF
--- a/iris/setup/FirstBoot.cls
+++ b/iris/setup/FirstBoot.cls
@@ -43,6 +43,12 @@ Parameter ARCHIVEDIR = "/tmp/labdemo/hl7-archive/";
 /// The Ensemble credential Cloud API uses to POST to the REST app.
 Parameter CREDENTIAL = "LabDemoREST";
 
+/// The stock monitoring web application, and the application role granted on it (#166).
+/// See `GrantMonitorSecureRole()` for what this buys and what it costs.
+Parameter MONITORAPP = "/api/monitor";
+
+Parameter MONITORROLE = "PGMonitorSecure";
+
 /// Stand the whole thing up. `pSourceDir` is where the .cls files live, as seen from
 /// inside IRIS; pass "" to skip compilation if the classes are already loaded.
 ClassMethod Run(pSourceDir As %String = "", pStartGenerator As %Boolean = 1) As %Status
@@ -67,6 +73,7 @@ ClassMethod Run(pSourceDir As %String = "", pStartGenerator As %Boolean = 1) As 
     do ..MakeDirectories()
     do ..RegisterWebApps()
     do ..ClearPasswordChangeFlag()
+    do ..GrantMonitorSecureRole()
     do ..CreateCredential()
 
     set sc = ..EnableMetrics()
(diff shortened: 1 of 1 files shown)


tanifgit:
Self-review (Ari is out; recording what I checked and the one thing I chose not to change).

**Checked and correct:**
- `if $$$ISERR(sc) quit` after role creation — aborts rather than granting a role that does not exist, which would have left `MatchRoles` naming a phantom.
- `new $namespace` / `%SYS` matches the `RegisterWebApps()` pattern, so the caller's namespace is restored.
- `Security.Applications.Modify()` with a partial props array preserves the other 30-odd properties — verified by behaviour, not assumption: the app still served 200 with the full payload afterwards.
- `Security.Roles.Create()` arg 3 really is Resources — read back as `%Admin_Secure:U`.
- The missing-app branch reports instead of creating. `/api/monitor` ships with the instance; inventing it would be worse than saying it is absent.

**Known limitation, deliberately not fixed:** line 366 tests `before [ ..#MONITORROLE`, a substring containment. A role whose name merely *contains* `PGMonitorSecure` would make this skip the grant and silently leave the flood in place — a quiet-wrong failure of the kind this repo is usually strict about. I left it because we own the name and the prefix, so the probability is effectively zero, and an exact-piece walk would add a loop to a method whose whole value is being obvious. Flagging it so the trade is on the record rather than implicit.

**Not claimed:** this was not tested from `compose down -v`. The create branch is proven by running it from genuine stock state (role deleted, `MatchRoles` restored), and step 4c is proven to run in the real boot sequence via `compose restart iris`. What a cold boot would add is `Run()`'s ordering on an empty volume. I did not destroy the volume because it holds the audit evidence #166 rests on — worth doing before the next demo, per `iris/CLAUDE.md`.

Issue #177, opened 2026-08-31
AI Detective recommends a bigger pool for a backlog that is already draining: no cross-host reasoning about an upstream bottleneck that was just fixed


tanifgit:
Requested by @tanifgit:

> We need to try and implement a little smarter reasoning for the Investigate ("Detective") part. For example: if simulated bad folder/path issue in incoming EMR Source, and then fixed, the Cloud API will suddenly get a surge of messages and have a big queue, but need to reason that those are due to an issue (a "bottleneck") in some other host, which was just fixed, and the current flow is "back to normal" so it probably does not need to increase the pool size further (say it was already at 4). Unlike a situation where there is a more constant pickup of incoming message volume.

## The scenario, and what the Detective does with it today

1. `MissingFolder()` points `EMR Source`'s `FilePath` at a path that does not exist. Inbound stalls.
2. The path is fixed. `EMR Source` releases everything that accumulated, in a burst.
3. `Cloud API` — `PoolSize` already 4 from an earlier Smart Resolve — queues behind the burst. `queue_buildup` fires.
4. Investigate runs. Every mandatory tool call it makes is scoped to `Cloud API`, and the queue is genuinely deep and genuinely above threshold, so the diagnosis is "Cloud API cannot keep up" and the recommendation is a larger pool.

That recommendation is wrong in the way that matters: the backlog is **draining through** a pipeline that is already fixed, and enlarging the pool a second time treats a transient as a capacity problem. Nothing about the finding on `Cloud API` distinguishes it from a sustained rise in inbound volume, which is the case where the same recommendation is right.

## Why it cannot currently reason otherwise

**The evidence it is handed is single-host.** `services/detection-engine/src/detect/investigate.ts:150` — `buildSnapshot(host, projection)` carries the finding's host only: status, queued, messagesPerSec, errored, the two timings, and its threshold. No neighbour, no upstream, no topology.

**The mandatory tool calls are single-host.** `iris/labdemo/REST/AgentDispatcher.cls:762` (`BuildGoal`) requires `get_recent_errors`, `get_host_settings` and `GetRecentConfigChanges` — each takes a `host` and each is called for the finding's host. Every method on `Tools.Read` is host-scoped (`GetHostStatus`, `GetQueueDepth`, `GetPoolSize`, `GetHostSettings`, `GetRecentErrors`).

[...]

Issue #165, opened 2026-08-30
get_active_findings caps at 25 silently: count reports the capped size and the contract does not mention a cap


tanifgit:
Found reviewing #162 post-merge (that PR requested my review as a `contracts/` change and merged two minutes later).

`get_active_findings` (`contracts/mcp-tools.md` §3.13) caps its result at 25 and publishes no signal that it did.

Two independent caps:

| where | code |
|---|---|
| engine, before sending | `services/detection-engine/src/detect/chat.ts:349` — `snapshot.findings.slice(0, MAX_FINDINGS)`, `MAX_FINDINGS = 25` (`:194`) |
| IRIS, on receipt | `iris/labdemo/Tools/Findings.cls:161-163` — `if findings.%Size() >= ..#MAXFINDINGS { quit }`, `MAXFINDINGS = 25` (`:91`) |

Then `Findings.cls:167` is `set out.count = findings.%Size()` — **`count` is the post-cap size, not the true number** — and `note` is only set when `count = 0` (`:178`). So at the boundary the payload is byte-shaped exactly like a complete one.

`§3.13` does not mention a cap: grepping the section for `truncat|cap|25|max` returns nothing. A consumer implementing from the contract cannot know findings can be dropped.

## Why it matters more than the `count: 0` cases §3.13 already guards

That section goes to real trouble over `count: 0` meaning four different things, and rightly. This is the same failure in the other direction and it is *more* likely to fire: findings multiply exactly when a production is in a bad state, which is exactly when someone asks *"how many issues are there right now?"*. The agent receives `count: 25` with no signal and will state 25 as a fact.

`§3.12` already solves this for its sibling — `get_recent_config_changes` publishes `truncated` because it scans `TOP 400`. The two sections are inconsistent on exactly the axis the rest of the contract is scrupulous about.

## Suggested fix

Follow §3.12's precedent rather than invent a shape: add `truncated` to §3.13's output (optionally `suppliedCount` for the true length), set it engine-side where the real length is known, and forward it alongside `findingsAsOf` / `findingsState`.

Cheapest correct version if you'd rather not change the wire: set `note` when `count = MAXFINDINGS` as well as when it is `0`. The field already exists for "this number does not speak for itself", and the `warming` / `stale` notes are the precedent.

## Related, and the reason this is filed rather than just commented

[...]

Issue #161, opened 2026-08-27
A raw control character in a source file silently disables review for that file — add a CI sweep


Ari-Glikman:
Split out of #157 rather than fixed inside it, because it is a process gap and not a dashboard bug.

## What happened

`apps/dashboard/src/api/mockClient.ts` contained a **literal U+0000** byte. Git classifies any file containing a NUL as binary, so every diff of it rendered as:

```
$ git show --stat 9a44346 -- apps/dashboard/src/api/mockClient.ts
 apps/dashboard/src/api/mockClient.ts | Bin 21868 -> 27238 bytes
 1 file changed, 0 insertions(+), 0 deletions(-)
```

On GitHub the file showed **no diff at all**. #140 introduced the character (`Bin 15885 -> 21868`) and #144 then changed a further 5,370 bytes behind it. Neither was reviewable.

What survived in that blind spot was a **one-line disagreement between two adjacent functions** — `recordSeries` wrote keys with a `\u0000` separator, `getHostSeries` read them with a space — so demo-mode host graphs served empty series for every host and metric and had **never worked** since the feature shipped. Measured on `main` before the fix, via the real client: `polledAt=null` and `queued=0 avgProcessingTime=0 messagesPerSec=0` on all three hosts.

That defect class is exactly what a diff catches and prose does not.

## Why this needs a mechanism and not a note

The fix in #157 removed the only NUL in the repo (swept every tracked file — it was the only one). But nothing stops the next one, and the failure mode is **silent in the direction that matters**: the file keeps compiling, tests keep passing, CI stays green, and the only casualty is human review. Nobody gets an error. That is the same shape as #84 — a correctness property that depends on someone noticing.

It also has a second-order cost. `core.autocrlf` is `true` here and every sibling source file is stored with LF, but autocrlf deliberately skips files it believes are binary, so this one accumulated 616 CRLF line endings. Removing the NUL let normalisation run and produced a 616-line diff that had to be split into its own commit (`4fd83d2`) to keep the real 3-line change reviewable.

## Suggested fix

A CI step in the existing `CI` workflow — cheap, no new dependency, and it fails loudly:

```yaml
- name: No control characters in tracked text files
  run: |
    bad=$(git grep -lIP '\x00' -- . || true)
    if [ -n "$bad" ]; then
      echo "::error::NUL byte in tracked file(s) — git will treat these as binary and suppress their diffs:"
      echo "$bad"
      exit 1
    fi
```

[...]

PR #185, opened 2026-09-01
contracts: a tool for the interface map, and trend gains the field that says a queue is draining (#177)

tanifgit:
Contract half of #177, separate per root `CLAUDE.md` §4. **Two files, both additive** — `mcp-tools.md` gains one read tool (§3.14, `get_interface_path`), `investigation-api.md` §2.2 gains one nullable field on `trend`. No existing field changes type or meaning.

The topology source is @tanifgit's suggestion: **IRIS's own Interface Maps feature.** It turned out to be fully queryable.

## Why a tool at all

Every read tool in `mcp-tools.md` is scoped to **one host**, and so is the snapshot §2.2 sends. So AI Detective could describe the host it was asked about in detail and had no way to know another host existed.

Measured: an upstream fault was fixed, 281 accumulated messages flushed through at once, `Cloud API` queued 296, and the Detective recommended the **maximum** pool of 8 at 0.85 confidence — on a queue that drained to zero unaided while it answered. Its own evidence read *"No errors recorded in the last 60 minutes"*, true of `Cloud API` and false of the production: `EMR Source` had been erroring every poll for twenty minutes inside that window.

A host-scoped tool cannot be *wrong* about another host. It can only be silent, and a model reads silence as absence.

## Why IRIS's map rather than reading config

`Ens.InterfaceMaps.Utils` **resolves routing rules**, and that is the whole reason. On this production:

```
EMR Source -> Lab Router     a TargetConfigNames setting on the service
Lab Router -> Cloud API      a <send target="Cloud API"/> inside RoutingRule.cls
```

The second is invisible to anything that reads settings. The utility walks rules, DTLs and BPL to find it. **A topology hand-built from `TargetConfigNames` would have been missing exactly the edge this tool exists for, and would have looked correct.** It is also derived from the production *definition*, so it still answers when a host is dead — which is when an investigation runs.

§3.14 records the two surfaces of that utility I tried and rejected: `FindAllPaths` returns a control-character-delimited string (undocumented internal encoding, and #161 is open about control chars in source), and `FindSequentialPath` wants a JSON spec whose shape isn't documented — a path string returns `ERROR #5035 ... 'Parsing error'`.

## `trend.recentDirection`, and a deviation it works around

This is the part worth a careful read. §2.2 has **always** specified `slope` as *"may be zero or negative here … because a queue that is draining is a fact the agent should see"*, and its §4 example carries a non-null slope beside `thresholdCrossed: true`.

**The engine has never delivered that.** It reads `slope` from Early Warning's `projection`, which is `null` for `already_crossed` — i.e. for *every* condition this endpoint is called about. So the one fact needed to tell a recovery from a runaway was specified, documented, and absent, and the measured consequence was a recommendation to enlarge a pool on a queue falling 261 → 181.

`recentDirection` (from `earlywarning-api.md` §1.5) is the half fixable without reopening that contract's §1.4 prohibition on a bare slope beside a withheld forecast: **a direction is not a rate**, so it carries no forecast to mislabel — the same argument that lets `kind: 'projection'` stay absent.

**The `slope` deviation is not fixed here** and is filed separately: honouring it needs the window slope published or refitted, which is `earlywarning-api.md`'s decision rather than this one's.

## Stated rather than left to be discovered

- **Cost: 33.7 ms measured**, on three hosts and one path. It walks rules and transforms, so that says nothing about a fifty-host production.
- **The `Processes` separator for a chain of 2+ processes is UNVERIFIED.** LABDEMO has exactly one business process, and the utility's row-building source is not shipped. The implementation splits defensively; getting it wrong degrades to one unsplit name in `upstream`, not a wrong answer about who is upstream of whom.
- **Truncation is reported** (`truncated`, `pathCount` vs `pathsReturned`), because the opposite choice is an open defect in this same contract (#165).

## Consumer impact: none required

Both changes are additive. A consumer already handling a `null` `slope` — which is what has always arrived — needs no change.

`npm run validate` — all checks passed (3 samples, 15 accept, 26 reject, 7 capture claims).

Per §4 this needs review by every other developer. **Not merging this without a nod** — yesterday's self-review arrangement was explicitly for yesterday. Implementation follows in a paired PR.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -29,10 +29,19 @@ as a `<send>` inside a routing rule class. A topology hand-built from settings w
 exactly the edge the tool exists for, and would have looked correct. It is also derived from the
 production DEFINITION, so it still answers when a host is dead — which is when an investigation runs.
 
-Two surfaces of that utility were tried and rejected, and §3.14 records why: `FindAllPaths` returns a
-control-character-delimited string (an undocumented internal encoding, and #161 is open about raw
-control characters in source), and `FindSequentialPath` wants a JSON spec whose shape is not documented
-— passing it a path string returns `ERROR #5035 ... 'Parsing error'`.
+Two surfaces of that utility were tried and rejected, and §3.14 records why. `FindAllPaths` is an
+**internal method** — its own dictionary description opens with that word — returning a `%Status` with
+the paths handed back **byref** as `$listbuild` lists, `$lb(Service,Processes,Rules,DTLs,Operations)`,
+per that same description. `FindSequentialPath` wants a JSON spec whose shape is not documented
+anywhere reachable; passing it a path string returns `ERROR #5035 ... 'Parsing error'`.
+
+**An earlier draft of this entry said `FindAllPaths` returned a control-character-delimited string in
+an undocumented encoding. That was wrong** (@Ari-Glikman, #185): the `$c(12)`, `$c(11)` and `$c(1)`
+bytes measured at offsets in the value are `$list` length and type headers, not delimiters, and
+`$listget` reads it in one line — verified, `$listvalid` is 1 with five elements. Corrected here rather
+than quietly dropped, because a wrong-but-plausible rationale is the kind a later reader trusts instead
+of re-checking, and it would have told them the richer surface was undecodable when it is documented.
+"Internal method" is the firmer reason and does not depend on decoding anything.
 
 ### `trend.recentDirection`, and the deviation it works around
 
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,71 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-09-01 — a tool for the interface map, and `trend` gains the field that says a queue is draining
+
+**Two files, both additive.** `mcp-tools.md` gains one read tool (§3.14, `get_interface_path`);
+`investigation-api.md` §2.2 gains one nullable field on `trend` (`recentDirection`). No existing field
+changes type or meaning, nothing is removed, and a consumer that ignores both behaves as before.
+
+#177, and the topology source was @tanifgit's suggestion: IRIS's own Interface Maps feature.
+
+### Why a tool at all
+
+Every read tool in `mcp-tools.md` is scoped to ONE host, and so is the snapshot §2.2 sends. So AI
+Detective could describe the host it was asked about in detail and had no way to know another host
+existed. Measured: an upstream fault was fixed, 281 accumulated messages flushed through at once,
+`Cloud API` queued 296, and the Detective recommended the MAXIMUM pool of 8 at 0.85 confidence on a
+queue that drained to zero unaided while it answered — its own evidence reading "No errors recorded in
+the last 60 minutes", true of `Cloud API` and false of the production.
+
+### Why IRIS's map rather than reading config
+
+`Ens.InterfaceMaps.Utils` resolves ROUTING RULES, and that is the whole reason. On this production
+`EMR Source -> Lab Router` is a `TargetConfigNames` setting, and `Lab Router -> Cloud API` exists only
+as a `<send>` inside a routing rule class. A topology hand-built from settings would have been missing
+exactly the edge the tool exists for, and would have looked correct. It is also derived from the
+production DEFINITION, so it still answers when a host is dead — which is when an investigation runs.
+
+Two surfaces of that utility were tried and rejected, and §3.14 records why: `FindAllPaths` returns a
+control-character-delimited string (an undocumented internal encoding, and #161 is open about raw
+control characters in source), and `FindSequentialPath` wants a JSON spec whose shape is not documented
+— passing it a path string returns `ERROR #5035 ... 'Parsing error'`.
+
+### `trend.recentDirection`, and the deviation it works around
+
+§2.2 has always specified `slope` as "may be zero or negative here ... because a queue that is draining
+is a fact the agent should see", and its §4 example carries a non-null slope beside
+`thresholdCrossed: true`. **The engine has never delivered that.** It reads `slope` from Early Warning's
+`projection`, which is `null` for `already_crossed` — i.e. for every condition this endpoint is called
+about. So the one fact needed to tell a recovery from a runaway was specified, documented, and absent.
+
+`recentDirection` (from `earlywarning-api.md` §1.5) is the half fixable without reopening that
+contract's §1.4 prohibition on publishing a bare slope beside a withheld forecast: a direction is not a
+rate, so it carries no forecast to mislabel. **The `slope` deviation itself is NOT fixed here** and is
+filed separately — honouring it needs the window slope published or refitted, which is
+`earlywarning-api.md`'s decision rather than this one's.
+
+### Verified end to end, live agent
+
+Same armed scenario, before and after:
+
+| | Before | After |
+|---|---|---|
+| tool calls | 4 | 6 |
+| upstream hosts read | none | both, via `get_recent_errors` |
+| recommendation | `set_pool_size` -> 6 (conf 0.9) | **null** |
+| root cause | "pool size 4 is insufficient" | "`EMR Source` errors #5021, 229 s ago, since ended — backlog draining through" |
+
+And the ordinary rising case still recommends a pool increase (`1 -> 2`, conf 0.9), which is #108's
+standing check and the thing a directive about transients could most easily have broken.
+
+### Consumer impact
+
+**None required.** Both changes are additive. A consumer already handling a `null` `slope` — which is
+what has always arrived — needs no change; `recentDirection` is the field to read instead.
+
+---
+
 ## 2026-08-31 — Early Warning publishes `recentDirection`, so a recovering queue stops reading like a runaway one
 
 **`earlywarning-api.md` only.** Additive: one nullable field on `HostProjection`, plus a
--- a/contracts/investigation-api.md
+++ b/contracts/investigation-api.md
@@ -125,6 +125,7 @@ The engine builds this from its own in-memory state. It is a closed allowlist:
     "metric": "queued",
     "slope": 124.2,
     "slopeUnit": "items/minute",
+    "recentDirection": "rising",
     "thresholdValue": 50,
     "thresholdCrossed": true,
     "secondsToThreshold": null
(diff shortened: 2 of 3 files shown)


Ari-Glikman (changes requested):
Reviewed as the other developer per §4. **The design is right and the central justification verified exactly** — I checked every claim I could reach rather than reading the body. Two text fixes before merge, both cheap, neither touching the shape.

## Verified against the live instance and the repo

- **The routing-rule asymmetry is exactly as stated.** `TargetConfigNames` appears once, on the service (`iris/labdemo/Production.cls:67`), and `Lab Router -> Cloud API` exists only as `<send transform="...HL7ToPID" target="Cloud API"/>` at `iris/labdemo/RoutingRule.cls:23`. A topology built from settings really would have missed that edge and really would have looked correct. This is the strongest argument in the PR and it holds.
- **`EnumeratePaths` matches §3.14's documented shape.** Live: 5 columns — `Service`, `Processes`, `Rules`, `Transforms`, `Operation` — and one row, `EMR Source | Lab Router | ...RoutingRule | ...Transform.HL7ToPID | Cloud API`. Byte-identical to the §3.14 example, `pathCount: 1` included.
- **Cost corroborated:** 30.6 ms here against your 33.7 ms.
- **The `slope` deviation is real, and worse than the PR says.** `investigate.ts:181` reads `projection.projection?.slope` while that same function's comment (lines 169-175) *explains* that Early Warning sets `projection` to `null` for `already_crossed`. The file documents the reason its own value is always null. Filing it separately is right; #187 covers it.
- **Adopting `recentDirection` here is consistent with the sibling contract, not novel** — `earlywarning-api.md:298` already tells an `already_crossed` consumer to read it.
- **The #165 citation is fair.** I wrote §3.13 and it genuinely never mentions the 25 cap. No objection to being quoted as the counterexample.
- #186 does populate the field, so this is not documenting a phantom.

## 1. §3.14 is wrong about `FindAllPaths`, and the true reason is better

The contract says it "returns a control-character-delimited string (an undocumented internal encoding)". Both halves are wrong. Live signature:

```
FindAllPaths(*pPaths, *pCategoryItems, pProdName="", pCategory="", pTempPortalNode="", pDeDup=1) -> %Library.Status
```

It returns a `%Status` and outputs **by reference** into `pPaths`. Calling it with `"Cloud API"` as the first argument passes a host name into the output slot, which is why a bare `1` comes back. And the structure is documented — in the method's own dictionary description:

> Internal method for finding all of the routes through the active production
> pPaths is an array of lists with each list describing one path through the production $lb(Service,Processes,Rules,DTLs,Operations)

So it is a standard `$listbuild` list, read with `$listget`. Raw `$list` bytes written to a terminal look precisely like an undocumented control-character encoding, which I assume is what happened.

**Avoiding it is still correct** — the description opens with "Internal method", which is a firmer reason than the one recorded and does not depend on decoding anything. But as written the note tells a future developer that the richer surface (it also yields `pCategoryItems` and a `pDeDup` flag) is undecodable, when it is a `$list` with its layout in the class dictionary. That is the kind of wrong-but-plausible rationale someone trusts instead of re-checking, and the CHANGELOG repeats it, so it lands twice.

Please reword to "an internal method (its own description says so), returning a `%Status` with paths byref as `$lb(...)` lists" — or drop the encoding claim entirely.

Not disputing the `FindSequentialPath` rejection: `(pSpec:%String, pProduction="", *pSC:%Status) -> %String` is consistent with a spec string being parsed, and #5035 on a path string fits.

## 2. `recentDirection` arrives here without the two caveats that make it safe

`investigation-api.md` §2.2's new row calls it "the **sign** of the tail fit, measured" and "**the field that says a queue is draining**". `earlywarning-api.md` §1.5 attaches two conditions to that field which did not come across:

- **it lags by design** — measured there at ~35 seconds and 46 messages of real drain before the sign flipped
- **"a consumer must not build a 'recovered' claim on it — only a 'coming down' one"**

This contract's consumer is a model, and it is the one already measured recommending the maximum pool at 0.85 confidence off an incomplete picture. Both caveats bite:

- for the first ~35 s of a genuine drain the agent reads `rising`, so #177 is narrowed from 110 s to ~35 s rather than closed — worth stating so the next person measuring a stale recommendation does not think this change should have prevented it
- `falling` on a queue still 200 deep and above threshold is "coming down", not "recovered", and nothing in §2.2 stops the agent concluding no action is needed

Everything else in this PR is scrupulous about exactly this failure mode — "WIRING, NOT TRAFFIC", "`known: false` is not an all-clear" — so the gap reads as an oversight against your own standard. Two sentences carried over from §1.5 would close it.

## 3. Two notes, no change requested

**`known` deserves a word on how the match is made.** `EnumeratePaths`'s real signature is a term search, not a host lookup:

```
EnumeratePaths(pSearchTerm="", pSearchIn="{""svc"":1,""op"":1,""proc"":1,""rule"":1,""xform"":1}", pCaseSensitive=0, pProdName="", ...)
```

It searches services, operations, processes, **rules and transforms**. So unless #186 filters rows to exact config-item equality after the search, `known: true` can be reached by a host name that is a substring of a rule or DTL class name — and §3.14 defines `known: false` as "appears in no path" without saying what "appears" means. I will check the filtering on #186; if it is exact, consider saying so in §3.14, since `known` is load-bearing for a not-an-all-clear claim.

**`npm run validate` passing does not exercise either change.** Nothing in `contracts/` mentions `trend` — no schema covers the §2.2 request object, and `earlywarning-api.md` §7 already records that nothing validates that endpoint. You did not overclaim it; flagging so the green tick is not later read as coverage of this diff.

---

Fix 1 and 2 and this is an approve from me — the shape, the source choice and the additive-only framing are all right, and the UNVERIFIED separator note plus the reported truncation are the standard I would want held. Ping me and I will re-read the same day.

tanifgit:
Both fixed in `8fbcce1`. Thank you for verifying rather than reading — point 1 was a real error and I would not have caught it.

## 1. `FindAllPaths` — you are right, and I reproduced it before rewriting

```
SC=1
VALID=1 len=5
EL 1 = EMR Source
EL 2 = Lab Router
EL 3 = ProductionGuardian.LabDemo.RoutingRule
EL 4 = ProductionGuardian.LabDemo.Transform.HL7ToPID
EL 5 = Cloud API
```

`$listvalid` is 1, five elements, and the description confirms it: *"Internal method for finding all of the routes through the active production"*. The `$c(12)`/`$c(11)`/`$c(1)` bytes I measured at offsets are `$list` length and type headers. Your diagnosis of what happened is exactly what happened — I `$ascii`'d a `$list` and read the headers as delimiters.

Reworded to "internal method (its own description says so), `%Status` with paths byref as `$lb(...)`", and **kept the correction visible** rather than quietly swapping it, for the reason you gave: the wrong version told a reader the richer surface was undecodable when its layout is in the class dictionary, and that is precisely the kind of rationale someone trusts instead of re-checking.

**One locator correction:** the claim was never in §3.14 — it lived in the CHANGELOG entry and in `Topology.cls`'s class comment. Both are fixed. §3.14 had no note on the surface choice at all, which is a gap given you plainly expected one there, so it now carries an accurate one — and it says the honest thing rather than the flattering one: **all three surfaces are internal, and the query is merely the least so.**

## 2. The two §1.5 caveats — carried over, and you are right that they bite harder here

Added to §2.2 as their own block, both with the measurements:

- **it lags ~35 s / 46 messages of real drain**, and — your point, which I had not stated anywhere — **#177 is narrowed from ~110 s to ~35 s, not closed.** That belongs in the record precisely so the next person measuring a stale recommendation in that window does not read it as this change having failed.
- **`falling` is "coming down", never "recovered"**, with the note that §2.2 already carries depth and threshold for that judgement, so a producer must weigh the direction *against* them rather than instead of them.

You are also right about why it reads as an oversight against my own standard: I was careful about exactly this failure mode for "WIRING, NOT TRAFFIC" and `known: false`, and then imported a field without importing its conditions.

## 3. Both notes answered

**`known` is exact-match, and §3.14 now says so.** You are right that it needed defining — `pSearchTerm` searches rules and transforms too, so "appears in a path" was doing unexamined work. The implementation passes it empty and filters on `hops.%Get(i) = host`, so §3.14 now states the match is on the config item name and that a host name which is a substring of a DTL class cannot reach `known: true`.

**`npm run validate` coverage** — agreed and worth having on the record. Nothing in `contracts/` covers `trend` or this endpoint, so the green tick means the three existing samples still validate and nothing more. `earlywarning-api.md` §7 already says this for its own endpoint; it is now true of two.

---

Ready for your re-read. #186 also has its fixes pushed (`e17cd99`) — including your point 1 there, which was a genuine bug and is the more important of the two sets. **Landing order as you said: this one first.**

I am handing both back to you rather than merging — the self-review arrangement was for the day you were out, and you are back.

Ari-Glikman (approved):
Re-reviewed `8fbcce1b`. All three points are addressed, and the first one is addressed better than I asked for. **Approving with one word to fold in before you merge** — see the end.

## The `FindAllPaths` correction is right, and correcting in place was the right call

I verified this independently before writing the original comment down as a finding, and your reading matches mine exactly: `$listvalid` is 1, five elements, `$listget` reads each, and the dictionary `Description` states the layout. So `$c(12)`/`$c(11)`/`$c(1)` are length and type headers.

Keeping the wrong rationale visible with the correction attached is the more useful choice than deleting it, for the reason you give: the failure mode of a wrong-but-plausible rationale is that a later reader trusts it instead of re-checking, and this one would have told them the richer surface was undecodable. "Internal method" is also the firmer reason because it does not depend on decoding anything — the rejection now survives IRIS changing the encoding.

One thing worth noting for whoever reads §3.14 next: you kept `SplitList`'s caveat and said why the correction does not settle it — element 2 is the plain string `Lab Router`, not a nested list. I checked the same thing and got the same answer. That is the honest place to leave it.

## The two §1.5 conditions match §1.5's own measurements

Checked against `earlywarning-api.md` §1.5 rather than against the prose here:

- **~35 s and 46 messages** — verbatim what §1.5 records for the lag ("reported `rising` for a further ~35 seconds and 46 messages of real drain before flipping").
- **~110 s → ~35 s narrowing** — consistent with the two measurements §1.5 carries (22 polls / 110 s of indistinguishable `already_crossed`, versus ~35 s of lag). Saying **#177 is narrowed, not closed** is the accurate claim and I would have pushed back if this had said closed.
- **`falling` is never "recovered"** — §1.5's wording is "a consumer must not build a 'recovered' claim on it — only a 'coming down' one". Same claim, and pointing at §2.2 for the depth and threshold gives a producer somewhere to go instead of just a prohibition.

Restating both here rather than cross-referencing is correct for the reason you give: this contract's consumer has no reason to open the sibling. Given the model already measured recommending the maximum pool at 0.85 confidence off an incomplete picture, a caveat one file away is a caveat that will not be read.

## §3.14's `known` answer is accurate against the implementation

I checked this against `Topology.cls` at #186's head rather than taking the contract's word for it: `pSearchTerm` is passed empty, `hops` is built from the `Service`, `Processes` and `Operation` columns only, and the match is `hops.%Get(i) = host` — exact equality. So a host name that is a substring of a rule or DTL class name cannot reach `known: true`, exactly as §3.14 now says. Rules and transforms are returned as path *content* but never matched against.

## One word before merge

> `pathsReturned` is what came back, `pathCount` is the true total, and `truncated` says whether they differ.

**"the true total" needs to say "for this host".** This is not pedantry — the ambiguity in that sentence is what #186's implementation read as license, and it shipped `pathCount` counting production-wide rows while `pathsReturned` counted the host's. #186 has now fixed the code to be host-scoped and its commit message says this is "the reading §3.14 already defines", but §3.14 does not actually say so; it says "the true total", which is exactly as readable as production-wide. The sample's `pathCount: 1` cannot disambiguate it either, because LABDEMO collapses both readings to 1.

Suggested: *"`pathCount` is the true total **of paths this host sits on** — including any the cap withheld"*.

Everything else is verified and I have no other changes. Fold that in and I will merge, or say the word and I will push it myself to save you the round trip.

tanifgit:
Folded in as `84bb6b5`. Ready to merge from my side.

Took your wording, and added **one sentence rather than only the clause**, because the clause alone leaves the sample able to contradict it:

> `pathCount` is the true total **of paths this host sits on** — including any the cap withheld — and `truncated` says whether they differ.
>
> **Both counts are scoped to the host asked about**, never to the production. Spelled out because "the true total" alone reads just as naturally as production-wide, and that is the reading the first implementation shipped (@Ari-Glikman, #186); LABDEMO cannot catch it, because one path collapses both readings to `1`.

The second sentence is there for the reason you gave for asking: the failure mode is a reader who takes the sentence at face value, and `pathCount: 1` in the §3.14 sample is consistent with *both* meanings. Naming that the sample cannot disambiguate it is what stops the next reader resolving the ambiguity against the example.

`npm run validate` — all checks passed (3 samples, 15 accept, 26 reject, 7 capture claims). Unchanged from `8fbcce1`, and it still means only that the three existing samples validate: nothing in `contracts/` exercises this endpoint.

#186's `Note()` fix is pushed too, and it is **rebased onto `main` past #190** — details over there, including one open question from your sequencing note. **Landing order as agreed: this one first.**

Issue #175, opened 2026-08-31
The chat's own descriptions are two capabilities behind: the UI says activity only, and the canned "what can you do" omits findings and config changes


tanifgit:
Reported by @tanifgit:

> the chat ("ask") descriptions seem outdated re latest features. It says: *"Ask about activity — Ask about message volume, throughput and latency over time. An AI agent inside IRIS answers by reading the interoperability activity tables through governed, audited tools…"* But now the chat knows also to check about the Event log and the Findings for example.

Confirmed, and it is **two** capabilities behind in the UI and **two** in the agent's own answer — different two.

## What the chat can actually do

`iris/labdemo/Tools/Governance.cls:245` is the authoritative membership:

```
chat    ACTIVITYTOOLS  EVENTLOGTOOLS  CHANGELOGTOOLS  FINDINGTOOLS    ChatDispatcher
```

Four families: recorded activity, the event log, configuration changes, and active findings.

## What each description claims

| Where | Owner | Claims | Omits |
|---|---|---|---|
| `apps/dashboard/src/components/ActivityChat.tsx:137` intro | Dev B | activity only | event log, config changes, findings |
| heading "Ask about activity" (`ActivityChat.tsx`) | Dev B | activity only | same three |
| `SUGGESTIONS`, `ActivityChat.tsx:63-69` — all five prompts | Dev B | activity only | same three |
| `iris/labdemo/REST/ChatDispatcher.cls:603` canned capability answer | Dev A | activity + event log | config changes, findings |

The canned answer is the one a user is most likely to hit, because `'What can you do?'` is the **first** suggestion chip (`ActivityChat.tsx:64`).

## Measured, last session, live agent

Asked through `POST /api/chat` on the running stack:

```
"What can you do?"
  -> source: "static"   ... "recorded activity and its event log" ... (findings and config changes absent)

"Are there any active findings on the production right now, and on which hosts?"
  -> source: "agent", toolCalls: 1, confidence: 1
     "There are no active findings on the production right now."
```

So the findings capability **works and nothing advertises it.** The config-change capability is read from the family table above rather than measured — worth one curl when the stack is next up, but `CHANGELOGTOOLS` is registered on the same row.

## Why this is worth an issue rather than a copy tweak

[...]

Issue #171, opened 2026-08-31
get_recent_config_changes: misses a change made seconds ago and calls it "no changes found", and reports no-op saves as changes


tanifgit:
`get_recent_config_changes` has two independent defects that both mislead AI Detective: it misses a change made seconds earlier and reports it as "no changes found", and it reports no-op saves as real changes.

Reported by @tanifgit from three live investigations: the `MissingFolder` scenario (change missed), the `HTTPPort` scenario (worked), and `PoolBottleneck` (phantom changes reported).

## Defect 1 — a change made seconds ago is invisible, and the model states the negative as fact

Arming `MissingFolder()` audits the `FilePath` change correctly. The Investigate run 16 s later reported:

> "**No recent configuration changes were found** that could also contribute to this issue."

The tool was called — this is not the "registered but never called" failure of #164's lesson. From `Audit.Entry`:

```
10:00:22 | GetRecentConfigChanges | {"host":"EMR Source"} | executed
```

The audit row it should have found was written at **10:00:06**, sixteen seconds earlier, and is correct in every field:

```
Event      ModifyConfiguration   (%Ensemble / %Production)
Namespace  LABDEMO
Routine    AuditSettingChange+8^ProductionGuardian.LabDemo.Triggers.1
Desc       item EMR Source of ProductionGuardian.LabDemo.Production
EventData  FilePath:/tmp/labdemo/hl7-in/>>/tmp/labdemo/hl7-in-missing/
```

Neither `HostOf()` nor `ParseChange()` is at fault — calling the tool now returns that exact row correctly. **`%SYS.Audit` rows are not queryable for tens of seconds after the write.**

Measured, by arming the trigger and polling for the specific new timestamp:

```
armed 10:50:04
+1s … +32s   newest visible change = 10:49:22   (the previous row)
+36s         newest visible change = 10:50:04   <- appears
```

**36 seconds.** It is variable, not a fixed delay: the `HTTPPort` scenario that worked was read at +24 s. That is why this looks intermittent rather than broken, and why it will fail a demo at random.

This is the same buffering that produced three wrong readings on #166 — recorded there as ">15 s and <110 s". It was a measurement nuisance then; here it is a correctness bug, because a consumer treats "empty" as "nothing happened".

[...]

PR #186, opened 2026-09-01
feat: AI Detective can see what feeds the host it is investigating (#177)

### LINKED ISSUE #177: AI Detective recommends a bigger pool for a backlog that is already draining: no cross-host reasoning about an upstream bottleneck that was just fixed

tanifgit:
Closes #177. Contract went in separately as #185. Topology source suggested by @tanifgit — IRIS's own Interface Maps feature, which turned out to be fully queryable.

## The defect

Every read tool was scoped to **one host**, and so is the snapshot. Measured: an upstream `MissingFolder` fault was fixed, 281 accumulated messages flushed through at once, `Cloud API` queued 296, and the Detective recommended the **maximum** pool of 8 at 0.85 confidence — on a queue that drained to zero unaided while it answered. Its own evidence read *"No errors recorded in the last 60 minutes"*: true of `Cloud API`, false of the production.

## `Tools.Topology.GetInterfacePath`

Reads `Ens.InterfaceMaps.Utils`, which **resolves routing rules** — the whole reason to use it rather than reading config:

```
EMR Source -> Lab Router     a TargetConfigNames setting
Lab Router -> Cloud API      a <send> inside RoutingRule.cls   <- invisible to settings
```

A hand-built topology would have been missing exactly the edge this exists for, and would have looked correct. Derived from the production *definition*, so it still answers when a host is dead — which is when an investigation runs.

Registered for `"all"` only, the mirror of `FINDINGTOOLS`. **Not given to `"chat"` yet**, and not for lack of use: the chat's capability text and chips enumerate what it can read (#175), so a fifth family there is a copy change in two more places and belongs with that work rather than smuggled in here.

## `trend.recentDirection` — without it the directive pointed at a field that is always null

§2.2 specifies `slope` as *"may be zero or negative here"*, and the engine reads it from Early Warning's `projection` — `null` for `already_crossed`, i.e. every condition this endpoint is called about. So the draining fact the contract promises had **never reached the agent**. The `slope` deviation itself is not fixed here; filed separately.

## Two prompt corrections, both measured rather than reasoned

- **The upstream read is its own `MUST`.** As one clause of a longer sentence, the model called `GetInterfacePath`, correctly named both upstream hosts, and then **never looked at either** — exactly what `iris/CLAUDE.md` §7 predicts about descriptions versus directives.
- **Each upstream host, not the nearest.** With "the nearest" it read `Lab Router` (genuinely clean) and reached the right verdict by the weaker route of "no fault anywhere", never seeing that `EMR Source` two hops up had been in Error for twenty minutes and was fixed three minutes earlier. Same answer, wrong reason.

**Positive evidence only** — `DropPoolFixDuringOutage`'s existing rule applied to the model. A directive about transients must not make the Detective timid about the ordinary sustained rise.

## Verified live, same armed scenario, before and after

| | Before | After |
|---|---|---|
| tool calls | 4 | 6 |
| upstream hosts read | none | both |
| recommendation | `set_pool_size` → 6, conf 0.9 | **`null`** |
| root cause | "pool size 4 is insufficient" | "`EMR Source` … error #5021 … occurred 229 seconds ago and has since ended, the backlog is likely draining through" |
| manual remediation | none | names the real fix (the directory) |

**#108 regression check, same build:** the ordinary rising queue still recommends `1 → 2` at conf 0.9 and still calls the map. That was the thing this change could most easily have broken.

| Check | Result |
|---|---|
| `npm test` (engine) | 415/415 |
| `tsc --noEmit` | clean |
| Boot tool-count guard | **15 discovered, expected 15**, names read back per §7 — `Tools.Topology` exposes exactly one tool, both helpers private |

## Two follow-ups I am filing rather than folding in

- **The sizing input.** `messagesPerSec` is inflated by a drain-through burst, so the Little's-law block is fed a bad rate. You chose to keep this separate, and it stays separate.
- **A `#171`-adjacent observation.** In both of today's runs the model wrote "No recent configuration changes found" in an evidence entry — the claim `BuildGoal` explicitly forbids. The tool behaved correctly (the change was on another host). It may be displacement from the ~25 lines this PR adds nearby, which is a mechanism this very file documents. I have no clean before/after run to attribute it, so I am filing it with the evidence rather than guessing.

**Not merging without a nod** — yesterday's self-review arrangement was explicitly for yesterday.

### DIFF
--- a/iris/labdemo/REST/AgentDispatcher.cls
+++ b/iris/labdemo/REST/AgentDispatcher.cls
@@ -746,6 +746,34 @@ ClassMethod SystemPrompt() As %String [ Private ]
     set p = p _ "errors. A host with a queue, a normal processing time and only stale errors is the "
     set p = p _ "pool case, not the connectivity case. "
 
+    // THE INTERFACE MAP, AND IT IS THE FIRST TOOL HERE THAT IS NOT ABOUT ONE HOST (#177).
+    //
+    // Named in the form the runtime accepts -- `GetInterfacePath` -- for the reason recorded above the
+    // activity tools: `%AI.ToolMgr` derives the tool name from the ClassMethod name, and the
+    // snake_case spelling in the contract's section titles is not callable. Naming a tool the model
+    // cannot call fails silently, looking like a model that chose not to call anything.
+    //
+    // WHAT IT FIXES. Every other read tool is host-scoped, and so is the snapshot, so the model had no
+    // way to know another host existed. Measured: an upstream `MissingFolder` fault was fixed, 281
+    // accumulated messages flushed through at once, `Cloud API` queued 296, and the Detective
+    // recommended pool 8 -- the maximum -- at 0.85 confidence on a queue that drained to zero unaided
+    // while it answered. Its own evidence read "No errors recorded in the last 60 minutes", true of
+    // `Cloud API` and false of the production: `EMR Source` had been erroring every poll for twenty
+    // minutes inside that window.
+    //
+    // THE WIRING/TRAFFIC DISTINCTION IS SPELLED OUT because it is the available misreading. The map is
+    // built from configuration, so "upstream" means "is configured to feed this host", not "is sending
+    // it anything now" -- and a model that conflated the two would report an idle upstream host as a
+    // cause.
+    set p = p _ "GetInterfacePath returns this production's interface map for one host: which hosts "
+    set p = p _ "feed it (upstream) and which it feeds (downstream), NEAREST FIRST, with the routing "
+    set p = p _ "rules and transformations on each path. It is derived from the production's "
+    set p = p _ "configuration, so it says how the production is WIRED and not what is flowing now: an "
+    set p = p _ "upstream host is one configured to feed this one, which is not evidence that it sent "
+    set p = p _ "anything. Use it whenever a finding could have been caused somewhere else, and use the "
+    set p = p _ "host names it returns as the argument to the other read tools -- that is how you look "
+    set p = p _ "at a host other than the one under investigation. "
+
     // SIZE THE POOL, DO NOT DOUBLE IT. The model reliably suggested `current + 1` -- 1 -> 2 -- which
     // on the demo scenario is EXACTLY break-even and therefore fixes nothing. Measured live at pool 2
     // with 2 msg/sec arriving and 1.01s of work each: queue 112, 112, 114 over 50 seconds. It was not
@@ -1022,6 +1050,85 @@ ClassMethod BuildGoal(body As %DynamicObject) As %String [ Private ]
     // that sees a non-zero number next to an empty list must not narrate it as one.
     set g = g _ " noOpSaves counts saves where the value did not move. It is NOT a change: never"
     set g = g _ " report it as one, and never treat a non-zero count as evidence of anything."
+
+    // A FOURTH MANDATORY CALL, AND IT IS MANDATORY FOR THE REASON THE OTHER THREE ARE (#177).
+    //
+    // `iris/CLAUDE.md` §7: register it, describe it, and name it in a MUST -- and the third is the one
+    // that keeps being skipped. Measured twice already, on `get_recent_errors` and then again on
+    // `GetRecentConfigChanges`, which was registered with its prompt paragraph verified present in the
+    // compiled routine and still went uncalled because the model had a satisfying answer after two
+    // calls and stopped.
+    //
+    // UNCONDITIONAL, rather than gated on the finding type or on the model's judgement -- the same
+    // argument the config-change directive above makes, arriving with more force. "Might this have been
+    // caused elsewhere?" is a judgement about the DIAGNOSIS, and the model cannot make it before
+    // gathering evidence, which is the circularity that leaves a tool uncalled. One bounded query,
+    // measured at 33.7 ms, against the one time it mattered and was skipped.
+    set g = g _ " You MUST also call GetInterfacePath on this host before concluding, to find out what"
+    set g = g _ " feeds it. Every other tool here answers about ONE host, so without this you cannot"
+    set g = g _ " tell whether the cause is upstream."
+
+    // AND THE REASONING IT IS FOR, which is the whole point of the tool and is stated as a QUESTION TO
+    // ANSWER rather than as a conclusion to reach. The failure being designed out is confident sizing
+    // arithmetic on a backlog that is already draining: measured at pool 4, queue 296 falling to 0
+    // unaided, and a recommendation of 8 at 0.85 confidence.
+    //
+    // POSITIVE EVIDENCE ONLY, and this is load-bearing. `DropPoolFixDuringOutage` in this class already
+    // establishes the rule -- it "only ever fires on POSITIVE evidence, never on absent evidence" --
+    // and the same discipline has to bind the model here, or a directive about transients becomes a
+    // reason to hedge on the ordinary sustained rise. That rise is the flagship scenario and #108's
+    // standing acceptance check; making the Detective timid about it would be a worse defect than the
+    // one being fixed.
+    //
+    // THE THREE SIGNALS ARE NAMED because "consider whether it is transient" is a description, and
+    // descriptions do not change behaviour. Each is already obtainable: the trend is in this goal, the
+    // upstream host names come from the tool above, and `secondsAgo` is already in `get_recent_errors`.
+    // AND THEN READ THE UPSTREAM HOST, AS ITS OWN MUST. This was one clause of a longer sentence in
+    // the first version and the model did exactly what §7 predicts: it called GetInterfacePath,
+    // correctly reported that `Lab Router` and `EMR Source` feed `Cloud API`, and then never looked at
+    // either of them. Measured on the armed transient -- `toolCalls: 4`, upstream named in evidence,
+    // and a recommendation to raise the pool 4 -> 6 on a queue falling 261 -> 181.
+    //
+    // Knowing the name and reading the host are different acts, and only the second is a MUST here.
+    set g = g _ " Having the upstream names is not enough: you MUST then call get_recent_errors with"
+    set g = g _ " sinceMinutes 60 on EACH host in the upstream list, nearest first, and read secondsAgo"
+    set g = g _ " on what comes back. An error on THIS host and an error on the host feeding it are"
+    set g = g _ " different faults with different fixes, and 'no errors' about this host says nothing"
+    set g = g _ " about the production."
+    // EACH, NOT THE NEAREST, and this is a measured correction rather than a thoroughness reflex. With
+    // "the nearest upstream host" the model read `Lab Router` -- which was genuinely clean -- and
+    // reached the right verdict by the weaker route of "no ongoing fault anywhere", never seeing that
+    // `EMR Source` two hops up had been in Error for twenty minutes and was fixed three minutes
+    // earlier. Same conclusion, but the diagnosis an operator needs is the second one, and on a
+    // scenario where the nearest host DID have a stale error the weaker route would have misread it.
+    //
+    // Bounded by `#MAXITERATIONS` (8) rather than by a cap here: the measured run used 5 of 8, so a
+    // second upstream read fits, and a chain long enough to exhaust the budget stops on its own.
+    set g = g _ " A fault two hops upstream is still the cause of your queue."
+
+    set g = g _ " Then decide whether this host is the CAUSE or the VICTIM, using evidence and never"
+    set g = g _ " assumption. A queue can build on a host that is keeping up perfectly, because"
+    set g = g _ " something upstream stopped and was then fixed, releasing a backlog that is now"
+    set g = g _ " draining THROUGH this host. Three things distinguish that from a genuine sustained"
+    set g = g _ " rise, and you must check all three: trend.recentDirection, where ""falling"" means"
+    set g = g _ " the queue is already coming down on the pool it already has; whether the upstream"
+    set g = g _ " host you just read has errors that have ENDED, which secondsAgo tells you; and"
+    set g = g _ " whether the pool was already enlarged shortly before now, which"
+    set g = g _ " GetRecentConfigChanges shows."
+    set g = g _ " If those say the backlog is draining through after an upstream fault that has ended,"
+    set g = g _ " say so in rootCause, name the upstream host, and set recommendedAction to null --"
+    set g = g _ " enlarging a pool that is already clearing its backlog fixes nothing and is a"
+    set g = g _ " configuration change nobody needed. Only conclude this on POSITIVE evidence: a queue"
+    set g = g _ " that is rising, or an upstream host with no ended fault, is the ordinary case and"
+    set g = g _ " still wants the pool sized properly. Absence of evidence is not evidence of a"
+    set g = g _ " transient."
+    // THE LAG IS STATED, because the field it points at has one. `earlywarning-api.md` §1.5 measured
+    // the tail fit taking ~35s to turn after a real peak, so a burst that has just landed still reads
+    // "rising" while it drains. A model told only "falling means draining" would read the inverse as
+    // "rising means growing" and reach the confident wrong answer faster.
+    set g = g _ " trend.recentDirection is fitted over the recent past and takes tens of seconds to"
+    set g = g _ " turn, so ""rising"" shortly after a burst arrived does not prove the queue is still"
+    set g = g _ " growing. Weigh it with the upstream errors rather than on its own."
 
     // ATTRIBUTION, RESTATED HERE, and it is a repair rather than a belt-and-braces addition. The
     // directive above cost the reply its evidence attribution: the first version closed with "report
--- a/iris/labdemo/Tools/Governance.cls
+++ b/iris/labdemo/Tools/Governance.cls
@@ -115,6 +115,27 @@ Parameter CHANGELOGTOOLS = "ProductionGuardian.LabDemo.Tools.ChangeLog";
 /// is set rather than after it is read.
 Parameter FINDINGTOOLS = "ProductionGuardian.LabDemo.Tools.Findings";
 
+/// The interface-map tool — which host feeds which, from the production's own configuration.
+///
+/// REGISTERED FOR `"all"` ONLY, which makes it the mirror image of `FINDINGTOOLS` above. AI Detective
+/// is the caller that needs it: every other read tool in this family is scoped to ONE host, and so is
+/// the snapshot `/api/investigate` sends, so before this the agent could not tell that any other host
+/// existed. Measured on #177 -- an upstream fault was fixed, the flushed backlog queued on
+/// `Cloud API`, and the Detective recommended doubling a pool on a queue that was draining unaided,
+/// having never looked at the host that caused it.
+///
+/// NOT GIVEN TO `"chat"` YET, deliberately and not for lack of use: "what feeds Cloud API" is a fair
+/// chat question. It is held back because the chat's capability text and its suggestion chips
+/// enumerate what it can read (#175), so a fifth family there is a copy change in two more places and
+/// belongs with that work rather than smuggled in with a Detective fix.
+///
+/// ITS OWN CLASS, on `Tools.EventLog`'s and `Tools.ChangeLog`'s reason: a new tool FAMILY in a new
+/// file moves `Setup.AIHub.ReportTools()`'s count by an amount attributable to one file. The stronger
+/// reason here is subject matter -- this is the only tool that describes RELATIONSHIPS between hosts
+/// rather than the state of one, and putting it on `Tools.Read` would leave a model one docstring away
+/// from reading "upstream" as a live traffic claim.
+Parameter TOPOLOGYTOOLS = "ProductionGuardian.LabDemo.Tools.Topology";
+
 /// The write-tool class, in its own class so the RBAC requirement does not leak onto the reads.
 Parameter WRITETOOLS = "ProductionGuardian.LabDemo.Tools.Resolve";
 
@@ -162,6 +183,12 @@ ClassMethod ToolManager(pIncludeWrite As %Boolean = 1, Output pStatus As %Status
         quit $$$NULLOREF
     }
     set pStatus = mgr.RegisterToolSet(..#CHANGELOGTOOLS)
+    if $$$ISERR(pStatus) {
+        quit $$$NULLOREF
+    }
+    // Context-free, unlike `FINDINGTOOLS`: the interface map is read from the production definition, so
+    // it answers the same question through this manager as through an agent's.
+    set pStatus = mgr.RegisterToolSet(..#TOPOLOGYTOOLS)
     if $$$ISERR(pStatus) {
         quit $$$NULLOREF
     }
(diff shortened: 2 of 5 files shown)


Ari-Glikman (changes requested):
Reviewed alongside #185. **The prompt wiring is the best in the repo so far** and I want to say that before the defects: you applied `iris/CLAUDE.md` §7 completely, and then caught the sub-failure I had not anticipated — that naming the upstream host and reading it are different acts, so the second needs its own MUST. The measured evidence for `EACH` over `the nearest` is the kind of thing that makes the next tool cheaper.

One real defect, one narrower one, and a correction that lands here as well as on #185.

## 1. `pathCount` counts the whole production, not this host's paths

`total = total + 1` runs at the top of the row loop, above the host filter:

```objectscript
while rs.%Next() {
    set total = total + 1               // every row in the production
    ...
    if host '= "" {
        ...
        if at < 0 { continue }          // rows not involving `host` are skipped AFTER counting
```

So with a `host` argument `pathCount` is production-wide while `pathsReturned` is host-scoped. §3.14 defines them against each other — *"`pathsReturned` is what came back, `pathCount` is the true total, and `truncated` says whether they differ"* — so on a production with 60 paths where the host sits on 3, the payload is `pathCount: 60, pathsReturned: 3, truncated: false`. The only honest reading available to a model is "57 paths were withheld from me", which is the small-production-versus-clipped-list confusion of #165 that this design explicitly cites, reintroduced in a subtler form. Your class header makes the contract authoritative, so this is a bug by your own rule.

Cheapest demonstration on LABDEMO as it stands today, no fixture needed — any host not in the map:

```
GetInterfacePath("Ens.MonitorService")
  -> known: false, paths: [], pathsReturned: 0, pathCount: 1, truncated: false
```

**It is invisible in every test on this production** because LABDEMO has exactly one path and the demo host sits on it, so `total`, `pathsReturned` and the truthful count all collapse to 1. That is why CI and both live runs passed, and it is why this wants a fixture.

Fix is one line — move the increment below the `at < 0` continue. If you would rather keep the production-wide number, it needs a third field and a §3.14 change, i.e. back through #185.

## 2. `position` is inferred from hop index, and hop 0 is not always a service

```objectscript
set position = $case(at, 0: "service", (hops.%Size() - 1): "operation", : "process")
```

`svc` is pushed only `if svc '= ""`, so on any row with an empty `Service` column hop 0 is a process or an operation, and the requested host is then reported as `position: "service"`. `position` is a claim about the host's ROLE that the model will repeat in evidence, and it is the one field here that can be confidently wrong rather than merely absent.

I could not produce an empty-`Service` row to prove it — LABDEMO has one fully populated path — so this is conditional on the query being able to emit one. But comparing `host` against `svc`, `op` and the `procs` members directly is the same few lines, needs no assumption about column population, and cannot drift. Worth doing unconditionally rather than verifying.

## 3. The `FindAllPaths` rationale is wrong, and I verified it rather than inferring it

Same correction as #185 point 1; repeated here because the class comment is where an implementer will read it. Live on `pg-iris`:

```objectscript
kill paths,cats
set sc = ##class(Ens.InterfaceMaps.Utils).FindAllPaths(.paths, .cats, "ProductionGuardian.LabDemo.Production")
// sc=1, and paths(1) has $listvalid = 1 with five elements:
//   EMR Source / Lab Router / ...LabDemo.RoutingRule / ...Transform.HL7ToPID / Cloud API
```

Signature, and its own description from the class dictionary:

```
FindAllPaths(*pPaths, *pCategoryItems, pProdName="", pCategory="", pTempPortalNode="", pDeDup=1) -> %Library.Status

  Internal method for finding all of the routes through the active production
  pPaths is an array of lists with each list describing one path through the production
  $lb(Service,Processes,Rules,DTLs,Operations)
```

It returns a `%Status` and writes paths **byref** into `pPaths`. Passing `"Cloud API"` as argument 1 puts a host name into the output slot, which is why a bare `1` comes back and looks like a string result. The `$c(12)`, `$c(11)`, `$c(1)` bytes measured at offsets in a 122-character value are `$list` length and type headers — not delimiters, not an undocumented encoding, and read with `$listget` in one line.

**Your decision to use the SQL query is still the right one** — the description opens with "Internal method", which is a firmer reason than the encoding and does not depend on decoding anything. Please swap the rationale for that, in the class comment and in `SplitList`'s.

Two things this does *not* undermine, which I checked because they rest on the same premise:

- **`SplitList`'s job is unchanged.** Element 2 of that `$list` is the plain string `Lab Router`, not a nested list, so a multi-process chain is still a joined value with an unknown separator. The UNVERIFIED caveat is honest and I would keep it worded exactly as it is.
- **"the implementation source is not shipped" is accurate.** `Ens.InterfaceMaps.Utils.1.INT` does not exist on the instance — `%ExistsId` returns 0 — so the row-building code genuinely cannot be read.

## Verified, and worth recording as verified

- **`known` is exact-match, not substring.** I raised this as a question on #185 because `EnumeratePaths(pSearchTerm, pSearchIn, ...)` is a term search across services, operations, processes, rules and transforms. You sidestep it entirely — `pSearchTerm` is `""` and the filter is `hops.%Get(i) = host` in ObjectScript. Right call, and it makes `known: false` trustworthy. Consider one clause in §3.14 saying the match is on the config item name, since `known` carries a not-an-all-clear claim.
- **The lag is in the goal.** `trend.recentDirection ... takes tens of seconds to turn` is exactly the §1.5 caveat I asked #185 to carry into the contract. The live risk is therefore already covered; my request there is about the durable record, not about this behaviour.
- **Positive-evidence-only is load-bearing and correctly placed.** A transient directive that made the Detective hedge on the ordinary rising queue would break #108's standing check, and the explicit "absence of evidence is not evidence of a transient" is what stops that.
- **`#MAXPATHS` applies to returned paths, not scanned rows**, so the cap itself behaves. Only the count is wrong.
- **Dedupe spans paths** — `seenUp`/`seenDown` killed once outside the loop — so a host on several paths appears once, nearest-first order preserved.
- **No new SQL grant is needed.** `AIHub.cls:74` records that these roles grant the `PG_*` resources and not database privileges, deferred from the #94 review, so this matches `Ens_Activity_Data.*`, `Ens_Util.Log` and `%SYS.Audit` before it. The comment-only `+6/-3` to `AIHub.cls` is correct, not an omission.
- **Governance placement is right**: `"all"`-only, and registered in `ToolManager()` too because it is context-free, unlike `FINDINGTOOLS`. The reason for holding it back from `"chat"` — #175's capability text and chips in two more places — is a better reason than "not needed".

## Order

Not stacked in GitHub terms (both base `main`), so nothing will auto-close. But #186 implements §3.14 and §3.14 has changes requested, so **#185 should land first** — otherwise this ships against a contract still being edited, possibly including a third field from point 1.

Fix 1 and 3, decide on 2, and this is an approve. I will re-read both together the same day.

tanifgit:
All three addressed in `e17cd99`. Point 1 was a real bug and your reproduction was exact — I confirmed it before fixing:

```
before  GetInterfacePath("Ens.MonitorService") -> known 0, returned 0, pathCount 1, truncated 0
after                                          -> known 0, returned 0, pathCount 0, truncated 0
```

## 1. `pathCount` — fixed, and it had a sibling one line down

The increment moved below the `at < 0` continue, so `pathCount` is now "paths this host sits on" — the reading §3.14 already defines, so **no third field and no return trip through #185**.

**Your framing is the part I want to acknowledge:** a payload that cites #165 as the thing it avoids, reintroducing #165 in a subtler form. That is worse than the original, because the surrounding prose invites trust.

While fixing it I found **the same bug one line further down**: the `#MAXPATHS` check also sat above the filter, so a non-matching row could set `truncated: true`. Same cause, same fix — the cap moved below the filter too.

One decision that needed making while moving it: `upstream` and `downstream` are still filled for **every** matching row, including ones the cap withholds from `paths`. Only `paths` is capped. Clipping the neighbour lists to honour a display cap would make the primary answer — "who feeds this host" — wrong in order to protect the secondary one. Stated in the code so the next reader does not "tidy" it into consistency.

And you are right about why it survived CI and two live agent runs: LABDEMO has one path and the demo host sits on it, so `total`, `pathsReturned` and the truthful count all collapse to 1. **The class of defect this exposes is that every count in this tool is untested on a production with more than one path** — which is worth a fixture rather than a note, and I have not built one.

## 2. `position` — fixed unconditionally, as you suggested

Now compared against the `Service`, `Operation` and `Processes` columns directly in a `PositionOf` helper, rather than `$case` on a hop index. Your reasoning is what decided it: I could not produce an empty-`Service` row either, but `position` is a claim about the host's **role** that a model repeats in evidence, so it is the one field here that can be confidently wrong rather than merely absent — and the comparison costs the same few lines and cannot drift with how `hops` is built. Verifying first would have bought nothing.

Re-checked all three roles live after the change: `EMR Source` → `service`, `Lab Router` → `process`, `Cloud API` → `operation`.

## 3. `FindAllPaths` — corrected in the class comment, verified first

`$listvalid` 1, five elements, `$listget` reads each; description opens "Internal method". Swapped to that rationale and kept the correction visible rather than deleting it, since the wrong version would have told an implementer the richer surface was undecodable.

**`SplitList`'s caveat is unchanged and now says why the `$list` correction does not settle it** — you checked the same thing and reached the same place: element 2 is the plain string `Lab Router`, not a nested list, so a two-process chain is still a joined value with an unknown separator. Also recorded that `Ens.InterfaceMaps.Utils.1.INT` genuinely does not exist, since that is what makes "source is not shipped" a checked claim rather than an assumption.

## On your opening note

The `EACH` versus `the nearest` correction came from watching it fail, not from foresight — it read `Lab Router`, found it clean, and reached the right verdict for the wrong reason. Worth saying because it is the second time on this issue that a directive which *looked* right produced a plausible answer from incomplete evidence, and only re-running the scenario showed it.

---

Ready for re-read. #185's fixes are pushed too (`8fbcce1`), and **it should land first** as you said — this PR implements §3.14 and §3.14 was being edited.

Not merging either; handing back to you now that you are back.

Ari-Glikman:
Heads-up on sequencing, unrelated to the review above: **#190 has landed on `main` and touches `BuildGoal()` in the block immediately before your insertion point** (this PR's hunk starts at old line 878, which is the `noOpSaves` sentence). Expect a textual conflict on rebase — the region is much longer now.

The semantics do not fight. #190 fixes #189 by removing the `retention.newest` comparison from the prohibition and refusing an absence claim in a new `DropUnsupportedConfigAbsence` guard instead. That guard only ever fires when `GetRecentConfigChanges` returns **nothing at all** for the finding's host, so this PR's positive use of it — "whether the pool was already enlarged shortly before now" — is exactly the case it leaves alone. Verified 6/6 that a real change is still named with old value, new value and timestamp.

One thing worth knowing while you rebase: the guard checks the **finding's** host only, so an absence claim about an *upstream* host is out of its reach. That is stated as a known limit in the method comment, and this PR is what makes per-upstream tool calls routine — so if you want the upstream case covered, it is worth deciding here rather than later.

Ari-Glikman (approved):
Re-reviewed `e17cd996`, and verified the fixes against the **merged** tree rather than against this branch alone, since `main` moved under you. Both review findings are fixed correctly, the `pathCount` one is fixed at the right line, and I found **one more defect of the same family** — a truncation note that reads as complete. Approving; see the end for the one line.

## First, a correction to my own comment above

I said an hour ago to **expect a textual conflict** with #190 on rebase. **That was wrong** — I checked rather than assumed this time:

```
git merge-tree --write-tree main e17cd996   -> clean
git merge main <- e17cd996                  -> Auto-merging AgentDispatcher.cls, no conflict
```

GitHub agrees (`MERGEABLE / CLEAN`). Your `BuildGoal` insertion is far enough from the block #190 rewrote that the 3-way merge handles it. **No rebase needed.** I also checked the merged result is semantically right, which is the part a clean merge does not guarantee:

- #190's new prohibition survives intact (merged line 1041, `"NEVER state that no configuration change was found... an empty changes list is not evidence of anything"`).
- The old `retention.newest` comparison is **gone**, present only in #190's historical comment. A clean merge cannot resurrect lines `main` deleted, and it did not.
- Your `You MUST also call GetInterfacePath on this host before concluding` lands at merged line 1067, after it.

And the semantics do not fight, as I said on the other thread: #190's guard only fires when `GetRecentConfigChanges` returns **nothing at all** for the finding's host, so this PR's positive use — "whether the pool was already enlarged shortly before now" — is the case it leaves alone.

## Verified on the merged tree

```
npx tsc --noEmit -p tsconfig.json   clean
npm test                            415 pass, 0 fail, 80 suites
```

Same numbers you report, on `main` + this branch rather than on this branch.

## `pathCount` — fixed at the right line, and the reasoning is right

Moving both the counter and the cap below the `at < 0` filter is the correct fix, and keeping `pathCount` host-scoped rather than adding a third field is the right call: a third number would have to be explained to a model, and "how many paths does the rest of the production have" is not a question this tool is answering.

Your reading of the severity matches mine. The bad part was never the number, it was that **the only reading available to a model was "57 paths were withheld from me"** — a payload that would have taught the agent to hedge every topology conclusion, in a tool whose whole value is that it answers when a host is dead.

Two things I checked specifically:

- **`truncated` moved with it.** This mattered as much as the count: a non-matching row could set `truncated`, so a host on 1 of 60 paths would publish `truncated: true` with 1 path returned and a note claiming the list was clipped. That is worse than the wrong count, because it is a false claim rather than a confusing one.
- **`upstream`/`downstream` stay uncapped past the cap.** Agreed, and the comment states the reason at the line where someone would "fix" it. Clipping the primary answer to honour a display cap on the secondary one is the wrong trade.
- **The unfiltered path (`host = ""`) still counts every row**, since the whole filter block is inside `if host '= ""`. Confirmed by reading, not assumed.

## `PositionOf` — right fix, and right to fix it unconditionally

The bug was real and your description of it is precise: `svc` is only pushed when the `Service` column is non-empty, so `at = 0` means "service" only on rows that have one.

Fixing it without first proving the query emits such a row is the right instinct for the reason you name — `position` is the one field here that can be **confidently wrong** rather than merely absent, and a model repeats it in evidence as a claim about the host's role. A wrong `position` on an operation reported as a `service` would send an investigation up the wrong side of the path.

Note that the `""` fallback is currently **unreachable**, and that is fine — `hops` is built from exactly the three values `PositionOf` compares against, so agreement is structural. Worth keeping as written; it is a guard against the two drifting apart, which is the coupling the old `$case` had.

The `RETURN, not QUIT` comment inside the `while` is worth having. That is a real trap in this codebase and it has bitten this class family before.

## The new finding: the truncation note says "50 of 50"

`Note()` prints:

```objectscript
set note = note _ " The path list is TRUNCATED at " _ returned _ " of " _ ..#MAXPATHS
    _ "; pathCount is the true total."
```

`truncated` is only set when `paths.%Size() >= ..#MAXPATHS`, so **`returned` is always exactly `#MAXPATHS` at this line**. The note therefore always reads:

> The path list is TRUNCATED at **50 of 50**; pathCount is the true total.

A model reading "50 of 50" has been handed a complete-looking ratio in the same sentence as the word TRUNCATED, and the number that would actually inform it — `total` — is the one value `Note()` is not passed. So the sentence points at a field it cannot show and shows a field that says nothing.

This is the same family as the bug you just fixed and the same family as #165, which is why it is worth catching now rather than later: the prose is what the model consumes, so a misleading note is not cosmetic here the way it would be in a dashboard string.

Fix is mechanical — pass `total` in and print the pair that differ:

```objectscript
do ..Note(host, out.known, paths.%Size(), truncated, total)
...
set note = note _ " The path list is TRUNCATED at " _ returned _ " of " _ total
    _ " paths this host sits on; pathCount carries the true total."
```

Unreachable on LABDEMO (one path), like the two before it — which is the third time on this class that the interesting branch is the one this production cannot produce. Worth a line in the class comment saying so, if you want the next reader to check the >50 case by reading rather than by running.

## One matching change on the contract side

#185 §3.14 says *"`pathCount` is the true total"* without saying **for this host**. Your commit message calls that "the reading §3.14 already defines", but it does not — that sentence is exactly as readable as production-wide, which is the reading this code shipped with. I have asked for the one word over on #185; flagging it here so the two land consistent.

Fix the `Note()` line and I will merge both. Or say the word and I will push both one-liners myself.

tanifgit:
`Note()` fixed and the branch is **rebased onto `main` past #190** — `e5a7e5e`. The paired one-word fix is on #185 (`84bb6b5`).

## The `Note()` fix

`Note()` now takes `total` and prints the pair that differ:

```objectscript
set note = note _ " The path list is TRUNCATED at " _ returned _ " of " _ total
    _ " paths this host sits on; pathCount carries the true total."
```

Your diagnosis was exact — `returned` is always `#MAXPATHS` at that line, so the sentence could only ever read "50 of 50", and `total` was the one value not passed in. Verified in the **compiled** `.INT` rather than the `.cls`, since that branch cannot be reached from outside:

```
set note = note _ " The path list is TRUNCATED at " _ returned _ " of " _ total
```

**I also took the class-comment line you offered**, and made it stronger than a line, because your count was the thing that persuaded me: all **three** defects on this class were in the same dead zone — production-wide `pathCount`, a `truncated` a non-matching row could set, and this note — and every one would have published a confident wrong number to a model. The comment now says the counting and truncation branches are unreachable on LABDEMO and that anything touching the row loop needs a >50-path production in your head, because there is not one to run against.

I tried to exercise the branch anyway and it is properly shut: `Note()` returns `<PRIVATE METHOD>` from a session, which is incidentally another confirmation of the §7 helper rule.

## The rebase came out clean, which I did not take as good news

**No conflict at all**, contrary to both our expectations — git placed the ~80 lines after #190's `noOpSaves` sentence and before the attribution block, which is the right place. Since a clean auto-merge of two edits to the same method is exactly the case that can be textually fine and semantically wrong, I checked rather than trusted:

- `diff origin/main...HEAD` is **only** this PR's five files, +556/−9. #190's work is untouched.
- `DropUnsupportedConfigAbsence` intact and still registered at the dispatch site; the `retention.newest` mentions that remain are all inside your explanatory comment, not in live prose.
- Both classes **compile clean**, and the boot guard reads **15 discovered, expected 15**, `Topology: GetInterfacePath` exposing exactly one tool.
- Live `GetInterfacePath("Cloud API")` still correct: `upstream ["Lab Router","EMR Source"]` nearest-first, `position: "operation"`, `pathCount: 1`.
- Engine **415/415**, `tsc --noEmit` clean — now against post-#190 `main` rather than against the old base.

## Your open question — and I do not think it should be decided here

You wrote that the guard checks the finding's host only, so an absence claim about an *upstream* host is out of its reach, and that this PR is what makes per-upstream calls routine — "worth deciding here rather than later."

**Agreed that this PR creates the exposure, and I still think folding the fix in is the wrong call.** Reading the merged prose together, there is a narrower and more concrete version of the problem than "upstream absence claims":

> …and you must check all three: `trend.recentDirection` … whether the upstream host you just read has errors that have ENDED … and **whether the pool was already enlarged shortly before now, which `GetRecentConfigChanges` shows**.

That third clause asks a question of `GetRecentConfigChanges` whose **negative answer is the claim #190 forbids**. On the finding's host your guard catches it, so the exposure is exactly the upstream case you named. The mitigating half is that "Only conclude this on POSITIVE evidence" means an empty list contributes nothing rather than licensing a transient verdict — so the risk is a bad *evidence entry*, not a bad recommendation.

Two reasons to file it:

1. **It is unmeasured.** Every number in this PR is a before/after on an armed scenario, and I have no run showing the model making an absence claim about an upstream host. Adding a guard for a failure I have not observed is the thing your own review of the `PositionOf` fallback called out as acceptable only because that one was structural — this would not be.
2. **Widening the guard means guessing which host an entry refers to**, which is the prose recognition #190 deliberately chose a `tool`-match to avoid. That is a design decision for the guard's own PR, and #190's method comment already states the limit honestly.

So my preference is **merge these two and file the upstream case with the mechanism written down**, and I would measure it first — arm the missing-folder scenario, run the Detective on `Cloud API`, and count absence claims attributed to an upstream host. Happy to do exactly that and file it with numbers. **Say the word if you would rather it went in here** and I will add it.

Also still happy to file the three you found while measuring #190 — the constant `diagnostics.toolCalls: 3` is the one I would take first, since it is an assertion in the #108 pre-demo check and it is currently vacuous.

Issue #174, opened 2026-08-31
Early Warning cannot say "recovering": the direction gates sit below the threshold, so a draining queue still reads as "Threshold reached"


tanifgit:
Reported by @tanifgit from a live run:

> there is the "early warning" feature — this works well for example when a queue starts to build up. The problem is, that this happens, falsely also when the queue starts diminishing while it's "cooling down". If we see the general direction is — the queue is growing smaller, not bigger (though it might be above the threshold of the warning, and below the critical) we should not say this is an early warning.

This is the **second** sighting of "early warning during a drain". The first was fixed in #142 / 9d1deb0 and that fix is on `main`. It does not cover the state described here, and the contract says so in as many words.

## Where the direction gates are, and why neither one runs

`earlywarning.ts` asks "is it rising?" twice — once of the whole fit window, once of its tail — and declines with `not_rising` unless both agree:

- `services/detection-engine/src/detect/earlywarning.ts:338` — window slope
- `services/detection-engine/src/detect/earlywarning.ts:360` — tail slope (`RECENT_FIT_FRACTION`)

**Both sit below an earlier return.** `services/detection-engine/src/detect/earlywarning.ts:333`:

```ts
// BEFORE the slope, deliberately. A crossed threshold makes the slope irrelevant, and a
// draining-but-still-crossed queue would otherwise report `not_rising`, which reads as
// "nothing to see" about a queue that is over its limit.
if (currentValue >= threshold.value) return decline('already_crossed', threshold);
```

So neither direction gate is reachable while the queue is above the threshold. The module's own comment states the consequence outright (`earlywarning.ts:115`):

> Note the drain only reaches this code once the queue is UNDER the threshold — above it, `already_crossed` answers first.

And `contracts/earlywarning-api.md:273-276` records it as a deliberate decision rather than an oversight:

> **This adds no eighth reason and does not move the precedence.** […] so a queue draining while still **above** its threshold keeps reporting `already_crossed` — draining-but-over-limit is still a problem.

## What the operator sees, and why it reads as a warning

`apps/dashboard/src/components/EarlyWarning.tsx:60`:

```ts
already_crossed: 'Threshold reached — see the finding below',
```

That is the whole of what distinguishes a queue at 58 and climbing from a queue at 58 and draining to zero: **nothing**. Both render the same sentence, in the same strip, above an active `queue_buildup` finding. "Threshold reached" is true in both cases and useful in only one.

[...]

Issue #178, opened 2026-08-31
Smart Resolve approval label reads "increase Cloud API pool ? -> 8": recommendedAction.currentValue is null on every investigation


tanifgit:
Found while reproducing #177 on the live stack. Independent of that issue's reasoning question, and
cheap to fix.

## The operator is asked to approve a write with a `?` in the label

A live investigation (`source: "agent"`, `gpt-4o-mini`, `toolCalls: 3`) returned:

```json
"recommendedAction": {
  "action": { "type": "set_pool_size", "host": "Cloud API", "size": 8 },
  "currentValue": null,
  "bounds": { "min": 2, "max": 8 },
  "reversible": true,
  "requiresApproval": true,
  "summary": "increase Cloud API pool ? -> 8"
}
```

`contracts/investigation-api.md:389` makes `summary` "One line for the approval button's label.
**Authoritative, render as-is.**" So the button on the approval path for a write to a live production
reads **"increase Cloud API pool ? -> 8"**.

## It is unconditional, not an edge case

`services/detection-engine/src/index.ts:183` passes the parameter as a literal `null`:

```ts
return investigate(finding, host, projection, null, {
```

with the reasoning stated just above it:

> Pool size comes from the projection's own threshold basis where available; the authoritative read is
> `get_pool_size`, which the AGENT calls. Passing null rather than guessing keeps this service out of
> the business of reading production config.

`investigate.ts:301` then renders the fallback:

```ts
summary: `increase ${finding.host} pool ${currentPoolSize ?? '?'} -> ${size as number}`,
```

There is no other call site. **Every investigation that recommends an action produces
`currentValue: null` and a `?` in the summary** — this is not a rare path, it is the only path.

The premise about not reading production config is reasonable, and the value was still available:
the agent had already read it in the same exchange and reported it as evidence —
`Current Pool Size | mcp_tool | GetHostSettings | 4 workers`. The engine never extracts it from the
reply it is already parsing.

## Three consequences, and the label is the least of them

`contracts/investigation-api.md:385` specifies the field as more than display:

> `currentValue` | **integer** | Pool size now, as read by `get_pool_size`. Not the engine's guess.
> **Feeds `precondition.poolSize` on the resolve call and the reversal target.**

So with it always null:

1. **The approval label reads `?`** — an operator is asked to approve a change from an unstated value.
2. **`reversible: true` has no reversal target.** `:387` defines reversible as "whether re-applying
[...]

PR #191, opened 2026-09-01
fix(detection-engine): trend.slope was null on every investigation ever served (#187)

### LINKED ISSUE #187: investigation-api §2.2 promises a draining slope in trend, and the engine has never sent one: trend.slope is null on every investigation
Found while building #177, and it is the reason that issue needed a new field rather than just a prompt.

## The contract says the slope may be negative here, deliberately

`contracts/investigation-api.md` §2.2, on `trend.slope`:

> Same fit and same units as `Projection.slope` — OLS over the trailing 300 s. **May be zero or negative here**, unlike in `earlywarning-api.md`, because a queue that is draining is a fact the agent should see rather than a forecast to withhold.

And §4's own example carries it beside a crossed threshold:

```json
"trend": { "metric": "queued", "slope": 124.2, ..., "thresholdCrossed": true }
```

That is a deliberate, argued decision: the contract explicitly carries the slope *past the point `earlywarning-api.md` stops publishing it*.

## The engine does not do that

`services/detection-engine/src/detect/investigate.ts`, `buildTrend`:

```ts
const slope = projection.projection?.slope ?? null;
```

`projection.projection` is `null` whenever `projectionUnavailable` is `already_crossed`. And `already_crossed` is returned **exactly when a queue has crossed its threshold** — which is the only condition an investigation is ever requested for. `buildTrend`'s own doc comment says as much, one paragraph above the line:

> Early Warning publishes `projection: null` with reason `already_crossed` exactly when a queue has crossed its threshold — which is the state that made `queue_buildup` fire, i.e. the only condition an investigation is ever requested for.

So the field is **null on every investigation**, and the contract's stated reason for carrying it — "a queue that is draining is a fact the agent should see" — has never once been delivered.

## The measured cost

This is not academic. On #177's armed scenario the Detective recommended enlarging a pool on a queue falling **261 → 181**, because nothing in its input said "falling". #177 worked around it by adding `trend.recentDirection` (the *sign* of the tail fit), which is the half fixable without reopening the other contract. The rate itself is still missing.

## Why it was not simply fixed

Publishing the window slope from `/api/earlywarning` would contradict `earlywarning-api.md` §1.4, which is equally deliberate:

> publishing "rising ~0.5/min" next to no ETA still implies a forecast we refused to make

So the engine cannot just forward a field Early Warning declines to publish. The options each cost something:

1. **Publish the window slope on `HostProjection`** as a measured field, the way `recentDirection` now is. Needs §1.4's boundary redrawn — a rate is exactly what that section is about, so this is the expensive one.
2. **Refit inside `buildTrend`.** It would need the baseline store, which `investigate()` is not given (it receives `HostProjection | undefined`). Duplicates the fit, and two fits of the same samples can disagree.
3. **Amend §2.2 to say `slope` is null on a crossed threshold** and that `recentDirection` is the draining signal. Cheapest, honest, and loses the rate — an agent could then say "draining" but not "draining at 4/min", so it cannot estimate how long the backlog will take to clear.

(3) is probably right for now and (1) is the real answer if a rate turns out to matter. Either way it is a contract decision, not a code fix, which is why #177 did not take it.

## Not urgent, and worth doing before the sizing work

The sizing formula issue (filed alongside this) wants to know whether an arrival rate is sustained or a burst. A signed queue slope is one of the better inputs for that, so these two are likely to be resolved together.

Ari-Glikman:
Closes #187.

## The defect

`investigation-api.md` §2.2 carries a **signed** slope to the agent on purpose:

> May be zero or negative here … because a queue that is draining is a fact the agent should see rather than a forecast to withhold.

It has never been delivered. `buildTrend` read `projection.projection.slope`, and Early Warning sets that object to `null` for `already_crossed` — which is the **only** state a `queue_buildup` investigation is ever requested in. So `trend.slope` was null on every investigation the product has ever served, and the agent recommended enlarging a pool for a queue falling 261 → 181 because nothing in its input said "falling".

## Why this is none of the three options the issue proposed

#187 frames it as a contract decision: publish the slope on `HostProjection` and redraw §1.4, refit inside `buildTrend`, or amend §2.2 to admit null. There is a fourth path that needs **no contract change**, because the two contracts already say the right thing for their own consumer:

| | consumer | rule | after this PR |
|---|---|---|---|
| `earlywarning-api.md` §1.4 | the panel | no slope outside `projection`, *even where we decline to forecast* | unchanged — `publishedProjection()` strips it |
| `investigation-api.md` §2.2 | the model | a signed slope, present whenever there is a usable fit | honoured as ratified |

§1.4's reasoning is not being bent: "a visible *rising ~0.5/min* next to no ETA still implies a forecast we refused to make" is a statement about a **panel**, and the panel still cannot see this field.

## What changed

**The window fit is hoisted, not duplicated.** It already existed as one least-squares fit inside `projectHost` — it was merely computed at step 6, *below* the `already_crossed` decline. It now sits above the precedence chain and lands on `base`, and the ETA path reads it instead of refitting.

- Arithmetically identical on every path that previously reached it: `fitSampleCount >= minFitSamples` is already checked by then, and the rounding and the `<= 0` test are unmoved.
- Same "computed once, used twice" arrangement the tail fit has had since #174, for the same reason — a second derivation is how the published rate and the agent's rate come to disagree.

**`publishedProjection()` is a whitelist, not a delete.** So the wire shape of `/api/earlywarning` is decided in one place and a future internal field cannot reach the endpoint by being forgotten. Typed as `Omit<HostProjection, 'windowSlopePerMinute'>`, so a rename fails to compile rather than silently vanishing from the payload.

**A second, contract-derived fix rides along.** `buildTrend`'s gate was `slope && threshold`, reachable only because `slope` was unconditionally null — so `insufficient_samples` served a trend object whose only real content was a threshold. §2.2 states the null case as *"no usable fit at all — a warming baseline, or fewer than 12 samples in the fit window"*, which is a **disjunction**, so the gate is now `||`. Nothing is lost: `thresholdValue`/`thresholdBasis` are in the snapshot too, and the agent gets one meaning for a present `trend` instead of two.

**`recentDirection` (#177) is kept, and the comment claiming this bug was unfixable there is gone.** The two are measured over different spans and the disagreement is the point — measured, not supposed: a queue that climbed for eight minutes and has drained for 100 seconds fits **+26.6/min over the window with a `falling` tail**. That is exactly the state that decides whether a second pool increase is warranted, and one number flattens it.

**IRIS needs no change.** `AgentDispatcher.BuildGoal` dumps `body.trend.%ToJSON()` wholesale and never names `slope`, so a real value reaches the model without editing the goal — and #190's displacement risk stays untouched.

## Verification

`tsc --noEmit` clean, **428/428 tests** (415 on `main`, 13 new).

The new tests assert the **request handed to `callAgent`**, not the served response — nothing in the response changes, so an end-to-end test would have passed against the defect.

| Test | Pins |
|---|---|
| slope negative on a crossed draining queue | the #187 case, and the magnitude (`-36`/min for a 3-per-poll drain) — a right-signed wrong-magnitude rate is a lie the agent cannot detect |
| `windowSlopePerMinute === projection.slope` on the happy path | the one-fit invariant that makes the hoist safe |
| null below `minFitSamples`, null when unmeasurable | no rate fitted through three points |
| `publishedProjection()` strips it **and keeps the other ten keys** | both directions — a whitelist that dropped a published field is the opposite defect |
| `GET /api/earlywarning` carries no `windowSlopePerMinute` | that the **route** calls the mapper. A mapper that works and a mapper that is reached are different claims; this route served `projections` raw for all of MVP 2 |
| `trend` null for `insufficient_samples` **and** for `warming` | both halves of §2.2's disjunction |

Live end-to-end check against the running stack is in a comment below.

### DIFF
--- a/services/detection-engine/src/api/server.ts
+++ b/services/detection-engine/src/api/server.ts
@@ -13,6 +13,7 @@
 
 import { createServer, type IncomingMessage, type Server, type ServerResponse } from 'node:http';
 import type { EngineSnapshot } from '../detect/engine.ts';
+import { publishedProjection } from '../detect/earlywarning.ts';
 // The off-state payload, imported rather than written out here. It was inlined twice below and
 // gained a third field in #135; two literals that must match a type in another file is the
 // stale-copy shape, and both copies would have been silently short an `activating: {}`.
@@ -342,7 +343,13 @@ export function createFindingsServer(options: ServerOptions): Server {
           // state header could disagree with the first, and a consumer would have no rule for
           // which to believe -- the projections come from the same poll as the findings, so they
           // share its freshness by construction.
-          sendJson(res, 200, snapshot.projections, snapshot.state);
+          //
+          // MAPPED, not served raw: `HostProjection` carries one internal field the contract does not
+          // publish -- the measured window slope, which §1.4 keeps off this endpoint and
+          // `investigation-api.md` §2.2 requires for the agent (#187). `publishedProjection` is a
+          // whitelist, so the endpoint's shape is decided there rather than by whatever the type
+          // happens to hold.
+          sendJson(res, 200, snapshot.projections.map(publishedProjection), snapshot.state);
           return;
         case '/api/hostseries': {
           // `host` IS REQUIRED AND IS A 400, not an empty series. Every other read here answers
--- a/services/detection-engine/src/detect/earlywarning.ts
+++ b/services/detection-engine/src/detect/earlywarning.ts
@@ -70,6 +70,51 @@ export interface HostProjection {
   threshold: Threshold | null;
   projection: Projection | null;
   projectionUnavailable: ProjectionUnavailableReason | null;
+  /**
+   * The window fit's slope in `SLOPE_UNIT`, or null when there is no usable fit.
+   *
+   * **INTERNAL. Not part of `earlywarning-api.md`, and `publishedProjection()` strips it** — §1.4
+   * forbids a slope outside `projection` on that endpoint even where we decline to forecast, because
+   * "rising ~0.5/min" next to no ETA implies the forecast we withheld. That rule is not being bent
+   * here: the panel still cannot see this.
+   *
+   * It exists because `investigation-api.md` §2.2 ratifies the opposite decision for the *agent*:
+   * `trend.slope` "may be zero or negative here ... because a queue that is draining is a fact the
+   * agent should see rather than a forecast to withhold". Two contracts, two consumers, and the
+   * difference is deliberate on both sides — a panel renders to an operator who acts on a rate, and
+   * the model has already been measured recommending a bigger pool for a queue falling 261 -> 181
+   * because nothing in its input carried one (#187).
+   *
+   * Sourced from the SAME fit the ETA uses, computed once below. A second derivation of it is how
+   * the published rate and the agent's rate would come to disagree — #174's argument for the tail.
+   */
+  windowSlopePerMinute: number | null;
+}
+
+/**
+ * One projection as `earlywarning-api.md` §1.1 defines it, for the endpoint to serve.
+ *
+ * A WHITELIST rather than a delete, so the wire shape is decided here and an internal field added to
+ * `HostProjection` cannot reach the endpoint by forgetting about it. Publishing a new field becomes a
+ * deliberate edit to this function, which is the right amount of friction for a ratified contract.
+ * Typed as an `Omit` so a renamed field fails to compile rather than silently vanishing from the
+ * payload.
+ */
+export function publishedProjection(
+  p: HostProjection,
+): Omit<HostProjection, 'windowSlopePerMinute'> {
+  return {
+    host: p.host,
+    metric: p.metric,
+    currentValue: p.currentValue,
+    measuredAt: p.measuredAt,
+    fitSampleCount: p.fitSampleCount,
+    fitSpanSeconds: p.fitSpanSeconds,
+    recentDirection: p.recentDirection,
+    threshold: p.threshold,
+    projection: p.projection,
+    projectionUnavailable: p.projectionUnavailable,
+  };
 }
 
 /**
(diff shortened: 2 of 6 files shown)


Ari-Glikman:
## Live check against the running stack

Engine rebuilt from this branch (`agent=live`, `demo-triggers=ON`), `pool_bottleneck` armed, `queue_buildup` fired at 77 queued, one metered investigation, triggers reset.

**The state the defect lived in, reproduced live** — `GET /api/earlywarning`, `Cloud API`:

```json
{ "currentValue": 87, "fitSampleCount": 38, "fitSpanSeconds": 185,
  "recentDirection": "rising", "projection": null,
  "projectionUnavailable": "already_crossed" }
```

Ten keys, **no `windowSlopePerMinute`** — §1.4 holds on the wire, and this is exactly the row where the old code had a real 38-sample fit available and threw it away below the decline.

**One live investigation, `POST /api/investigate`** — the #108 assertions, on the new code:

```
state: complete    source: agent    model: gpt-4o-mini    toolCalls: 3    durationMs: 6974
confidence: 0.9    summary: "increase Cloud API pool 1 -> 4"    currentValue: 1
```

Correct root cause (capacity, not connectivity), five evidence bullets with three `mcp_tool` attributions, and #178's before-value intact. No regression from the change on the live path.

**What this does and does not prove.** It proves the deployed code serves the `already_crossed` state and the whole agent chain end to end. It does *not* observe the request body — nothing persists it, and the served response is unchanged by design. That link is covered by `test/investigationTrend.test.ts`, which asserts the object handed to `callAgent`; `AgentDispatcher.BuildGoal:913` then does `" Trend: " _ body.trend.%ToJSON()`, so whatever is in that object reaches the model verbatim.

## One follow-up worth filing separately

`BuildGoal`'s "three things distinguish a draining backlog from a sustained rise" block (line ~1112) names `trend.recentDirection` and never `trend.slope`, and the comment at 1125 states the tail fit's ~35s lag as a caveat the model has to weigh. A signed window slope is now available to put beside it. That is a **prompt** change in `iris/**` with #190's measured displacement risk — three separate regressions have come from editing that block — so it does not ride along here. Say the word and I will file it.

PR #193, opened 2026-09-01
fix(detection-engine): the pool was sized from throughput, which measures the throttle and not the load (#188)

### LINKED ISSUE #188: Pool sizing is computed from messagesPerSec, which is throughput not arrival rate: it understates the pool on a bound host and overstates the need on a draining one
Split out of #177 deliberately, so the topology change could be measured against #108 one variable at a time.

## The formula is sound and its input is not

`AgentDispatcher.SystemPrompt` tells the model to size the pool from Little's law rather than by doubling, and that block earned its place — it was added because the model reliably suggested `current + 1`, which on the demo scenario is exactly break-even and therefore fixes nothing:

```
workersToKeepUp = messagesPerSec x avgProcessingTime
```

The arithmetic is right. **`messagesPerSec` is not a reliable arrival rate**, and the case where it is worst is precisely the case #177 is about.

## Measured

On #177's armed transient — an upstream fault fixed, 281 accumulated messages flushing through at once — the snapshot handed the agent `messagesPerSec: 4` while the queue was *draining* from 261 to 181. That 4/sec is throughput *through* the host as it clears a backlog, not the rate at which new work is arriving. Feeding it to the formula gives:

```
workersToKeepUp = 4 x 1.01 = 4.04  ->  "pool 4 is insufficient"
```

which is how a queue that needed nothing produced a confident recommendation of 6. #177 fixed the *conclusion* by giving the model upstream evidence and a `falling` direction; it did not fix the *input*, and the formula still reads as supporting evidence for a change nobody needs.

The same distortion runs the other way on the flagship scenario. At `PoolSize 1` against a ~1s downstream, `messagesPerSec` reads about **1**, because that is all the host can clear — while inflow is ~2.9/sec. Measured on the #108 check: the model computed `1 x 0.82 ≈ 1`, recommended `1 -> 2`, and 2 is break-even at best. `PoolBottleneck`'s own arming output states the real arithmetic (`net ~+1.9/sec into the queue`), and the agent cannot see it.

**So a throughput-bound host systematically understates its own arrival rate**, and the pool the model computes is systematically too small — which is the defect the sizing block was written to prevent, arriving through its input instead of its wording.

## There is already a field for this, and it is not used here

`investigation-api.md` §2.2 defines `snapshot.inboundRatePerSec`:

> The engine computes `messagesPerSec + trend.slope/60` … It sits in `snapshot` rather than `trend` because it is a rate *now*, not a forecast

That is exactly the right quantity: throughput plus the rate the queue is growing equals arrival rate. Two things to check before building on it:

- whether the engine actually populates it (the §4 example shows `inboundRatePerSec: 3.04`, but #187 is an instance of this same contract promising a `trend` field the engine does not send — worth verifying rather than assuming);
- it depends on `trend.slope`, which **#187 shows is null on every investigation**. So this is blocked on #187, or on `recentDirection` being insufficient for it — a direction gives the sign, not the magnitude, and this needs the magnitude.

## What to do

1. Verify whether `inboundRatePerSec` is populated and correct. If it is, the fix may be as small as naming it in the sizing block instead of `messagesPerSec`.
2. If it is not, #187 has to land first: the arrival rate cannot be computed without a signed queue slope.
3. Re-measure **both** scenarios after any change. The flagship rise and the drain-through transient distort `messagesPerSec` in *opposite* directions, so a change that fixes one can easily worsen the other, and #108's standing check only covers the first.

## Not a regression

Nothing here is newly broken — #177 made the failure visible by giving the model enough context to reach the right conclusion *despite* the arithmetic. Worth fixing so the stated reasoning and the recommendation agree, because a rootCause that shows its working and gets the working wrong is harder for an operator to challenge than one that just asserts.

Ari-Glikman:
Closes #188.

## The defect

`snapshot.inboundRatePerSec` has been in `investigation-api.md` §2.2 since MVP 2 and **no engine build ever populated it** — #188's first step confirmed the only three mentions of the field in the repo were in that contract. So the agent sized the pool from `messagesPerSec`, which counts messages **completed**:

- on a **throughput-bound** host it is the throttle, not the load. The flagship scenario clears ~1/sec while ~2.9/sec arrives, so break-even comes out at 1.01 workers and Little's law says "one is enough" about a host 1.9/sec behind.
- on a host **clearing a backlog** after a fix it overstates the load instead — 4/sec of completions while ~1.7/sec arrives.

Wrong in both directions, on the one piece of arithmetic the Detective shows its working with.

## What changed

| Half | Files |
|---|---|
| The engine derives the field, and `trend` carries the term it is derived from | `detect/investigate.ts`, `detect/earlywarning.ts` |
| The prompt names it in the sizing formula, and `RaiseUndersizedPool` prefers it | `iris/labdemo/REST/AgentDispatcher.cls` |
| §2.2 amended: formula, the new `trend.recentSlope` row, the clamp | `contracts/investigation-api.md`, `contracts/CHANGELOG.md` |

`iris/CLAUDE.md`'s rule that **a tool that is registered and described is not a tool that gets called** applies to a snapshot field inside a stated calculation for the same reason: the model uses the term the formula spells out. Measured under the *old* prompt, which never named the field — the model read `inboundRatePerSec` off the snapshot JSON anyway and reported "Receiving approximately 4.69 messages per second", presenting a derived estimate as an observation. That is an argument for naming and caveating it, not for hiding it.

## Re-measuring both scenarios found a second defect

#188 was explicit that **both** scenarios be re-measured, because they distort `messagesPerSec` in opposite directions and #108's standing check covers only the first. Doing it is what caught this: §2.2 specified `messagesPerSec + trend.slope/60`, and the **window** fit is 300 s, so minutes into a drain it still leans up. The estimate then lands *above* the raw completion rate on the one state where completions are already an overstatement.

| Arrival rate from | Value | Agent's recommendation |
|---|---|---|
| `messagesPerSec` alone | 4 | `set_pool_size 4 -> 8` |
| `+ slope/60`, queue 94 | 4.57 | `set_pool_size 4 -> 6` |
| `+ recentSlope/60`, queue 108 | **3.82** | **none** |

The two derived rows are separate runs a few polls apart, not the same sample — the transient is short. What decides the outcome is which side of `messagesPerSec` each estimate lands on. The `4 -> 8` row is `main`'s behaviour, measured by stashing the `.cls` and recompiling `main`'s dispatcher against the same state, so the first build is an improvement rather than a regression I am reporting as one.

So the fix is the **tail** slope — the fit already computed for `recentDirection`'s sign, now published beside it as `trend.recentSlope`. Computed once, used twice: the arrangement #174 and #187 both landed.

**Attribution, so the table above does not overclaim.** The `none` outcome needs the corrected estimate **and** a new prompt directive ("if the queue is already draining, recommend nothing"). Neither alone is sufficient — the directive cannot fire on an estimate reading 4.57, and the estimate alone leaves the sizing formula free to justify a second increase.

## Contract change, per root `CLAUDE.md` §4

**One PR rather than two, stated rather than assumed.** The change is additive for `trend` and a genuine *semantic* change for `inboundRatePerSec` — the same field name can now hold a different number for the same inputs. It is safe to land in one step because no build has ever populated it, so nothing changes meaning underneath a consumer; and the CHANGELOG entry argues the two halves must land together, since the contract's own formula is what produced `4 -> 6`. Splitting it would land a contract describing a field no build populates, which is the state #188 was filed about. `contracts/CHANGELOG.md` carries the entry and the measurements.

`earlywarning-api.md` is untouched: §1.4 forbids a bare slope beside a withheld forecast and does not distinguish the two spans, so `publishedProjection()` strips both, and `api.test.ts` asserts the wire payload still carries exactly its ten keys.

## Also in here

`trend` now declines when the **tail** fit declines, not only the window fit. The two share a sample-count gate but not an outcome: a poll gap — the engine records nothing while the proxy is unreachable — can leave one sample inside the trailing 120 s while the 300 s window still holds twelve. Without that arm `trend` would carry a window slope beside two nulls and a null `inboundRatePerSec`, and §2.2's promise that `recentSlope` is non-null whenever `trend` is would be nearly true rather than true. Unit-tested; not reachable in the measured scenario.

## Verification

`tsc --noEmit` clean, **439/439** (415 was the floor before #187, 428 after it).

Three live runs, each `state: complete` / `source: agent` / `model: gpt-4o-mini` / `toolCalls: 6`, with every `evidence[]` entry attributed — the `#190` displacement risk re-checked, since this edits `SystemPrompt`:

1. **rising**, queued ~174, `messagesPerSec` 1 -> `set_pool_size 1 -> 4`, "Estimated inbound rate is 2 messages per second"
2. **15 s post-apply**, queued 176, `messagesPerSec` 4, direction still `rising` -> inbound 4.84, `4 -> 6`. **The honest limit of this fix**: at 15 s the 120 s tail genuinely still leans up, so no derivation can know the queue is draining yet.
3. **at `recentDirection: falling`**, queued 108, `messagesPerSec` 4 -> inbound 3.82, **`recommendedAction: null`**, confidence 0.85, `rootCause` describing the recovery and its rate.

Demo triggers reset afterwards; `Cloud API` pool restored to 1.

### DIFF
--- a/contracts/CHANGELOG.md
+++ b/contracts/CHANGELOG.md
@@ -4,6 +4,77 @@ Every contract change, dated, with the reason. Newest first.
 
 ---
 
+## 2026-09-01 — `inboundRatePerSec` is derived from the TAIL slope, not the window slope
+
+**One file, one formula corrected and one field added.** `investigation-api.md` §2.2:
+`snapshot.inboundRatePerSec` was defined as `messagesPerSec + trend.slope/60` and now reads
+`messagesPerSec + trend.recentSlope/60`; `trend` gains `recentSlope`, the signed magnitude whose sign
+has been published as `recentDirection` since #177.
+
+**Additive for `trend`, a genuine semantic change for `inboundRatePerSec`** — the same field name can
+now hold a different number for the same inputs. That is the honest description, and the reason this is
+a contract change rather than an implementation detail. It is nonetheless safe to land in one step,
+because the field had **never been populated by any engine build**: #188 found the only three mentions
+of it in the repo were in this contract. Nothing has consumed it, so nothing changes meaning underneath
+a consumer. The two things that will read it — `AgentDispatcher`'s prompt and `RaiseUndersizedPool` —
+are in the same PR.
+
+#188.
+
+### Why the original formula was wrong
+
+Arrival rate is completions plus the rate the backlog is growing, and **"is growing" has to mean now**.
+The window fit is 300 s. Minutes into a drain it still leans up, so the estimate lands *above* the raw
+completion rate on the one state where completions are already an overstatement.
+
+Measured on the live drain-through transient: `set_pool_size 1 -> 4` applied, `recentDirection`
+reporting `falling`, `messagesPerSec` reading 4 because four workers are clearing a backlog.
+
+| Arrival rate from | Value | Agent's recommendation |
+|---|---|---|
+| `messagesPerSec` alone | 4 | `set_pool_size 4 -> 8` |
+| `+ slope/60`, queue 94 | 4.57 | `set_pool_size 4 -> 6` |
+| `+ recentSlope/60`, queue 108 | **3.82** | **none** |
+
+The two derived rows are separate runs a few polls apart, not the same sample — the transient is short.
+What decides the outcome is which side of `messagesPerSec` each estimate lands on.
+
+So this is not a refinement of the original formula: it moved the number the wrong way. The flagship
+rise is the only scenario that ever exercised it, and there the two spans agree — which is exactly why
+the defect was invisible until #188 required both scenarios be measured.
+
+One caveat on attribution, because the table above would otherwise overclaim. The `none` outcome is
+produced by the corrected estimate **and** a new prompt directive that says to recommend nothing when
+`recentDirection` is `falling` and `inboundRatePerSec` is below `messagesPerSec`. Neither alone is
+sufficient: the directive cannot fire on an estimate that reads 4.57, and the estimate alone leaves the
+sizing formula free to justify a second increase. The contract change is the half that lives here.
+
+The field was ratified in the MVP 2 design, before either scenario existed to measure it against. That
+is not a criticism of the original spec; it is the argument for #188's instruction to re-measure both.
+
+### Why `recentSlope` is published rather than kept internal
+
+The prompt asks the model to state its sizing arithmetic. A term it cannot see is a term it either
+omits or invents — and the field it reaches for instead is `slope`, which is the wrong span. Measured
+under the *previous* prompt, which never named the derived field at all: the model read
+`inboundRatePerSec` off the snapshot JSON anyway and reported "Receiving approximately 4.69 messages
+per second", presenting a derived estimate as an observation. A number the model will use regardless is
+better named and caveated than hidden.
+
+It stays off `earlywarning-api.md` — §1.4's rule that no bare slope accompanies a withheld forecast does
+not distinguish the two spans, and `publishedProjection()`'s whitelist strips both.
+
+### Also corrected
+
+The §2.2 paragraph telling consumers to "treat a `null` `slope` on a crossed threshold as expected" was
+stale: #187 delivered `slope` as specified and did not update the prose. Both slopes are now populated
+on every investigation, and the section says so.
+
+`inboundRatePerSec` is now stated to clamp at `0`. The two terms span different windows, so a steep
+enough drain can put the sum negative; `null` would claim there was no fit.
+
+---
+
 ## 2026-09-01 — a tool for the interface map, and `trend` gains the field that says a queue is draining
 
 **Two files, both additive.** `mcp-tools.md` gains one read tool (§3.14, `get_interface_path`);
--- a/contracts/investigation-api.md
+++ b/contracts/investigation-api.md
@@ -124,6 +124,7 @@ The engine builds this from its own in-memory state. It is a closed allowlist:
   "trend": {
     "metric": "queued",
     "slope": 124.2,
+    "recentSlope": 124.2,
     "slopeUnit": "items/minute",
     "recentDirection": "rising",
     "thresholdValue": 50,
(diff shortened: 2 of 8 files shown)


Ari-Glikman:
Self-review (sole ownership; #144's precedent).

**Read the diff as a reviewer, three things checked and one changed.**

1. **Did the contract amendment stay inside §2.2's consumer?** Yes. `earlywarning-api.md` is untouched and `publishedProjection()`'s `Omit` now strips both slopes, so the compiler refuses a leak rather than a test catching one — which is how `api.test.ts`'s `PROJECTION` literal failed to typecheck the moment the field was added, and why that test now asserts the absence of both names plus the surviving ten keys.

2. **Is the measured table honest?** Corrected during review: the first draft of the contract and CHANGELOG tables quoted `+41/min`, `-132/min` and `1.8` — slopes I predicted before measuring. The four copies of the table (both contract files, both doc comments) now carry the measured set, and each states that the two derived rows are separate runs a few polls apart rather than one sample. The `4 -> 8` row is `main`'s behaviour, measured against the same state by recompiling `main`'s dispatcher, so the claim is comparative rather than asserted.

3. **Does `recentSlope` actually keep the promise its contract row makes?** It did not, and that is the one code change made during review. `buildTrend` gated on the window fit only, and the two fits share a sample-count gate but not an outcome — a poll gap can leave one sample inside the trailing 120 s while the 300 s window still holds twelve, in which case the tail fit and `recentDirection` are both null. `trend` would then have carried a window slope beside two nulls and a null `inboundRatePerSec`: exactly the two-meanings object the `||` in that gate was introduced to refuse. Third arm added, doc comment corrected from the overclaim it started as, and a test pins the state. **439/439.**

**What I would flag to a second reviewer, unresolved by choice:**

- Scenario 2 at 15 s post-apply still returns `4 -> 6`. That is in the PR body rather than smoothed over, because it is the limit of the approach and not a bug: 15 s in, the 120 s tail genuinely leans up, and any estimate that claimed otherwise would be guessing. The finding clears on its own well before an operator would act on it.
- The prompt directive and the corrected estimate are jointly, not separately, responsible for the `none` outcome. Both the CHANGELOG and the `.cls` comment say so.
- Unfiled drift noticed while reading §2.2, out of scope here: the engine omits `capturedAt` and all four `*Baseline` snapshot fields the section promises, and §2.2 calls the baselines load-bearing for the "~1s from the downstream throttle" evidence bullet.

PR #240, opened 2026-09-02
feat(dashboard): drag-resizable panels, and the triggers and evidence collapse

### LINKED ISSUE #239: dashboard: move the demo triggers up, collapse them and the evidence list, and make both panels drag-to-resize
## What

Four UI changes, all in `apps/dashboard/`, all requested from driving the live demo (@Ari-Glikman, 2026-09-02):

1. **Demo triggers move up the rail**, from pinned-to-the-bottom to directly under Production Guardian › Thresholds.
2. **The trigger list collapses behind a disclosure arrow**, collapsed by default.
3. **The left rail and the right-hand panel are drag-to-resize**, by their inner edge.
4. **The evidence list on a finding retracts**, so the recommended fix is visible without the evidence.

The exact ask on 3 was *"if we are talking about the compactability I'd prefer if it could be dragged left/right with the mouse"* — so this is a drag handle, not a collapse toggle. The exact ask on 4 was *"the evidence itself should be hidden if retracted and not shown below. if expanded one should be able to see as is the case now."*

## Why each one

**1 and 2 go together.** `.pg-triggers` carried `margin-top: auto` with the comment *"an audience should read the navigation above it first"*. That reasoning is unchanged and is now served by the list being collapsed rather than by distance, which is the stronger version of it — absent beats lower down, and it also stops four buttons that break a production on purpose being the tallest thing in the rail.

**Collapsing must not hide state.** `pool_bottleneck` reads `trigger activating…` for ~75s — the longest wait in the demo and the whole reason the middle phase exists (#135). A collapsed rail that said nothing about it would re-create that defect one layer up. So the closed heading carries a count and the phase word, with the same dot the buttons use, and `activating` wins over `activated` because it is the one describing a wait somebody is sitting through.

**3 is one hook for two panels.** They differ in which custom property they drive and which edge the handle sits on; the clamp, the pointer capture, the keyboard fallback and the persistence are shared, and the half that drifts between two copies is always the keyboard half.

**4's initial state is the load-bearing part.** The evidence sits *between* the root cause and the recommended action — deliberately, because it is what the recommendation rests on — and four to six bullets is enough to push Approve below the fold of a 420px drawer. An expand-by-default toggle would leave the original complaint exactly where it was. What stays visible while closed: the whole badge row (live-vs-canned, model, tool count, confidence), plus the item count and how many were **read from IRIS** rather than asserted by the model. `InvestigationPanel`'s point 3 is that provenance is part of the evidence; a disclosure that hid the count silently would let a canned narrative and a governed one look identical while closed.

## Implementation notes worth reviewing

- **The width is written to a custom property on `<html>`, not as an inline `width`.** Neither target would obey an inline width correctly: the rail is a *grid column* on `.pg-shell`, so a width on the rail element does nothing at all; and `.pg-drawer` narrows to `min(var(--pg-drawer-width), 60vw)` under 1100px, which an inline width would override and so cover the grid on a 1024px laptop. Same arrangement as `data-theme` — JS decides a value, CSS decides what it means.
- **Default, min and max are tokens**, read back off the document rather than written in TS, per §9. That is what makes the reset target unable to disagree with the width the page loads at. The read is cached per property *and* taken during the first render, before the hook has written anything — after that the computed value is our own override.
- **Pointer capture, not window listeners.** Nothing is attached to `window`, so there is no listener to leak and no drag that outlives its element.
- **`role="separator"`, the one exception to §7.3's "real `<button>`".** This is the ARIA window-splitter pattern; a button would announce the wrong thing and have no way to report a width. Arrow keys, Shift+arrows, Home/End, and Enter/Space or double-click to reset — a drag-only control would not meet the bar.
- **All three right-edge panels share one width.** `FindingDetail`, `HostDetail` and `ThresholdSettings` are mutually exclusive by construction, so `DrawerResizeHandle` exists to keep them from naming three different storage keys.
- **The `min(…, 60vw)` clamp outranks a drag, deliberately.** On a 1024px viewport the last stretch of a drag has no visible effect. The reason the drawer is not a modal does not stop applying because someone pulled its edge.
- **The rail handle is hidden under 1100px**, where the media query pins the rail to an icon strip and the width token is not being read. An absent handle says that; an inert one invites the drag and reports nothing.
- **Neither disclosure state is persisted**, unlike the widths. A remembered "expanded" would put the scaffolding — or the below-the-fold fix — back for whoever opens the dashboard next, which is the state this change exists to avoid.

## Verification

- `node node_modules/typescript/bin/tsc --noEmit` — clean, exit 0. (`npx tsc` cannot run on this machine; `node_modules/.bin` has no `tsc` symlink.)
- `node tools/validate-architecture.mjs` — ok, and the fixture validators pass.
- `docker compose up -d --build dashboard` — the build that actually ships (§6.1) — succeeds, `http://localhost:5173/` serves 200, and the shipped bundle contains `pg-resize--rail`, `pg-resize--drawer`, `pg-triggers__toggle`, `pg-disclose__count`, `read from IRIS`, `data-pg-resizing`, `--pg-rail-width-min`, `--pg-drawer-width-max`, and **zero** occurrences of `margin-top:auto`.
- `npm run build` on the host still fails on `@rollup/rollup-win32-x64-msvc`, which is the pre-existing half-installed `node_modules` on this machine and unrelated to this change.

## Not in scope

No engine, contract, or `iris/**` change — this is display only. No new dependency.

Ari-Glikman:
Closes #239.

Four display-only changes in `apps/dashboard/`, all from driving the pool-bottleneck demo (@Ari-Glikman, 2026-09-02):

1. **Demo triggers move up the rail**, to directly under Production Guardian › Thresholds.
2. **The trigger list collapses behind an arrow**, closed by default.
3. **The rail and the right-hand panel drag by their inner edge.**
4. **The evidence list on a finding collapses**, so the recommended fix is readable without it.

The ask on 3 was explicitly a drag, not a toggle — *"I'd prefer if it could be dragged left/right with the mouse"* — and on 4, *"the evidence itself should be hidden if retracted and not shown below."*

## Files

| File | |
|---|---|
| `hooks/useResizable.ts` | **new** — clamp, persistence, pointer capture, keyboard, reset |
| `components/ResizeHandle.tsx` | **new** — presentational `role="separator"`, renders `null` when disabled |
| `components/DrawerResizeHandle.tsx` | **new** — the drawer's hook + handle, so three panels name one key |
| `AppShell.tsx` | the rail's handle, one inline hook call |
| `FindingDetail.tsx`, `HostDetail.tsx`, `ThresholdSettings.tsx` | `<DrawerResizeHandle />` as each `<aside>`'s first child |
| `TriggerRail.tsx` | heading becomes a disclosure button carrying a phase summary |
| `InvestigationPanel.tsx` | evidence behind a disclosure carrying the counts |
| `styles/tokens.css` | `--pg-rail-width-min/-max`, `--pg-drawer-width-min/-max` |
| `styles/app.css` | `.pg-resize*`, `.pg-triggers__toggle*`, `.pg-disclose*`; `.pg-triggers` loses `margin-top: auto` |

## Review notes — the decisions, not the diff

**The width is a custom property on `<html>`, not an inline `width`.** Neither target would obey an inline width correctly. The rail is a *grid column* on `.pg-shell` (`grid-template-columns: var(--pg-rail-width) minmax(0,1fr)`), so a width on the rail element does nothing at all. And `.pg-drawer` narrows to `min(var(--pg-drawer-width), 60vw)` under 1100px — an inline width would outrank that clamp and cover the grid on a 1024px laptop, the exact viewport §7.3 sets as the floor. Same split as `data-theme` in `lib/theme.ts`: JS picks a value, CSS decides what it means.

**Default, min and max are tokens, read back off the document** rather than written in TS, per §9's no-magic-numbers rule. That is what makes the double-click reset target unable to disagree with the width the page loads at. `tokenPx()` caches per property in a module `Map`, and the read happens during the **first render, before the hook has written anything** — after that the computed value is our own override, so a later read would return whatever was last dragged to. If a `-min`/`-max` token is ever dropped, `enabled` goes false, the handle stops rendering and a `console.warn` names the property: resizing disappears loudly rather than clamping to `NaN`.

**Pointer capture, not window listeners.** `setPointerCapture` routes subsequent moves to the handle and the browser releases implicitly on `pointerup`, so nothing is attached to `window` — no listener to leak, no drag that outlives its element. `onLostPointerCapture` also ends the drag so `dragging` cannot latch if capture is stolen. Deltas are measured from the drag **origin** rather than accumulated per event, so hitting a clamp edge does not desync the handle from the pointer.

**`role="separator"` is the one documented exception to §7.3's "a real `<button>` for anything clickable."** This is the ARIA window-splitter pattern (`aria-orientation`, `aria-valuenow/min/max`, `tabIndex={0}`); a button would announce the wrong role and have no way to report a width. Arrow keys move by `--pg-space-4` so a keyboard step lands on the 4px scale, Shift moves by 64, Home/End jump to the bounds with the sign keyed off which edge the handle is on, and Enter/Space or a double-click resets. A drag-only control would not clear the bar.

**All three right-edge panels share one width.** `FindingDetail`, `HostDetail` and `ThresholdSettings` are mutually exclusive by construction — `App` clears the other two selections on each — so `DrawerResizeHandle` exists to keep three call sites from naming three storage keys. The custom property is deliberately **never** cleaned up on unmount, so switching between them keeps the width.

**The `min(…, 60vw)` clamp outranks a drag, on purpose.** On a narrow viewport the last stretch of a drag has no visible effect. The reason the drawer is not a modal — the operator should keep seeing the grid — does not stop applying because someone pulled its edge. And the **rail handle is hidden under 1100px**, where the media query pins the rail to an icon strip and the width token is not being read: an absent handle says that, where an inert one would invite the drag and report nothing.

**Collapsing must not hide state — twice.** `pool_bottleneck` reads `trigger activating…` for ~75s, the longest wait in the demo and the entire reason the middle phase exists (#135); the closed heading therefore carries the count and the phase word with the same dot the buttons use. And on the evidence side, `InvestigationPanel`'s point 3 is that provenance is part of the evidence — so the closed control states the item count *and* how many were **read from IRIS** rather than asserted by the model, and the badge row above it (live-vs-canned, model, tool count, confidence) is never collapsed. A disclosure that hid the count would let a canned narrative and a governed one look identical while closed.

**Closed-by-default is the load-bearing choice for the evidence.** It sits *between* the root cause and the recommended action, deliberately, because it is what the recommendation rests on — and four to six bullets is enough to push Approve below the fold of a 420px drawer, which is the complaint being fixed. Expand-by-default would leave it exactly where it was. Neither disclosure state is persisted, unlike the widths: a remembered "expanded" would put the scaffolding, or the below-the-fold fix, back for whoever opens the dashboard next.

**Both lists use `hidden`, not unmounting.** `TriggerNote`'s live regions are marked ALWAYS MOUNTED so a phase change announces, and `aria-controls` has to point at an element that exists.

**`.pg-evidence`'s base rule is untouched** because `ActivityChat` renders it too; the spacing fix is scoped as `.pg-disclose + .pg-evidence`. The chevron rotates on an explicit `--open` class rather than an `[aria-expanded='true']` selector, matching AppShell's stated preference — the attribute selector would catch any expandable control added here later.

## Verification

- `node node_modules/typescript/bin/tsc --noEmit` — clean, exit 0. (`npx tsc` cannot run here: `node_modules/.bin` has no `tsc` symlink.)
- `node tools/validate-architecture.mjs` — `ok (8 labels and 12 paths, no overlaps, all within the viewBox)`; `validate-fixtures` and `validate-fixture-claims` exit 0.
- **`docker compose up -d --build dashboard`** — the build that ships (§6.1), which runs the real `tsc && validate-architecture && vite build`. Succeeds; `http://localhost:5173/` serves 200; all four containers up.
- The shipped bundle contains `pg-resize--rail`, `pg-resize--drawer`, `data-pg-resizing`, `pg-triggers__toggle`, `pg-triggers__chevron`, `pg-disclose__count`, `read from IRIS`, `--pg-rail-width-min`, `--pg-drawer-width-max`, and **zero** occurrences of `margin-top:auto`.
- `npm run build` on the host still fails on `@rollup/rollup-win32-x64-msvc` — pre-existing half-installed `node_modules` on this machine (npm/cli#4828), unrelated.

**Not verifiable from here:** the interaction itself. Dragging both edges and the two arrows need a human on `http://localhost:5173/`; there is no headless browser in this environment, and saying otherwise would be the green-claim-over-a-red-build failure §6 names.

## Not in scope

No engine, `contracts/` or `iris/**` change. No new dependency. No host name or count in `src/` (§9). No colour or spacing literal — the four new geometry values are tokens.

### DIFF
--- a/apps/dashboard/src/components/AppShell.tsx
+++ b/apps/dashboard/src/components/AppShell.tsx
@@ -47,8 +47,10 @@ import {
   IconShield,
   type IconProps,
 } from './icons';
+import { ResizeHandle } from './ResizeHandle';
 import { ThemeToggle } from './ThemeToggle';
 import { TriggerRail } from './TriggerRail';
+import { useResizable } from '../hooks/useResizable';
 
 /** Which top-level view is on screen. */
 export type View = 'dashboard' | 'brochure' | 'architecture';
@@ -137,6 +139,16 @@ export function AppShell({
   settingsOpen,
   children,
 }: AppShellProps): JSX.Element {
+  /* Drives `--pg-rail-width`, which `.pg-shell`'s grid column reads — see `useResizable`. Owned here
+     rather than by the rail's contents because the rail is not the element that is sized; the shell
+     is, and this is the component that renders both. */
+  const rail = useResizable({
+    variable: '--pg-rail-width',
+    storageKey: 'pg.railWidth.v1',
+    edge: 'end',
+    label: 'Navigation rail width',
+  });
+
   return (
     <div className="pg-shell">
       <nav className="pg-rail" aria-label="Health Connect">
@@ -211,6 +223,12 @@ export function AppShell({
             rail so an audience reads the product's own navigation first and the demo scaffolding
             after it, and so its absence leaves the rail visually unchanged. */}
         <TriggerRail />
+
+        {/* Last of all, and absolutely positioned on the rail's own right edge — so it tracks the
+            column whatever the rail contains, including the demo triggers being off. Hidden under
+            1100px, where the media query pins the rail to an icon strip and a drag would appear to
+            do nothing. */}
+        <ResizeHandle {...rail} className="pg-resize--rail" />
       </nav>
 
       <header className="pg-header">
--- a/apps/dashboard/src/components/DrawerResizeHandle.tsx
+++ b/apps/dashboard/src/components/DrawerResizeHandle.tsx
@@ -1,0 +1,28 @@
+/**
+ * The right-hand drawer's resize handle, options and all.
+ *
+ * A WRAPPER BECAUSE THERE ARE THREE CALL SITES. `FindingDetail`, `HostDetail` and `ThresholdSettings`
+ * are the three panels that render `.pg-drawer`, they are mutually exclusive by construction (see
+ * `App.tsx`), and they share `--pg-drawer-width` — so whichever is open must resize the same thing to
+ * the same remembered value. Three copies of the same `useResizable` options is three chances for one
+ * of them to name a different storage key and quietly stop agreeing. The rail has one call site and
+ * so calls the hook directly in `AppShell`.
+ *
+ * `edge: 'start'` — the handle sits on the drawer's LEFT edge, so dragging right narrows it. Dragging
+ * right always moves the separator right; which side gains is a consequence of which side the panel
+ * is on.
+ */
+
+import { ResizeHandle } from './ResizeHandle';
+import { useResizable } from '../hooks/useResizable';
+
+export function DrawerResizeHandle(): JSX.Element | null {
+  const drawer = useResizable({
+    variable: '--pg-drawer-width',
+    storageKey: 'pg.drawerWidth.v1',
+    edge: 'start',
+    label: 'Panel width',
+  });
+
+  return <ResizeHandle {...drawer} className="pg-resize--drawer" />;
+}
--- a/apps/dashboard/src/components/FindingDetail.tsx
+++ b/apps/dashboard/src/components/FindingDetail.tsx
@@ -34,6 +34,7 @@ import {
   formatRate,
   formatRelative,
 } from '../lib/format';
+import { DrawerResizeHandle } from './DrawerResizeHandle';
 import { SeverityBadge } from './SeverityBadge';
 import { IconClose } from './icons';
 
@@ -117,6 +118,11 @@ export function FindingDetail({
       aria-modal="false"
       aria-labelledby="pg-drawer-title"
     >
+      {/* First child, on the drawer's own left edge, so it tracks the real rendered width — including
+          the `min(…, 60vw)` narrowing under 1100px, which a handle positioned from the viewport
+          would not. */}
+      <DrawerResizeHandle />
+
       <header className="pg-drawer__header">
         <div className="pg-drawer__heading">
           <SeverityBadge severity={finding.severity} />
--- a/apps/dashboard/src/components/HostDetail.tsx
+++ b/apps/dashboard/src/components/HostDetail.tsx
@@ -36,6 +36,7 @@ import { useEffect } from 'react';
 import type { FindingView, HostView, Severity } from '../types/healthscan';
 import type { HostSeriesView } from '../types/hostseries';
 import { formatCount, formatDuration, formatRate, formatRelative } from '../lib/format';
+import { DrawerResizeHandle } from './DrawerResizeHandle';
 import { MetricChart } from './MetricChart';
 import { SeverityBadge } from './SeverityBadge';
 import { StatusDot } from './StatusDot';
@@ -118,6 +119,9 @@ export function HostDetail({
       aria-modal="false"
       aria-labelledby="pg-host-drawer-title"
     >
+      {/* One width for all three right-edge panels — see `DrawerResizeHandle`. */}
+      <DrawerResizeHandle />
+
       <header className="pg-drawer__header">
         <div className="pg-drawer__heading">
           <h2 id="pg-host-drawer-title" className="pg-drawer__title">
--- a/apps/dashboard/src/components/InvestigationPanel.tsx
+++ b/apps/dashboard/src/components/InvestigationPanel.tsx
@@ -31,9 +31,17 @@
  * be explicit for the same reason: nothing about a write should be inferred.
  */
 
+import { useState } from 'react';
 import type { InvestigationView, ResolveActionView, ResolveMode, ResolveView } from '../types/mvp2';
 import { ABSENT, toolLabel } from '../lib/format';
 import { findingMeta, investigationScope } from '../lib/findingMeta';
+import { IconChevronRight } from './icons';
+
+/**
+ * `aria-controls` needs an id, and the three right-edge panels are mutually exclusive by construction
+ * (see `App.tsx`), so at most one of these lists is in the document at a time.
+ */
+const EVIDENCE_ID = 'pg-evidence-list';
 
 export interface InvestigationPanelProps {
   /** The finding's type, for the §2.4 scope check. Only the type is needed, so only it is passed. */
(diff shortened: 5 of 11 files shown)


Ari-Glikman:
Self-review under the standing single-owner arrangement (root `CLAUDE.md` §3 — two developers, and Dev A does not work in `apps/dashboard/`). Read as a reviewer rather than as the author: what follows is what I went looking for and what I found.

## What I checked, and what it showed

**Does the custom-property approach actually reach both targets?** Yes, and the two reasons in the file comment are both real, verified against the CSS rather than remembered:

- `.pg-rail` is `grid-area: rail`, and `.pg-shell` resolves the column from `var(--pg-rail-width)`. A `width` on the rail element genuinely does nothing.
- `.pg-drawer` is `position: fixed; width: var(--pg-drawer-width)` with `width: min(var(--pg-drawer-width), 60vw)` under 1100px. An inline width would have outranked that.

**Are the handles positioned against the right containing block?** `.pg-rail` gained `position: relative` in this diff and `.pg-drawer` was already `position: fixed`, so both `absolute` handles resolve against their own panel. Neither ancestor clips: the only `overflow: hidden` near the top of `app.css` is `.pg-visually-hidden`, not the shell.

**Does the 10px target cover anything clickable?** No. The rail's horizontal padding is `var(--pg-space-3)` = 12px and the drawer's body padding is larger, so the strip sits in padding on both. Rail items end 12px short of the edge the handle occupies.

**Can the rail handle (`z-index: 30`) paint over the drawer (`z-index: 20`)?** Only if they overlapped, and they cannot: at the narrowest viewport where the rail handle exists (1101px) a 720px drawer starts at 381px, well right of a rail capped at 420px — and below 1100px the handle is `display: none` anyway.

**Can `dragging` latch on?** Both `pointerup` and `lostpointercapture` call the same `endDrag`, which is idempotent (`origin.current === null` early-returns). The `data-pg-resizing` effect cleans up on unmount, so an unmount mid-drag cannot leave the page `user-select: none`.

**Can the width ever be written as `NaN`?** No. `enabled` is false unless all three tokens parsed, and every write goes through `clamp`. When it is false nothing is written at all and `ResizeHandle` returns `null`, so a missing token degrades to "this panel does not resize" with a `console.warn` naming the property.

**Does the token cache defeat itself?** This is the subtle one and it holds: `TOKEN_PX` is populated during the first render of each hook, before that hook's write effect runs. The rail reads at mount; the drawer reads the first time any drawer opens, which is also before any drawer write. Every later mount hits the cache and so still sees the token, not the override.

**Hook-order safety in `InvestigationPanel`.** `useState` sits above the `scope !== 'investigable'` early return, so the hook count does not change when the panel switches subject.

## Two things I'd flag to a reviewer as accepted, not overlooked

**`aria-valuenow` on the drawer can overstate below 1100px**, because the rendered width is `min(width, 60vw)` and the reported value is the variable. Left as is: the alternative is reporting a number that the token no longer explains, and the clamp is documented in both the media query and the hook comment. The rail case has no such gap — the handle is hidden exactly where its token stops being read.

**`STEP = 16` is a literal in TS rather than a read of `--pg-space-4`.** Deliberate asymmetry with the bounds: the bounds must agree with CSS or the reset lies, whereas the step is an input granularity that nothing in CSS consumes. Reading it would add a fourth `getComputedStyle` for no invariant. The comment says which 16 it is.

## Rules I checked this against

No host name or count in `src/` (§9). No new dependency. No colour, spacing or radius literal — the four new geometry values went to `tokens.css`, in `:root` only, because §7.1 keeps layout tokens out of the dark block. `contracts/` untouched. No `fetch` added. Nothing outside `apps/dashboard/`. Every new file is one component per file, named for the file, and under the ~150-line guidance.

## Verification, restated for the record

`tsc --noEmit` clean; `validate-architecture` and both fixture validators pass; `docker compose up -d --build dashboard` — the build that ships (§6.1) — succeeds and serves 200 on `:5173`; the shipped bundle carries every new class and token and has zero `margin-top:auto`. CI's **dashboard — typecheck + build** is green on this head.

**The interaction itself is unverified and I am not claiming otherwise.** There is no browser in this environment, so dragging both edges and clicking the two arrows needs a human. Everything above is structural.

Approving on that basis: the mechanism is verified, the behaviour needs eyes.

PR #242, opened 2026-09-02
fix(dashboard): the host and Thresholds panels ignored the width their handle drives

### LINKED ISSUE #241: dashboard: the host and Thresholds panels cannot be resized — .pg-drawer--host/--settings hardcode a width that outranks --pg-drawer-width
## Symptom

Click a host, and the right-hand panel cannot be dragged. Reported from the live demo (@Ari-Glikman, 2026-09-02).

The Thresholds panel has the same defect. The finding drawer is fine, which is why this got through — that is the one that was driven while building #239.

## Root cause

Not the handle, and not `useResizable` — the variable is being written correctly. Two CSS rules override the width the handle drives. From the shipped bundle, in source order:

```css
.pg-drawer{ … width:var(--pg-drawer-width) … }        /* what the handle drives */
.pg-drawer{ width:min(var(--pg-drawer-width),60vw) }  /* the <1100px narrowing */
.pg-drawer--host{ width:min(480px,92vw) }             /* last, and wins */
.pg-drawer--settings{ width:min(480px,92vw) }
```

`.pg-drawer--host` and `.pg-drawer--settings` are **the same specificity** as `.pg-drawer` (one class each) and **later in the file** — `app.css:2311` and `:2520` against `:871`. So source order decides, and the two variants win. `--pg-drawer-width` changes on every pointermove; those two panels have no rule that reads it.

Both overrides predate #239 by a long way and were entirely reasonable when written: each is a deliberately wider surface with a comment saying why (three charts and six metric rows for the host; a label, an input, two lines of prose and two reference values per row for the settings form). Neither could have anticipated a rule that needed to move the width from outside.

**A second thing falls out of the same source-order win, and it is not new.** Those two panels also outranked the `@media (max-width: 1100px)` rule, so they have been using `92vw` rather than `60vw` the whole time. At 1024px that is 942px — very nearly the entire viewport, which is what `.pg-drawer`'s own comment says a drawer exists *not* to do ("the operator must keep seeing the grid behind it"). Worth fixing in the same change since the fix is the same deletion.

## Why the review missed it

The self-review on #240 checked that the base `.pg-drawer` rule reads the variable, and it does. It did not check whether anything **later in the file** re-declares `width` on the same element at equal specificity. Reading a rule and confirming it is the *winning* rule are two different checks, and only the first was done.

The verification missed it for a narrower reason: the shipped bundle was grepped for the presence of every new class and token, which all passed. Presence is not precedence. A grep that had asked "what else sets `width` on `.pg-drawer`" would have found this in one line — and that grep is what actually located it now.

## Fix

Delete both `width` declarations and let all three panels read the token.

- The **reason** for 480px was right for two of the three panels, so it becomes the shared default in `tokens.css` (`--pg-drawer-width: 420px → 480px`). The finding drawer loses nothing at 480px, and an operator who wants it back at 420 can now drag it there — which is the whole point of the feature.
- `-min: 320px` / `-max: 720px` are unchanged; 480 sits inside them.
- The `92vw` goes with it, so all three narrow to `min(…, 60vw)` under 1100px as `.pg-drawer` always intended.

The per-panel default matters less than it looks: the three share one storage key by design (#239 — they are mutually exclusive in `App`), so after the first drag they are all the same width anyway. The token is the width they *start* at.

## Guard

Nothing in CI can catch a CSS precedence bug today, and a test for this specific pair would not generalise. The durable form is a lint over `app.css`: **no rule may re-declare a property that a resize handle drives.** Filed as a follow-up rather than bundled here, so the fix the demo needs is not gated on designing a checker.

Ari-Glikman:
Closes #241. Reported from the live demo: *"the right panel cannot be moved when i click on a host."*

Confirmed, and it affected the Thresholds panel too. The finding drawer was fine, which is why #240 shipped with it.

## Root cause — CSS precedence, not the handle

The handle and `useResizable` were doing their job; `--pg-drawer-width` updates on every pointermove. Two rules override the property it drives. From the bundle that was actually running, in source order:

```css
.pg-drawer{ … width:var(--pg-drawer-width) … }        /* what the handle drives */
.pg-drawer{ width:min(var(--pg-drawer-width),60vw) }  /* the <1100px narrowing */
.pg-drawer--host{ width:min(480px,92vw) }             /* last, and wins */
.pg-drawer--settings{ width:min(480px,92vw) }
```

`.pg-drawer--host` and `.pg-drawer--settings` are **one class each — the same specificity as `.pg-drawer`** — and sit at `app.css:2311` and `:2520` against `:871`. Source order decides, so the variants win and nothing in those two panels ever reads the variable.

Both overrides predate the resize handle by a long way and each had a stated, good reason. Neither could have anticipated a rule needing to set that width from outside.

**A second defect falls out of the same win, and it is not new.** Those two panels also outranked the `@media (max-width: 1100px)` rule, so they have been using `92vw` rather than `60vw` throughout. At 1024px that is 942px — very nearly the whole viewport, and precisely what `.pg-drawer`'s own comment says a drawer exists *not* to do. Fixed here because it is the same deletion.

## The fix

Delete both `width` declarations; the two rules had nothing else in them, so both are now comment-only blocks recording what was there and why it had to go.

- **`--pg-drawer-width: 420px → 480px`.** Not a new number — it is the one both variants hardcoded, and it was right for two of the three panels (charts and six metric rows; a form row with prose and two reference values). The finding drawer loses nothing at 480, and 420 is now a drag away, which is what the feature is for.
- `-min: 320px` / `-max: 720px` unchanged; 480 sits inside them.
- All three panels now clamp to `min(…, 60vw)` under 1100px.

A per-panel default was never really on the table: the three share one storage key by design (#239 — they are mutually exclusive in `App`), so after the first drag they are identical anyway. The token is only what they *start* at.

Also dropped three now-stale `420px` mentions in comments rather than re-pointing them at 480 — a width copied into prose is the value that goes stale next.

## Why #240's review missed it

Worth writing down, because the miss is more reusable than the bug. The self-review checked that `.pg-drawer` reads the variable. It does. It never asked whether a **later rule re-declares the same property at equal specificity**. And the bundle verification grepped for the presence of every new class and token — all present. *Presence is not precedence.* The grep that actually found this was "what else sets `width` on `.pg-drawer`", which is one line and now the one I'd run first.

## Verification — in the shipped bundle, since that is where it was proven

```
$ grep -o "\.pg-drawer[^{,]*{[^}]*width[^}]*}" index.html
.pg-drawer{ … width:var(--pg-drawer-width) … }
.pg-drawer{width:min(var(--pg-drawer-width),60vw)}
```

Only two, and both read the token. Also: `--pg-drawer-width: 480px`, `92vw` appears **zero** times, and both `pg-resize--drawer` occurrences are intact.

`node node_modules/typescript/bin/tsc --noEmit` clean · `validate-architecture` ok · `docker compose up -d --build dashboard` builds and serves 200 on `:5173`.

**Still needs eyes:** that the host panel now actually drags. The mechanism is proven from the CSS; the gesture needs a human, same caveat as #240.

## Follow-up not bundled here

Nothing in CI can catch a CSS precedence bug. The durable guard is a lint over `app.css` — *no rule may re-declare a property a resize handle drives* — filed separately so the fix the demo needs is not gated on designing a checker.

### DIFF
--- a/apps/dashboard/src/components/InvestigationPanel.tsx
+++ b/apps/dashboard/src/components/InvestigationPanel.tsx
@@ -108,8 +108,8 @@ export function InvestigationPanel({
    * COLLAPSED IS THE INITIAL STATE, and that is the load-bearing half of the request rather than a
    * default picked for tidiness. The evidence sits BETWEEN the root cause and the recommended action
    * in this panel — deliberately, because it is what the recommendation rests on — and four to six
-   * bullets is enough to push Approve below the fold of a 420px drawer. An expand-by-default toggle
-   * would leave that complaint exactly where it was.
+   * bullets is enough to push Approve below the fold at any width the drawer opens at. An
+   * expand-by-default toggle would leave that complaint exactly where it was.
    *
    * WHAT MUST NOT BE HIDEABLE IS PROVENANCE, so two things stay outside the collapse: the badge row
    * above (live-vs-canned, model, tool count, confidence) is never collapsed, and the closed toggle
--- a/apps/dashboard/src/styles/app.css
+++ b/apps/dashboard/src/styles/app.css
@@ -1085,7 +1085,7 @@ html[data-pg-resizing] {
     margin-right: 0;
   }
 
-  /* At 1024px a 420px drawer would cover most of the grid, defeating the reason
+  /* At 1024px a drawer at its default width covers most of the grid, defeating the reason
      it is a drawer and not a modal. Narrower, so the grid stays visible beside it.
 
      THIS CLAMP OUTRANKS A DRAG, deliberately. `--pg-drawer-width-max` is 720px, so on a
@@ -2307,13 +2307,22 @@ html[data-pg-resizing] {
 
 /* The two panels are mutually exclusive in `App`, so this shares the drawer's z-index
    rather than stacking above it -- if it ever needed a stacking order, that would mean two
-   were open at once, which is the defect the exclusivity prevents. */
-.pg-drawer--host {
-  /* A slightly wider surface than the finding drawer: three charts plus six metric rows do
-     not read at 420px, and the charts are the reason the panel exists. Still `min()` against
-     the viewport so the grid stays visible beside it, which is why this is a drawer. */
-  width: min(480px, 92vw);
-}
+   were open at once, which is the defect the exclusivity prevents.
+
+   THERE IS DELIBERATELY NO `.pg-drawer--host` RULE, and the absence is the fix for #241.
+   It used to carry `width: min(480px, 92vw)` -- wider than the finding drawer because three
+   charts and six metric rows did not read at the finding drawer's then-420px, which was a good
+   reason. But it is the
+   same specificity as `.pg-drawer` and later in this file, so source order made it the
+   winning rule, and it therefore ignored BOTH the width the resize handle drives and the
+   `min(..., 60vw)` narrowing below 1100px. The panel an operator most wants wider was the
+   one that could not be dragged at all.
+
+   The 480px was right, so it moved to `--pg-drawer-width` in `tokens.css` and all three
+   panels now start there. The 92vw did not move: at 1024px it resolves to 942px, which is
+   very nearly the whole viewport and exactly what `.pg-drawer` says a drawer exists not to
+   cover. A per-panel width cannot survive here anyway -- the three share one remembered
+   width by design, so a variant could only ever change what they start at. */
 
 .pg-host-detail__status {
   display: flex;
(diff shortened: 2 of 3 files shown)


Ari-Glikman:
Self-review under the standing single-owner arrangement (root `CLAUDE.md` §3).

**The one check that matters here, since it is the check #240 skipped:** after the change, what sets `width` on a `.pg-drawer` element in the shipped bundle?

```
.pg-drawer{ … width:var(--pg-drawer-width) … }
.pg-drawer{width:min(var(--pg-drawer-width),60vw)}
```

Two rules, both reading the token, nothing later re-declaring it. `92vw` is gone from the bundle entirely, so the 1100px narrowing now applies to all three panels as `.pg-drawer` always intended.

**Did anything else depend on those two rules?** No — each contained only `width`, so both rules are gone rather than emptied. `.pg-drawer--host` and `.pg-drawer--settings` remain as DOM classes with no CSS behind them, which is deliberate: they identify which panel is open, the comments that replaced the rules reference them by name, and removing them from the `className` would make those comments dangle.

**Is 480 in range?** `-min: 320` / `-max: 720`, so yes, and a double-click resets to it.

**One behaviour change worth naming rather than leaving to be discovered:** anyone who already dragged the finding drawer has a stored `pg.drawerWidth.v1`, and the host panel now honours it — so it will open at whatever they last dragged to, not at 480. That is the shared-width design working as specified (#239), not a regression, and a double-click on the handle resets all three.

**Rules checked:** no host name or count in `src/`. No new dependency. No new literal — the change removes two magic numbers and moves one into a token. `contracts/` untouched. Nothing outside `apps/dashboard/`.

`tsc --noEmit` clean, `validate-architecture` ok, `docker compose up -d --build dashboard` builds and serves 200, all six CI checks green on this head.

The gesture itself still needs a human, same as #240. The mechanism is proven; #243 is the guard so the mechanism does not need re-proving by hand next time.

PR #245, opened 2026-09-07
feat(demo): cut the pool_bottleneck settle 75s -> 30s for a 5-minute slot

### LINKED ISSUE #244: pool_bottleneck: cut the settle 75s -> 30s for a 5-minute slot, and record why it costs nothing
## Ask

> "the demo has been changed to be only 5 minutes so we need to demo the pol bottleneck quicker. can we have the baseline be 30 seconds instead of the 75?"
> — @Ari-Glikman, 2026-09-07

`PoolBottleneck()`'s `pWarmSeconds` default goes **75 → 30**.

## It costs nothing, and that took checking

The wait was documented as load-bearing in five places. It is not, and it had not been for a while — **both** justifications on the record are dead:

**Dead #1 — `queue_buildup`'s rolling baseline had to warm at zero.** Superseded when `thresholds.json` gained `referenceBaselines` stating `Cloud API`.`queued = 0`. `effectiveBaseline()` returns the *stated* value instead of the rolling mean, so there is no `minBaselineSamples` gate on that path at all. This was already recorded as dead in the method comment.

**Dead #2 — Early Warning's `minFitSamples` (12) × the 5s poll = 60s.** This one was still believed, and it is the interesting failure. `BaselineStore.recent(host, metric, now, spanMs)` filters **only** on `fitWindowSeconds` and **deliberately ignores `#regimeStart`** (its own comment says so). So `fitSampleCount` is a function of **engine uptime**, not of anything the trigger does, and it survives a `Reset all` (`beginRegime`) intact. No arming sequence can affect it.

Measured, both runs, at the instant the arm POST was sent:

```
run 1   fitSampleCount=20  fitSpanSeconds=95
run 2   fitSampleCount=49  fitSpanSeconds=240
```

Twelve samples were banked long before the button was pressed. The wait was never producing them.

**The lesson, which is why the autopsy goes in the code and not just here:** the comment reasoned from a threshold *name* (`minFitSamples`) to a delay, without checking where the samples come from. Both times, the delay outlived its reason by weeks — because the reason was written down and the check was not.

**The one genuinely cold case is a cold engine** — the first ~60s after `compose up` or an engine restart. 30s of settling does not cover that and is not meant to; it is a precondition on the stack, which #108's pre-demo check already requires for four other reasons. Now stated in the cue sheet's §1.2 as the only real precondition.

## So what does the 30s buy

Pacing, and that is now the stated reason: the room sees a healthy production for a beat before the fault, and the presenter gets to say what is about to happen while nothing is happening. `A NUMBER TO PUT BACK IF THE FIT IS EVER COLD ON PURPOSE: 60s, not 75` — the 75 was 60 plus margin for a reason that no longer holds, and nothing measured ever needed the extra 15.

## Measured

Two consecutive arms, containerised stack, warm engine, 2026-09-07. `t` is seconds from the arm POST, which returns in **0.02s** because the dispatcher jobs this scenario:

| | run 1 | run 2 |
|---|---|---|
| armed (`DispatcherMs` set) | +30s | +30s |
| first Early Warning projection | +40s | +40s |
| queue crosses the floor of 50 | +91s | +86s |
| confirmed `queue_buildup` | +96s | +91s |

The Early Warning strip, run 2, is now the cue sheet's §1.2b table verbatim:

```
+40s  q=4   Queue depth 4 rising ~3.8/min; at this rate it crosses 50 in ~12 min.
+50s  q=14  rising ~19.6/min; ... in ~2 min.
+61s  q=24  rising ~39.4/min; ... in ~40 s.
+71s  q=34  rising ~55.2/min; ... in ~18 s.
+81s  q=45  rising ~61.4/min; ... in ~5 s.
+86s  q=52  already_crossed
```

The rate converges on ~61/min — one message a second, exactly the net inflow the scenario creates.

## Two alternatives ruled out

- **Lower `POLL_INTERVAL_MS`.** `thresholds.json` documents the invariant `(sustainedSamples-1) × interval > sustainedSeconds × 1000`, i.e. `1 × 5000 > 4000`; 4500ms already breaks it. Declined twice already (#64, #65).
- **Lower `minFitSamples`.** Unnecessary once `recent()`'s behaviour is understood, and it would weaken a published slope for no gain.

## Blast radius — every site that quoted 75 or 77

The number was copied into nine files' worth of prose. All updated, and each updated to say *what is now true* rather than just swapping a digit:

| File | What changed |
|---|---|
| `iris/labdemo/Triggers.cls` | the default; the autopsy of both dead reasons; the measured table |
| `iris/labdemo/REST/TriggerDispatcher.cls` | `SLOWSCENARIOS`, `Status()`, `Scenarios()`, `Arm()`, and the **user-facing JSON `note`** |
| `services/detection-engine/src/detect/triggers.ts` | both `TriggerStatus` field comments; arm's 150s timeout rationale |
| `services/detection-engine/src/detect/series.ts` | `DEFAULT_SPAN_SECONDS`'s worked example |
| `apps/dashboard/src/components/TriggerRail.tsx` | five comments (all comments — no behaviour change) |
| `README.md` | the terminal-arming recipe |
| `docs/demo/cue-sheet.md` | §1.2 and §1.2b re-measured; the act table |

### Two things fixed on the way past, both self-inconsistencies the change exposed

1. **`TriggerDispatcher`'s jobbing block asserted the 504 was certain.** At a 30s settle it is not — 30 is under IRIS's measured 60s Apache limit, so a synchronous request would now come back. `pool_bottleneck` **stays** jobbed, and the comment now gives three reasons that do not depend on the settle: the `activating` state exists *because* the POST returns immediately; 30s against 60s is a 2× margin on a number that is now presentational and so gets nudged; and `pWarmSeconds` is a method default a terminal caller can override on a path the route cannot see.
2. **The cue sheet claimed sub-90s countdowns render as *"under a minute"*.** The shipped `humanDuration` rounds to whole seconds below 90 (`~40 s`, `~18 s`, `~5 s`) and to minutes above. The measured table contradicted the note directly beneath it, so the note is corrected.

## Verification

- `Triggers.cls` and `TriggerDispatcher.cls` compiled in the container from the `/pg-src` bind mount, `errcount=0` each.
- The new user-facing `note` read back off the live endpoint: `Settling for ~30s with nothing armed, so the room sees a healthy production before the fault.`
- Engine `tsc --noEmit` clean; dashboard `tsc --noEmit` clean.
- `npm test` in `services/detection-engine`: **459 pass, 0 fail**.
- `docker compose up -d --build dashboard` succeeded (§6.1 — the build that ships); `http://localhost:5173/` → 200.

## Out of scope, and it needs a decision

**The 30s settle does not make a 5-minute demo.** The cue sheet's act structure targets ~9–10 minutes and Act 1 is the only act this change shortens (~47s). The floor is Act 2.3: the drain is a real production catching up, measured at ~90s to an empty queue and ~200s to a clean board, and it is the beat the whole demo exists for. A 5-minute cut is a **different script** — most plausibly Acts 1+2 only, with 2.3 called at an empty queue rather than a clean board. Flagged in the cue sheet rather than done, because re-cutting the script is a product decision and would need its own measurement pass.

Ari-Glikman:
Closes #244.

`PoolBottleneck()`'s `pWarmSeconds` default goes **75 → 30**, because the demo slot went to 5 minutes. Requested by @Ari-Glikman, 2026-09-07.

**One line of code changes.** Everything else in this diff is prose — but the prose was the problem, because five files documented the wait as load-bearing and it has not been for weeks.

## Why 30 is safe: both documented reasons were already dead

**Dead #1 — the zero baseline (#43).** Superseded when `thresholds.json` gained `referenceBaselines` stating `Cloud API`.`queued = 0`; `effectiveBaseline()` returns the stated value *instead of* the rolling mean, so `minBaselineSamples` never gates that path. Already on the record as dead.

**Dead #2 — Early Warning's `minFitSamples` (12) × 5s poll = 60s.** This one was still believed, and it is the interesting failure. `BaselineStore.recent()` filters **only** on `fitWindowSeconds` and **deliberately ignores `#regimeStart`** — its own comment says so. So `fitSampleCount` is a function of **engine uptime**, not of anything this trigger does, and it survives a `Reset all` intact.

Measured at the instant the arm POST was sent, both runs:

```
run 1   fitSampleCount=20  fitSpanSeconds=95
run 2   fitSampleCount=49  fitSpanSeconds=240
```

Twelve samples were banked long before the button was pressed. **The wait was never producing them.**

The lesson is why the autopsy goes in the code and not only in the issue: the comment reasoned from a threshold *name* to a delay, without checking where the samples come from. Both times, the delay outlived its reason by weeks — because the reason was written down and the check was not.

**The one genuinely cold case is a cold engine** (~60s after `compose up`). 30s does not cover that and is not meant to; it is a stack precondition #108 already requires, now stated in the cue sheet as the only real one. `Triggers.cls` records `A NUMBER TO PUT BACK IF THE FIT IS EVER COLD ON PURPOSE: 60s, not 75` — the 75 was 60 plus margin for a reason that no longer holds.

## What the 30s buys, stated plainly

Pacing. The room sees a healthy production for a beat before the fault, and the presenter gets to say what is about to happen while nothing is happening. That is now the documented reason, in place of two that were not true.

## Measured

Two consecutive arms, containerised stack, warm engine, 2026-09-07. `t` is seconds from the arm POST, which returns in **0.02s** because the dispatcher jobs this scenario:

| | run 1 | run 2 |
|---|---|---|
| armed (`DispatcherMs` set) | +30s | +30s |
| first Early Warning projection | +40s | +40s |
| queue crosses the floor of 50 | +91s | +86s |
| confirmed `queue_buildup` | +96s | +91s |

The Early Warning strip, run 2 — now the cue sheet's §1.2b table verbatim:

```
+40s  q=4   Queue depth 4 rising ~3.8/min; at this rate it crosses 50 in ~12 min.
+50s  q=14  rising ~19.6/min; ... in ~2 min.
+61s  q=24  rising ~39.4/min; ... in ~40 s.
+71s  q=34  rising ~55.2/min; ... in ~18 s.
+81s  q=45  rising ~61.4/min; ... in ~5 s.
+86s  q=52  already_crossed
```

The rate converges on ~61/min — one message a second, exactly the net inflow the scenario creates.

## Two alternatives ruled out

- **Lower `POLL_INTERVAL_MS`** — breaks `thresholds.json`'s documented invariant `(sustainedSamples-1) × interval > sustainedSeconds × 1000` (`1 × 5000 > 4000`; 4500ms already fails). Declined twice: #64, #65.
- **Lower `minFitSamples`** — unnecessary once `recent()`'s behaviour is understood, and it would weaken a published slope for no gain.

## Two self-inconsistencies the change exposed, fixed here

**1. `TriggerDispatcher`'s jobbing block asserted the 504 was certain.** At a 30s settle it is not — 30 is under IRIS's measured 60s embedded-Apache limit, so a synchronous request would now come back. `pool_bottleneck` **stays** on `SLOWSCENARIOS`, and the comment now gives three reasons that do not depend on the settle: the `activating` state exists *because* the POST returns immediately; 30s against 60s is a 2× margin on a number that is now presentational and therefore gets nudged for a room; and `pWarmSeconds` is a method default a terminal caller can override on a path the route cannot see.

**2. The cue sheet claimed sub-90s countdowns render as *"under a minute"*.** The shipped `humanDuration` rounds to whole seconds below 90 (`~40 s`, `~18 s`, `~5 s`) and to minutes above. The measured table contradicted the note directly beneath it.

## Files

| File | What changed |
|---|---|
| `iris/labdemo/Triggers.cls` | **the default**; the autopsy of both dead reasons; the measured table |
| `iris/labdemo/REST/TriggerDispatcher.cls` | `SLOWSCENARIOS`, `Status()`, `Scenarios()`, `Arm()`, and the **user-facing JSON `note`** |
| `services/detection-engine/src/detect/triggers.ts` | both `TriggerStatus` field comments; arm's 150s timeout rationale |
| `services/detection-engine/src/detect/series.ts` | `DEFAULT_SPAN_SECONDS`'s worked example |
| `apps/dashboard/src/components/TriggerRail.tsx` | five comments — all comments, no behaviour change |
| `README.md` | the terminal-arming recipe |
| `docs/demo/cue-sheet.md` | §1.2 and §1.2b re-measured; the act table |

## Verification

- `Triggers.cls` and `TriggerDispatcher.cls` compiled in the container from the `/pg-src` bind mount, `errcount=0` each.
- The new user-facing `note` read back off the **live** endpoint: `Settling for ~30s with nothing armed, so the room sees a healthy production before the fault.`
- Engine `tsc --noEmit` clean; dashboard `tsc --noEmit` clean.
- `npm test` in `services/detection-engine`: **459 pass, 0 fail**.
- `docker compose up -d --build dashboard` succeeded (§6.1 — the build that ships); `http://localhost:5173/` → 200.
- Instance left reset: nothing armed, all three hosts `OK`, `queued 0`.

## Out of scope, flagged not decided

**The 30s settle does not make a 5-minute demo.** The cue sheet targets ~9–10 minutes and Act 1 is the only act this shortens (~47s). The floor is Act 2.3: the drain is a real production catching up — measured at ~90s to an empty queue and ~200s to a clean board — and it is the beat the whole demo exists for. A 5-minute cut is a **different script**, most plausibly Acts 1+2 only with 2.3 called at an empty queue. Recorded in the cue sheet's act table rather than done, because re-cutting the script is a product decision needing its own measurement pass.


Ari-Glikman:
## Self-review

Whole project is under one owner (@Ari-Glikman's standing authorization), so this is a self-review rather than a second pair of eyes. Reviewed as a reader who does not already believe the conclusion.

**Did I actually verify the claim the change rests on, or reason from a name?** That is the exact failure this PR indicts, so it is the first thing to check on myself. The claim is `recent()` ignores the regime boundary. I read `window.ts`, and I also confirmed it behaviourally: `fitSampleCount` read 49 with `fitSpanSeconds` 240 on an engine that had been up ~4 minutes, on a queue that had been flat at zero throughout — and the count kept climbing monotonically through a `Reset all` earlier in the session. Both the code and the observable agree.

**One behavioural line, and I checked the blast radius of the number rather than the code.** `pWarmSeconds As %Integer = 30` is the whole functional diff. Everything else is prose — verified by `git diff` reading as comment bodies only in the two `.ts` files and `TriggerRail.tsx`, and the engine's 459 tests are unchanged and pass, which they would not need to be if any of it were live.

**The `note` string was the one that could not be caught by a typecheck.** It is a user-facing JSON literal inside an ObjectScript method — no test reads it and no compiler cares what it says. It said "so Early Warning has enough samples to project the crossing", which was both the wrong number *and* the dead reason, shown to a presenter mid-demo. Verified by recompiling and reading it back off the live endpoint rather than off the source.

**Where I deliberately did NOT change behaviour, twice:**
- `SLOWSCENARIOS` keeps `pool_bottleneck` even though the 504 argument no longer forces it. Un-jobbing would collapse the three-state rail to two — that is #135's whole subject — and 30s against a measured 60s limit is a 2× margin on a number that is now presentational, i.e. the kind that gets nudged.
- Arm's 150s timeout stays at 150s. Its stated reason is dead but a generous ceiling on a once-per-demo route costs nothing, and `pWarmSeconds` is a *method default* a terminal caller can override on a path the engine cannot see. Corrected the prose, not the value — retuning a timeout to a measurement I took on one build is how you get an abort on a path you did not measure.

**What the three-state model got right, and it is worth naming as the counter-example.** `TriggerDispatcher`'s phases are keyed on witnesses the scenario sets (`PoolWas` = request landed, `DispatcherMs` = armed), not on a duration. So the UI adapted to 75 → 30 with **no code change at all** — only the prose describing it was stale. Every place that had to be touched here is a place that wrote a duration down. That is the argument for witnesses over timers, and it now says so in the class.

**Two stale claims found while measuring, both fixed rather than left:** the jobbing block's certainty about the 504, and the cue sheet's "under a minute" rendering (`humanDuration` rounds to whole seconds below 90 — the measured table sat directly above a note contradicting it). Neither was caused by this change; both were exposed by measuring instead of trusting.

**Measured, not computed.** Two runs, both in the PR body. I did not derive "+86s" from "30 + 56" — the arithmetic would have been close and wrong, since run 1 crossed at +91s and run 2 at +86s on identical settings. Timed from the host over HTTP, not through an MCP tool call, which carries ~15s of round trip and would have poisoned every number.

**Not in this PR, and it is the thing @Ari-Glikman actually needs to decide:** a 5-minute demo. This buys ~47s of Act 1 against a ~9–10 minute script whose floor is a real production draining. Recorded in the cue sheet's act table with the one plausible cut named, and left there.

Merging.

PR #246, opened 2026-09-14
fix(dashboard): let the image override the Management Portal URL

Ari-Glikman:
## The bug

The header's **Management Portal** link opened `http://localhost:52773` from every container ever built from `apps/dashboard/Dockerfile`. Reported from the EKS deployment (`guardian.intersystemsisrael.com`), where the button pointed the audience at their own machine.

## Root cause

Not the default — it is *correct* for `docker compose up`, which publishes 52773 on the host. The bug is that nothing could override it.

Vite substitutes `import.meta.env.VITE_IRIS_PORTAL_URL` while `npm run build` runs, i.e. **inside the build stage**, so the value is frozen into the emitted HTML. nginx cannot envsubst it at container start the way it does `HEALTHSCAN_UPSTREAM`. The Dockerfile declared no matching `ARG`, so the documented variable had **no path into a containerised build at all** — and the two things you would reach for both look right and do nothing:

| attempt | why it silently fails |
|---|---|
| `apps/dashboard/.env.local` | root `.dockerignore` excludes `**/.env.*` from the context |
| `environment:` on the serve stage | Vite already ran; there is nothing left to substitute |

## The change

- **`apps/dashboard/Dockerfile`** — `ARG VITE_IRIS_PORTAL_URL=""` + `ENV`, declared *after* `npm ci` so changing the URL does not invalidate the install layer. Empty default, because `portalUrl()` treats unset and empty alike: an unpassed build keeps today's compose-correct behaviour, and **no hostname enters this repo** (`apps/dashboard/CLAUDE.md` §5/§9, #25). The deployment that knows its own IRIS hostname supplies it.
- **`docker-compose.yml`** — passes it through as `PG_IRIS_PORTAL_URL`, as a `build.args` entry rather than an `environment:` one, with the rebuild-not-restart consequence recorded.
- **`.env.example`** / **`AppShell.tsx`** — record that this is a build input, in both places someone will look.

## Consequence worth knowing

Passing the ARG makes the image **environment-specific**: two builds of one commit with different values are not interchangeable, so they cannot share a tag. The EKS deployment tags accordingly.

## Verification

- `docker compose config` resolves `args: VITE_IRIS_PORTAL_URL: <value>` from `PG_IRIS_PORTAL_URL`.
- `https://iris.intersystemsisrael.com/ui/interop/interop-editor/index.html?%24NAMESPACE=LABDEMO&%24PRODUCTION=ProductionGuardian.LabDemo.Production` → **200** through the ALB and Web Gateway (`[APP_PATH:/ui]` is enabled in the gateway's CSP.ini), so the value the deployment will pass is a measured URL, not a guess.
- The legacy `/csp/healthshare/labdemo/EnsPortal.ProductionConfig.zen` fallback named in the code comments returns **404** on this base image (community IRISHealth, not healthshare) — worth knowing before anyone reaches for it.
- Image build + on-screen check of the rendered `href` follow this merge; the built bundle is grepped for the URL.


Ari-Glikman:
**Correction to one line in the description above.** It claimed the legacy fallback named in `AppShell.tsx`'s comments — `/csp/healthshare/labdemo/EnsPortal.ProductionConfig.zen` — returns 404 "on this base image (community IRISHealth, not healthshare)". That is wrong, and nothing in this PR depends on it.

The 404 came from the EKS Web Gateway config, not from the image. `Security.Applications` on the deployed instance lists `/csp/healthshare/labdemo` (ns=`LABDEMO`, enabled) — it *is* an IRIS for Health build with HSLIB/HSSYS present, so LABDEMO's web application lives under `/csp/healthshare`, exactly where the code comment says. The hand-derived `CSP.conf` simply had no `<Location "/csp/healthshare">`, so Apache never handed the request to the Web Gateway module and answered off disk:

```
404 Not Found -- Apache/2.4.58 (Ubuntu) Server at iris.intersystemsisrael.com Port 80
```

Apache's own page, naming Apache, with no `CSP.log` entry. `CSP.ini` already had `/csp=Enabled` and an `[APP_PATH:/csp]` section, which is what made the coverage look complete — but `CSP.ini` only says *where* to route a request that Apache has already handed over, so it cannot rescue a path with no `<Location>`.

Added the `<Location>` and a matching `[APP_PATH:/csp/healthshare]` in the deployment manifests (not in this repo). That URL now returns **200** with the IRIS login page, and the fallback the code comments describe is genuinely available again. No regressions: `/csp/sys/cache_status.cxw`, `/csp/sys/UtilHome.csp`, `/ui/interop/interop-editor/`, `/api/monitor/metrics` and the dashboard's `/api/healthscan/health` all still 200.

Reported by @Ari-Glikman.

Issue #251, opened 2026-09-15
chat: "which host is handling the most messages" names the upstream hosts — CompareHostActivity publishes a zero-width window and no rate


Ari-Glikman:
Reported by the owner: with `pool_bottleneck` armed, asking the chat **"which host is handling the most messages"** answers **EMR Source and Lab Router** — the two hosts *feeding* the bottleneck — and describes the window as "the last second".

The model is not fabricating. `CompareHostActivity` hands it a payload with no window in it.

## Root cause

`Ens_Activity_Data.*.TimeSlotUTC` names the **instant a bucket starts**, and `Period` is how wide that bucket is: **10 seconds** at `seconds` resolution, not 1.

`CompareHostActivity` published `from` and `to` as bucket **start** instants, plus `bucketsRequested`, and **nothing else**:

- the stated window is short by exactly one bucket, always;
- at `buckets = 1`, `from == to` — a **zero-width instant** published as a window;
- no bucket width and no window duration anywhere in the payload, so a consumer cannot tell 10 seconds from an hour;
- no per-host rate, though `GetActivityTrend` publishes `messagesPerSecond` per bucket and a private `Rate()` helper already exists in the same class.

**The width was already being queried.** The SQL selects `MIN(Period) AS Secs` and the row loop never reads it.

## Evidence

Two identical chat runs, `toolCalls: 1`, `evidence: []`, both answering *"Cloud API, EMR Source, and Lab Router, each with 7 messages processed in the last second… Cloud API having an average processing time of 1.01 seconds"*.

Sessions are not persisted and chat tool calls are not audited, so the transcript could not be recovered. Instead the tool was enumerated across every plausible argument combination on the live pod and matched to the answer exactly:

```
("seconds",  1)  from 2026-09-15T08:00:50Z -> to 2026-09-15T08:00:50Z    <- from == to
                 EMR Source msgs=8 proc=0.00275
                 Lab Router msgs=8 proc=0.001
                 Cloud API  msgs=7 proc=1.011286
("seconds",  2)  07:00:40Z -> 08:00:50Z     18  | 18  | 17
("seconds",  6)  08:00:00Z -> 08:00:50Z     57  | 57  | 57
("seconds", 24)  07:57:00Z -> 08:00:50Z     235 | 235 | 234
("hours",    1)  08:00:00Z -> 08:00:00Z     57  | 57  | 57
```

[...]

Issue #196, opened 2026-09-01
A dispatcher guard refusal leaves no durable trace: droppedEvidence never reaches the audit trail its own comment claims, and all four guards are affected


Ari-Glikman:
Found while measuring #192, and it is narrower than what I first offered to file there — the reach of these fields is documented and deliberate. What is wrong is one half of the documented claim, and the consequence is that **a guard refusal leaves no durable trace anywhere.**

## The claim

`AgentDispatcher.DropUnsupportedConfigAbsence`, lines 470-474:

> `droppedEvidence` **REACHES THE AUDIT TRAIL AND THIS INSTANCE'S OWN REPLY, NOT THE API.** Same reach as `droppedAction` and `droppedRemediation` … `investigate.ts` builds its response field by field from a parsed subset, so an extra key here is ignored rather than rejected — no contract change, and no route to the dashboard either.

The second half is true: `parsed.%ToJSON()` at line 143 writes the whole object, `droppedEvidence` included, as the reply from `/labdemo/investigate`, and `investigate.ts` does discard it by construction.

## The first half is not true

Only two things in `iris/**` write to `Audit.Entry`:

```
Tools/AuditPolicy.cls:76    do ...Audit.Entry).Record(   <- a tool execution
Tools/AuthPolicy.cls:147    do ...Audit.Entry).Record(   <- an authorization denial
```

Both are tool-level and both fire *before* the dispatcher inspects the reply. Nothing writes a row when a guard refuses a claim. So the four guards —

| guard | what it refuses |
|---|---|
| `DropUnapplicableAction` | an action the whitelist cannot apply |
| `DropPoolFixDuringOutage` | a pool fix recommended during an outage |
| `DropStaleConnectivityClaim` | a connectivity claim the evidence does not support |
| `DropUnsupportedConfigAbsence` | a config-absence claim the audit log cannot support |

— all record their refusal into an object whose only consumer discards it, and none of them into anything that outlives the request. The engine does not log the raw reply either. A reviewer asking "did the dispatcher refuse anything during that demo run?" has nothing to read, and "the AI recommended X and we suppressed it" is not reconstructable after the fact.

[...]

PR #256, opened 2026-09-15
audit: record the chat turn -- the question, the answer and a turn id (#255)

### LINKED ISSUE #255: chat: the audit table records every tool call but never the question or the answer
## The premise I opened this on was wrong, and the measurement is worth keeping

I claimed in #251 that "nothing records what the chat model was handed" and that the bad run behind that issue was permanently unreproducible. Both false. `Tools.Governance.GovernAgent` attaches `Tools.AuditPolicy` **before** any early return (`Governance.cls:251`), and the chat path calls it like every other — so every chat tool call has been writing an `executed` row all along. Measured on the live pod:

```
COMPAREHOSTACTIVITY  EXECUTED n=18   newest 2026-09-15T08:53:27Z
GETACTIVEFINDINGS    EXECUTED n=11
GETRECENTERRORS      EXECUTED n=41
                                        total 139 rows
```

And the specific run I called unrecoverable is row **132**:

```
132  2026-09-15T07:52:59Z  args={"buckets":60,"resolution":"seconds"}  -> messages 764
```

60 buckets at `seconds` is 600 seconds. The model reported it as "the last minute". The entire #251 defect chain is reconstructable from this table — 132 (600s as a minute), 133/134 (`buckets:1`, 10s as a second), 135 (`{}` with **no `window` field**, the pg4 run that said "over the last hour"), 136 (`window: "23.8 hours"`, pg5), 137 (`buckets:1` -> `"10 seconds"`, right label wrong span), 139 (`buckets:6` -> `"60 seconds"`, fixed).

Keeping this here because the failure was not the absence of a record, it was **not querying the record that existed** — and the fix I proposed off that mistake would have added a second recorder next to a working one.

## What is actually missing

The table records what the model was **handed**, in full: tool, arguments, result JSON, actor, roles, timestamp. It records nothing about the **turn**:

1. **The question.** No row holds the operator's text. Rows 138 and 139 are one turn each and nothing says which question produced either.
2. **The answer.** The row carries the tool's return value, never the sentence the model built from it. #251 was a defect *in the prose over correct data* — row 135 is a perfectly good 48588 next to an answer that said "the last hour", and the audit table can only show you the 48588. The wrong sentence survived solely because I was the one who made the HTTP call.
3. **Turn correlation.** Rows 128/129/130 are three identical `buckets:10` calls two seconds apart. One turn asking three times, or three turns? Unanswerable. `RecordedAt` has one-second resolution and CSP pools processes across requests, so neither time nor process ordering is sound.

(3) is what makes (1) and (2) worth more than a log line: without it a turn row cannot be tied to the calls that fed it.

## Proposed

One row per chat turn, through the existing storage, plus a correlation column.

- **`TurnId`**, new property. Minted per request in `ChatDispatcher.Ask`, stashed process-private (the pattern `StashFindings` already uses), read by `AuditPolicy.%LogExecution` — which runs in the caller's own process, so it can see it. Stamped on tool rows and on the turn row; "" on everything written outside a chat turn, which is every existing row.
- **`Disposition = "answered"`**, a fourth value. Its own writer, by the rule the property comment already sets: `executed` / `denied` / `dropped` each exist because the writer before could not see the event, and a *turn* is not visible to any of the three — the audit hook fires per tool, inside the turn.
- **`Tool = "Ask"`**, the ObjectScript method name, as `dropped` rows already hold a guard's method name rather than a tool's.
- `publishHandle = 0`, by `RecordRefusal`'s argument: a turn is not the call this process just made, and clobbering the process-private handle would hand a `SetPoolSize` response the turn's id as its `auditId`.
- Swallows its own failure. A turn that answered correctly must not 500 because its record could not be written.

## The boundary question, since this writes new text to storage

The **question** is operator free text and may contain anything typed. The **answer** is model prose over tool output. Neither leaves the instance — this table is never sent anywhere, unlike a tool return value (`Audit.Entry` class comment) — and `root CLAUDE.md` §2.1's rule is about what *leaves*. So this is in bounds, but it is the first row content not derived from a tool's guaranteed-clean return, and that is worth stating in the class rather than inferred.

Ari-Glikman:
Closes #255.

## The correction first, because it is the more useful half

I told the owner that nothing recorded what the chat model was handed, and that the run behind #251 was permanently unreproducible. Both were false, and one `SELECT` disproved them. `Tools.Governance.GovernAgent` attaches `Tools.AuditPolicy` **before any early return** (`Governance.cls:251`), and the chat path calls it like every other — so every chat tool call has been writing an `executed` row since the chat shipped.

```
COMPAREHOSTACTIVITY  executed  n=18   newest 2026-09-15T08:53:27Z
GETACTIVEFINDINGS    executed  n=11
GETRECENTERRORS      executed  n=41            total 139 rows
```

The specific run I called unrecoverable is row 132:

```
132  2026-09-15T07:52:59Z  {"buckets":60,"resolution":"seconds"}  ->  messages 764
```

60 buckets at `seconds` is **600 seconds**, reported as "the last minute". The whole #251 chain is legible in six rows — 132 (600s as a minute), 133/134 (`buckets:1`, 10s as a second), 135 (`{}`, **no `window` field**, the pg4 run that said "over the last hour"), 136 (`"23.8 hours"`, pg5), 137 (`buckets:1` → `"10 seconds"`, right label wrong span), 139 (`buckets:6` → `"60 seconds"`, fixed).

The fix I originally proposed would have added a second recorder beside a working one. The rule is now in `iris/CLAUDE.md` §7: **a record nobody queries is indistinguishable from one nobody wrote.**

## What this actually adds

The table recorded what the model was *handed*, completely, and nothing about the *exchange*:

| Question | Before |
|---|---|
| what tool ran, with what arguments, returning what | recorded — never a gap |
| which **question** caused it | no |
| what the assistant **said** | no |
| which calls shared a **turn** | no |

For a defect **in the prose over correct data** that is the half that matters. Row 135 is a correct `48588` next to a wrong sentence, and only the 48588 survived.

- **`TurnId`** — read inside `Record` rather than passed as an argument, which is what leaves the other three writers untouched. `%AI.Policy.Audit.%LogExecution`'s signature is fixed by `%AI.ToolMgr`, so there is no parameter to add to the writer that most needed correlating; a process-private node set by `Ask` is visible to all of them, and a fifth writer is correlated for free.
- **`answered`**, a fourth disposition, by the rule the `Disposition` comment already sets: a writer exists because the one before it could not see the event. The audit hook fires per *tool*, from inside the turn.
- **`DescribeTurn`** and **`RecentTurns`** — because a turn id you cannot query is the thing that opened this issue.

**Why a column and not an ordering.** Rows 128/129/130 are identical `{"buckets":10,"resolution":"seconds"}` calls spanning two seconds. One turn or three is unanswerable: `RecordedAt` has one-second resolution and CSP pools processes across requests.

## Verification

Compiled in-pod (the storage-map addition is the risky part — 139 existing rows read back unchanged, `turnId` empty), then end to end against `https://guardian.intersystemsisrael.com/api/chat`:

```
Q: which host is handling the most messages in the last five minutes?

143  2026-09-15T09:56:30Z  CompareHostActivity  executed  chat-20260915095628-1112
144  2026-09-15T09:56:32Z  Ask                  answered  chat-20260915095628-1112

DescribeTurn("chat-20260915095628-1112"):
  Q: which host is handling the most messages in the last five minutes?
  A: ...each processing 294 messages. All three are tied...
  model=gpt-4o-mini  toolCalls=1
  CALL pg-audit-143  args={"buckets":30,"resolution":"seconds"}  window="5 minutes"
```

Question, answer, arguments and window in one read. Three synthetic rows written during verification were deleted — a fabricated audit row is worse than none.

`459/459` tests pass, 86 suites.

## Deliberate asymmetries a reviewer should check rather than assume

- **A turn id with no `answered` row is a feature** — a turn that made tool calls and then failed to reply. The row is written on the success path only.
- **Small talk writes no turn row.** It reaches no tool and answers nothing about the production.
- **`answered` rows are not tool calls.** Counting calls from this table needs a `disposition` filter; this is the second value after `dropped` that makes that true.
- **`publishHandle = 0`**, by `RecordRefusal`'s argument: CSP reuses processes, so a resolve following a chat would otherwise report a *question* as its `auditId`.
- **First row content not derived from a tool's return value.** Operator free text and model prose are sanitised by nothing, because they came from no tool; the guarantee is that audit rows never leave the instance. Stated in §5.5 and in the class rather than left to inference.

## Contracts

§5.5 opened with **"Two mechanisms, not one"** and had been two writers short — `dropped` shipped with #196 and was never documented there. Now four, with the generating rule stated once so the next writer is not caught the same way. `resolve.schema.json` was checked first: `disposition` is not enum-constrained, so the fourth value breaks no validation. CHANGELOG entry included.


Ari-Glikman:
Self-review found a real defect in the first commit, fixed in `441c294`.

Clearing the stashed turn id only at the top of `Ask` protects the next **chat** request and nothing else. CSP pooling cuts both ways: the process that has just served a chat serves the next *resolve*, and those write audit rows through the same `Record`. A turn id left set would stamp them with an unrelated question — rows grouped under a question that did not cause them, which is worse than uncorrelated rows because the grouping looks authoritative. Precisely the failure the `TurnId` comment I had just written warns about.

The node now lives exactly as long as `agent.Run`, and `RecordChatTurn` takes the id explicitly. `Record` still reads the process node when none is supplied, because that is the only way the audit hook can be correlated at all — `%LogExecution`'s signature is fixed by `%AI.ToolMgr`. The top-of-request clear stays, since an exception inside `Run` skips the new one.

Verified on the case that distinguishes them:

```
145  CompareHostActivity  executed  chat-20260915100113-8060   during the turn
146  SetPoolSize          executed  []                         after the clear
147  Ask                  answered  chat-20260915100113-8060   id passed in
```

